PR #27, opened 2016-12-23
<Li> should be <li> in handlebars_templates/templates/index.html.hbs

rbalicki2:
* self explanatory.
* found this while going through the examples


SergioBenitez:
Thanks for catching that!

PR #39, opened 2016-12-24
Added IntelliJ IDEA exclusion to .gitignore. Fixed a bad URL in the c…

ereichert:
…ontrib template docs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,88 @@ scripts/upload-docs.sh
 
 # Backup files.
 *.bak
+
+## Text editor exclusions. All Rocket specific exclusions should be above this line. ##
+
+# Emacs files
+# -*- mode: gitignore; -*-
+\#*\#
+/.emacs.desktop
+/.emacs.desktop.lock
+*.elc
+auto-save-list
+tramp
+.\#*
+# Org-mode
+.org-id-locations
+*_archive
+# flymake-mode
+*_flymake.*
+# eshell files
+/eshell/history
+/eshell/lastdir
+# elpa packages
+/elpa/
+# reftex files
+*.rel
+# AUCTeX auto folder
+/auto/
+# cask packages
+.cask/
+dist/
+# Flycheck
+flycheck_*.el
+# server auth directory
+/server/
+# projectiles files
+.projectile
+# directory configuration
+.dir-locals.el
+
+# IntelliJ IDEA files
+.idea/
+
+# Sublime Text files
+# cache files for sublime text
+*.tmlanguage.cache
+*.tmPreferences.cache
+*.stTheme.cache
+# workspace files are user-specific
+*.sublime-workspace
+# project files should be checked into the repository, unless a significant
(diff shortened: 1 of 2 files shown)


saleemrashid:
I would suggest you add `.idea/` to your `.git/info/exclude` file

https://git-scm.com/docs/gitignore#_description (emphasis mine)
> Patterns which are specific to a particular repository but which **do not need to be shared with other related repositories** (e.g., auxiliary files that live inside the repository but are **specific to one user’s workflow**) should go into the $GIT_DIR/info/exclude file.

ereichert:
@saleemrashid In my experience that method is not used regularly despite the guidance. 

Further, and more importantly, it's a 6 character commit in a file that rarely changes. It seems like a rather benign change to cause a discussion of any kind.

saleemrashid:
@ereichert This was more intended for future reference than anything else.

ereichert:
@saleemrashid I understand. Thanks.

SergioBenitez:
I do think non-project related items shouldn't be in `.gitignore`. Is it alright if I merge this without that change?

ereichert:
I think it's an odd decision given the impact and that it's pretty standard to include exclusions for common editors.  But, it's your project to do with what you wish.

SergioBenitez:
Actually, that's a good idea. I'll merge this in and add other common editors as well. Can you rebase on master?

ereichert:
@SergioBenitez I rebased and added exclusions (grabbed from https://github.com/github/gitignore) for the editors I'm aware of being used in the Rust community. 

I fixed the URL in a separate commit to make it easy to work with should you change your mind about the exclusions.

saleemrashid:
@ereichert The reason I'm generally against this is there's no main editor for Rust so including certain editors starts flame wars and including all makes a messy `gitignore`. Anyway, if this gets merged I think only IDEA and VS should be included from the list.

ereichert:
@saleemrashid Based on what I saw at the Rust Belt Rust Conference I'd be willing to bet Vim and Emacs make up the vast majority of editors being used for Rust.

I've been using git since ~ late '08 and have never once heard or read of anyone arguing for not adding exclusions or trying to keep their gitignore file from being messy. 

I have, however, seen plenty of pull requests get rejected because of errant files and a lot of time wasted cleaning up a bad commit.

I can't understand why you wouldn't want to make a gitignore file as comprehensive as possible (within reason) so that contributors don't have to worry about which files should be included in a commit and which shouldn't. It seems to me the gitignore file is the right place to document that concern.

I have spent way, way too much time on this conversation. I've made my case clear. Sergio can do what he wishes.

SergioBenitez:
Merged in https://github.com/SergioBenitez/Rocket/commit/13cac7db4598f27376c1ea5fe67683eb8f42069b. I decided not to go with the gitignore changes; apologies for the back and forth. For some reason, I assumed there'd be ~3 to gitignore, which would be fine. But 85 was a bit much. I trust contributors will be aware of committing erroneous files or amend their PRs appropriately when it happens.

PR #59, opened 2016-12-26
Use ContentType::Plain for String responses

### LINKED ISSUE #49: `Content-Type` inconsistent between `String` and `&str`
The `Content-Type` for `String` is `text/html`, but it is `text/plain` for `&str`. This seems like an unintuitive behavior, one would expect them to behave the same. Is there a reason for this?

sethlopezme:
There is an inconsistency between `&str` and `String` responses. `&str` responses use `Content-Type: text/plain` and `String` responses use `Content-Type: text/html`, which may be unexpected.

This commit changes `String` responses to use `Content-Type: text/plain`.

Fixes #49

### DIFF
--- a/lib/src/response/responder.rs
+++ b/lib/src/response/responder.rs
@@ -188,12 +188,12 @@ impl<'r> Responder<'r> for &'r str {
     }
 }
 
-/// Returns a response with Content-Type `text/html` and a fized-size body
+/// Returns a response with Content-Type `text/plain` and a fixed-size body
 /// containing the string `self`. Always returns `Ok`.
 impl Responder<'static> for String {
     fn respond(self) -> Result<Response<'static>, Status> {
         Response::build()
-            .header(ContentType::HTML)
+            .header(ContentType::Plain)
             .sized_body(Cursor::new(self))
             .ok()
     }


LLBlumire:
I'm going to guess this decision was because of using format! for html templating, but I agree the behaviour should be the same

SergioBenitez:
I love that this is a single line change! Thanks! As it's a breaking change, this will be merged in as we approach 0.2.

sethlopezme:
@LegNeato Good catch! I've gone back and fixed it.

SergioBenitez:
We're now approaching 0.2! Looking to get this in.

Unfortunately, this won't be as simple as a one line change. :( This is because many of the tests and/or internal code assume that the Content-Type of a `String` is `HTML`. Of course, with this change, this will no longer be true.

I don't know the best way to do this, but we need to make sure that everything that currently returns a `String`, or anything where `String` is used to set the Content-Type, is modified to return a `content::HTML<String>` instead. I reran the testing suite to see if it would catch these in the examples, at least. Unfortunately it did not. So we should take this as an opportunity to improve the testing suite as well.

SergioBenitez:
I went ahead and implemented this in https://github.com/SergioBenitez/Rocket/commit/374593c2f0626bed6a205d0d42adb76b81289e27. Thank you for the initial push! :)

PR #60, opened 2016-12-26
Allow registration of handlebars helpers

philwhineray:


philwhineray:
I am pretty new to both rust and rocket, so if you would sooner see this done some other way or want me to make changes, I won't be offended.

ereichert:
I need this feature as well.

SergioBenitez:
Thanks for the pull request!

I'd like for Rocket to remain as engine agnostic as possible. I feel it does a good job today: Rocket is entirely agnostic as to whether you use Handlebars or Tera. I'd like to do the same for helpers.

In particular, this means that there should be _one_ `register_helper` method that registers a helper than can be used in either Tera _or_ Handlebars as well as any engines Rocket supports in the future. In other words, one API that registers a helper across all engines that Rocket supports. I'd imagine that this would be a static method under `Template`. So, you'd call `Template::register_helper(...)`, although it'd be good if there was a nicer name than `register_helper`. Maybe `add_helper`. Or something.

In any case, I'd like to stay from functions that work for only a single engine. This is a tall order, but I think it's important to create a consistent interface.

philwhineray:
I had a quick look round and as far as I can tell, Tera does not support helper functions at all, which makes this impossible.

I understand the desire to keep an agnostic interface but I do think that where you are allowing multiple implementations, not allowing the programmer access to the underlying functionality will lead to a lowest-common-denominator.

Which is not to say that this PR necessarily gets it right: I would be just as happy to have a mechanism to expose the underlying object (HANDLEBARS in this case) to manipulate it directly for those times when there is no equivalent functionality in the other permitted implementations.

ereichert:
I also think trying to remain agnostic and still support template libs is going to be really, really hard. Something like the Handlebars template helpers is essential.

> I would be just as happy to have a mechanism to expose the underlying object (HANDLEBARS in this case) 

I think this might be the better way to go.

SergioBenitez:
Tera supports the equivalent of handlebars inline helper functions via "filters", which should also function as block helper functions, though I'm not sure the latter has been implemented. It's entirely possible to expose this functionality in a template agnostic way, and it's the mechanism I would prefer. Being able to switch between template engines at no cost is a feature I'd like to preserve.

That being said, there will be features that exist in some engine and not others. It would be wrong for Rocket to prevent you from making use of those features.  Thus, Rocket should have a mechanism to interact with the underlying template engine directly. I'll have to think about the right way to expose this, but it is something I agree Rocket should do.

philwhineray:
OK, I don't think the Tera README mentioned how filters come into existence but I can see from the example what you mean.

That said, providing a templater-agnostic interface to the high quality of the remainder of the Rocket code is definitely beyond my current Rust ability and available time. When that changes, if there is no solution already, I'll come back to it.

I'll create a pair of issues and then close this PR.

PR #67, opened 2016-12-27
Add link to the homepage

clarus:
Add a direct link to the homepage from the README. I put it under the title but other positions could be a good fit too.


SergioBenitez:
How about a badge?

[Rocket Homepage](https://rocket.rs)

clarus:
Yes sounds good!

SergioBenitez:
Okay, I'll go ahead and add it! Thanks for the push to make this happen! :D

Issue #62, opened 2016-12-27
Every example should have a test suite


SergioBenitez:
At present, only about 25% of the crates in `examples/` have a test suite. The ideal, of course, is 100%. This is particularly important for examples as they can serve as a measure of understanding for new users. Let's get to 100%.

Below is a list of all of the examples. Those checked off have a complete test suite, while those that aren't do not.

- [X] config
- [X] content_types
- [X] cookies
- [X] errors
- [x] extended_validation
- [X] fairings
- [ ] form_kitchen_sink (**in progress**)
- [X] forms
- [X] from_request
- [X] handlebars_templates
- [X] hello_alt_methods
- [X] hello_person
- [X] hello_ranks
- [X] hello_world
- [X] json
- [X] managed_queue
- [X] manual_routes
- [X] optional_redirect
- [X] optional_result
- [X] pastebin
- [X] query_params
- [x] raw_upload
- [X] redirect
- [X] state
- [X] static_files
- [x] stream
- [x] state
- [X] testing
- [X] todo

This is a great opportunity for those looking to contribute. If you'd like to write a test suite for an example, please comment with the example's name and expected time of completion. The example will be annotated as "in progress". This will prevent overlapping work. Use the existing test suites as a basis for writing new tests; consistency is important.

P.S: While working on the test suites, be sure to record your experience with the testing library. I'd like for testing in Rocket to be a simple, straightforward, and enjoyable experience. Your feedback is essential to making this happen.

mattnenterprise:
@SergioBenitez I can do testing on the query_params example. I expect to have it completed by Friday.

sethlopezme:
@SergioBenitez I can take on the optional_redirect and optional_result examples, with the expectation that they will be complete by the end of tomorrow (12/28/2016).

chazelton331:
@SergioBenitez I can write tests for the `redirect` example. I expect it to be completed before 10:00pm (UTC) on Thursday (12/29/2016).

sethlopezme:
@SergioBenitez I will submit a PR for the cookies example shortly!

rbalicki2:
@SergioBenitez I will write a test for the `hello_alt_methods` example 👍

alanstoate:
@SergioBenitez I will write the test for the manual_routes by Monday (01/09/2016)

mike820324:
@SergioBenitez I would like to pick up the extended_validation test, if no one has working on this yet.
It should be finished before this weekend.

[...]

PR #68, opened 2016-12-27
Add test(s) for the config example

sethlopezme:
Related to #62.


SergioBenitez:
Testing this one is a bit tricky as you want to ensure that the config parameters are changing to what's in the `Rocket.toml` file depending on the active environment. The way to do this is to [programmatically set](https://doc.rust-lang.org/std/env/index.html) the `ROCKET_ENV` environment variable before doing `ignite()` and then using `config::active()` to get the active configuration and compare it against a manually constructed `Config` (see `config/mod.rs` for examples of constructing custom configs). You'll need to do this three times, one for each environment. Of course, you also want to test that the response is correct.

But configuration is global, so once it's set by one `ignite`, another `ignite` won't change it. Huh. Trickier than I thought. The only solution to this that I can think about at the moment is to spawn a new process for each environment to test. But that sounds like a silly way to do it. Rocket could also expose the non-active config information, but I'd rather it not. Tricky, tricky. Will come back to this. If you have any ideas, feel free to suggest them.

sethlopezme:
After discussing this with you in the #rocket irc channel, I went ahead and created environment tests in a `tests` directory so each test would be compiled as a separate crate during testing. This seems to have alleviated any issues with environment testing.

examples/config/tests/production.rs
```diff
@@ -0,0 +1,13 @@
+#![feature(plugin)]
+#![plugin(rocket_codegen)]
+
+extern crate rocket;
+use rocket::config::Environment;
+
+mod common;
+use common::test_config;
```
SergioBenitez: Don't use this. Just do `common::test_config` in the `#[test]` function. Same for the rest of the files.

examples/config/tests/common/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+    rocket::ignite().mount("/hello", routes![lib::hello]);
+
+    let config = config::active().unwrap();
+    match environment {
+        Environment::Development => {
+            assert_eq!(config.address, "localhost".to_string());
+            assert_eq!(config.port, 8000);
+            assert_eq!(config.log_level, LoggingLevel::Normal);
+            assert_eq!(config.env, config::Environment::Development);
+            assert_eq!(config.get_str("hi"), Ok("Hello!"));
```
SergioBenitez: Ensure there are only 2 extra parameters as well:

```rust
assert_eq!(config.extras().count(), 2);
```

examples/config/tests/common/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+            assert_eq!(config.get_bool("is_extra"), Ok(true));
+        }
+        Environment::Staging => {
+            assert_eq!(config.address, "0.0.0.0".to_string());
+            assert_eq!(config.port, 80);
+            assert_eq!(config.log_level, LoggingLevel::Normal);
+            assert_eq!(config.env, config::Environment::Staging);
```
SergioBenitez: Test the `extra` parameters.

examples/config/tests/common/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+        }
+        Environment::Production => {
+            assert_eq!(config.address, "0.0.0.0".to_string());
+            assert_eq!(config.port, 80);
+            assert_eq!(config.log_level, LoggingLevel::Critical);
+            assert_eq!(config.env, config::Environment::Production);
```
SergioBenitez: Test the `extra` parameters.

examples/config/tests/common/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+        }
+    }
+
+    assert_eq!(config.take_session_key(), None);
```
SergioBenitez: Add a comment explaining this.

examples/config/tests/common/mod.rs
```diff
@@ -0,0 +1,39 @@
+use std;
+
+extern crate rocket;
+use rocket::config::{self, Environment};
+use rocket::LoggingLevel;
+
+extern crate config as lib;
+
+pub fn test_config(environment: Environment) {
+    // set the environment for rocket
```
SergioBenitez: Make this comment more descriptive. Something like:

```rust
// Manually set the config environment variable so that Rocket initializes it in `init()`.
```

examples/config/tests/common/mod.rs
```diff
@@ -0,0 +1,39 @@
+use std;
+
+extern crate rocket;
```
SergioBenitez: Space between an `extern crate` and` use` line.

examples/config/tests/common/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+use rocket::config::{self, Environment};
+use rocket::LoggingLevel;
+
+extern crate config as lib;
```
SergioBenitez: Group `extern crate`s together.

examples/config/src/main.rs
```diff
@@ -2,14 +2,10 @@
 #![plugin(rocket_codegen)]
 
 extern crate rocket;
+extern crate config as lib;
```
SergioBenitez: Why rename it here?

SergioBenitez:
Also needs a rebase + squash of all commits with a commit message of: "Add tests for config example."

examples/config/src/main.rs
sethlopezme: I renamed it to `lib` because it was also renamed in the tests. I can remove the renaming–no biggie!

sethlopezme:
I was able to get my changes down to 2 commits, but that's about the best I can get it. Perhaps you have better git skills than I do?

SergioBenitez:
Merged in https://github.com/SergioBenitez/Rocket/commit/e650587159f90b15de6c9c2ec91b36f0d236204b.

PR #69, opened 2016-12-27
Add test(s) for the optional_redirect example

sethlopezme:
Relates to #62.

### DIFF
--- a/examples/optional_redirect/src/tests.rs
+++ b/examples/optional_redirect/src/tests.rs
@@ -1,0 +1,38 @@
+use super::rocket;
+use rocket::testing::MockRequest;
+use rocket::http::{Method, Status};
+
+fn test_200(uri: &str, expected_body: &str) {
+    let rocket = rocket::ignite()
+        .mount("/", routes![super::root, super::user, super::login]);
+    let mut request = MockRequest::new(Method::Get, uri);
+    let mut response = request.dispatch_with(&rocket);
+
+    assert_eq!(response.status(), Status::Ok);
+    assert_eq!(response.body().and_then(|b| b.into_string()),
+               Some(expected_body.to_string()));
+}
+
+fn test_303(uri: &str, expected_location: &str) {
+    let rocket = rocket::ignite()
+        .mount("/", routes![super::root, super::user, super::login]);
+    let mut request = MockRequest::new(Method::Get, uri);
+    let response = request.dispatch_with(&rocket);
+    let location_headers: Vec<_> = response.header_values("Location").collect();
+
+    assert_eq!(response.status(), Status::SeeOther);
+    assert_eq!(location_headers, vec![expected_location]);
+}
+
+#[test]
+fn test() {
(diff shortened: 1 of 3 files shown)


examples/optional_redirect/src/tests.rs
```diff
@@ -0,0 +1,27 @@
...
+#[test]
+fn test_200() {
+    test("/users/Sergio", Status::Ok, Some("Hello, Sergio!".to_string()));
+    test("/users/login",
+         Status::Ok,
+         Some("Hi! That user doesn't exist. Maybe you need to log in?".to_string()));
+}
+
+#[test]
+fn test_303() {
```
SergioBenitez: These should check the `Location` header to ensure the redirect goes to the right place. Something like:

```rust
let location: Vec<_> = response.header_values("Location").collect();
assert_eq!(location, vec!["/users/login"]);
```

sethlopezme: Good catch, I didn't think to test the Location header. Tests have been updated!

examples/optional_redirect/src/tests.rs
```diff
@@ -0,0 +1,38 @@
+use super::rocket;
+use rocket::testing::MockRequest;
+use rocket::http::{Method, Status};
+
+fn test_200(uri: &str, expected_status: Status, expected_body: String) {
```
SergioBenitez: This function is called `test_200`, so you shouldn't need the `expected_status` parameter. Test for Status::Ok directly below instead.

examples/optional_redirect/src/tests.rs
```diff
@@ -0,0 +1,38 @@
...
+    let rocket = rocket::ignite()
+        .mount("/", routes![super::root, super::user, super::login]);
+    let mut request = MockRequest::new(Method::Get, uri);
+    let mut response = request.dispatch_with(&rocket);
+
+    assert_eq!(response.status(), expected_status);
+    assert_eq!(response.body().and_then(|b| b.into_string()), Some(expected_body));
+}
+
+fn test_303(uri: &str, expected_status: Status, expected_location: &str) {
```
SergioBenitez: This function is called `test_303`, so you shouldn't need the `expected_status` parameter. Test for Status::SeeOther directly below instead.

examples/optional_redirect/src/tests.rs
SergioBenitez: Take `expected_body` as an `&str` and do a `to_string` inside this function.

sethlopezme: Good catch. Didn't mean to leave this in here.

examples/optional_redirect/src/tests.rs
sethlopezme: Good catch. Didn't mean to leave this in here.

SergioBenitez:
Merged in 9580d6cdfde8cdf947d9be3ea7186cb2680fae00.

PR #70, opened 2016-12-28
Add test(s) for the optional_result example

sethlopezme:
Relates to #62.

### DIFF
--- a/examples/optional_result/src/tests.rs
+++ b/examples/optional_result/src/tests.rs
@@ -1,0 +1,24 @@
+use super::rocket;
+use rocket::testing::MockRequest;
+use rocket::http::{Method, Status};
+
+#[test]
+fn test_200() {
+    let rocket = rocket::ignite().mount("/", routes![super::user]);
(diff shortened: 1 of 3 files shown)


examples/optional_result/src/tests.rs
```diff
@@ -0,0 +1,17 @@
...
+    let rocket = rocket::ignite().mount("/", routes![super::user]);
+    let mut request = MockRequest::new(Method::Get, uri);
+    let response = request.dispatch_with(&rocket);
+
+    assert_eq!(response.status(), expected);
+}
+
+#[test]
+fn test_user() {
+    test("/users/Sergio", Status::Ok);
```
SergioBenitez: This should check the body as well.

sethlopezme: A test was added for the body of the user route.

SergioBenitez:
Merged in https://github.com/SergioBenitez/Rocket/commit/ab94e344b43ae9532dac18ad5258e02dbbcbf979.

PR #74, opened 2016-12-28
autoescape vars in .html.tera files

### LINKED ISSUE #73: xss with contrib tera module
It looks like the contrib/tera module (as demonstrated with the todo example) is subject to script injection.  

For example add a Todo with: <script>alert(1); </script>

It appears that tera enables html autoescaping for [.html extension files](https://github.com/Keats/tera/search?utf8=%E2%9C%93&q=autoescape_extensions), but this is subverted by putting the tera template in a .tera file.

roller:
Add .html.tera to list of file extensions that tera should
autoescape.  This should allow the same default XSS protection
provided by tera defaults, with the added the .tera extension.

Fixes #73

### DIFF
--- a/contrib/src/templates/tera_templates.rs
+++ b/contrib/src/templates/tera_templates.rs
@@ -8,11 +8,10 @@ use super::{TemplateInfo, TEMPLATE_DIR};
 lazy_static! {
     static ref TERA: Result<tera::Tera, String> = {
         let path: PathBuf = [&*TEMPLATE_DIR, "**", "*.tera"].iter().collect();
-        let escape_ext = [".html.tera", ".htm.tera", ".xml.tera",
-                          ".html", ".htm", ".xml"];
+        let ext = [".html.tera", ".htm.tera", ".xml.tera", ".html", ".htm", ".xml"];
         tera::Tera::new(path.to_str().unwrap())
             .map(|mut tera| {
-                tera.autoescape_on(escape_ext.to_vec());
+                tera.autoescape_on(ext.to_vec());
                 tera
             })
             .map_err(|e| format!("{:?}", e))
--- a/contrib/src/templates/tera_templates.rs
+++ b/contrib/src/templates/tera_templates.rs
@@ -8,15 +8,12 @@ use super::{TemplateInfo, TEMPLATE_DIR};
 lazy_static! {
     static ref TERA: Result<tera::Tera, String> = {
         let path: PathBuf = [&*TEMPLATE_DIR, "**", "*.tera"].iter().collect();
+        let escape_ext = [".html.tera", ".htm.tera", ".xml.tera",
+                          ".html", ".htm", ".xml"];
         tera::Tera::new(path.to_str().unwrap())
-            .map(|mut t| {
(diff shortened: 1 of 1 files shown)


contrib/src/templates/tera_templates.rs
```diff
@@ -8,7 +8,17 @@ use super::{TemplateInfo, TEMPLATE_DIR};
 lazy_static! {
     static ref TERA: Result<tera::Tera, String> = {
         let path: PathBuf = [&*TEMPLATE_DIR, "**", "*.tera"].iter().collect();
-        tera::Tera::new(path.to_str().unwrap()).map_err(|e| format!("{:?}", e))
+        tera::Tera::new(path.to_str().unwrap())
```
SergioBenitez: I'd prefer something more succinct, like this:

```rust
lazy_static! {
    static ref TERA: Result<tera::Tera, String> = {
        let path: PathBuf = [&*TEMPLATE_DIR, "**", "*.tera"].iter().collect();
        let ext = [".html.tera", ".htm.tera", ".xml.tera", ".html", ".htm", ".xml"];
        tera::Tera::new(path.to_str().unwrap())
            .map(|mut tera| {
                tera.autoescape_on(ext.to_vec());
                tera
            })
            .map_err(|e| format!("{:?}", e))
    };
}
```

contrib/src/templates/tera_templates.rs
```diff
@@ -8,7 +8,14 @@ use super::{TemplateInfo, TEMPLATE_DIR};
 lazy_static! {
     static ref TERA: Result<tera::Tera, String> = {
         let path: PathBuf = [&*TEMPLATE_DIR, "**", "*.tera"].iter().collect();
-        tera::Tera::new(path.to_str().unwrap()).map_err(|e| format!("{:?}", e))
+        let escape_ext = [".html.tera", ".htm.tera", ".xml.tera",
```
SergioBenitez: Can we call this something so that it the line fits in the 85 cols?

roller: Haha, 85 cols is the cutoff, eh?  Might be we're clashing because I've been trying for 80.   I didn't realize you were optimizing for minimizing line count.

I like putting this on two lines just to reinforce that the two lines of extensions are intended repetition:  If you added a different escaped extension you would expect to add an item to each line.  (Not that I imagine anyone would ever have a file named ".htm.tera".)

SergioBenitez: Yeah, 85 is arbitrary. Rust's default cutoff is 100.

Definitely not optimizing for line count. But, things that are concise are sometimes easier to read. I think that's the case here.

roller: I'd normally consider explicit naming more important to readability than vertical space; but I don't think this is worth re-litigating. I'll resubmit with your original edit.  Thanks for the review!

SergioBenitez:
Sweet! Thanks so much for reporting and fixing this!

Merged in https://github.com/SergioBenitez/Rocket/commit/8deac6dba939e29ebe680b569b8d7a97f9deb90e.

Issue #42, opened 2016-12-24
Templates cannot be organized into subdirectories


sgrif:
To reproduce:

* Create a file like `templates/posts/new.html.tera`
* Have a route which performs `Template::render("posts/new", &context)`

Expected behavior:

The template is rendered

Actual behavior:

> Error: Template 'posts/new' does not exist.

SergioBenitez:
Confirmed. Looks like a simple fix.

PR #75, opened 2016-12-29
Properly resolve nested template names on Windows

marcusball:
Issue #42 was addressed, however, it did not fix the issue on Windows systems. Template names were being parsed using Windows paths, so a file like `templates/posts/new.html.tera`, was given the name `posts\new`, resulting in the template not being found for the name `posts/new`. 

To fix this, I just added a string replace, when compiling for Windows, to replace `\` characters with `/`, when generating the template name. 

I also included some basic unit tests, and split some of the name logic into separate functions so as to help facilitate said unit tests.

### DIFF
--- a/contrib/src/templates/mod.rs
+++ b/contrib/src/templates/mod.rs
@@ -185,6 +185,32 @@ fn remove_extension<P: AsRef<Path>>(path: P) -> PathBuf {
     }
 }
 
+/// Splits a path into a relative path from TEMPLATE_DIR, a name that
+/// may be used to identify the template, and the template's data type.
+fn split_path(path: &Path) -> (PathBuf, String, Option<String>) {
+    let rel_path = path.strip_prefix(&*TEMPLATE_DIR).unwrap().to_path_buf();
+    let path_no_ext = remove_extension(&rel_path);
+    let data_type = path_no_ext.extension();
+    let name = remove_extension(&path_no_ext);
+
+    (rel_path,
+     to_template_name(name.to_string_lossy().into_owned()),
+     data_type.map(|d| d.to_string_lossy().into_owned()))
+}
+
+/// Converts a Windows path, like "foo\bar\file" to "foo/bar/file".
+/// Used to standardize template path names for different operating systems.
+#[cfg(windows)]
+fn to_template_name(name: String) -> String {
+    name.replace("\\", "/")
+}
+
+/// Returns `name` unchanged, on non-Windows systems.
+#[cfg(not(windows))]
+fn to_template_name(name: String) -> String {
+    name
+}
+
 /// Returns a HashMap of `TemplateInfo`'s for all of the templates in
 /// `TEMPLATE_DIR`. Templates are all files that match one of the extensions for
 /// engine's in `engine_set`.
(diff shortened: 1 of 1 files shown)


contrib/src/templates/mod.rs
```diff
@@ -185,6 +185,32 @@ fn remove_extension<P: AsRef<Path>>(path: P) -> PathBuf {
...
+
+/// Converts a Windows path, like "foo\bar\file" to "foo/bar/file".
+/// Used to standardize template path names for different operating systems.
+#[cfg(windows)]
+fn to_template_name(name: String) -> String {
+    name.replace("\\", "/")
+}
+
+/// Returns `name` unchanged, on non-Windows systems.
+#[cfg(not(windows))]
```
SergioBenitez: Add `#[inline(always)]` so this definitely becomes a noop.

contrib/src/templates/mod.rs
```diff
@@ -185,6 +185,32 @@ fn remove_extension<P: AsRef<Path>>(path: P) -> PathBuf {
...
+
+    (rel_path,
+     to_template_name(name.to_string_lossy().into_owned()),
+     data_type.map(|d| d.to_string_lossy().into_owned()))
+}
+
+/// Converts a Windows path, like "foo\bar\file" to "foo/bar/file".
+/// Used to standardize template path names for different operating systems.
+#[cfg(windows)]
+fn to_template_name(name: String) -> String {
```
SergioBenitez: Add `#[inline(always)]` so the call becomes a noop.

contrib/src/templates/mod.rs
```diff
@@ -200,19 +226,67 @@ fn discover_templates() -> HashMap<String, TemplateInfo> {
...
-            let data_type = path_no_ext.extension();
-            let name = remove_extension(&path_no_ext);
-            templates.insert(name.to_string_lossy().into_owned(), TemplateInfo {
-                full_path: path.to_path_buf(),
-                path: rel_path,
-                extension: path.extension().unwrap().to_string_lossy().into_owned(),
-                data_type: data_type.map(|d| d.to_string_lossy().into_owned())
-            });
+            let (rel_path, name, data_type) = split_path(&path);
+            templates.insert(name, TemplateInfo {
```
SergioBenitez: Did `rustfmt` do this? Don't think this is right.

SergioBenitez:
This is awesome! Thanks for catching and fixing this, it's much appreciated!

contrib/src/templates/mod.rs
marcusball: What part of this are you asking about? I manually undid some `rustfmt` changes, so as to match the existing code.

contrib/src/templates/mod.rs
```diff
@@ -185,6 +185,32 @@ fn remove_extension<P: AsRef<Path>>(path: P) -> PathBuf {
     }
 }
 
+/// Splits a path into a relative path from TEMPLATE_DIR, a name that
+/// may be used to identify the template, and the template's data type.
+fn split_path(path: &Path) -> (PathBuf, String, Option<String>) {
+    let rel_path = path.strip_prefix(&*TEMPLATE_DIR).unwrap().to_path_buf();
+    let path_no_ext = remove_extension(&rel_path);
+    let data_type = path_no_ext.extension();
+    let name = remove_extension(&path_no_ext);
```
marcusball: Ah, yeah, that's a good idea. I'll go ahead and change that.

contrib/src/templates/mod.rs
SergioBenitez: The indentation of the fields for the `TemplateInfo`. I'm not sure what the canonical thing is, though. If `rustfmt` did this and the rustc source does it this way too, then this is okay.

marcusball: Oh, I see. I committed a coding-sin and mixed tabs and spaces in my attempt to revert `rustfmt` changes. In my next change, I'll leave the format unchanged.

SergioBenitez: :o

I should have the test script catch that.

contrib/src/templates/mod.rs
```diff
@@ -185,6 +185,24 @@ fn remove_extension<P: AsRef<Path>>(path: P) -> PathBuf {
...
+/// Splits a path into a relative path from TEMPLATE_DIR, a name that
+/// may be used to identify the template, and the template's data type.
+fn split_path(path: &Path) -> (PathBuf, String, Option<String>) {
+    let rel_path = path.strip_prefix(&*TEMPLATE_DIR).unwrap().to_path_buf();
+    let path_no_ext = remove_extension(&rel_path);
+    let data_type = path_no_ext.extension();
+    let name = remove_extension(&path_no_ext).to_string_lossy().into_owned();
+
+    // Ensure template name consistency on Windows systems
+    let name = match cfg!(windows) {
```
SergioBenitez: Why not:

```rust
let mut name = remove_extension(&path_no_ext).to_string_lossy().into_owned();
if cfg!(windows) {
    name = name.replace("\\", "/");
}

File: contrib/src/templates/mod.rs [hunk 6]
```diff
@@ -200,19 +217,71 @@ fn discover_templates() -> HashMap<String, TemplateInfo> {
...
-            let data_type = path_no_ext.extension();
-            let name = remove_extension(&path_no_ext);
-            templates.insert(name.to_string_lossy().into_owned(), TemplateInfo {
-                full_path: path.to_path_buf(),
-                path: rel_path,
-                extension: path.extension().unwrap().to_string_lossy().into_owned(),
-                data_type: data_type.map(|d| d.to_string_lossy().into_owned())
-            });
+            let (rel_path, name, data_type) = split_path(&path);
+            templates.insert(name,
```
SergioBenitez: Ah, I really dislike this. What you had before was great.

marcusball: Alright, I'll go ahead and revert the `rustfmt`, but this time I'll make sure it's spaces to prevent the original issues.

SergioBenitez: Sorry for the trouble. I appreciate your diligence. :)

PR #76, opened 2016-12-29
Adding tests for the query_params example

mattnenterprise:


examples/query_params/src/main.rs
```diff
@@ -3,6 +3,8 @@
 
 extern crate rocket;
 
+#[cfg(test)] mod tests;
```
SergioBenitez: Can we split these into two lines? I'm okay with this, actually, but it's just what the rest of the example files do.

examples/query_params/src/tests.rs
```diff
@@ -0,0 +1,46 @@
...
+use rocket::http::Status;
+use rocket::Response;
+
+macro_rules! run_test {
+    ($query:expr, $test_fn:expr) => ({
+        let rocket = rocket::ignite()
+            .mount("/", routes![super::hello]);
+
+        let mut request = MockRequest::new(Get, format!("/hello{}", $query));
+
```
SergioBenitez: Remove this new line here.

examples/query_params/src/tests.rs
```diff
@@ -0,0 +1,46 @@
...
+        $test_fn(request.dispatch_with(&rocket));
+    })
+}
+
+#[test]
+fn age_and_name_params() {
+  run_test!("?age=10&name=john", |mut response: Response| {
+    let body_str = response.body().and_then(|body| body.into_string());
+    assert_eq!(body_str, Some(String::from("Hello, 10 year old named john!")));
```
SergioBenitez: Use `.into()` or `.to_string()` instead of `String::from`.

examples/query_params/src/tests.rs
```diff
@@ -0,0 +1,46 @@
...
+fn name_param_only() {
+  run_test!("?name=John", |mut response: Response| {
+    let body_str = response.body().and_then(|body| body.into_string());
+    assert_eq!(body_str, Some(String::from("Hello John!")));
+  });
+}
+
+#[test]
+fn no_params() {
+  run_test!("", |response: Response| {
```
SergioBenitez: Just for completeness, can you add a test with just `?`, and one that uses parameters that aren't in the struct, say something like `?x=y`, and `?age=10&name=john&complete=true`?

examples/query_params/src/main.rs
SergioBenitez: Huh. Alrighty then, this is okay1

mattnenterprise:
I have added the requested updates.

SergioBenitez:
Merged in 21ddb7390f5e770b75e5e231dc2e14e831816fbc.

PR #80, opened 2016-12-30
write tests for hello_alt_methods

rbalicki2:
* self explanatory
#62

### DIFF
--- a/examples/hello_alt_methods/src/tests.rs
+++ b/examples/hello_alt_methods/src/tests.rs
@@ -1,0 +1,31 @@
+use super::rocket;
+use rocket::testing::MockRequest;
+use rocket::http::Method;
+use rocket::http::Status;
+
+#[test]
+fn hello_world_alt_methods() {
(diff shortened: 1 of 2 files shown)


examples/hello_alt_methods/src/tests.rs
```diff
@@ -0,0 +1,31 @@
...
+use rocket::http::Status;
+
+#[test]
+fn hello_world_alt_methods() {
+    test(Method::Get, Status::Ok, Some("<!DOCTYPE html>"));
+    test(Method::Put, Status::Ok, Some("Hello, PUT request!"));
+    test(Method::Post, Status::NotFound, None);
+}
+
+fn test(
```
SergioBenitez: These are really long parameter names. Can we make them shorter while keeping them fairly descriptive? Maybe just `status` and `body_prefix`?

rbalicki2: sgtm, will change

SergioBenitez:
Merged in e4006c56c83b1fde21a637393253754e3e7d35f1.

PR #84, opened 2016-12-30
Add config option to disable emojis

LauriM:
I personally don't like the emojis on the console output, so I wrote a patch to make it possible to disable them.

Alternative path I was considering was to add some "replace emojis with blanks" on the logger part, but that would have affected the performance. As there were only couple of places where the emojis were used this seemed like a sensible way of handling this.

### DIFF
--- a/examples/config/Rocket.toml
+++ b/examples/config/Rocket.toml
@@ -4,7 +4,7 @@
 [development]
 address = "localhost"
 port = 8000
-disable_emojis = true
+disable_emojis = false
 log = "normal"
 hi = "Hello!"
 is_extra = true
--- a/examples/config/Rocket.toml
+++ b/examples/config/Rocket.toml
@@ -4,6 +4,7 @@
 [development]
 address = "localhost"
 port = 8000
+disable_emojis = true
 log = "normal"
 hi = "Hello!"
 is_extra = true
--- a/lib/src/config/config.rs
+++ b/lib/src/config/config.rs
@@ -20,6 +20,8 @@ pub struct Config {
     pub log_level: LoggingLevel,
     /// The environment that this configuration corresponds to.
     pub env: Environment,
+    /// Is rocket allowed to print emojis?
+    pub disable_emojis: bool,
     session_key: RwLock<Option<String>>,
     extras: HashMap<String, Value>,
     filepath: String,
@@ -53,6 +55,7 @@ impl Config {
                     session_key: RwLock::new(None),
                     extras: HashMap::new(),
                     env: env,
+                    disable_emojis: true,
                     filepath: filepath.to_string(),
                 }
             }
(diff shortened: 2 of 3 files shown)


RandomInsano:
I enjoy the emojis! Gives this guy some personality :D

I have no weight in this at all. Up to the maintainer.

SergioBenitez:
Thanks for the pull request.

I quite like the emojis! They're only on startup, so I can't imagine they're _too_ bothersome. Furthermore, this can be dealt with out of band with a tool like `sed`. The following will do exactly what your PR does:

```sh
cargo run | sed 's/[🚀👾🛰🔧]/=>/g'
```

LauriM:
That actually doesn't solve the original "problem" I was having. I don't want to see emojis on my fullscreen terminal windows. With this sed command I still haveto see them, but just one line above.

Yes, I could add them to my .zshrc file as alias, but then I have emojis on my config files!

I can look into making the patch more elegant, for example moving the emoji filtering into the logging system, making it less obscuring on the codebase.

RandomInsano:
What if there would be a way to just suppress the output altogether? I see more people benefitting from a quiet debug startup.

LauriM:
I could see benefit from that also. For the emojis, all the major applications that use them also provide a way to disable them. For example Brew provides environment variable to disable them.

As this feature is completely _optional_ and _not enabled by default_, I don't see a reason why the emojis could not be turned off.

reillysiemens:
FWIW, I **love** the emojis, but I would also appreciate a way to suppress them. I'm currently on an older system without the satellite emoji and would prefer to see nothing at all instead of the ugly tofu until I upgrade my system. I really do enjoy the emojis and I think they're a great personal touch. Just figured I'd share my experience for whatever it's worth. ¯\\\_(ツ)_/¯

theduke:
Show a serious ops person a server that puts emojis in their logs and they will scream at you.

I've been there.

This patch should really be merged.

flosse:
BTW: I really hate the emojis:
bildschirmfoto_2017-02-05_22-30-43

lukad:
I get these emojis look cool on OS X and I can probably live with broken emojis on linux, but I really don't want to have them in any production logs.

SergioBenitez:
For posterity, it should be known that the emojis don't appear under the `production` environment or when logging is set to `critical`. As a result, the emojis would not appear in production logs. The "serious enterprise
ops person" can rest at ease knowing that no emojis will infiltrate their logs.

lukad:
@SergioBenitez That's cool, didn't know that. The broken emojis on linux were just the first impression I got from running a Hello World.

flosse:
> the emojis don't appear under the `production` environment 

That is cool :) But here it seems not to work:

bildschirmfoto_2017-02-08_14-40-40

SergioBenitez:
@flosse The log setting needs to be set to "critical", which is the default log setting for production.

PR #85, opened 2016-12-30
add a command line parameter to test only files matching a certain string

rbalicki2:
I grew tired of waiting for all of the tests to finish when I was making my own test, so I added a flag to filter the tests by :)
* `./scripts/test.sh hello_w` will only build/bootstrap/run tests matching with `hello_w` in their full path
* `./scripts/test.sh` is unchanged.
* Right now, I just use the first parameter `$1` for the filter parameter, but I'm happy to make that `-f FILTER_PARAM` instead, if that future proofs this.
* Also, the `$FILTER_PARAM` expression is in two places. Instead, we should put the bootstrapping code inside of `build_and_test` so that we can only check in one spot whether the directory matches `$FILTER_PARAM`

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,10 @@ This will build and test the `core`, `codegen`, and `contrib` libraries as well
 as all of the examples. This is the script that gets run by Travis CI. To test a
 crate individually, run `cargo test --all-features`.
 
+You can also test only certain files by passing a parameter to the testing script.
+For example, `.scripts/test.sh hello_world` will run tests only in files that have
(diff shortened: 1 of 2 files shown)


SergioBenitez:
Do you find it too cumbersome to `cd` into the appropriate directory and `cargo test --all-features`? That's what I typically do, and it hasn't been too bad. I worry that if this is allowed, some might miss testing thing that are affected by their changes.

rbalicki2:
ah, didn't know that was a thing!

PR #87, opened 2016-12-31
Fix security issue with _method to block GET requests.

greglearns:
Security risk:

Currently, it is possible to use _method override to change a GET request into a POST/DELETE/PUT/etc. which is unsafe.
```
curl -H "Content-Type: application/x-www-form-urlencoded" -d "_method=DELETE" localhost:8000
```

Fix:

This PullRequest checks to ensure that only a POST is allowed when using _method override.

(This is also my first Rust PR, so, be gentle if I didn't go the best job :-)

### DIFF
--- a/lib/src/rocket.rs
+++ b/lib/src/rocket.rs
@@ -148,6 +148,7 @@ impl Rocket {
 
             let mut form_items = FormItems(form);
             if let Some(("_method", value)) = form_items.next() {
+                if req.method() != Method::Post { return; }
                 if let Ok(method) = value.parse() {
                     req.set_method(method);
                 }
--- a/lib/src/rocket.rs
+++ b/lib/src/rocket.rs
@@ -141,14 +141,13 @@ impl Rocket {
         // field which we use to reinterpret the request's method.
         let data_len = data.peek().len();
         let (min_len, max_len) = ("_method=get".len(), "_method=delete".len());
-        if req.content_type().is_form() && data_len >= min_len {
+        if req.method() == Method::Post && req.content_type().is_form() && data_len >= min_len {
             let form = unsafe {
                 from_utf8_unchecked(&data.peek()[..min(data_len, max_len)])
             };
 
             let mut form_items = FormItems(form);
             if let Some(("_method", value)) = form_items.next() {
-                if req.method() != Method::Post { return; }
                 if let Ok(method) = value.parse() {
                     req.set_method(method);
                 }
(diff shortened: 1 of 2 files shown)


lib/src/rocket.rs
```diff
@@ -148,6 +148,7 @@ impl Rocket {
 
             let mut form_items = FormItems(form);
             if let Some(("_method", value)) = form_items.next() {
+                if req.method() != Method::Post { return; }
```
SergioBenitez: This is a great catch! The correct change is to amend line 144 in rocket.rs so that it reads:

```rust
        if req.method() == Method::Post && req.content_type().is_form() && data_len >= min_len {
```

Don't change this line.

lib/tests/form_method-issue-45.rs
```diff
@@ -24,11 +24,24 @@ use rocket::http::ContentType;
...
         .header(ContentType::Form)
         .body("_method=patch&form_data=Form+data");
 
     let mut response = req.dispatch_with(&rocket);
     let body_str = response.body().and_then(|b| b.into_string());
     assert_eq!(body_str, Some("OK".to_string()));
 }
+
+#[test]
+fn method_eval_with_Get_should_fail() {
```
SergioBenitez: Name this `get_passes_through` or something like that.

lib/tests/form_method-issue-45.rs
```diff
@@ -24,11 +24,24 @@ use rocket::http::ContentType;
...
+    let rocket = rocket::ignite().mount("/", routes![bug]);
+
+    let mut req = MockRequest::new(Get, "/")
+        .header(ContentType::Form)
+        .body("_method=patch&form_data=Form+data");
+
+    let mut response = req.dispatch_with(&rocket);
+    let body_str = response.body().and_then(|b| b.into_string());
```
SergioBenitez: Don't look at the body. We should get a 404, so test for `response.status(), Status::NotFound`. That suffices.

lib/tests/form_method-issue-45.rs
greglearns: Should it be `get_passes_through` or `get_does_not_pass_through` ?

greglearns:
Suggestions incorporated. Thank you so much for providing the feedback!

SergioBenitez:
Merged inhttps://github.com/SergioBenitez/Rocket/commit/d19cb0349c33582d94870e3061e82c47f5846668. Thanks again!

PR #103, opened 2017-01-03
Form failure status

### LINKED ISSUE #46: HTTP 400 is returned when it is the incorrect response code.
If deserializing the response data into a struct using `FromForm` fails, Rocket by default will return `400 BAD REQUEST`, regardless of the reason that deserialization failed. This is the incorrect response code, unless deserialization failed due to malformed JSON. It is clearly defined in [RFC 2616](https://tools.ietf.org/html/rfc2616#page-65) as "The request could not be understood by the server due to malformed syntax. The client SHOULD NOT repeat the request without modifications.". For failures due to missing fields, unrecognized fields, fields of the wrong type, or pretty much any failure that isn't the result of the body being syntactically incorrect for the declared content-type, [`422 UNPROCESSABLE ENTITY`](https://tools.ietf.org/html/rfc4918#page-78) should be used instead.

gsquire:
Fixes #46 

I may have misunderstood the scope of the issue but I am happy to update it if needed.

### DIFF
--- a/examples/forms/src/tests.rs
+++ b/examples/forms/src/tests.rs
@@ -40,6 +40,6 @@ fn test_bad_login() {
 #[test]
 fn test_bad_form() {
     // Mess with the form formatting.
-    test_login("Sergio&other=blah&", "password", 0, Status::UnprocessableEntity, None);
-    test_login("&&&===&", "password", 0, Status::UnprocessableEntity, None);
+    test_login("Sergio&other=blah&", "password", 0, Status::BadRequest, None);
+    test_login("&&&===&", "password", 0, Status::BadRequest, None);
 }
--- a/examples/forms/src/tests.rs
+++ b/examples/forms/src/tests.rs
@@ -40,6 +40,6 @@ fn test_bad_login() {
 #[test]
 fn test_bad_form() {
     // Mess with the form formatting.
-    test_login("Sergio&other=blah&", "password", 0, Status::BadRequest, None);
-    test_login("&&&===&", "password", 0, Status::BadRequest, None);
+    test_login("Sergio&other=blah&", "password", 0, Status::UnprocessableEntity, None);
+    test_login("&&&===&", "password", 0, Status::UnprocessableEntity, None);
 }
--- a/lib/src/catcher.rs
+++ b/lib/src/catcher.rs
@@ -219,6 +219,8 @@ pub mod defaults {
                 teapot.", handle_418,
             421, "Misdirected Request", "The server cannot produce a response for this
                 request.", handle_421,
+            422, "Unprocessable Entity", "The server cannot process this request body.",
+                handle_422,
             426, "Upgrade Required", "Switching to the protocol in the Upgrade header
                 field is required.", handle_426,
             428, "Precondition Required", "The server requires the request to be
--- a/lib/src/catcher.rs
+++ b/lib/src/catcher.rs
@@ -219,8 +219,8 @@ pub mod defaults {
                 teapot.", handle_418,
             421, "Misdirected Request", "The server cannot produce a response for this
                 request.", handle_421,
-            422, "Unprocessable Entity", "The server cannot process this request body.",
-                handle_422,
+            422, "Unprocessable Entity", "The request was well-formed but was unable to be
+                followed due to semantic errors.", handle_422,
             426, "Upgrade Required", "Switching to the protocol in the Upgrade header
                 field is required.", handle_426,
             428, "Precondition Required", "The server requires the request to be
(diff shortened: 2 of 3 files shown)


SergioBenitez:
Just want to chime in and say I've not forgotten about you! Will have some time mid-next week to consider this thoroughly, and I'll get back to you then.

gsquire:
It's no problem! I know this was slated for 0.2.

SergioBenitez:
Alright! This is one of two remaining issues for 0.2!

So, looking at the [RFC, section 8.2.1](https://tools.ietf.org/html/rfc1866), we should return a `400` when any of the following occurs:

  1. A form string has no `=` at all or has no `=` after an `&`.
  2. The form string contains spaces.
  3. The form string contains unencoded URL characters.

The first one is pretty easy to detect but requires a change to the `FormItems` iterator. The second one means we have to check every string for spaces, which we can do efficiently if we change the `FormItems` decorator again. The last one...seems expensive, and I don't really know what the point of it is besides being able to send it in the URL.

Would be good to know what other frameworks do here. My intuition is that we should just check for 1 and ignore 2 and 3. This can be done by simply adding a method to the `FormItems` structure to see if there is any remaining, unparsed text. But even then: should Rocket just let this fly?

It's not clear to me what the correct approach is here.

gsquire:
Thanks for doing the research. I don't know if the original issue author would want to weigh in at all, but I can adjust this PR when you settle on the approach.

lib/src/catcher.rs
```diff
@@ -219,6 +219,8 @@ pub mod defaults {
                 teapot.", handle_418,
             421, "Misdirected Request", "The server cannot produce a response for this
                 request.", handle_421,
+            422, "Unprocessable Entity", "The server cannot process this request body.",
```
SergioBenitez: Change this to: "The request was well-formed but was unable to be followed due to semantic errors."

examples/forms/src/tests.rs
```diff
@@ -40,6 +40,6 @@ fn test_bad_login() {
 #[test]
 fn test_bad_form() {
     // Mess with the form formatting.
-    test_login("Sergio&other=blah&", "password", 0, Status::BadRequest, None);
-    test_login("&&&===&", "password", 0, Status::BadRequest, None);
+    test_login("Sergio&other=blah&", "password", 0, Status::UnprocessableEntity, None);
```
SergioBenitez: This should be a `BadRequest`.

examples/forms/src/tests.rs
```diff
@@ -40,6 +40,6 @@ fn test_bad_login() {
...
+    test_login("&&&===&", "password", 0, Status::UnprocessableEntity, None);
```
SergioBenitez: This should also be a `BadRequest`.

examples/forms/src/tests.rs
```diff
@@ -40,6 +40,6 @@ fn test_bad_login() {
 #[test]
```
SergioBenitez: We should add some tests that should be `UnprocessableEntity`s.

SergioBenitez:
Let's only worry about 1.

How do we go about implementing this? I see two possibilities:

  1. Change the `FromData` implementation for `Form` so that it validates the form string before passing it to the `FromForm` implementation and returns `BadRequest` on failed validation. This means that we iterate through the string twice when parsing forms.
  2. Change the `FromForm` trait so that it can inform the caller of a bad form string value somehow. Change that `FormItems` iterator so that it exposes a method to check if there is unparsed (i.e, invalid) input remaining. Change the `FromForm` derivation in codegen so that it calls this new method in `FormItems` and does the appropriate thing to inform the caller of a bad form string value. Finally, change the `Form` and `FromData` impl for `Form` to take this new value into account, returning the appropriate error.

I think **2** is the more "correct" approach, and it will be faster since we only go through the string once. But it's also a lot more complex. Maybe there's way to modify `FromForm` so that the changes are mostly backwards compatible? For instance, returning an `Option<Result<Self, Self::Error>>`, where `None` means that the form string was bad. Or something else, like changing the input type to be something mutable and inspecting it afterwards.

SergioBenitez:
I went with a form of 2. I'm going to take your changes and merge them in with mine. :)

examples/forms/src/tests.rs
SergioBenitez: This should be a BadRequest.

gsquire:
Ok, thanks for the help. I will fix this up tonight when I get home.

gsquire:
Are you going to add some tests for handling the 422 case? I guess I don't know how to do this without your changes :)

SergioBenitez:
I will indeed. In the process of handling this now. Thanks so much for pushing these so quickly!

SergioBenitez:
Merged in https://github.com/SergioBenitez/Rocket/commit/11e90f525f51c45eefb2acaef77414cfc24210b6. Thanks so much, again! :)

PR #112, opened 2017-01-05
Fix clippy lints within rocket_codegen

shssoichiro:
Some of these warnings were in code that rocket's macros would produce,
and so would also produce extraneous clippy warnings in crates using rocket's macros.


SergioBenitez:
Thanks so much for this! It's greatly appreciated. :D

Merged in 17bbd41f73a67584665a3f7e40f48e655b3037e6.

PR #115, opened 2017-01-06
add tests for examples/pastebin

ttdonovan:
Adding tests for examples/pastebin #62 

Any feedback is welcomed.


SergioBenitez:
I haven't had a chance to look over this yet, but I want to say thank you for taking this on! It's _very much_ so appreciated. :D

examples/pastebin/.gitignore
```diff
@@ -0,0 +1 @@
+/upload/
```
SergioBenitez: I want to keep this folder around so that user's don't have to create it manually.

examples/pastebin/src/main.rs
```diff
@@ -4,6 +4,8 @@
 extern crate rocket;
 extern crate rand;
 
+#[cfg(test)] mod tests;
+
```
SergioBenitez: We can kill this empty line.

examples/pastebin/src/tests/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+
+    let body_str = response.body().and_then(|body| body.into_string());
+    assert_eq!(body_str, Some(include_str!("usage.txt").to_string()));
+}
+
+#[test]
+fn upload() {
+    let rocket = rocket::ignite().mount("/", routes![super::upload]);
+    let mut req = MockRequest::new(Method::Post, "/")
+        .body(include_str!("file.txt").to_string());
```
SergioBenitez: You shouldn't need to call `to_string` here.

examples/pastebin/src/tests/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+    let mut response = req.dispatch_with(&rocket);
```
SergioBenitez: Add an empty line above and kill the empty one below.

examples/pastebin/src/tests/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+
+    assert_eq!(response.status(), Status::Ok);
+
+    match response.body().and_then(|body| body.into_string()) {
+        Some(url) => {
+            match url.trim().split("/").last() {
+                Some(paste_id) => {
+                    let filename = format!("upload/{}", paste_id);
+                    assert!(Path::new(&filename).exists());
```
SergioBenitez: This is assuming a particular implementation, which I don't think these tests should do. I think instead we should combine this with the next test and do a `POST` followed by a `GET`. Would be cool to do several of them, in fact, to make sure everything looks good.

examples/pastebin/src/tests/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+                },
+                _ => unreachable!(),
+            }
+        },
+        _ => unreachable!(),
+    }
+}
+
+#[test]
+fn retrieve_file_found() {
```
SergioBenitez: This is interesting because it's peering into the implementation. I think it'd be better to do a `POST` and then a `GET` instead, as mentioned above. What if we decide not to use files in the future, for instance? Or if the file structure changes?

examples/pastebin/src/tests/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+    assert_eq!(response.status(), Status::Ok);
+
+    let body_str = response.body().and_then(|body| body.into_string());
+    assert_eq!(body_str, Some("Hello from Rocket!".to_string()));
+}
+
+#[test]
+fn retrieve_file_not_found() {
+    let rocket = rocket::ignite().mount("/", routes![super::retrieve]);
+    let mut req = MockRequest::new(Method::Get, format!("/{}", "file-not-found"));
```
SergioBenitez: Why use `format` here?

ttdonovan:
@SergioBenitez I am out of town until next week. I briefly read over some
of your comments but will most likely not be able to dig into them anytime
soon. Feel free to alter or scrap my pull request however you feel fit.
Rust is fairly new language to me and coming from Ruby there is still a lot
I need to learn. My one observation or commit about testing the
implementation is - I'm not certain I like the idea of combining the POST
with the GET. Wouldn't there be double testing going on with also testing
the POST? Sometimes I find it hard to draw the line between integration and
unit tests. This example I feel like some of the internal implementation
knowledge is required. I would be interested in seeing an approach from
another Rust developer.

SergioBenitez:
Went ahead and implemented this in e6615af7e6aa1f71f7a6a69f4777a4530310272c. Thanks so much again for your contribution! :)

PR #121, opened 2017-01-09
Config port to u16

### LINKED ISSUE #119: Rocket configuration should accept u16 for a port rather than a usize
`rocket::config::Config.port` is currently of type `usize`. I'd like to propose that the configuration class for Rocket be changed to a `u16` to more accurately reflect desired type of the configuration parameter.

- [x] Update configuration type
- [x]  Update docs
- [x] Make sure the config parsing reports a nice warning when above `u16::max`
- [x] Update tests

drusellers:
Converts the Config struct's `port` field to a `u16` from a `usize` to better represent the correct field options to be used.

- Changed initial field type
    - corrected type errors
- updated docs to use "Unsigned 16 bit Integer"
- tests
    - added a test for the edge case bad value
    - added a test for the edge case good value
- generated a good error message `"an integer larger than 65535"`

closes #119

### DIFF
--- a/lib/src/config/config.rs
+++ b/lib/src/config/config.rs
@@ -120,7 +120,7 @@ impl Config {
             }
 
             if port > (u16::max_value() as i64) {
-                return Err(self.bad_type(name, val, "too large"))
+                return Err(self.bad_type(name, val, "an integer larger than 65535"))
             }
 
             self.port = port as u16;
--- a/lib/src/config/config.rs
+++ b/lib/src/config/config.rs
@@ -99,7 +99,7 @@ impl Config {
     /// returned:
     ///
     ///   * **address**: String
-    ///   * **port**: Integer
+    ///   * **port**: Integer (16-bit unsigned)
     ///   * **session_key**: String (192-bit base64)
     ///   * **log**: String
     ///
--- a/lib/src/config/config.rs
+++ b/lib/src/config/config.rs
@@ -99,7 +99,7 @@ impl Config {
     /// returned:
     ///
     ///   * **address**: String
-    ///   * **port**: Unsigned 16bit Integer
+    ///   * **port**: Integer
     ///   * **session_key**: String (192-bit base64)
     ///   * **log**: String
     ///
@@ -120,7 +120,7 @@ impl Config {
             }
 
             if port > (u16::max_value() as i64) {
-                return Err(self.bad_type(name, val, "an integer larger than 65535"))
+                return Err(self.bad_type(name, val, "a 16-bit unsigned integer"))
             }
 
             self.port = port as u16;
--- a/lib/src/config/config.rs
+++ b/lib/src/config/config.rs
@@ -15,7 +15,7 @@ pub struct Config {
     /// The address to serve on.
     pub address: String,
     /// The port to serve on.
-    pub port: usize,
+    pub port: u16,
     /// How much information to log.
     pub log_level: LoggingLevel,
     /// The environment that this configuration corresponds to.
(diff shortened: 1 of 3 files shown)


lib/src/config/config.rs
```diff
@@ -119,7 +119,11 @@ impl Config {
                 return Err(self.bad_type(name, val, "an unsigned integer"));
             }
 
-            self.port = port as usize;
+            if port > (u16::max_value() as i64) {
```
drusellers: I'm not pleased with how this turned out and blame my newness to rust.

lib/src/config/config.rs
```diff
@@ -99,7 +99,7 @@ impl Config {
     /// returned:
     ///
     ///   * **address**: String
-    ///   * **port**: Integer
+    ///   * **port**: Unsigned 16bit Integer
```
SergioBenitez: This should stay as `Integer` as it's referring to the `Value::Integer` variant.

lib/src/config/config.rs
mgattozzi: I think you should just be able to do 

```
if !(port <= u16::max_value())
```

for logic equivalence. That way you don't need to cast to an `i64`. Also with the declaration of `port` if you want it as a `u16` do

```
let port: u16 = parse!(self, name, val, as_integer, "an integer")?;
```

Which should allow you to do:

```
self.port = port;
```

lib/src/config/config.rs
```diff
@@ -119,7 +119,11 @@ impl Config {
                 return Err(self.bad_type(name, val, "an unsigned integer"));
             }
 
-            self.port = port as usize;
+            if port > (u16::max_value() as i64) {
+                return Err(self.bad_type(name, val, "an integer larger than 65535"))
```
SergioBenitez: The string should say what was expected, not what was parsed. In this case, it should say `a 16-bit unsigned integer`.

lib/src/config/mod.rs
```diff
@@ -33,7 +33,7 @@
 //!   * **address**: _[string]_ an IP address or host the application will
 //!     listen on
 //!     * examples: `"localhost"`, `"0.0.0.0"`, `"1.2.3.4"`
-//!   * **port**: _[integer]_ a port number to listen on
+//!   * **port**: _[u16]_ a port number to listen on
```
SergioBenitez: This should also say `integer` as it's referring to the TOML value, which doesn't know about `u16`. The description should say `a 16-bit unsigned integer` somewhere. Or maybe nothing. Probably better to just leave it alone.

lib/src/config/mod.rs
```diff
@@ -589,6 +596,11 @@ mod test {
             [staging]
             port = -1
         "#.to_string(), TEST_CONFIG_FILENAME).is_err());
+
+        assert!(RocketConfig::parse(r#"
```
SergioBenitez: Let's add another test for an even higher port number, just for posterity.

lib/src/config/config.rs
drusellers: Ok, I was curious about this one.

lib/src/config/config.rs
drusellers: excellent, will adjust

lib/src/config/mod.rs
drusellers: Ok, I'll revert that back to integer

lib/src/config/mod.rs
drusellers: boosh

lib/src/config/config.rs
drusellers: > Also with the declaration of port if you want it as a u16 do

But then I can't add a test for ever u16::max, and would instead rely on the parse error which I'm thinking is worse?? Also, I couldn't test for the `<0` and would have to rely on the parse error. Hmm, I'm seeing what you are saying though.

SergioBenitez: @mgattozzi If you do `let port: u16`, you'll truncate the value. This means it can never exceed `u16::max`, and values above `u16::max` will be accept unexpectedly.

mgattozzi: @SergioBenitez now that you mention it I totally see the problem. Kind of makes this unavoidable then.

SergioBenitez:
Merged in https://github.com/SergioBenitez/Rocket/commit/a6084ab3e2ef9b5d26188ee01816316e0805c938.

PR #123, opened 2017-01-10
Fix downstream clippy warnings

### LINKED ISSUE #117: Warning for each route annotation from clippy
[Clippy](https://github.com/Manishearth/rust-clippy) gives the following warning at each route annotation:
this expression borrows a reference that is immediately dereferenced by the compiler

It would be nice if this was fixed, it makes output from clippy that I can actually fix harder to find and it's always nice to get empty output from clippy.

shssoichiro:
Fixes #117


SergioBenitez:
I hope we got them all! Thanks again!

Merged in 59043e262a6b0f63b984bc224e152100234e3ec9.

PR #126, opened 2017-01-10
Support FromParam on Uuid

lholden:
Simple patch to alllow FromParam on the Uuid type. :)


drusellers:
I love that this is hidden behind a feature. uuid does this with a feature toggle for serde for instance and postgres crate has a feature for uuid as well called `with-uuid` - so I think it makes sense to have something similar. https://github.com/sfackler/rust-postgres/blob/master/postgres/Cargo.toml#L30

lholden:
Moving to a version of this for contrib instead.

PR #127, opened 2017-01-10
Fix a typo in the documentation.

Eijebong:


lib/src/response/redirect.rs
```diff
@@ -44,7 +44,7 @@ impl Redirect {
 
     /// Construct a "permanent" (308) redirect response. This redirect must only
     /// be used for permanent redirects as it is cached by clients. This
-    /// response instructs the client to reissue requests tot he current URL to
+    /// response instructs the client to reissue requests to the current URL to
```
SergioBenitez: I think this would read better if it said, "for the current" instead of "to the current".

Eijebong: Agreed and fixed

SergioBenitez:
Thanks so much! Merged in 1398626710858e1de370d39e57431721ed36d989.

PR #128, opened 2017-01-10
Add UUID type

lholden:
Provide a UUID type for contrib.


contrib/src/lib.rs
```diff
@@ -43,8 +45,14 @@ mod json;
 #[cfg(feature = "templates")]
 mod templates;
 
+#[cfg(any(test, feature = "uuid"))]
```
SergioBenitez: This shouldn't be enabled on `test`.

contrib/src/lib.rs
```diff
@@ -43,8 +45,14 @@ mod json;
...
+mod uuid;
+
 #[cfg(feature = "json")]
 pub use json::JSON;
 
 #[cfg(feature = "templates")]
 pub use templates::Template;
+
+#[cfg(any(test, feature = "uuid"))]
```
SergioBenitez: Same with the `test` `cfg` here.

contrib/src/uuid.rs
```diff
@@ -0,0 +1,117 @@
+extern crate uuid as uuid_ext;
+
+use std::fmt;
+use std::str::FromStr;
+use std::ops::{Deref, DerefMut};
+use rocket::request::FromParam;
+
+
+/// The UUID type, which implements `FromParam`. This type allows you to accept
+/// values from the [Uuid](https://github.com/rust-lang-nursery/uuid) crate in
```
SergioBenitez: maybe say "...as a dynamic parameter in your request handlers".

contrib/src/uuid.rs
```diff
@@ -0,0 +1,117 @@
...
+use std::str::FromStr;
+use std::ops::{Deref, DerefMut};
+use rocket::request::FromParam;
+
+
+/// The UUID type, which implements `FromParam`. This type allows you to accept
+/// values from the [Uuid](https://github.com/rust-lang-nursery/uuid) crate in 
+/// your request handlers.
+///
+/// Make sure to enable the `uuid` feature in your Cargo.toml to use this type.
```
SergioBenitez: Provide the TOML to do so, as with JSON. Also use the wording that the JSON/Template documentation uses. Something like, "The `uuid` feature must be enabled to...".

contrib/src/uuid.rs
```diff
@@ -0,0 +1,117 @@
...
+/// 
+/// ```rust,ignore
+/// // A function expecting the Uuid type from the uuid crate.
```
SergioBenitez: This is confusing. I think it suffices to have the route below.

contrib/src/uuid.rs
```diff
@@ -0,0 +1,117 @@
...
+///
+/// #[get("/users/<id>")]
+/// fn user(id: UUID) -> String {
+///     // Use Deref to access the underlying Uuid type.
+///     expects_uuid(*id);
+///
+///     format!("We found: {}", id)
+/// }
+/// ```
+
```
SergioBenitez: No empty line here.

contrib/src/uuid.rs
```diff
@@ -0,0 +1,117 @@
...
+#[derive(Debug)]
+pub struct UUID(uuid_ext::Uuid);
+
+impl UUID {
+    /// Consumes the UUID wrapper returning the underlying Uuid type.
```
SergioBenitez: Provide an example.

contrib/src/uuid.rs
```diff
@@ -0,0 +1,117 @@
...
+    }
+}
+
+impl fmt::Display for UUID {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+       self.0.fmt(f)
+    }
+}
+
+impl<'a> FromParam<'a> for UUID {
```
SergioBenitez: Document what happens. When does this return `Ok`, when does it return `Err`?

contrib/src/uuid.rs
```diff
@@ -0,0 +1,117 @@
...
+    type Error = uuid_ext::ParseError;
+    fn from_param(p: &'a str) -> Result<UUID, Self::Error> {
```
SergioBenitez: Empty line above here.

contrib/src/uuid.rs
```diff
@@ -0,0 +1,117 @@
...
+        p.parse()
+    }
+}
+
+impl FromStr for UUID {
+    type Err = uuid_ext::ParseError;
+    fn from_str(s: &str) -> Result<UUID, Self::Err> {
```
SergioBenitez: Empty line above here.

contrib/src/uuid.rs
```diff
@@ -0,0 +1,117 @@
...
+    use super::uuid_ext;
+    use super::FromParam;
+    use super::UUID;
+
+    #[test]
+    fn test_from_str() {
+        let uuid_str = "c1aa1e3b-9614-4895-9ebd-705255fa5bc2";
+        let uuid_result = uuid_str.parse();
+        assert!(uuid_result.is_ok());
+        let uuid:UUID = uuid_result.unwrap();
```
SergioBenitez: Need space after colon.

contrib/src/uuid.rs
```diff
@@ -0,0 +1,117 @@
...
+        assert!(uuid_str.to_string() == uuid.to_string())
```
SergioBenitez: Why can't we compare directly? Does the UUID crate's Uuid type implement PartialEq for str? If so, this should work here too.

contrib/src/uuid.rs
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    #[test]
+    fn test_from_param() {
+        let uuid_str = "c1aa1e3b-9614-4895-9ebd-705255fa5bc2";
+        let uuid_result = UUID::from_param(uuid_str);
+        assert!(uuid_result.is_ok());
```
SergioBenitez: We can just unwrap above to get the same check.

contrib/src/uuid.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+impl fmt::Display for UUID {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+       self.0.fmt(f)
+    }
+}
+
+impl<'a> FromParam<'a> for UUID {
+    type Error = uuid_ext::ParseError;
```
SergioBenitez: If the type is public, it must be reexported.

contrib/src/uuid.rs
```diff
@@ -26,32 +33,44 @@ use rocket::request::FromParam;
...
-#[derive(Debug)]
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Debug)]
 pub struct UUID(uuid_ext::Uuid);
 
 impl UUID {
     /// Consumes the UUID wrapper returning the underlying Uuid type.
-    pub fn unwrap(self) -> uuid_ext::Uuid {
+    ///
+    /// # Example
+    /// ```rust,ignore
```
SergioBenitez: This shouldn't be ignore, right? Let's get this tested.

contrib/src/uuid.rs
```diff
@@ -26,32 +33,44 @@ use rocket::request::FromParam;
...
-
-#[derive(Debug)]
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Debug)]
 pub struct UUID(uuid_ext::Uuid);
 
 impl UUID {
     /// Consumes the UUID wrapper returning the underlying Uuid type.
-    pub fn unwrap(self) -> uuid_ext::Uuid {
+    ///
+    /// # Example
```
SergioBenitez: Empty line after here, just for consistency.

contrib/src/uuid.rs
```diff
@@ -26,32 +33,44 @@ use rocket::request::FromParam;
...
+    /// ```rust,ignore
+    /// # extern crate uuid;
+    /// # use rocket_contrib::UUID;
+    /// # use std::str::FromStr;
+    ///
+    /// let uuid_str = "c1aa1e3b-9614-4895-9ebd-705255fa5bc2";
+    /// let my_inner_uuid = UUID::from_str(uuid_str).unwrap().into_inner();
+    /// assert_eq!(uuid::Uuid::from_str(uuid_str), my_inner_uuid)
+    /// ```
+    pub fn into_inner(self) -> uuid_ext::Uuid {
```
SergioBenitez: Make this `#[inline(always)]`

contrib/src/uuid.rs
```diff
@@ -26,32 +33,44 @@ use rocket::request::FromParam;
...
         self.0
     }
 }
 
 impl fmt::Display for UUID {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
```
SergioBenitez: Make this `#[inline(always)]`

contrib/src/uuid.rs
```diff
@@ -26,32 +33,44 @@ use rocket::request::FromParam;
...
-       self.0.fmt(f)
+        self.0.fmt(f)
     }
 }
 
 impl<'a> FromParam<'a> for UUID {
     type Error = uuid_ext::ParseError;
+
     fn from_param(p: &'a str) -> Result<UUID, Self::Error> {
```
SergioBenitez: Make this `#[inline(always)]`

contrib/src/lib.rs
```diff
@@ -6,13 +6,15 @@
...
-//! that implement the `FromRequest` trait, `Responder` trait, or both.
+//! implementing a combination of the `FromRequest`, `FromParam`, and 
+//! `Responder` traits.
 //!
 //! Each module in this library is held behind a feature flag, with the most
 //! common modules exposed by default. The present feature list is below, with
 //! an asterisk next to the features that are enabled by default:
 //!
 //! * [json*](struct.JSON.html)
+//! * [uuid](struct.UUID.html)
```
SergioBenitez: Let move this to the bottom of the list. I guess that's arbitrary, but I want to keep them in order of addition.

contrib/src/uuid.rs
```diff
@@ -0,0 +1,133 @@
...
+/// ```
+///
+/// The UUID type implements the Deref trait allowing you to access the
+/// underlying Uuid type.
+///
+/// ```rust,ignore
+/// #[get("/users/<id>")]
+/// fn user(id: UUID) -> String {
+///     // Use Deref to access the underlying Uuid type.
+///     expects_uuid(*id);
```
SergioBenitez: Sorry, I was totally vague last time. I don't think we need to mention that it implements `Deref` here. The docs take care of that. So can just keep the `format!` line here.

contrib/src/uuid.rs
```diff
@@ -0,0 +1,133 @@
...
+/// of your `Cargo.toml`:
+///
+/// ```toml,ignore
+/// [dependencies.rocket_contrib]
+/// version = "*"
+/// default-features = false
+/// features = ["uuid"]
+/// ```
+///
+/// The UUID type implements the Deref trait allowing you to access the
```
SergioBenitez: Instead of saying this, say that the fact that it implements `FromParam` means you can use it directly as a target of a dynamic parameter. And the example is clear.

contrib/src/uuid.rs
```diff
@@ -0,0 +1,133 @@
...
+/// }
+/// ```
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Debug)]
+pub struct UUID(uuid_ext::Uuid);
+
+impl UUID {
+    /// Consumes the UUID wrapper returning the underlying Uuid type.
+    ///
+    /// # Example
+    /// ```rust,ignore
```
SergioBenitez: We should be able to doctest this, so don't ignore.

contrib/src/uuid.rs
```diff
@@ -0,0 +1,133 @@
...
+    /// let my_inner_uuid = UUID::from_str(uuid_str).unwrap().into_inner();
+    /// assert_eq!(uuid::Uuid::from_str(uuid_str), my_inner_uuid)
+    /// ```
+    pub fn into_inner(self) -> uuid_ext::Uuid {
+        self.0
+    }
+}
+
+impl fmt::Display for UUID {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
```
SergioBenitez: Let's `#[inline]` this.

contrib/src/uuid.rs
```diff
@@ -0,0 +1,133 @@
...
+impl Deref for UUID {
+    type Target = uuid_ext::Uuid;
+
+    fn deref<'a>(&'a self) -> &'a Self::Target {
+        &self.0
+    }
+}
+
+impl DerefMut for UUID {
+    fn deref_mut<'a>(&'a mut self) -> &'a mut uuid_ext::Uuid {
```
SergioBenitez: Does `Uuid` have any mut methods? Doesn't really matter, but just wondering.

contrib/src/uuid.rs
```diff
@@ -0,0 +1,133 @@
...
+
+#[cfg(test)]
+mod test {
+    use super::uuid_ext;
+    use super::FromParam;
+    use super::UUID;
+
+    #[test]
+    fn test_from_str() {
+        use std::str::FromStr;
```
SergioBenitez: Let's move this `use` to the top of the module.

contrib/src/uuid.rs
lholden: I was originally going to let this be a doctest... but for whatever reason I can't make this actually use the uuid crate.

contrib/src/uuid.rs
lholden: Only one that I can see. `fn clone_from(&mut self, source: &Self)` It would be pretty silly to deref_mut and then clone_from though... :).

contrib/src/lib.rs
lholden: You don't think order by type of thing is better? It's a little weird to have them split up, but I can certainly do so.

contrib/src/uuid.rs
SergioBenitez: Have you tried using `extern crate uuid` inside the doctest? That _should_ work.

lholden: I did...  It ends up somehow breaking things and it suddenly stops being able to find rocket_contrib and etc. Feel free to give it a try, but I wasn't able to make it work.

SergioBenitez: This works:

```rust
    /// # Example
    /// ```rust
    /// # extern crate rocket_contrib;
    /// # extern crate uuid;
    /// # use rocket_contrib::UUID;
    /// # use std::str::FromStr;
    /// # use uuid::Uuid;
    /// # fn main() {
    /// let uuid_str = "c1aa1e3b-9614-4895-9ebd-705255fa5bc2";
    /// let real_uuid = Uuid::from_str(uuid_str).unwrap();
    /// let my_inner_uuid = UUID::from_str(uuid_str).unwrap().into_inner();
    /// assert_eq!(real_uuid, my_inner_uuid);
    /// # }
    /// ```
```

SergioBenitez:
Merged in 8f39d3399eb35d0e033a398617fe405b099aa8c9.  🎉

This is the first contrib contrib! :D Congrats!

lholden:
*cheer* :)

PR #130, opened 2017-01-11
Add clippy as an optional dependency

shssoichiro:
This is a PR to add clippy as an optional dependency which can be enabled via `cargo build --features clippy`. The reason for this PR is that adding clippy as a dependency appears to scan through more of the code than `cargo clippy` (which may be a bug in `cargo clippy`).

This PR does not fix any lints, just adds clippy as a dependency to all three crates within Rocket.

### DIFF
--- a/codegen/src/lib.rs
+++ b/codegen/src/lib.rs
@@ -95,6 +95,9 @@
 #![feature(i128_type)]
 #![allow(unused_attributes)]
 #![allow(deprecated)]
+#![cfg_attr(feature="clippy", feature(plugin))]
+#![cfg_attr(feature="clippy", plugin(clippy))]
+#![cfg_attr(feature="clippy", allow(needless_lifetimes))]
 
 #[macro_use] extern crate syntax;
 #[macro_use] extern crate log;
--- a/codegen/Cargo.toml
+++ b/codegen/Cargo.toml
@@ -18,6 +18,10 @@ plugin = true
 rocket = { version = "0.1.4", path = "../lib/" }
 log = "^0.3"
 
+[dependencies.clippy]
+version = ">=0.0.106"
+optional = true
+
 [dev-dependencies]
 compiletest_rs = "^0.2"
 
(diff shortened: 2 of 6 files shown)


codegen/src/lib.rs
```diff
@@ -95,6 +95,9 @@
 #![feature(i128_type)]
 #![allow(unused_attributes)]
 #![allow(deprecated)]
+#![cfg_attr(feature="clippy", feature(plugin))]
+#![cfg_attr(feature="clippy", plugin(clippy))]
+#![cfg_attr(feature="clippy", allow(needless_lifetimes))]
```
shssoichiro: The reason for allowing this lint is because it was producing many false positives on Rocket's code base (tracked by https://github.com/Manishearth/rust-clippy/issues/740)

SergioBenitez:
What's the benefit to doing this? Presumably it's to get clippy warnings on compilation. But what's wrong with simply running `cargo clippy`?

The amount of noise clippy generates on the Rocket codebase is phenomenal. I don't thnk that's something we want to see on each compilation.

SergioBenitez:
It also seems to interact poorly with `compiletest-rs`, resulting in failed compilation.

shssoichiro:
The benefit is that integrating the clippy library scans all of the modules, whereas `cargo clippy` does not seem to. However, if it's breaking the test suite, that would be a blocker. The better route may be for me to try and fix `cargo clippy` to scan properly (or at least file an issue on `clippy` for this). Thanks for the feedback.

jhpratt:
@SergioBenitez Is this something you'd reconsider two and a half years on? Running `cargo clippy` yields a few warnings and a pair of errors for identical blocks in `if` statements. The only noise I'm seeing is for redundant pattern matching on the `switch` macro from the pear crate, which I see is yours (so I'm assuming that warning can be disabled).

SergioBenitez:
@jhpratt I think it'd be nice to throw this into the CI as a warning, though I still don't think we should modify the code base to support clippy specifically.

jebrosen:
The main difficulty with clippy is that it is not necessarily available with every nightly release (`rustup component add clippy` will sometimes fail). I see no harm in having the CI tests run clippy when it is installed, as long as it's not too slow.

jhpratt:
Clippy itself runs at a decent speed. It's the installation that would take a minute. Jeb is right, though, not every nightly has it available, and Rust doesn't provide an easy way to install the most recently available.

PR #138, opened 2017-01-16
Example for the UUID type

lholden:
This is an example for the UUID type. :)

### DIFF
--- a/contrib/src/uuid.rs
+++ b/contrib/src/uuid.rs
@@ -3,6 +3,7 @@ extern crate uuid as uuid_ext;
 use std::fmt;
 use std::str::FromStr;
 use std::ops::Deref;
+use std::cmp::Ordering;
 
 use rocket::request::FromParam;
 
@@ -97,6 +98,13 @@ impl PartialEq<uuid_ext::Uuid> for UUID {
     #[inline(always)]
     fn eq(&self, other: &uuid_ext::Uuid) -> bool {
         self.0.eq(other)
+    }
+}
+
+impl PartialOrd<uuid_ext::Uuid> for UUID {
+    #[inline(always)]
+    fn partial_cmp(&self, other: &uuid_ext::Uuid) -> Option<Ordering> {
+        Some(self.0.cmp(other))
     }
 }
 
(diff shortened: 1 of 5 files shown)


examples/uuid/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+}
+
+#[get("/people/<id>")]
+fn people(id: UUID) -> Result<String, String> {
+    // Because UUID implements the Deref trait, we use Deref coercion to 
+    // convert rocket_contrib::UUID to uuid::Uuid.
+    let person = PEOPLE.get(&id)
+        .ok_or(format!("Person not found for UUID: {}", id))?;
+
+    Ok(format!("We found: {}", person))
```
SergioBenitez: I think we can roll this into the `PEOPLE.get`call with a `map` before the `ok_or`.

examples/uuid/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+#![plugin(rocket_codegen)]
+
+#[macro_use]
+extern crate lazy_static;
+extern crate uuid;
+extern crate rocket;
+extern crate rocket_contrib;
+
+use std::collections::HashMap;
+use uuid::Uuid;
```
SergioBenitez: Why do we need this?

examples/uuid/src/main.rs
lholden: You want this to look something like this?
```rust
Ok(PEOPLE.get(&id)
    .map(|person| format!("We found: {}", person))
    .ok_or(format!("Person not found for UUID: {}", id))?)
```

examples/uuid/src/main.rs
lholden: Common use case is going to be inter-op with the uuid crate. Ex, it's what the Postgres and Diesel crates use for uuid support.

So, it makes sense to actually show how one does this. (Ex, the HashMap is of the uuid::Uuid type and not UUID)

contrib/src/uuid.rs
```diff
@@ -100,6 +101,13 @@ impl PartialEq<uuid_ext::Uuid> for UUID {
     }
 }
 
+impl PartialOrd<uuid_ext::Uuid> for UUID {
```
SergioBenitez: This should be in a separate PR.

examples/uuid/src/main.rs
SergioBenitez: Okay, cool, that makes sense.

SergioBenitez:
Merged in f230d43fd5a7836e263de9977e20fd715ac0502a.

Issue #25, opened 2016-12-23
Lack of handling for CORS/OPTIONS


LLBlumire:
In order for a server to provide an external API it needs to be able to deal with Cross Origin Resource Sharing. These are done by web browsers sending a preflight OPTIONS request asking what resources it is allowed to access on the server, then after being given a resonse containing certain information in it's header a browser will send the actual GET/PUT/POST/etc query to the server, with no response it will not do this and cause Cross Site Scripting errors to be reported. 

There are two obvious ways this can be handled, the first and simplest is to add an "OPTIONS" annotation of the same ilk as the existing HTTP Methods, this will allow users of Rocket to manually implement the correct preflight request handling for api endpoints that need to provide CORS.

The other is to provide some form of automatic CORS handling, this can be seen in a number of other libraries used for web server development (outside the realm of rust) such as spring.io (http://docs.spring.io/spring/docs/current/spring-framework-reference/html/cors.html) or flasks (https://pypi.python.org/pypi/Flask-Cors).

fabricedesre:
Having a "native" built in support would be better. We did something for iron, which was quite simple (https://github.com/fxbox/iron-cors/).

kybishop:
More actionable and human-readable info than the w3 spec: http://enable-cors.org/server.html

SergioBenitez:
e74f5ce8648d393f64ecf247aef6fbfc5746c3eb implements the ability to have an `options` decorator, so you can now implement this manually via an `options` route and a `CORS` `Responder`. I agree that Rocket should have a nicer to handle this, and it's on my list of things to determine.

flosse:
I'd love to see a full CORS example :) Did someone already made experiences?

sebasmagri:
This is what I'm currently doing to handle the preflight request:

```
#[route(OPTIONS, "/endpoint/")]
fn options_handler<'a>() -> Response<'a> {
    Response::build()
        .raw_header("Access-Control-Allow-Origin", "http://host.tld")
        .raw_header("Access-Control-Allow-Methods", "OPTIONS, POST")
        .raw_header("Access-Control-Allow-Headers", "Content-Type")
        .finalize()
}
```
However, for a handler returning data, I can't still find a way to wrap the existing response in a `Response` to add the headers. The existing handler looks like this:

```
#[post("/endpoint/", format = "application/json", data = "<config>")]
fn post_handler(config: JSON<Config>) -> Option<JSON<Wrapper>> {
[...]

PR #141, opened 2017-01-19
Very simple CORS implementation

impowski:
Due to this issue #25, I rewrote what @SergioBenitez posted in this [post](https://github.com/SergioBenitez/Rocket/issues/25#issuecomment-271065434). There is still no middleware-like functionality and this is a very basic implementation.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,7 @@ members = [
   "codegen/",
   "contrib/",
   "examples/cookies",
+  "examples/cors",
   "examples/errors",
   "examples/extended_validation",
   "examples/forms",
--- a/contrib/Cargo.toml
+++ b/contrib/Cargo.toml
@@ -12,6 +12,7 @@ license = "MIT/Apache-2.0"
 
 [features]
 default = ["json"]
+cors = []
 json = ["serde", "serde_json"]
 tera_templates = ["tera", "templates"]
 handlebars_templates = ["handlebars", "templates"]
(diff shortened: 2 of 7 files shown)


mehcode:
I feel I must comment on how nested these Responders are getting. Does Rocket not have a way to add middleware to a group of routes?

Something like CORS is something you likely want to enable or disable with similar options for the vast majority of your routes and having to add `CORS` everywhere is a bit intense. 

I'd vote to work on middleware support in Rocket before adding this.

impowski:
@mehcode Yeah, I had a discussion about middlewares with @SergioBenitez on IRC, he said that it will be available soon, but I guess we must take on it and help him, at least he could mentor us a little so it would be idiomatic with Rocket and how he is seeing it.

SergioBenitez:
@impowski Thank you for packaging up the `CORS` implementation in #25 into a nice PR!

Unfortunately, I don't feel that this is a good fit for `contrib` for exactly the reason discussed: middleware (fairings) provide a much better solution here. Let's instead work to find a great design for fairings instead of encouraging a suboptimal one. As a stop-gap measure, feel free to create a `rocket-cors` crate with this functionality.

Closing this PR for the above reasons.

Enether:
Please do create a `rocket-cors` crate for a way to get around the missing functionality for now

onelson:
@Enether I think I'm about to drop this into a toy project -- I'll see about publishing a crate if @impowski doesn't beat us to it. Seems like a decent workaround for the time being, and it's essential to making my app (and yours?) work.

onelson:
Just got a drive-by PR on it from @Enether, so I guess the cat is out of the bag. https://github.com/onelson/rocket_cors is a thing, at least for now. There are some typos in the readme, etc, but I'll aim to get a couple tests/examples added before pushing to crates.io. Perhaps @impowski will want to take it off my hands and publish it himself? That'd be fine by me.

PR #165, opened 2017-01-31
Default T in JSON to improve ergonomics of returning arbitrary JSON

mehcode:
This ...

```rust

use rocket_contrib::json::{JSON, Value};

type SimpleMap = HashMap<&'static str, Value>;

#[derive(Debug, Deserialize)]
struct UserForm {
    username: String,
    email: String,
    password: String,
}

#[post("/user", data = "<body>")]
fn post_user(body: JSON<UserForm>) -> JSON<SimpleMap> {
    println!("{:?}", body);

    JSON(map!(
        "status" => "ok",
    ))
}
```

Becomes ...

```rust
use rocket_contrib::json::JSON;

#[derive(Debug, Deserialize)]
struct UserForm {
    username: String,
    email: String,
    password: String,
}

#[post("/user", data = "<body>")]
fn post_user(body: JSON<UserForm>) -> JSON {
    println!("{:?}", body);

    JSON(map!(
        "status" => "ok",
    ))
}
```

Thoughts?

### DIFF
--- a/contrib/src/json/mod.rs
+++ b/contrib/src/json/mod.rs
@@ -129,20 +129,23 @@ impl<T> DerefMut for JSON<T> {
 /// ```
 ///
 /// The return type of a macro invocation is
-/// [Value](/rocket_contrib/enum.Value.html). A value created with this macro
+/// [Value](/rocket_contrib/enum.Value.html), and is the default type parameter 
+/// of the JSON type. A value created with this macro
 /// can be returned from a handler as follows:
 ///
 /// ```rust,ignore
-/// use rocket_contrib::{JSON, Value};
+/// use rocket_contrib::JSON;
 ///
 /// #[get("/json")]
-/// fn get_json() -> JSON<Value> {
+/// fn get_json() -> JSON {
 ///     JSON(json!({
 ///         "key": "value",
 ///         "array": [1, 2, 3, 4]
 ///     }))
 /// }
 /// ```
+/// 
+/// 
 ///
 /// # Examples
 ///
--- a/contrib/src/json/mod.rs
+++ b/contrib/src/json/mod.rs
@@ -3,6 +3,7 @@ extern crate serde_json;
 
 use std::ops::{Deref, DerefMut};
 use std::io::Read;
+use std::collections::HashMap;
 
 use rocket::outcome::Outcome;
 use rocket::request::Request;
@@ -53,7 +54,7 @@ pub use self::serde_json::to_value;
 /// ```
 ///
 #[derive(Debug)]
-pub struct JSON<T>(pub T);
+pub struct JSON<T = HashMap<&'static str, Value>>(pub T);
 
 impl<T> JSON<T> {
     /// Consumes the JSON wrapper and returns the wrapped item.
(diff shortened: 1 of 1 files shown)


impowski:
Neat.

SergioBenitez:
I actually just got rid of the `map!` macro in favor of `json!` in https://github.com/SergioBenitez/Rocket/commit/5fabb43a1b62e00c1fdeecd6ed8e5d057be83ff1. Now you simply need to return `JSON<Value>`. Do you think this change is still worth it?

mehcode:
Yes. I still think so. It's not as big as improvement as before but it still reads like an improvement.

---

Hmm.. I like how it nicely supports arrays now. But `JSON(json!( ... ))` reads a bit funny. I was thinking (before now) of adding a `json!` macro that expands to `JSON(map!( ... ))` but I can see why you'd want to keep the consistency of the responder wrapper.

Perhaps it can change again to `o!(...)` (or something) that is content-agnostic so you could do `MsgPack(o!(...))` for instance?

SergioBenitez:
That means we'd have to rewrite the macro. Currently, we just use the implementation from `serde` directly. It's really nice not to have to maintain it.

We could make `json!` expand to a `JSON<Value>` directly, however.

mehcode:
> That means we'd have to rewrite the macro. Currently, we just use the implementation from serde directly. It's really nice not to have to maintain it.

Why? Sorry if I wasn't clear. I just meant rename it. Because it just returns a thing that is deserializable shouldn't in theory work anywhere you work arbitrarily with deserializables? Or do I not understand how this is done internally?

SergioBenitez:
> Because it just returns a thing that is deserializable [...]

It's returns a thing that's `Serialize`, not `Deserialize`. Anyway, I see what you're saying. Let's revisit this when there's more than one serialization format in contrib.

mehcode:
> It's returns a thing that's Serialize, not Deserialize. Anyway, I see what you're saying. Let's revisit this when there's more than one serialization format in contrib.

Heh. Yeah I meant `Serialize`.

---

@SergioBenitez Fixed this branch.

SergioBenitez:
Why isn't Travis running this? :\

SergioBenitez:
Alright, I've come around on this.

Can you modify the docs and example to inform users about the default parameter? In particular, the `json!` macro code example should just return `JSON` and mention why that's okay.

contrib/src/json/mod.rs
```diff
@@ -129,20 +129,23 @@ impl<T> DerefMut for JSON<T> {
 /// ```
 ///
 /// The return type of a macro invocation is
-/// [Value](/rocket_contrib/enum.Value.html). A value created with this macro
+/// [Value](/rocket_contrib/enum.Value.html), and is the default type parameter 
+/// of the JSON type. A value created with this macro
```
SergioBenitez: This needs to be word-wrapped at 80 columns.

SergioBenitez: Also, I'd like a slightly longer explanation. Kill the `,` and add another sentence explaining that it's the default value and as such you can return `JSON` without specifying the type.

SergioBenitez:
Merged in 5086f0eb01bcd2a789993ab0444dbd14ede272d0.

PR #168, opened 2017-02-03
Add feature list on the documentation

adelarsq:
Add nightly feature list on the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,9 +61,24 @@ guide.
 
 ### Nightly
 
-Rocket requires a nightly version of Rust as it makes heavy use of syntax
-extensions. This means that the first two unwieldly lines in the introductory
+Rocket requires a nightly version of Rust as it makes heavy use of syntax extensions.
+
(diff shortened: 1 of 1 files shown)


SergioBenitez:
I'm not sure that this belongs in the `README`. Can you tell me why you think it should be there?

adelarsq:
@SergioBenitez Most likely. Are there any other place where this information can be available? I just think that is a relevant information about why nightly is needed actually.

shssoichiro:
You can always find the list of feature flags in the [`lib.rs`](https://github.com/SergioBenitez/Rocket/blob/v0.2.0/lib/src/lib.rs#L1-L7) at the root of the crate.

SergioBenitez:
https://github.com/SergioBenitez/Rocket/issues/19 should be the canonical source for this. I've updated the initial comment with the full list of features.

adelarsq:
@SergioBenitez thanks

PR #176, opened 2017-02-06
[WIP] Connection pools in `contrib`

### LINKED ISSUE #167: Implement connection pools in `contrib`
Rocket's `contrib` should make it ease to retrieve a connection to a database from a connection pool. My initial thoughts on a design for this can be summarized as:

  * All configuration, including the pool size, which database to use, credentials, etc, should be done in the `Rocket.toml` file. It would be nice if configuration issues were reported at Rocket initialization, ~~but this likely requires a more involved change to Rocket~~.
  * Dependencies should be included on an as-needed basis. For instance, `contrib` shouldn't pull in `postgres` if the user only wants `sqlite`.
  * A single type should be exposed per configured database. The type should implement `FromRequest`, which results in a connection being taken from the connection pool.
  * Multiple databases, including several of the same type, should be allowed in a single project.
  * All major databases with Rust adapters should be supported, this includes:
    - postgres
    - redis
    - mysql
    - sqlite
    - cypher/neo4j
    - couchdb
    - diesel (postgres, sqlite)

The list of "major databases" is just that from the [r2d2](https://github.com/sfackler/r2d2) adapters page. The `r2d2` library should likely be used as the backing connection pool. ~~The [state](https://github.com/SergioBenitez/state) crate should likely be used to make the connection pool accessible globally to the `contrib` database module.~~

Edit: To clarify, it should be possible, and indeed a requirement of a good implementation, to use multiple databases in a single project, and even multiple different connections to a single type of database. For instance, I should be able to get handles to two different MySQL databases, a Redis store, and a PostgreSQL database without issue.

impowski:
So this is it, this is the start of implementation, and basically where we will discuss everything controversial about this feature. And this will fix #167.

Remember that this is only a basic structure of it, like an example, and I'll fix everything what isn't right.

TODO:
- [x] Initial design
- [x] Clarify names for our databases.
- [x] Clarify `#cfg()` features structure
- [x] Clarify configuration for connection
- [ ] Clarify logging for connection
- [ ] Fix tests
- [ ] Fix examples
- [ ] etc.

Right now my goal is to set myself and get into documentation of Rocket and read it as much as possible to get some small things like `Config` and everything else what I missed in past days especially a really deep journey into `State` and `Storage`.


contrib/src/lib.rs
```diff
@@ -66,3 +69,6 @@ mod uuid;
 
 #[cfg(feature = "uuid")]
 pub use uuid::{UUID, UuidParseError};
+
+#[cfg(feature = "database")]
```
shssoichiro: Probably the wrong feature flag here, should be `diesel_pg`

contrib/Cargo.toml
```diff
@@ -15,6 +15,12 @@ default = ["json"]
 json = ["serde", "serde_json"]
 tera_templates = ["tera", "templates"]
 handlebars_templates = ["handlebars", "templates"]
+database = []
+diesel_pg = ["diesel/postgres", "r2d2", "r2d2-diesel"]
```
shssoichiro: I think it would be nice to have each of the implementation features automatically use the `database` feature. The less rocket's users have to add to their Cargo.toml, the better.

contrib/src/lib.rs
impowski: I was in a rush, and didn't change it. Working on that already.

contrib/Cargo.toml
impowski: Didn't thought of that. Thanks!

impowski:
Just will leave a little note for myself here, maybe someone will be interested too.

So far I've been thinking about proper configuration design and multiple databases configuration from config. Here is couple of my thoughts:

- I'll define database by `database_{name}` and then configure it further something like `[dev.{name}]` or `[dev.database_{name}]`.
- Inside of `Database` or `DatabaseBuilder` (it's more related to config stuff), I'll set different variables for credentials and other things, and for different kinds I'll just check if they presented for some database type or not. For example:
```
struct Database {
    url: Option<Url>,
    creds: Option<(String, String)>
// etc.
}
```
Then inside of config:
```
database = "diesel"
cred = { user: "user", pass: "user" } // Will give a error on Rocket.toml that url is not presented
```
- I'll expose to user `State` types which will be an alias for `State<SomeConnection>`, and he will create his connection manually building with builder or automatically from config (if he will not present anything he will receive an error whether it's from config or manually), and then he will use that alias inside of his request guards.
- I'll give some sort of database testing interface and other things like try to connect and give an error if unable to connect. With testing I wanted to implemented because I'll use it myself inside of my project and it's such a pain to test database, and I'll think about migrations for diesel how could I play around with them.
- I'll think about some sort of logging for databases where if client wants he could see his queries and other things, like how much it took and etc.
- I'll think that having an ability to pass things through environment variables for databases would make sense too.

impowski:
@SergioBenitez I hope to get in touch with you, to clarify some things out, or if you prefer it here just tell me about it.

SergioBenitez:
@impowski I'm going to close this out simply because it's not in a PR-like state yet. I would absolutely love any help here, of course, so if you'd like, please continue working on this! I'm available on IRC. :)

PR #181, opened 2017-02-09
Add MsgPack implementation in contrib

### LINKED ISSUE #149: Add support for MsgPack in contrib
MsgPack is very common de/serialization protocol that is widely being used for backend services.
Serde already provides support for MsgPack through a user contributed crate.
Can we please add it to our TODO list?

shssoichiro:
Fixes #149. This is very similar to the existing JSON implementation, since they provide the same type of functionality.

### DIFF
--- a/contrib/src/lib.rs
+++ b/contrib/src/lib.rs
@@ -14,6 +14,7 @@
 //! an asterisk next to the features that are enabled by default:
 //!
 //! * [json*](struct.JSON.html)
+//! * [msgpack*](struct.MsgPack.html)
 //! * [handlebars_templates](struct.Template.html)
 //! * [tera_templates](struct.Template.html)
 //! * [uuid](struct.UUID.html)
@@ -54,6 +55,13 @@ pub mod json;
 
 #[cfg(feature = "json")]
 pub use json::{JSON, SerdeError, Value};
+
+#[cfg(feature = "msgpack")]
+#[doc(hidden)]
+pub mod msgpack;
+
+#[cfg(feature = "msgpack")]
+pub use msgpack::{MsgPack, MsgPackError};
 
 #[cfg(feature = "templates")]
 mod templates;
--- a/contrib/src/lib.rs
+++ b/contrib/src/lib.rs
@@ -14,7 +14,7 @@
 //! an asterisk next to the features that are enabled by default:
 //!
 //! * [json*](struct.JSON.html)
-//! * [msgpack*](struct.MsgPack.html)
+//! * [msgpack](struct.MsgPack.html)
 //! * [handlebars_templates](struct.Template.html)
 //! * [tera_templates](struct.Template.html)
 //! * [uuid](struct.UUID.html)
--- a/contrib/src/msgpack/mod.rs
+++ b/contrib/src/msgpack/mod.rs
@@ -1,0 +1,118 @@
+extern crate rmp_serde;
+
+use std::ops::{Deref, DerefMut};
+use std::io::Read;
+
+use rocket::outcome::Outcome;
+use rocket::request::Request;
+use rocket::data::{self, Data, FromData};
+use rocket::response::{self, Responder, content};
+use rocket::http::Status;
+
+use serde::{Serialize, Deserialize};
+
+pub use self::rmp_serde::decode::Error as MsgPackError;
+
+/// The MsgPack type: implements `FromData` and `Responder`, allowing you to easily
+/// consume and respond with MsgPack.
+///
+/// If you're receiving MsgPack data, simple add a `data` parameter to your route
+/// arguments and ensure the type o the parameter is a `MsgPack<T>`, where `T` is
+/// some type you'd like to parse from MsgPack. `T` must implement `Deserialize`
+/// from [Serde](https://github.com/3Hren/msgpack-rust). The data is parsed from the
+/// HTTP request body.
+///
+/// ```rust,ignore
+/// #[post("/users/", format = "application/msgpack", data = "<user>")]
+/// fn new_user(user: MsgPack<User>) {
+///     ...
+/// }
+/// ```
+/// You don't _need_ to use `format = "application/msgpack"`, but it _may_ be what
+/// you want. Using `format = application/msgpack` means that any request that
+/// doesn't specify "application/msgpack" as its first `Content-Type:` header
+/// parameter will not be routed to this handler.
+///
+/// If you're responding with MsgPack data, return a `MsgPack<T>` type, where `T`
+/// implements `Serialize` from [Serde](https://github.com/3Hren/msgpack-rust). The
+/// content type of the response is set to `application/msgpack` automatically.
+///
+/// ```rust,ignore
+/// #[get("/users/<id>")]
+/// fn user(id: usize) -> MsgPack<User> {
+///     let user_from_id = User::from(id);
+///     ...
+///     MsgPack(user_from_id)
+/// }
+/// ```
+///
+#[derive(Debug)]
+pub struct MsgPack<T>(pub T);
+
+impl<T> MsgPack<T> {
+    /// Consumes the MsgPack wrapper and returns the wrapped item.
+    ///
+    /// # Example
+    /// ```rust
+    /// # use rocket_contrib::MsgPack;
+    /// let string = "Hello".to_string();
+    /// let my_msgpack = MsgPack(string);
+    /// assert_eq!(my_msgpack.into_inner(), "Hello".to_string());
+    /// ```
+    pub fn into_inner(self) -> T {
+        self.0
+    }
+}
+
+impl<T: Deserialize> FromData for MsgPack<T> {
+    type Error = MsgPackError;
+
+    fn from_data(request: &Request, data: Data) -> data::Outcome<Self, Self::Error> {
+        if !request.content_type().map_or(false, |ct| ct.is_msgpack()) {
+            error_!("Content-Type is not MsgPack.");
+            return Outcome::Forward(data);
+        }
+
+        let mut buf = Vec::new();
+        if let Err(e) = data.open().read_to_end(&mut buf) {
+            let e = MsgPackError::InvalidDataRead(e);
+            error_!("Couldn't read request data: {:?}", e);
+            return Outcome::Failure((Status::BadRequest, e));
+        };
+        match rmp_serde::from_slice(&buf).map(|val| MsgPack(val)) {
+            Ok(value) => Outcome::Success(value),
+            Err(e) => {
+                error_!("Couldn't parse MsgPack body: {:?}", e);
+                Outcome::Failure((Status::BadRequest, e))
+            }
+        }
+    }
+}
+
+// Serializes the wrapped value into MsgPack. Returns a response with Content-Type
+// MsgPack and a fixed-size body with the serialization. If serialization fails, an
+// `Err` of `Status::InternalServerError` is returned.
+impl<T: Serialize> Responder<'static> for MsgPack<T> {
+    fn respond(self) -> response::Result<'static> {
+        rmp_serde::to_vec(&self.0).map(|buf| {
+            content::MsgPack(buf).respond().unwrap()
+        }).map_err(|e| {
+            error_!("MsgPack failed to serialize: {:?}", e);
+            Status::InternalServerError
+        })
+    }
+}
+
+impl<T> Deref for MsgPack<T> {
+    type Target = T;
+
+    fn deref<'a>(&'a self) -> &'a T {
+        &self.0
+    }
+}
+
+impl<T> DerefMut for MsgPack<T> {
+    fn deref_mut<'a>(&'a mut self) -> &'a mut T {
+        &mut self.0
+    }
+}
--- a/contrib/src/msgpack/mod.rs
+++ b/contrib/src/msgpack/mod.rs
@@ -34,7 +34,7 @@ pub use self::rmp_serde::decode::Error as MsgPackError;
 /// doesn't specify "application/msgpack" as its first `Content-Type:` header
 /// parameter will not be routed to this handler. By default, Rocket will accept a
 /// Content Type of any of the following for MessagePack data:
-/// `application/msgpack`, `application/x-msgpack`, or `bin/msgpack`.
+/// `application/msgpack`, `application/x-msgpack`, `bin/msgpack`, or `bin/x-msgpack`.
 ///
 /// If you're responding with MessagePack data, return a `MsgPack<T>` type, where `T`
 /// implements `Serialize` from [Serde](https://github.com/serde-rs/serde). The
--- a/contrib/src/msgpack/mod.rs
+++ b/contrib/src/msgpack/mod.rs
@@ -7,7 +7,7 @@ use rocket::outcome::Outcome;
 use rocket::request::Request;
 use rocket::data::{self, Data, FromData};
 use rocket::response::{self, Responder, Response};
-use rocket::http::Status;
+use rocket::http::{ContentType, Status};
 
 use serde::{Serialize, Deserialize};
 
@@ -70,15 +70,18 @@ impl<T> MsgPack<T> {
 /// TODO: Determine this size from some configuration parameter.
 const MAX_SIZE: u64 = 1048576;
 
+/// Accepted content types are:
+/// `application/msgpack`, `application/x-msgpack`, `bin/msgpack`, and `bin/x-msgpack`
+fn is_msgpack_content_type(ct: &ContentType) -> bool {
+    (ct.ttype == "application" || ct.ttype == "bin")
+        && (ct.subtype == "msgpack" || ct.subtype == "x-msgpack")
+}
+
 impl<T: Deserialize> FromData for MsgPack<T> {
     type Error = MsgPackError;
 
     fn from_data(request: &Request, data: Data) -> data::Outcome<Self, Self::Error> {
-        // Accepted content types are:
-        // `application/msgpack`, `application/x-msgpack`, `bin/msgpack`, and `bin/x-msgpack`
-        if !request.content_type().map_or(false, |ct|
-            (ct.ttype == "application" || ct.ttype == "bin") &&
-            (ct.subtype == "msgpack" || ct.subtype == "x-msgpack")) {
+        if !request.content_type().map_or(false, |ct| is_msgpack_content_type(&ct)) {
             error_!("Content-Type is not MessagePack.");
             return Outcome::Forward(data);
         }
@@ -100,19 +103,18 @@ impl<T: Deserialize> FromData for MsgPack<T> {
     }
 }
 
-// Serializes the wrapped value into MessagePack. Returns a response with Content-Type
-// MessagePack and a fixed-size body with the serialization. If serialization fails, an
-// `Err` of `Status::InternalServerError` is returned.
+/// Serializes the wrapped value into MessagePack. Returns a response with Content-Type
+/// MessagePack and a fixed-size body with the serialization. If serialization fails, an
+/// `Err` of `Status::InternalServerError` is returned.
 impl<T: Serialize> Responder<'static> for MsgPack<T> {
     fn respond(self) -> response::Result<'static> {
-        rmp_serde::to_vec(&self.0).map(|buf| {
-            Response::build()
-                .sized_body(Cursor::new(buf))
-                .ok::<Status>()
-                .unwrap()
-        }).map_err(|e| {
+        rmp_serde::to_vec(&self.0).map_err(|e| {
             error_!("MsgPack failed to serialize: {:?}", e);
             Status::InternalServerError
+        }).and_then(|buf| {
+            Response::build()
+                .sized_body(Cursor::new(buf))
+                .ok()
         })
     }
 }
--- a/contrib/src/msgpack/mod.rs
+++ b/contrib/src/msgpack/mod.rs
@@ -28,6 +28,7 @@ pub use self::rmp_serde::decode::Error as MsgPackError;
 ///     ...
 /// }
 /// ```
+///
 /// You don't _need_ to use `format = "application/msgpack"`, but it _may_ be what
 /// you want. Using `format = application/msgpack` means that any request that
 /// doesn't specify "application/msgpack" as its first `Content-Type:` header
@@ -47,7 +48,6 @@ pub use self::rmp_serde::decode::Error as MsgPackError;
 ///     MsgPack(user_from_id)
 /// }
 /// ```
-///
 #[derive(Debug)]
 pub struct MsgPack<T>(pub T);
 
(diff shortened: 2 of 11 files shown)


contrib/src/lib.rs
```diff
@@ -14,6 +14,7 @@
 //! an asterisk next to the features that are enabled by default:
 //!
 //! * [json*](struct.JSON.html)
+//! * [msgpack*](struct.MsgPack.html)
```
SergioBenitez: This shouldn't have a `*` on it.

contrib/src/msgpack/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+use rocket::request::Request;
+use rocket::data::{self, Data, FromData};
+use rocket::response::{self, Responder, content};
+use rocket::http::Status;
+
+use serde::{Serialize, Deserialize};
+
+pub use self::rmp_serde::decode::Error as MsgPackError;
+
+/// The MsgPack type: implements `FromData` and `Responder`, allowing you to easily
```
SergioBenitez: It looks like you just replaced `JSON` with `MsgPack` here, but the wording doesn't work in some cases. Can you customize this for `MsgPack`? In particular, you should likely refer to it as `MessagePack`.

lib/src/http/content_type.rs
```diff
@@ -104,6 +104,7 @@ impl ContentType {
         "HTML", HTML, is_html => "text", "html" ; "charset=utf-8",
         "Plain", Plain, is_plain => "text", "plain" ; "charset=utf-8",
         "JSON", JSON, is_json => "application", "json",
+        "MsgPack", MsgPack, is_msgpack => "application", "msgpack",
```
SergioBenitez: The internet seems to think that the correct Content-Type is `application/x-msgpack`. Why did you choose this?

lib/src/response/content.rs
```diff
@@ -82,6 +82,7 @@ macro_rules! ctrs {
 ctrs! {
     JSON: "JSON", "application/json",
     XML: "XML", "text/xml",
+    MsgPack: "MsgPack", "application/msgpack",
```
SergioBenitez: Same thing here.

lib/src/response/responder.rs
```diff
@@ -201,6 +201,16 @@ impl Responder<'static> for String {
     }
 }
 
+/// Returns a response with a fixed-size body
+/// containing the byte vec `self`. Always returns `Ok`.
+impl Responder<'static> for Vec<u8> {
```
SergioBenitez: Why do we need this? Why can't we just use `Cursor::new(vec)` in the `MsgPack` `Responder`?

lib/src/http/content_type.rs
shssoichiro: The reason I used `application/msgpack` is because [the x- prefix is deprecated](https://tools.ietf.org/html/rfc6648) on MIME types. The most recent/most upvoted answer on [this StackOverflow question](http://stackoverflow.com/questions/6888900/mime-type-for-msgpack#25886130) recommends using `application/msgpack` for this reason. However, I've found `application/x-msgpack` and even `bin/msgpack` used in other places on a Google search, and it seems that msgpack [never registered](https://www.iana.org/assignments/media-types/media-types.xhtml) or [decided on](https://github.com/msgpack/msgpack/issues/194) an official MIME type. So, I'll defer to your judgment on which MIME type we should use here.

lib/src/response/responder.rs
shssoichiro: Because the `Responder` wants to return a `Result` with `rocket::Response`, so `Cursor<Vec<u8>>` won't fit that type definition.

contrib/src/msgpack/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub fn into_inner(self) -> T {
+        self.0
+    }
+}
+
+impl<T: Deserialize> FromData for MsgPack<T> {
+    type Error = MsgPackError;
+
+    fn from_data(request: &Request, data: Data) -> data::Outcome<Self, Self::Error> {
+        if !request.content_type().map_or(false, |ct| ct.is_msgpack()) {
```
SergioBenitez: I think the right thing to do here is to match against all of the commonly used MessagePack content types, and accept all of them. Document which are accepted in the top-level docs _and_ here.

contrib/src/msgpack/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+            error_!("Content-Type is not MessagePack.");
+            return Outcome::Forward(data);
+        }
+
+        let mut buf = Vec::new();
+        if let Err(e) = data.open().read_to_end(&mut buf) {
```
SergioBenitez: This is bad as it opens the application up to DOS attack via an out-of-memory problem. The client can just stream infinite data until the machine runs out of memory. This needs to be limited.

contrib/src/msgpack/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+            let e = MsgPackError::InvalidDataRead(e);
+            error_!("Couldn't read request data: {:?}", e);
+            return Outcome::Failure((Status::BadRequest, e));
+        };
+        match rmp_serde::from_slice(&buf).map(|val| MsgPack(val)) {
```
SergioBenitez: Will `rmp_serde` not operate on a stream directly?

contrib/src/msgpack/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+    }
+}
+
+// Serializes the wrapped value into MessagePack. Returns a response with Content-Type
+// MessagePack and a fixed-size body with the serialization. If serialization fails, an
+// `Err` of `Status::InternalServerError` is returned.
+impl<T: Serialize> Responder<'static> for MsgPack<T> {
+    fn respond(self) -> response::Result<'static> {
+        rmp_serde::to_vec(&self.0).map(|buf| {
+            content::MsgPack(buf).respond().unwrap()
```
SergioBenitez: Construct the response here directly so we don't need to `Vec` `Responder` implementation.

lib/src/http/content_type.rs
SergioBenitez: `application/msgpack` seems like what we should use here. But see my inline comments about what we should accept.

lib/src/response/content.rs
SergioBenitez: Name is `MessagePack` not "MsgPack`.

lib/src/response/responder.rs
SergioBenitez: See my inline comments.

contrib/src/msgpack/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+use rocket::data::{self, Data, FromData};
+use rocket::response::{self, Responder, content};
+use rocket::http::Status;
+
+use serde::{Serialize, Deserialize};
+
+pub use self::rmp_serde::decode::Error as MsgPackError;
+
+/// The `MsgPack` type: implements `FromData` and `Responder`, allowing you to easily
+/// consume and respond with MessagePack.
```
SergioBenitez: Should say "MessagePack data."

contrib/src/msgpack/mod.rs
shssoichiro: It does not appear to have a method for that, no

shssoichiro:
Pushed the next set of changes. I'm not sure that I did the content type checking in the best possible way, since the `ctrs!` macro doesn't support checking for multiple types, but I needed to leave the MsgPack type in the macro or other parts of the feature would break.

contrib/src/msgpack/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+/// TODO: Determine this size from some configuration parameter.
+const MAX_SIZE: u64 = 1048576;
+
+impl<T: Deserialize> FromData for MsgPack<T> {
+    type Error = MsgPackError;
+
+    fn from_data(request: &Request, data: Data) -> data::Outcome<Self, Self::Error> {
+        // Accepted content types are:
+        // `application/msgpack`, `application/x-msgpack`, and `bin/msgpack`
+        if !request.content_type().map_or(false, |ct|
```
SergioBenitez: This is really hard to read. Can we split this into another function and reformat it so it's obvious what's happening? Maybe use a `match`? Due to the `UncasedAscii`, I don't know if Rust will allow something like:

```rust
match (ct.ttype, ct.subtype) {
    ("application", "msgpack") | ("application", "x-msgpack") => true,
    ("bin", "msgpack") | ("bin", "x-msgpack") => true,
    _ => false
}
```

But play around with it; maybe something like that will work.

contrib/src/msgpack/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+///
+/// ```rust,ignore
+/// #[get("/users/<id>")]
+/// fn user(id: usize) -> MsgPack<User> {
+///     let user_from_id = User::from(id);
+///     ...
+///     MsgPack(user_from_id)
+/// }
+/// ```
+///
```
SergioBenitez: Kill this empty line here.

contrib/src/msgpack/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+/// some type you'd like to parse from MessagePack. `T` must implement `Deserialize`
+/// from [Serde](https://github.com/serde-rs/serde). The data is parsed from the
+/// HTTP request body.
+///
+/// ```rust,ignore
+/// #[post("/users/", format = "application/msgpack", data = "<user>")]
+/// fn new_user(user: MsgPack<User>) {
+///     ...
+/// }
+/// ```
```
SergioBenitez: Add an empty line below this.

contrib/src/msgpack/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+            return Outcome::Forward(data);
+        }
+
+        let mut buf = Vec::new();
+        if let Err(e) = data.open().take(MAX_SIZE).read_to_end(&mut buf) {
+            let e = MsgPackError::InvalidDataRead(e);
+            error_!("Couldn't read request data: {:?}", e);
+            return Outcome::Failure((Status::BadRequest, e));
+        };
+        match rmp_serde::from_slice(&buf).map(|val| MsgPack(val)) {
```
SergioBenitez: Add an empty line above here.

contrib/src/msgpack/mod.rs
shssoichiro: Looks like `match` needs the type to be exactly the same, so even though `UncasedAscii` implements `PartialEq` for `str`, Rust's typechecker won't allow this.

I did modify it slightly in a way that improves readability, although I don't think it's as nice as this.

contrib/src/msgpack/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+            Ok(value) => Outcome::Success(value),
+            Err(e) => {
+                error_!("Couldn't parse MessagePack body: {:?}", e);
+                Outcome::Failure((Status::BadRequest, e))
+            }
+        }
+    }
+}
+
+// Serializes the wrapped value into MessagePack. Returns a response with Content-Type
```
SergioBenitez: This should be a real docstring, not a comment. Use three `///`. I've fixed this in the `JSON` docs.

contrib/src/msgpack/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+// MessagePack and a fixed-size body with the serialization. If serialization fails, an
+// `Err` of `Status::InternalServerError` is returned.
+impl<T: Serialize> Responder<'static> for MsgPack<T> {
+    fn respond(self) -> response::Result<'static> {
+        rmp_serde::to_vec(&self.0).map(|buf| {
+            Response::build()
+                .sized_body(Cursor::new(buf))
+                .ok::<Status>()
```
SergioBenitez: Can Rust not infer `Status`?

contrib/src/msgpack/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+        }
+    }
+}
+
+// Serializes the wrapped value into MessagePack. Returns a response with Content-Type
+// MessagePack and a fixed-size body with the serialization. If serialization fails, an
+// `Err` of `Status::InternalServerError` is returned.
+impl<T: Serialize> Responder<'static> for MsgPack<T> {
+    fn respond(self) -> response::Result<'static> {
+        rmp_serde::to_vec(&self.0).map(|buf| {
```
SergioBenitez: I think if you use `and_then` instead of `map` and remove the `unwrap`, things might work out okay and you'd be able to get rid of the type annotation. You'd need to move the `map_err` before the `and_then` call for this to work, I believe.

shssoichiro:
Fixed that round of comments and added examples. I also needed to add a method to get a response body as a byte Vec, since MessagePack data isn't representable as a UTF-8 String.

examples/msgpack/src/tests.rs
```diff
@@ -0,0 +1,105 @@
...
+
+use rocket;
+use rocket::testing::MockRequest;
+use rocket::http::Method::*;
+use rocket::http::{Status, ContentType};
+use rocket::Response;
+use std::collections::HashMap;
+
+#[derive(Deserialize)]
+#[allow(dead_code)]
```
SergioBenitez: Why allow dead code?...

examples/msgpack/src/tests.rs
```diff
@@ -0,0 +1,105 @@
...
+    run_test!(&rocket, req, |mut response: Response| {
+        assert_eq!(response.status(), Status::Ok);
+        let body = rmp_serde::from_slice::<Message>(&response.body().unwrap().into_bytes().unwrap()).unwrap();
+        assert!(body.contents == "Hello, world!");
+    });
+
+    // Change the message contents.
+    let req = MockRequest::new(Put, "/message/1")
+        .header(ContentType::MsgPack)
+        .body_binary(rmp_serde::to_vec(&[("contents", "Bye bye, world!")].iter().cloned().collect::<HashMap<&'static str, &'static str>>()).unwrap());
```
SergioBenitez: This line is way too long.

examples/msgpack/src/tests.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    run_test!(&rocket, req, |response: Response| {
+        assert_eq!(response.status(), Status::Ok);
+    });
+
+    // Check that the message exists with the correct contents.
+    let req = MockRequest::new(Get, "/message/1") .header(ContentType::MsgPack);
+    run_test!(&rocket, req, |mut response: Response| {
+        assert_eq!(response.status(), Status::Ok);
+        let body = rmp_serde::from_slice::<Message>(&response.body().unwrap().into_bytes().unwrap()).unwrap();
```
SergioBenitez: This line is way too long.

examples/msgpack/src/tests.rs
```diff
@@ -0,0 +1,105 @@
...
+    // Check that a message with ID 1 doesn't exist.
+    let req = MockRequest::new(Get, "/message/1").header(ContentType::MsgPack);
+    run_test!(&rocket, req, |response: Response| {
+        assert_eq!(response.status(), Status::NotFound);
+    });
+
+    // Add a new message with ID 1.
+    let req = MockRequest::new(Post, "/message/1")
+        .header(ContentType::MsgPack)
+        .body_binary(rmp_serde::to_vec(&[("contents", "Hello, world!")].iter().cloned().collect::<HashMap<&'static str, &'static str>>()).unwrap());
```
SergioBenitez: Line length. This should probably be split out in a help function or something.

examples/msgpack/src/tests.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    run_test!(&rocket, req, |response: Response| {
+        assert_eq!(response.status(), Status::Ok);
+    });
+
+    // Check that the message exists with the updated contents.
+    let req = MockRequest::new(Get, "/message/1") .header(ContentType::MsgPack);
+    run_test!(&rocket, req, |mut response: Response| {
+        assert_eq!(response.status(), Status::Ok);
+        let body = rmp_serde::from_slice::<Message>(&response.body().unwrap().into_bytes().unwrap()).unwrap();
```
SergioBenitez: Line length.

lib/src/testing.rs
```diff
@@ -229,6 +229,28 @@ impl<'r> MockRequest<'r> {
...
+    /// use rocket::testing::MockRequest;
+    /// use rocket::http::ContentType;
+    ///
+    /// # #[allow(unused_variables)]
+    /// let req = MockRequest::new(Post, "/")
+    ///     .header(ContentType::MsgPack)
+    ///     .body_binary(&[0x92, 0x2a, 0xaa, 0x74, 0x68, 0x65, 0x20, 0x41, 0x6e, 0x73, 0x77, 0x65, 0x72]);
+    /// ```
+    #[inline]
+    pub fn body_binary<S: AsRef<[u8]>>(mut self, body: S) -> Self {
```
SergioBenitez: I don't like the idea of having two `body` methods. `str` and `String` are `AsRef<[u8]>`, so we can just use those bounds instead.

lib/src/response/response.rs
```diff
@@ -54,7 +54,7 @@ impl<T> Body<T> {
 }
 
 impl<T: io::Read> Body<T> {
-    /// Attepts to read `self` into a `String` and returns it. If reading or
+    /// Attempts to read `self` into a `String` and returns it. If reading or
     /// conversion fails, returns `None`.
     pub fn into_string(self) -> Option<String> {
         let (mut body, mut string) = match self {
```
SergioBenitez: If you're going to add `into_bytes()`, you should just call `into_bytes().map(String::from_utf8)` here.

SergioBenitez: I've added an `into_bytes` in master.

examples/msgpack/src/tests.rs
shssoichiro: This should be just for the id field. The example returns an id field, but that field is never used in the tests.

shssoichiro:
The primary use case of Msgpack is sending data from one backend service to another, e.g. from one Rocket app to another using a library such as `reqwest`. However, the setup on the receiving side doesn't change from what's shown in the example. The difference would be that instead of sending data from a browser, it's sent from another app, but it's still just through an HTTP request. So I don't think there's much benefit in us showing that in the example.

I'm about to push a commit with the other changes mentioned.

SergioBenitez:
My main point is that the example is a lot more complex than it needs to be. Let's just parse some msgpack and return a message proving that it was parsed correctly and have that be it.

SergioBenitez:
Ah, nice! That's much better.

Can you add something that shows deserialization as well? Maybe just an `echo` that return the ID and contents as a string or something?

shssoichiro:
How's that?

examples/msgpack/src/tests.rs
```diff
@@ -0,0 +1,48 @@
+extern crate rmp_serde;
```
SergioBenitez: Can we not depend on the crate here? I liked that the msgpack binary was in the test case. Just the binary with a note of what the bytes are representing would be great.

examples/msgpack/src/main.rs
```diff
@@ -0,0 +1,39 @@
...
+#[get("/<id>", format = "application/msgpack")]
+fn get(id: usize) -> MsgPack<Message> {
+    MsgPack(Message {
+        id: id,
+        contents: "Hello, world!".to_string(),
+    })
+}
+
+#[post("/", data = "<data>", format = "application/msgpack")]
+fn create(data: MsgPack<Message>) -> Result<(), ()> {
```
SergioBenitez: This isn't testable. Just return the contents directly (so return type of this function is `String`) so you can test for them later:

```
data.into_inner().contents
```

examples/msgpack/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "msgpack"
+version = "0.0.1"
+workspace = "../../"
+
+[dependencies]
+rocket = { path = "../../lib" }
+rocket_codegen = { path = "../../codegen" }
+rmp-serde = "0.12"
```
SergioBenitez: Let's remove this dep.

examples/msgpack/src/tests.rs
shssoichiro: I don't believe we can decode the response from the `msgpack_get` test, on line 26 of tests.rs, without this crate, and that whole test is rather pointless without that ability.

SergioBenitez: No need to decode. We should know the bytes, right? Just check against the bytes.

shssoichiro:
That should be everything.

contrib/src/msgpack/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+///     ...
+/// }
+/// ```
+///
+/// You don't _need_ to use `format = "application/msgpack"`, but it _may_ be what
+/// you want. Using `format = application/msgpack` means that any request that
+/// doesn't specify "application/msgpack" as its first `Content-Type:` header
+/// parameter will not be routed to this handler. By default, Rocket will accept a
+/// Content Type of any of the following for MessagePack data:
+/// `application/msgpack`, `application/x-msgpack`, or `bin/msgpack`.
```
SergioBenitez: We also accept `bin/x-msgpack`.

contrib/src/msgpack/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+    /// assert_eq!(my_msgpack.into_inner(), "Hello".to_string());
+    /// ```
+    pub fn into_inner(self) -> T {
+        self.0
+    }
+}
+
+/// Maximum size of MessagePack data is 1MB.
+/// TODO: Determine this size from some configuration parameter.
+const MAX_SIZE: u64 = 1048576;
```
SergioBenitez: Is this a good default limit?

shssoichiro: This is the same that we're using for JSON, and since MessagePack is more efficient, this actually gets us a little more data.

SergioBenitez: I was thinking that maybe MessagePack encoded messages tend to contain more information, but let's go with this for now.

examples/msgpack/src/main.rs
```diff
@@ -0,0 +1,38 @@
...
+fn get(id: usize) -> MsgPack<Message> {
+    MsgPack(Message {
+        id: id,
+        contents: "Hello, world!".to_string(),
+    })
+}
+
+#[post("/", data = "<data>", format = "application/msgpack")]
+fn create(data: MsgPack<Message>) -> Result<String, ()> {
+    Ok(data.contents.clone())
```
SergioBenitez: Why clone? `data.into_inner().contents` would be better.

SergioBenitez:
Thanks so much for this, again! Merged in d43678c35ea3cbaab9343faf1565638f44f94570.

PR #194, opened 2017-02-13
updated json example with managed state

Crazy-Owl:
Reading the docs and trying out examples, I've noticed that JSON example uses `lazy_static` way to handle state. Which I've changed to managed state as this seems to be the preferred way to handle state in Rocket.

I've also had to update tests since `run_test!` macro was not properly isolating objects that were created inside: `lazy_static`-initialized objects were leaking outside the test runs.

### DIFF
--- a/examples/json/src/main.rs
+++ b/examples/json/src/main.rs
@@ -70,10 +70,13 @@ fn not_found() -> JSON<Value> {
     }))
 }
 
-fn main() {
+fn rocket() -> rocket::Rocket {
     rocket::ignite()
         .mount("/message", routes![new, update, get])
         .catch(errors![not_found])
         .manage(Mutex::new(HashMap::<ID, String>::new()))
-        .launch();
+}
+
+fn main() {
+    rocket().launch();
 }
--- a/examples/json/src/main.rs
+++ b/examples/json/src/main.rs
@@ -3,13 +3,13 @@
 
 extern crate rocket;
 extern crate serde_json;
-#[macro_use] extern crate lazy_static;
 #[macro_use] extern crate rocket_contrib;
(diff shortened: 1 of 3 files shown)


examples/json/src/main.rs
```diff
@@ -76,5 +74,6 @@ fn main() {
     rocket::ignite()
```
SergioBenitez: Move this into a `rocket` function like the `forms` example does, and then call that function from `main` and from the tests as well.

Crazy-Owl: Sure thing! I'll submit the change tomorrow

Crazy-Owl:
Actually, the change is so minor I've decided to give it a try right now ;)

examples/json/src/tests.rs
```diff
@@ -5,18 +5,16 @@ use rocket::http::{Status, ContentType};
 use rocket::Response;
 
 macro_rules! run_test {
-    ($req:expr, $test_fn:expr) => ({
-        let rocket = rocket::ignite()
-            .mount("/message", routes![super::new, super::update, super::get])
-            .catch(errors![super::not_found]);
-
+    ($req:expr, $test_fn:expr, $rocket:expr) => ({
```
SergioBenitez: Can we make `rocket` the first argument to this macro? It looks a bit odd hanging off at the end there.

Crazy-Owl: Sure, done.

SergioBenitez:
This is awesome. Thank you!

SergioBenitez:
Merged in https://github.com/SergioBenitez/Rocket/commit/3e063af965ece6fac3376eb78646eebad92ab439. Thank you again!

PR #197, opened 2017-02-16
Add Font Content Types

mmstick:
Adds support for TTF, OTF, WOFF, and WOFF2 content types, in addition to fixing a few Clippy concerns in the file. Also added WEBP.

### DIFF
--- a/lib/src/http/content_type.rs
+++ b/lib/src/http/content_type.rs
@@ -10,7 +10,7 @@ use http::ascii::{uncased_eq, UncasedAscii};
 ///
 /// # Usage
 ///
-/// ContentTypes should rarely be created directly. Instead, an associated
+/// `ContentTypes` should rarely be created directly. Instead, an associated
 /// constant should be used; one is declared for most commonly used content
 /// types.
 ///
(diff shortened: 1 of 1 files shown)


SergioBenitez:
Can you point me to an authoritative resource on the selection of these Content-Types?

mmstick:
- https://www.iana.org/assignments/media-types/media-types.xhtml
- TTF/OTF: http://www.iana.org/assignments/media-types/application/font-sfnt
- WOFF: http://www.iana.org/assignments/media-types/application/font-woff
- WOFF2: http://www.iana.org/assignments/media-types/font/woff2

Although it seems there's also support for `font/ttf` and `font/otf`.

SergioBenitez:
According to [this IETF document](https://datatracker.ietf.org/doc/draft-ietf-justfont-toplevel/), the `font` top-level MIME type is still being reviewed; it is a _draft_.

mmstick:
The newish WOFF2 format requires the new top level MIME type, hence why I'm
using it for WOFF2.

SergioBenitez:
Hmm, yeah.

Can you commit to monitoring that proposal and updating these content-types when it goes through?

mmstick:
Sure thing

SergioBenitez:
Thanks!

Merged in https://github.com/SergioBenitez/Rocket/commit/4161949a1cbd16fa9f3ceb228528c105087529f6.

PR #198, opened 2017-02-17
Add tests for examples/extended_validation.

mike820324:
This PR add tests for examples/extended_validation #62.

This is my first time contribute to a rust project, any advice and recommendation will be helpful.

### DIFF
--- a/examples/extended_validation/src/main.rs
+++ b/examples/extended_validation/src/main.rs
@@ -79,7 +79,7 @@ fn user_page(username: &str) -> String {
     format!("This is {}'s page.", username)
 }
 
-pub fn rocket() -> rocket::Rocket {
+fn rocket() -> rocket::Rocket {
     rocket::ignite()
         .mount("/", routes![files::index, files::files, user_page, login])
 }
--- a/examples/extended_validation/src/main.rs
+++ b/examples/extended_validation/src/main.rs
@@ -4,6 +4,8 @@
 extern crate rocket;
 
 mod files;
+#[cfg(test)]
+mod tests;
 
 use rocket::response::Redirect;
 use rocket::request::{Form, FromFormValue};
@@ -77,8 +79,11 @@ fn user_page(username: &str) -> String {
     format!("This is {}'s page.", username)
 }
 
-fn main() {
+pub fn rocket() -> rocket::Rocket {
     rocket::ignite()
         .mount("/", routes![files::index, files::files, user_page, login])
-        .launch();
+}
+
+fn main() {
+    rocket().launch()
 }
--- a/examples/extended_validation/src/tests.rs
+++ b/examples/extended_validation/src/tests.rs
@@ -1,0 +1,89 @@
+use rocket::testing::MockRequest;
+use rocket::http::Method::*;
+use rocket::http::{ContentType, Status};
+
+use super::rocket;
+
+fn test_login(username: &str,
+              password: &str,
+              age: &str,
+              status: Status,
+              body: Option<&'static str>) {
+    let rocket = rocket();
+    let query = &format!("username={}&password={}&age={}", username, password, age);
+
+    let mut req = MockRequest::new(Post, "/login")
+        .header(ContentType::Form)
+        .body(&query);
+
+    let mut response = req.dispatch_with(&rocket);
+    let body_str = response.body().and_then(|body| body.into_string());
+
+    println!("Checking: {:?}/{:?}/{:?}/{:?}", username, password, age, body_str);
+    assert_eq!(response.status(), status);
+
+    if let Some(string) = body {
+        assert!(body_str.map_or(true, |s| s.contains(string)));
+    }
+}
+
+const OK: Status = self::Status::Ok;
+
+#[test]
+fn test_good_login() {
+    test_login("Sergio", "password", "30", Status::SeeOther, None);
+}
+
+#[test]
+fn test_invalid_user() {
+    test_login("-1", "password", "30", OK, Some("Unrecognized user"));
+    test_login("Mike", "password", "30", OK, Some("Unrecognized user"));
+}
+
+#[test]
+fn test_invalid_password() {
+    test_login("Sergio", "password1", "30", OK, Some("Wrong password!"));
+    test_login("Sergio", "ok", "30", OK, Some("Password is invalid: Too short!"));
+}
+
+#[test]
+fn test_invalid_age() {
+    test_login("Sergio",
+               "password",
+               "20",
+               OK,
+               Some("Age is invalid: Must be at least 21."));
+    test_login("Sergio",
+               "password",
+               "-100",
+               OK,
+               Some("Age is invalid: Must be at least 21."));
+    test_login("Sergio",
+               "password",
+               "hello",
+               OK,
+               Some("Age is invalid: Age value is not a number"));
+}
+
+fn check_bad_form(form_str: &str, status: Status) {
+    let rocket = rocket();
+    let mut req = MockRequest::new(Post, "/login")
+        .header(ContentType::Form)
+        .body(form_str);
+
+    let response = req.dispatch_with(&rocket);
+    assert_eq!(response.status(), status);
+}
+
+#[test]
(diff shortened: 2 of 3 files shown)


examples/extended_validation/src/tests.rs
```diff
@@ -0,0 +1,89 @@
...
+    let query = &format!("username={}&password={}&age={}", username, password, age);
+
+    let mut req = MockRequest::new(Post, "/login")
+        .header(ContentType::Form)
+        .body(&query);
+
+    let mut response = req.dispatch_with(&rocket);
+    let body_str = response.body().and_then(|body| body.into_string());
+
+    println!("Checking: {:?}/{:?}/{:?}/{:?}", username, password, age, body_str);
```
SergioBenitez: We can get rid of this `println`and kill the empty line above.

examples/extended_validation/src/tests.rs
```diff
@@ -0,0 +1,89 @@
...
+    assert_eq!(response.status(), status);
+
+    if let Some(string) = body {
+        assert!(body_str.map_or(true, |s| s.contains(string)));
```
SergioBenitez: Shouldn't this be `false` since the test expected a body but there was none?

SergioBenitez: We can remove the `body_str` declaration a few lines above and write this as something like:

```rust
assert_eq!(response.body().map(|body| body.contains(string)), Some(true))
```

examples/extended_validation/src/tests.rs
```diff
@@ -0,0 +1,89 @@
+use rocket::testing::MockRequest;
+use rocket::http::Method::*;
+use rocket::http::{ContentType, Status};
+
+use super::rocket;
+
+fn test_login(username: &str,
```
SergioBenitez: I dislike the formatting here. Maybe we can use shorter variable names like `user` and `pass` to try to get this to fit?

SergioBenitez: We can also use the signature:

```rust
fn test_login(..., body: T)
    where T: Into<Option<&'static str>
{
...
}
```

so that we don't have to write `Some` every time we call this function with a `Some`.

examples/extended_validation/src/tests.rs
```diff
@@ -0,0 +1,89 @@
...
+
+#[test]
+fn test_invalid_password() {
+    test_login("Sergio", "password1", "30", OK, Some("Wrong password!"));
+    test_login("Sergio", "ok", "30", OK, Some("Password is invalid: Too short!"));
+}
+
+#[test]
+fn test_invalid_age() {
+    test_login("Sergio",
```
SergioBenitez: Play around with the `test_login` signature and see if we can get these to fit in one line.

mike820324:
Thanks for the advice, it's very helpful. I'll deal with this when my day-job is done. :)

mike820324:
Sorry I accidentally close the PR. reopen it. T_T

examples/extended_validation/src/tests.rs
mike820324: Sure, I'll use some shorter name and see it can fit into one line perfectly.

This code format was originally generated by the rustfmt. 
One small question, is there any recommend rust format or more properly coding style for this project.

examples/extended_validation/src/main.rs
```diff
@@ -77,8 +79,11 @@ fn user_page(username: &str) -> String {
     format!("This is {}'s page.", username)
 }
 
-fn main() {
+pub fn rocket() -> rocket::Rocket {
```
mike820324: Yes, your right. I'll change this into non pub later.

examples/extended_validation/src/tests.rs
SergioBenitez: Well, we have a `rustfmt` file that kinda does what I want it to do, but otherwise, no. `rustfmt` still isn't super dependable, unfortunately.

examples/extended_validation/src/tests.rs
```diff
@@ -0,0 +1,89 @@
...
+    let mut req = MockRequest::new(Post, "/login")
+        .header(ContentType::Form)
+        .body(form_str);
+
+    let response = req.dispatch_with(&rocket);
+    assert_eq!(response.status(), status);
+}
+
+#[test]
+fn test_bad_form() {
```
mike820324: It's the same as the forms example.
To my understanding, these tests seems to be testing some malformed form data or form data that missed several required fields.

And the fields name between these two examples are the same, and hence are duplicated.

IMHO, all the tests in each example should be isolated and independent to each other and that is why I copy and paste test_bad_form to this test.

mike820324:
@SergioBenitez I have add several new commits to improve this PR based on your suggestions. 
Thanks.

examples/extended_validation/src/tests.rs
```diff
@@ -0,0 +1,86 @@
...
+    let query = &format!("username={}&password={}&age={}", user, pass, age);
+
+    let mut req = MockRequest::new(Post, "/login")
+        .header(ContentType::Form)
+        .body(&query);
+
+    let mut response = req.dispatch_with(&rocket);
+    assert_eq!(response.status(), status);
+    if let Some(string) = body.into() {
+        assert!(response.body().and_then(|body| body.into_string()).map_or(false, |s| s.contains(string)));
```
SergioBenitez: Can't this be the following?

```rust
response.body().map_or(false, |body| body.into_string().contains(string))
```

examples/extended_validation/src/tests.rs
```diff
@@ -0,0 +1,86 @@
...
+    }
+}
+
+const OK: Status = self::Status::Ok;
```
SergioBenitez: Just do a `use rocket::http::Status::{Ok, BadRequest, UnprocessableEntity}` at the top and remove this `const`.

examples/extended_validation/src/tests.rs
```diff
@@ -0,0 +1,86 @@
...
+use rocket::http::Method::*;
+use rocket::http::{ContentType, Status};
+
+use super::rocket;
+
+fn test_login<T>(user: &str, pass: &str, age: &str, status: Status, body: T)
+    where T: Into<Option<&'static str>>
+{
+    let rocket = rocket();
+    let query = &format!("username={}&password={}&age={}", user, pass, age);
```
SergioBenitez: There's an `&` here...

examples/extended_validation/src/tests.rs
```diff
@@ -0,0 +1,86 @@
...
+
+    let mut req = MockRequest::new(Post, "/login")
+        .header(ContentType::Form)
+        .body(&query);
```
SergioBenitez: And an `&` here? Probably remove the first one.

examples/extended_validation/src/tests.rs
mike820324: The return value of ```into_string``` is a ```Option<String>``` type. As a result, the above suggestion code may lead to the following compiler error.

```
error: no method named `contains` found for type `std::option::Option<std::string::String>` in the current scope
  --> src/tests.rs:21:73
   |
21 |         assert!(response.body().map_or(false, |body| body.into_string().contains(string)));
   |   
```

That's why the ```and_then().map_or``` chain is needed. The return value of the closure within ```and_then``` can be ```Option::None```. If it does, the ```map_or``` will use the default value, which is false.

SergioBenitez: Oh, that's right. This is pretty hard to read. Can we split it up into two lines?

```rust
let body_str = response.body().and_then(|b| b.into_string());
assert!(body_str.map_or(false, |s| s.contains(string));
```

mike820324: Sure, no problem at all.

mike820324:
Rebase to upstream master and force push remote.

SergioBenitez:
Awesome! Merged in https://github.com/SergioBenitez/Rocket/commit/7e7c31b9e7f61400e17bb794ffeaf743524435f1.

PR #200, opened 2017-02-17
Make the info message 'Rocket has launched from http://localhost:8000…' clickable in shells.

spytheman:
If there is no space after the URL, then the '…' character is treated as part of the URL, so that when you right click it or double click it to select/copy it, most terminal emulators become confused.

### DIFF
--- a/lib/src/rocket.rs
+++ b/lib/src/rocket.rs
@@ -555,7 +555,7 @@ impl Rocket {
             }
         };
 
-        info!("🚀  {} {}{}...",
+        info!("🚀  {} {}{} ...",
               White.paint("Rocket has launched from"),
               White.bold().paint("http://"),
               White.bold().paint(&full_addr));


spytheman:
I have tested the new behaviour in gnome-terminal, xterm and konsole - when there is space before the '…' character, the url becomes selectable/clickable by default.

SergioBenitez:
The intention is definitely for the link to be clickable, and the few terminals I've used do make the link clickable with the space. Unfortunately I don't like the aesthetics of the space there. Can we do something better? What if we use a [zero-width space](http://www.fileformat.info/info/unicode/char/200B/index.htm)? Does the terminal make it clickable then?

SergioBenitez:
@spytheman ping.

SergioBenitez:
@spytheman Pinging once again.

spytheman:
Inserting a zero width space utf character instead of a normal space does NOT make the link easily selectable by double clicking/right clicking. I have tried the following, where the last "empty" string, contain the zero width space:
`
        info!("🚀  {} {}{}{}...",
              White.paint("Rocket has launched from"),
              White.bold().paint("http://"),
              White.bold().paint(&full_addr),
              "");
`

SergioBenitez:
Okay. I just killed the '...' in d4b9360f57cc9a5825019edcc1a04dedaa8dc677. Thanks for this. :)

PR #201, opened 2017-02-18
Add tests for examples/static_files.

mike820324:
This PR add tests for examples/static_files #62.

Should I also add test for the *rocket-icon.jpg* ?

### DIFF
--- a/examples/static_files/src/tests.rs
+++ b/examples/static_files/src/tests.rs
@@ -3,11 +3,27 @@ use std::io::Read;
 use rocket::testing::MockRequest;
 use rocket::http::Method::*;
 use rocket::http::Status;
+use rocket::response::Body;
 
 use super::rocket;
 
+fn body_to_bytes(body: Body<&mut Read>) -> Option<Vec<u8>> {
+    match body {
+        Body::Sized(b, _)  => {
+            let mut data: Vec<u8> = vec![];
+            b.read(&mut data).expect("Read from response failed");
+            return Some(data);
+        },
+        Body::Chunked(b, _) => {
+            let mut data: Vec<u8> = vec![];
+            b.read(&mut data).expect("Read from response failed");
+            return Some(data);
+        }
+    }
+}
+
 fn test_query_file<T> (path: &str, status: Status, content: T)
-    where T: Into<Option<String>>
+    where T: Into<Option<Vec<u8>>>
 {
     let rocket = rocket();
     let mut req = MockRequest::new(Get, &path);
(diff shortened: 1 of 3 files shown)


mike820324:
Sure, separate the open file and read content code into a function. One question about the implementation idea is that, when the file is not exist or the reading operation is failed should I

1. panic inside the function
2. return the File Error code and let the caller decide what to do.

Thanks.

SergioBenitez:
If we expect it to exist, it's totally fine to panic so that the test will fail.

mike820324:
Ok, I'll push a new commit to improve this. :D

mike820324:
I have push new commit to add the following.

- add new function to read file content
- add new function to transform body into Vec<u8>

Not quite sure is there any other better way to do the transform body into Vec<u8>.

examples/static_files/src/tests.rs
```diff
@@ -0,0 +1,71 @@
...
+use rocket::response::Body;
+
+use super::rocket;
+
+fn body_to_bytes(body: Body<&mut Read>) -> Option<Vec<u8>> {
+    match body {
+        Body::Sized(b, _)  => {
+            let mut data: Vec<u8> = vec![];
+            b.read(&mut data).expect("Read from response failed");
+            return Some(data);
```
SergioBenitez: Just as a note, you don't need a `return` here. Just remove the semicolon.

examples/static_files/src/tests.rs
```diff
@@ -0,0 +1,71 @@
...
+        },
+        Body::Chunked(b, _) => {
+            let mut data: Vec<u8> = vec![];
+            b.read(&mut data).expect("Read from response failed");
+            return Some(data);
```
SergioBenitez: Same here.

examples/static_files/src/tests.rs
```diff
@@ -0,0 +1,71 @@
+use std::fs::File;
+use std::io::Read;
+use rocket::testing::MockRequest;
```
SergioBenitez: Add an empty new line above here.

examples/static_files/src/tests.rs
```diff
@@ -0,0 +1,71 @@
...
+use rocket::http::Method::*;
+use rocket::http::Status;
+use rocket::response::Body;
+
+use super::rocket;
+
+fn body_to_bytes(body: Body<&mut Read>) -> Option<Vec<u8>> {
+    match body {
+        Body::Sized(b, _)  => {
```
SergioBenitez: Just as a note, you can match on both variants at the same time to avoid the duplication like this:

```rust
match body {
    Body::Sized(b, _) | Body::Chunked(b, _) => {
        ...
    }
}
```

examples/static_files/src/tests.rs
```diff
@@ -0,0 +1,71 @@
+use std::fs::File;
+use std::io::Read;
+use rocket::testing::MockRequest;
+use rocket::http::Method::*;
+use rocket::http::Status;
+use rocket::response::Body;
+
+use super::rocket;
+
+fn body_to_bytes(body: Body<&mut Read>) -> Option<Vec<u8>> {
```
SergioBenitez: `Body` implements read. You should use `read_to_end` directly, making this entire function 3 lines.

```rust
let mut data = vec![];
body.read_to_end(&mut data);
data
```

examples/static_files/src/tests.rs
```diff
@@ -0,0 +1,71 @@
...
+    let body_data = response.body().and_then(|body| body_to_bytes(body));
+    if let Some(expected_data) = content.into() {
+        assert!(body_data.map_or(false, |s| s == expected_data));
+    }
+}
+
+fn read_file_content(path: &str) -> Vec<u8> {
+    let mut fp = File::open(&path).expect(&format!("Can not open {}", path));
+
+    let mut file_content:Vec<u8> = vec![];
```
SergioBenitez: You shouldn't need to specify the type here.

examples/static_files/src/tests.rs
```diff
@@ -0,0 +1,71 @@
...
+    fp.read(&mut file_content).expect(&format!("Reading {} failed.", path));
```
SergioBenitez: Use `read_to_end`. Using `read` can result in incomplete reads.

examples/static_files/src/tests.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    let body_data = response.body().and_then(|body| body_to_bytes(body));
+    if let Some(expected_data) = content.into() {
+        assert!(body_data.map_or(false, |s| s == expected_data));
+    }
+}
+
+fn read_file_content(path: &str) -> Vec<u8> {
+    let mut fp = File::open(&path).expect(&format!("Can not open {}", path));
+
```
SergioBenitez: Kill this empty line. Move the `file_content` declaration to the top of the function.

examples/static_files/src/tests.rs
```diff
@@ -0,0 +1,71 @@
...
+        Body::Chunked(b, _) => {
+            let mut data: Vec<u8> = vec![];
+            b.read(&mut data).expect("Read from response failed");
+            return Some(data);
+        }
+    }
+}
+
+fn test_query_file<T> (path: &str, status: Status, content: T)
+    where T: Into<Option<Vec<u8>>>
```
SergioBenitez: Can't we just make this `Into<Option<&'static str>>` and then pass the filename and do the file reading here?

examples/static_files/src/tests.rs
mike820324: When I change to the above code snippets, the compiler give the following error.

```
error: no method named `read_to_end` found for type `rocket::response::Body<&mut std::io::Read>` in the current scope
  --> src/tests.rs:13:10
   |
13 |     body.read_to_end(&mut data);
   |          ^^^^^^^^^^^

```
Based on the code
- https://github.com/SergioBenitez/Rocket/blob/master/lib/src/response/response.rs#L13 
- https://github.com/SergioBenitez/Rocket/blob/master/lib/src/response/response.rs#L527

```Body``` type is a enum with generic T where T implement the ```std::io::Read``` trait.
And inside ```Response```, the T type is a ```Box<std::io::Read>``` which is a trait object.

The question is, is there any better way to get the content of enum without using match.

mike820324:
Thanks for the review, I have made several changes based on your suggestions. One thing that I haven't changed is the body_to_bytes that do not use matches.

examples/static_files/src/tests.rs
SergioBenitez: Sorry, I was mistaken. `Body` doesn't implement `Read`. I've added an `into_bytes` method for `Body` that does exactly this.

mike820324: Thanks for the fix, I'll change this part right away.

mike820324:
@SergioBenitez I have used the ```into_bytes``` method. Thanks.

SergioBenitez:
Excellent! This is really nice and clean. :)

SergioBenitez:
Merged in f0836e22fa87610eeafa934311cd879ccda2fd6d. :)

PR #204, opened 2017-02-21
Add `get_` methods to config

impowski:
Added two methods for `Config` to retrieve table and array from `extras`.

### DIFF
--- a/lib/src/config/config.rs
+++ b/lib/src/config/config.rs
@@ -550,6 +550,64 @@ impl Config {
         parse!(self, name, value, as_float, "a float")
     }
 
+    /// Attempts to retrieve the extra named `name` as an array.
+    ///
+    /// # Errors
+    ///
+    /// If an extra with `name` doesn't exist, returns an `Err` of `NotFound`.
+    /// If an extra with `name` _does_ exist but is not an array, returns a
+    /// `BadType` error.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// use std::collections::BTreeMap;
+    /// use rocket::config::{Config, Environment};
+    ///
+    /// let mut table = BTreeMap::new();
+    /// table.insert("my_value".to_string(), 1);
+    ///
+    /// let my_array = vec![table];
+    ///
+    /// let config = Config::build(Environment::Staging)
+    ///     .extra("my_array", my_array)
+    ///     .unwrap();
+    ///
+    /// assert!(config.get_slice("my_array").is_ok());
+    /// ```
+    pub fn get_slice(&self, name: &str) -> config::Result<&[Value]> {
+        let value = self.extras.get(name).ok_or_else(|| ConfigError::NotFound)?;
+        parse!(self, name, value, as_slice, "a slice")
+    }
+
+    /// Attempts to retrieve the extra named `name` as a table.
+    ///
+    /// # Errors
+    ///
+    /// If an extra with `name` doesn't exist, returns an `Err` of `NotFound`.
+    /// If an extra with `name` _does_ exist but is not a table, returns a
+    /// `BadType` error.
+    ///
+    /// # Example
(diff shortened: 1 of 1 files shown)


lib/src/config/config.rs
```diff
@@ -550,6 +550,64 @@ impl Config {
...
+    ///
+    /// let my_array = vec![table];
+    ///
+    /// let config = Config::build(Environment::Staging)
+    ///     .extra("my_array", my_array)
+    ///     .unwrap();
+    ///
+    /// assert!(config.get_array("my_array").is_ok());
+    /// ```
+    pub fn get_array<'a>(&'a self, name: &'a str) -> config::Result<&'a [Value]> {
```
SergioBenitez: This should return a `config::Array`.

lib/src/config/config.rs
```diff
@@ -550,6 +550,64 @@ impl Config {
...
+    /// let mut my_table = BTreeMap::new();
+    /// my_table.insert("my_value".to_string(), 1);
+    /// 
+    /// let config = Config::build(Environment::Staging)
+    ///     .extra("my_table", my_table)
+    ///     .unwrap();
+    ///
+    /// assert!(config.get_table("my_table").is_ok());
+    /// ```
+    pub fn get_table<'a>(&'a self, name: &'a str) -> config::Result<&'a BTreeMap<String, Value>> {
```
SergioBenitez: This should return a `config::Table`.

SergioBenitez:
Also, the lifetimes are incorrect. You shouldn't specify the same lifetime for `self` and `name`. I think specifying no lifetimes at all and letting Rust infer them will work here and will be the right thing.

lib/src/config/config.rs
```diff
@@ -550,6 +550,64 @@ impl Config {
...
+    ///
+    /// let my_array = vec![table];
+    ///
+    /// let config = Config::build(Environment::Staging)
+    ///     .extra("my_array", my_array)
+    ///     .unwrap();
+    ///
+    /// assert!(config.get_slice("my_array").is_ok());
+    /// ```
+    pub fn get_slice<'a>(&'a self, name: &str) -> config::Result<&'a [Value]> {
```
SergioBenitez: Do we actually need the `'a`? I think Rust will correctly infer it.

lib/src/config/config.rs
```diff
@@ -550,6 +550,64 @@ impl Config {
...
+    /// let mut my_table = BTreeMap::new();
+    /// my_table.insert("my_value".to_string(), 1);
+    /// 
+    /// let config = Config::build(Environment::Staging)
+    ///     .extra("my_table", my_table)
+    ///     .unwrap();
+    ///
+    /// assert!(config.get_table("my_table").is_ok());
+    /// ```
+    pub fn get_table<'a>(&'a self, name: &str) -> config::Result<&'a config::Table> {
```
SergioBenitez: Same here.

SergioBenitez:
Waiting on some changes here.

lib/src/config/config.rs
```diff
@@ -550,6 +550,64 @@ impl Config {
...
+    /// use std::collections::BTreeMap;
+    /// use rocket::config::{Config, Environment};
+    ///
+    /// let mut table = BTreeMap::new();
+    /// table.insert("my_value".to_string(), 1);
+    ///
+    /// let my_array = vec![table];
+    ///
+    /// let config = Config::build(Environment::Staging)
+    ///     .extra("my_array", my_array)
```
SergioBenitez: Can we make `my_array` a simpler vector? Maybe just `vec![1, 2, 3]` or something?

SergioBenitez:
A made the requested changes and merged in c465109fb43c15f56e6c2585382a4dc58b3ca480. Thanks again!

PR #208, opened 2017-02-22
Moving `toml` to a newer version

impowski:
Right now everything works just like that, and I'm thinking about implementing it a `serde` way with `Deserialize` and everything else, to be more idiomatic and to drop such functions as `set`, because they look really ugly.

Note for @SergioBenitez : I want this PR to be merged before #204 , because I'm planning to change methods inside of `Config`

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -22,8 +22,10 @@ term-painter = "0.2"
 log = "0.3"
 url = "1"
 hyper = { version = "0.10.4", default-features = false }
-toml = { version = "0.2", default-features = false }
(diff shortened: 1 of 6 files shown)


SergioBenitez:
What's the advantage of this? It looks like it introduces more code. What are we winning?

SergioBenitez:
Ping!

impowski:
I'll take a little bit time to check wether I able to simplify it or not.

SergioBenitez:
@impowski Can you state why we should upgrade?

SergioBenitez:
Closing this out.

PR #210, opened 2017-02-24
Port auto config

anderejd:
Hi,

This PR is not ready to merge yet, I just wanted to throw the code up on github to get some initial feedback before putting more work into this. 

Overview:
- [ ] Find consensus and approval
- [x] Write a new example hello_auto_port.
- [ ] Write a test for hello_auto_port example.
- [X] Do not break the public API (existing examples and tests work without modification)
- [ ] Write documentation
- [ ] Use rustfmt

This is just a rough initial attempt at providing auto configuration of the port. My motivation is that this is a feature I need for a project where I want to use Rocket as backend for a desktop application with an embedded browser as GUI and I would like to let Rocket allocate an available port and then pass the resulting address on to the browser.

Please feel free to close if this is totally out of scope.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ members = [
   "examples/errors",
   "examples/extended_validation",
   "examples/forms",
+  "examples/hello_auto_port",
   "examples/hello_person",
   "examples/query_params",
   "examples/hello_world",
--- a/examples/hello_auto_port/Cargo.toml
+++ b/examples/hello_auto_port/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "hello_auto_port"
+version = "0.0.1"
+workspace = "../../"
+
+[dependencies]
+rocket = { path = "../../lib" }
+rocket_codegen = { path = "../../codegen" }
+
+[dev-dependencies]
+rocket = { path = "../../lib", features = ["testing"] }
--- a/examples/hello_auto_port/src/main.rs
+++ b/examples/hello_auto_port/src/main.rs
@@ -1,0 +1,34 @@
+#![feature(plugin)]
+#![plugin(rocket_codegen)]
+
+extern crate rocket;
+
+use rocket::config::{Config, Environment};
+use rocket::config::ConfigError;
+use rocket::logger::LoggingLevel;
+
+#[get("/")]
+fn index() -> &'static str {
+    "Hello, world!"
+}
+
+fn main() {
+    try_config().unwrap()
+}
+
+#[allow(dead_code)]
+fn try_config() -> Result<(), ConfigError> {
+    let config = Config::build(Environment::Development)
+        .address("127.0.0.1")
+        .port(0) // 0 will request that the OS assigns a port.
+        .log_level(LoggingLevel::Debug)
+        .finalize()?;
+    let logging = true;
+    let on_success = |addr| {
+            println!("Resulting address: {}", addr);
+    };
+    rocket::custom(config, logging)
+        .mount("/", routes![index])
+        .launch(on_success);
+    Ok(())
+}
--- a/examples/hello_auto_port/src/main.rs
+++ b/examples/hello_auto_port/src/main.rs
@@ -24,11 +24,12 @@ fn try_config() -> Result<(), ConfigError> {
         .log_level(LoggingLevel::Debug)
         .finalize()?;
     let logging = true;
-    let on_success = |addr| {
+    let callback = Box::new(|addr| {
             println!("Resulting address: {}", addr);
-    };
+    });
     rocket::custom(config, logging)
         .mount("/", routes![index])
-        .launch(on_success);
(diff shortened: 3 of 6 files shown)


SergioBenitez:
There seem to be two things going on here:

  1. Using a port of `0` means Rocket randomly selects and unused port.
  2. Rocket issues a callback once it has bound to an address/port.

I'm totally okay with **1**, but I'm not convinced about the utility of **2**. I would accept a well constructed PR implementing just **1**.

anderejd:
Thanks for the quick feedback!

> Edit: Actually, does Hyper already do this for port 0?

Yes, hyper does this by in turn using std::net::TcpListener, docs: https://doc.rust-lang.org/std/net/struct.TcpListener.html#method.bind
In the new example hello_auto_port I'm using port 0 for this reason.

> There seem to be two things going on here:
>
>   
>   1. Using a port of 0 means Rocket randomly selects an unused port.
>   2. Rocket issues a callback once it has bound to an address/port.
>
>I'm totally okay with 1, but I'm not convinced about the utility of 2. I would accept a well constructed PR implementing just 1.

I can spend the time needed to polish this PR, no problem.
But for my use case I need to in some way get notified which port was allocated. How would you suggest to solve that notification in a better way?

anderejd:
I removed the callback parameter to Rocket::launch and moved it to Rocket::on_success to extend the existing builder pattern and allow existing examples to build and test ok.

Does the latest commit https://github.com/SergioBenitez/Rocket/pull/210/commits/b8c6ab0e48791762c66e327026757c812df9feba  look more promising?

mehcode:
To throw another idea in, perhaps `launch_and`? I agree we shouldn't modify the existing `launch` method.

anderejd:
@mehcode Ah interesting. Would you prefer .launch_and or the .on_success builder chain?

SergioBenitez:
Can you share why you need to know the port after launching?

mehcode:
@SergioBenitez Short term, I imagine @anderejd simply wants to use https://github.com/amodm/webbrowser-rs to auto-open the web browser when the server is run in development mode.

Long term, I'd imagine something like [godo](https://github.com/go-godo/godo) for Rust. This type of success callback would be important for a task runner.

---

@anderejd I don't have a strong preference either way but if I had to choose I'd want `.launch_and` because it feels _sort of_ like `.launch().and` without needing the `.launch` method to return.

anderejd:
@mehcode That is precisely what I have in mind. :) Short term.

@SergioBenitez The next step in my project will be to launch a new CEF process (or maybe a Servo process when that project matures) to display a HTML based GUI that connects to the Rocket server on the automatically allocated port. The task runner example by @mehcode is also interesting to me. But why not simply specify a port in a config? The motivation for that is that I want to allow multiple instances of the application to be able to start and run correctly side by side. The specific port number is not important in this case when running both server and client on localhost, but it is important to find and use a port that is available.

I could also imagine some micro service situation where multiple versions and or multiple instances of each service could be overlapping and in flight at any time where each service need to register itself to some manager service and report address, port, version etc. before new requests can be sent to the new service and the old service initiating shutdown.

SergioBenitez:
I'm quite wary of adding another `launch` method to Rocket. It can be very confusing to new users: "Which `launch` am I supposed to use?".

Rocket needs a mechanism to perform actions just before or after launching the server. This mechanism is important for extensions such as `Template` and the upcoming `Database` contrib package. I'd like for these extensions to be able to perform sanity checks _before_ they're actually used. As a concrete, existing example, consider `Template`: at present, Rocket doesn't emit errors about invalid templates until a template is rendered. It would be nice if, instead, you received these errors before the server even starts serving requests. This type of thing is even more important when it comes to databases: it would be a bummer to discover that you've entered the wrong URL only after you try to execute a SQL statement.

At the moment, my preferred mechanism is via a `#[rocket_init]` annotation. Rocket would ensure that every function marked with this annotation is called at the appropriate time. In your particular example, we might imagine that it looks something like:

```rust
#[rocket_init]
fn start_cef_process(rocket: &Rocket) {
    spawn_cef(rocket.config().address, rocket.config().port);
}
```

We can also imagine having different forms of this init annotation for different stages of the launch process. For instance, we might have `#[rocket(pre_launch)]` and `#[rocket(post_launch)]` init function which execute just before the server starts and just after the server starts, respectively.

I believe this solves the problem in an elegant manner. Unfortunately, I can't think of a clean way to do this without a new feature landing in Rust (https://github.com/rust-lang/rust/pull/39987). And of course, there are a lot of questions to answer: what is the running environment of these init functions? What are the parameters to these init functions? What capabilities do the functions have? Which data does the function have access to? Which of that can it mutate? And so on.

anderejd:
Ok, thanks to both of you @SergioBenitez, @mehcode for your feedback and time. I will keep an eye on Rocket and try Iron in the short term. Please feel free to close this PR if it is unlikely to merge.

PR #211, opened 2017-02-25
Add UrlDecoder trait for str.

### LINKED ISSUE #43: The usage of the `&str` type is potentially misleading.
## Steps to reproduce:

- Have a route with the following code:

    ```rust
       #[derive(FromForm)]
    pub struct FormData<'a> {
        form_data: &'a str,
    }

    #[post("/reproduce_bug", data = "<form>")]
    fn reproduce_bug<'a>(form: Form<'a, FormData<'a>>) -> &'static str {
        assert_eq!("A value with spaces", form.get().form_data);
        "Everything is fine"
    }
    ````
 
- Create a template with the following HTML, and submit it using a browser

    ```html
    <form action="reproduce_bug" method="post">
        <input type="text" name="form_data" value="A value with spaces" />
        <input type="submit" value="Submit" />
    </form>
    ````

## Expected behavior

The response "Everything is fine" is rendered

## Actual behavior

The value of the field is "A+value+with+spaces", causing the assertion to fail and a 500 to be returned.

## Additional details

This only applies to values of `&str` and not `String`, so I'm assuming that `&str` is unconditionally pointing at the raw request data to avoid any allocation. Even if this behavior is documented (I couldn't find it mentioned), I think this is a surprising behavior and likely to be a common source of confusion. At absolute minimum this should produce some kind of error, but even that is a sub-optimal solution as it would lead to unexpected gotchas for users who do not test their inputs with spaces.

I'd expect the framework to be allocating internally if needed to properly handle decoding entities, and pointing at the raw request data if possible. I would never expect the framework to force me to care about encoding or content-type specific details unless I've specifically asked for some form of "raw" data.

I have not checked, but I suspect there are likely similar issues if the request is submitted with a charset other than UTF-8, and for JSON requests that would require additional decoding (such as an escaped backslash character).

mike820324:
This PR try to resolve #43 .

The original proposal is trying to add a new type called ```RawStr``` that act exactly as str type but add a new additional method ```url_decode```.

After playing for a while, I think adding a new trait might be a good idea.

I have added a new trait called ```UrlDecoder``` which currently implement to the ```str``` type.

The following code snippets shows how the user can use this trait and the difference between two return value.
```rust
use rocket::http::uri::UrlDecoder;

#[derive(FromForm)]
pub struct FormData<'a> {
    form_data: &'a str,
}

#[post("/reproduce_bug", data = "<form>")]
fn reproduce_bug<'a>(form: Form<'a, FormData<'a>>) -> &'static str {
    assert_eq!("A+value+with+spaces", form.get().form_data);
    assert_eq!("A value with spaces", form.get().form_data.url_decode().unwrap());
    "Everything is fine"
}
```

Any advices and recommendations are helpful. Thanks.

### DIFF
--- a/lib/src/http/uri.rs
+++ b/lib/src/http/uri.rs
@@ -314,6 +314,20 @@ impl<'a> fmt::Display for URI<'a> {
 
 unsafe impl<'a> Sync for URI<'a> { /* It's safe! */ }
 
+pub trait UrlDecoder<'a> {
+    fn url_decode(&'a self) -> Result<String, &'a str>;
+}
+
+impl<'a> UrlDecoder<'a> for &'a str {
+    fn url_decode(&'a self) -> Result<String, &'a str> {
+        let replaced = self.replace("+", " ");
+        match URI::percent_decode(replaced.as_bytes()) {
+            Err(_) => Err(self),
+            Ok(string) => Ok(string.into_owned())
+        }
+    }
+}
+
 /// Iterator over the segments of an absolute URI path. Skips empty segments.
 ///
 /// ### Examples
--- a/lib/src/request/form/from_form_value.rs
+++ b/lib/src/request/form/from_form_value.rs
@@ -2,7 +2,7 @@ use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddrV4, SocketAddrV6, SocketAdd
 use std::str::FromStr;
 
 use error::Error;
-use http::uri::URI;
+use http::uri::UrlDecoder;
 
 /// Trait to create instance of some type from a form value; expected from field
 /// types in structs deriving `FromForm`.
@@ -167,10 +167,9 @@ impl<'v> FromFormValue<'v> for String {
 
     // This actually parses the value according to the standard.
     fn from_form_value(v: &'v str) -> Result<Self, Self::Error> {
-        let replaced = v.replace("+", " ");
-        match URI::percent_decode(replaced.as_bytes()) {
+        match v.url_decode() {
             Err(_) => Err(v),
-            Ok(string) => Ok(string.into_owned())
+            Ok(string) => Ok(string)
         }
     }
 }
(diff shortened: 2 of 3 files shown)


mehcode:
I'm not a fan of this as it keeps the strange difference between `String` and `&str`.

Wanting the _raw_ string should be an explicit operation.

SergioBenitez:
Indeed, as @mehcode says, the intention behind the explicit `RawStr` type is that it forces the user to ask, "Wait, what's a `RawStr`?" and thus consider the ramifications of handling a raw HTTP string more deliberately.

mike820324:
@mehcode @SergioBenitez Sorry for the late reply, the past few weeks are a bit busy for me.

Indeed, the trait maybe a bit implicit for the user and causing confusion. I'll modify the PR to add a new type.

If I want to add a RawStr type, which file should be the proper place to add it? Thanks.

By the way, one small issue is that
it seems like the decode logic in ```FromFormValue``` and ```FromParam``` are different. 
In ```FromFormValue```, the '+' will be decode into a space while in  ```FromParam``` it will preserve the '+' sign.

https://github.com/SergioBenitez/Rocket/blob/master/lib/src/request/form/from_form_value.rs#L165-L176
https://github.com/SergioBenitez/Rocket/blob/master/lib/src/request/param.rs#L211-L216

SergioBenitez:
`RawStr` would likely live in a `http/raw.rs` (or something like that) and be in the `http` module.

Outside of query parameters and forms, a `+` character should be interpreted literally. As such, the code as it stand is correct.

SergioBenitez:
Let's close this PR and open one up with the `RawStr` implementation, if you're so inclined to give it a shot. If not, it's something I was planning on implementing, so please let me know. :)

mike820324:
No problem at all. 
I"m still very new to this project and rust, I think I'll pass this one and watch how you implement this as a learning experience.

Thanks. :)

SergioBenitez:
You've done a great job thus far!

PR #213, opened 2017-02-26
Removing hashing from router to hopefully increase route lookup effic…

craftytrickster:
I was looking at Rocket because it looks like a really cool and well done library, and I was just poking around.

I happened to notice that the Router is using a hashmap to store its routes. I would suspect that using a hashmap per REST method to perform a route lookup every time would be less efficient then giving concrete values inside of the Router struct. 

I would like to know how you are performing the benchmarks in the README.md, because if so, I can run them on Master and then compare them with the changes here. I suspect that removing the hashmap as I did here would be a tiny bit more efficient.

### DIFF
--- a/lib/src/router/mod.rs
+++ b/lib/src/router/mod.rs
@@ -1,43 +1,71 @@
 mod collider;
 mod route;
 
-use std::collections::hash_map::HashMap;
-
 use self::collider::Collider;
 pub use self::route::Route;
 
 use request::Request;
 use http::Method;
 
-// type Selector = (Method, usize);
-type Selector = Method;
-
 #[derive(Default)]
 pub struct Router {
-    routes: HashMap<Selector, Vec<Route>>, // using 'selector' for now
+    get_routes: Vec<Route>,
+    put_routes: Vec<Route>,
+    post_routes: Vec<Route>,
+    delete_routes: Vec<Route>,
+    options_routes: Vec<Route>,
+    head_routes: Vec<Route>,
+    trace_routes: Vec<Route>,
+    connect_routes: Vec<Route>,
+    patch_routes: Vec<Route>
 }
 
 impl Router {
     pub fn new() -> Router {
-        Router { routes: HashMap::new() }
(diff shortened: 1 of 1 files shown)


SergioBenitez:
There are no benchmarks for routing specifically; it's something I've been meaning to add. The benchmarks in the `README` are done simply by running the `hello_world` example with `ROCKET_ENV=prod cargo run --release` and then calling `wrk` via something like `wrk -t1 -c18 http://localhost:80`; you should adjust `18` so that it is something close to the number of cores on your machine.

Right now we look up potentially matching routes by `Method`, but I think we can separate routes into more effective buckets. For instance, routes with and without query strings can never match the same input, so they can be in distinct buckets. Also, routes with differing segment counts can't collide, so they, too, can be in different buckets. This brings down the cost of search for the correct match but necessitates a `HashMap`.

That being said, the cost to hash `Method` is about zero, so a `HashMap` lookup should cost about the same as your modifications modulo an extra dereference or two and collisions in the hash table. Since we know all of the keys, we should be able to use perfect hashing to get collisions to zero, an approach that will work even with further optimized keys.

In summary, I think we can do the above to get routing to be extremely quick. This PR makes the code less flexible and less clean.  Still, I'd be curious to see benchmarks; I'd be very surprised if this is a big improvement. I would be accepting of PRs that implement what I've detailed above or better, or of PRs to add benchmarks for routing specifically.

craftytrickster:
What you are saying makes sense, so I can close this.

I am not sure if it would make a marked difference, but one thing to maybe consider is storing the route uri segments in a tree, which would allow for efficient route lookups.

Right now in the route method you have the following code:

```rs
routes.iter()
  .filter(|r| r.collides_with(req))
  .collect()
```

That looks like you have to do a full collision check on every single route in that bucket (your idea of more effective buckets mitigate the issue though). With a tree based approach you would only need traverse the segment tree as far as the incoming route's segments would need, and you can detect collisions that way.  Just food for thought.

PR #215, opened 2017-03-01
Add Rocket version to Server header

mehcode:
The information disclosure issue of exposing the Server header to the internet are known and this makes it a touch worse but I feel if a deployment cares about this they are already filtering out this header at a reverse proxy.

This changes increases the debugging value by letting us known which Rocket version is deployed.

---

I'd additionally like to print the Rust version as well but it [seems non-trivial](https://github.com/Kimundi/rustc-version-rs) so I didn't attempt to.

### DIFF
--- a/lib/src/rocket.rs
+++ b/lib/src/rocket.rs
@@ -79,7 +79,7 @@ impl Rocket {
     fn issue_response(&self, mut response: Response, hyp_res: hyper::FreshResponse) {
         // Add the 'rocket' server header, and write out the response.
         // TODO: If removing Hyper, write out `Date` header too.
-        response.set_header(Header::new("Server", "rocket"));
+        response.set_header(Header::new("Server", "Rocket"));
 
         match self.write_response(response, hyp_res) {
             Ok(_) => info_!("{}", Green.paint("Response succeeded.")),


echochamber:
Cool feature, but I'd prefer if it could be toggled on/off when compiling as opposed to requiring a reverse proxy to filter it.

cbrewster:
@vitalyd the env! macro will be replaced with the contents of the environment variable during runtime.
https://doc.rust-lang.org/std/macro.env.html

vitalyd:
@cbrewster - doh, face palm! The rest of my comment stands though right?

SergioBenitez:
I'm not inclined to accept this PR. We're going to make mistakes; letting the world know that software is vulnerable because of those mistakes is a very bad idea. I don't believe leaving it up to others to filter that disclosure is reasonable.

Why does it matter which version of Rocket was used to deploy your application? Ideally we've have some mechanism by which you can add your own headers to every response; would that suffice?

mehcode:
My specific use case is I would like to internally know what versions of Rocket and Rust my various API nodes are using. Before it hits anything public-facing, the load balancer will filter out that header.

There are many other ways I can do this of course. I just thought this would be a simple one.

---

> I'm not inclined to accept this PR. We're going to make mistakes; letting the world know that software is vulnerable because of those mistakes is a very bad idea [...]

And the current value of the `Server` header ( `rocket` ) does that as well. Not as severely, sure. But still enough that the best practice would be to not have it at all in a public facing application.

---

Here are some options (as I see it):

1. Accept a version of this PR that statically creates (using a build.rs script to inject an env variable during compilation seems the most fun to me as it'd be much easier to get the rustc version) a `rocket_version` string and passes it as the `Server` header.

2. 1 _and_ a flag in config to disable the server header. Perhaps `secure` or a better name. Default this to `true` if the environment is `production`.

3. Remove the existing `Server` header.

SergioBenitez:
> And the current value of the Server header (`rocket`) does that as well. Not as severely, sure.

There is an incredible, measurable, and objective difference between saying which server you're using and which _version_ of that server you're using. Attackers want to maximize the value of their attacks. If they know certain versions of some server software are vulnerable, and the server is advertising a vulnerable version, that server will be a primary target.

> 3. Remove the existing Server header.

How would that give you what you're looking for? Did you consider my proposal that this may be solved by a future change?

mehcode:
> How would that give you what you're looking for? 

It wouldn't. But we're talking about the `Server` header being "bad". So perhaps that should be fixed if we agree on that.

---

> Also, I'm still not clear on why the version of Rocket you're running matters. What are you doing with that information? Why does it have to be on every response from the server?

To be honest.. I'm just very used to this being a thing. Nearly all the stuff I've played with in the past does it.

```
Server: Werkzeug/0.9.6 Python/2.7.6
```

---

Really I opened this because `Server: rocket` was really bothering me. `Server: Rocket` would make me feel better. I just added the version number because I thought, hey, I've seen this everywhere before so maybe it might be appreciated.

SergioBenitez:
> It wouldn't. But we're talking about the Server header being "bad". So perhaps that should be fixed if we agree on that.

The `Server` header isn't intrinsically bad, but disclosing information in the header that gives attackers an advantage is.

I'm okay with capitalizing the `r`. :D

mehcode:
>  Did you consider my proposal that this may be solved by a future change?

```rust
#[rocket(before_response)]
fn before_response(response: &mut Response) {
  response.set_header("Server", "Whatever/I/Want");
}
```

That feels sort of fun.

---

Anyway, changed this to just fixing my initial issue with the header. :crab:

shssoichiro:
> To be honest.. I'm just very used to this being a thing. Nearly all the stuff I've played with in the past does it.

Which is odd, because this is very widely regarded as a bad practice. But you are correct, it does seem to be common.

I think if this feature were to go in at some point, it should be off by default in production mode.

mehcode:
@shssoichiro ( and others )

> I think if this feature were to go in at some point, it should be off by default in production mode.

How do you feel about the currently existing Server header of `Server: rocket` ? Should that be off by default in production mode?

shssoichiro:
> How do you feel about the currently existing Server header of Server: rocket ? Should that be off by default in production mode?

I'm not a fan of having any `Server` header in production mode, although I feel less strongly about it if the version isn't shown. An attacker would have to find a 0-day in the latest version of rocket, as opposed to exploiting a vulnerability that may be known from an older version, if the server isn't running the latest version of rocket.

SergioBenitez:
The `r` has been capitalized in 4f704e95f2987cf365ef798c03c6a47d67c086a2!

PR #216, opened 2017-03-03
Add some emoji support in errors

Manishearth:
r? @SergioBenitez


SergioBenitez:
This PR simply doesn't live up to its name: there aren't enough emojis! Come back when you have more, @Manishearth. 🐰

Manishearth:
But it now has emojis in all the errors? :smile:

PR #221, opened 2017-03-09
Updated catcher example so that it returns Status::NotFound

alanstoate:
Noticed that this would return Status::Ok like the manual_routes example did


SergioBenitez:
Merged in 9d10aa23295377bf8c4d4638864f9b2acfdcd402. Thanks so much. :)

Issue #223, opened 2017-03-11
A list of headers is represented as a single comma separated value


scottschroeder:
### Description of Issue
I would expect that when multiple headers are passed in with the same key, the values are interpreted as a list.

So for a query like:
``` bash
curl -H "friend: adam" -H "friend: bob" localhost:8000
```
I would expect a list with two entries
``` bash
{"friends": ["adam", "bob"]}
```
But instead, I get a list with one entry, separated by a `, `
``` bash
{"friends": ["adam, bob"]}
```

This can get even worse if our original header has a `, ` inside one of the values
``` bash
curl -H "friend: adam" -H "friend: bob, carol" localhost:8000
{"friends": ["adam, bob, carol"]}
```

This is on the latest version of Rocket, both v0.2.2 and master branch as of this writing.

### Root Cause

When Rocket grabs the headers from Hyper, we use the `value_string()`
```
for hyp in h_headers.iter() {
    let header = Header::new(hyp.name().to_string(), hyp.value_string());
    request.add_header(header);
}
```
See [request/request.rs](https://github.com/SergioBenitez/Rocket/blob/16cb7297abeef3af03666247e2fbbf6d9aae683e/lib/src/request/request.rs#L481-L485) for exact listing.

According to the [Hyper Documentation](https://hyper.rs/hyper/v0.10.5/hyper/header/struct.HeaderView.html#method.value_string): 
> `fn value_string(&self) -> String`
> Get just the header value as a String.
> This will join multiple values of this header with a `,`.
> Warning: This may not be the format that should be used to send a Request or Response.

It seems like this is the incorrect interface to use for grabbing header values from Hyper.

SergioBenitez:
Fixed in https://github.com/SergioBenitez/Rocket/commit/47fe659ebe64e4079105798aa8850dc952c1bdd9.

PR #224, opened 2017-03-11
Preserve header list using Hyper's raw headers interface

scottschroeder:
### Changes
The hyper `value_string()` will take a list of headers and return a string which
is joined by commas. See Issue #223.

This change grabs the raw headers and loops over them, creating the proper
number of entries in the header map.

### Testing
* I tested this using a basic endpoint which prints out headers, and threw various combinations of headers at it with `curl`.
* I tried to pass in invalid utf8 to exercise the unhappy path, but it seemed like something at a lower level panic'd before this code even got run.
* `cargo test` passes.

### Thoughts
I also tried to fix this by implementing Hyper's [Header](https://hyper.rs/hyper/v0.10.5/hyper/header/trait.Header.html) and [Header Format](https://hyper.rs/hyper/v0.10.5/hyper/header/trait.HeaderFormat.html) traits on a simple struct wrapping a `Vec<String>` so that I could then make the direct call to `hyp.value()` which seemed recommended from the [docs](https://hyper.rs/hyper/v0.10.5/hyper/header/struct.HeaderView.html#method.value). This ended up being a lot more code, and I couldn't get it to work without introducing another allocation per header. I'm not well versed with Hyper and maybe someone more experienced than I can come up with a better solution.

### DIFF
--- a/lib/src/request/mod.rs
+++ b/lib/src/request/mod.rs
@@ -15,3 +15,97 @@ pub use self::state::State;
 /// Type alias to retrieve [Flash](/rocket/response/struct.Flash.html) messages
 /// from a request.
 pub type FlashMessage = ::response::Flash<()>;
+
+#[cfg(test)]
+mod tests {
+    /// These tests are related to Issue#223
+    /// The way we were getting the headers from hyper
+    /// was causing a list to come back as a comma separated
+    /// list of entries.
+
+    use super::Request;
+    use super::super::http::hyper::header::Headers;
+    use hyper::method::Method;
+    use hyper::uri::RequestUri;
+    use std::net::{IpAddr, Ipv4Addr, SocketAddr};
+    use std::collections::HashMap;
+
+    fn check_headers(test_headers: HashMap<String, Vec<String>>) {
+        let h_method: Method = Method::Get;
+        let h_uri: RequestUri = RequestUri::AbsolutePath("/test".to_string());
+        let h_addr: SocketAddr = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 8000);
+        let mut h_headers: Headers = Headers::new();
+
+        for (key, values) in &test_headers {
+            let raw_bytes: Vec<Vec<u8>> = values
+                .iter()
+                .map(|v| v.clone().into_bytes())
+                .collect();
+            h_headers.set_raw(key.clone(), raw_bytes);
+        }
+
+        let req = Request::from_hyp(h_method, h_headers, h_uri, h_addr).unwrap();
+        let r_headers = req.headers();
+
+        for (key, values) in &test_headers {
+            for (v1, v2) in values.iter().zip(r_headers.get(&key)) {
+                assert_eq!(v1, v2)
+            }
+        }
+    }
+
+    #[test]
+    fn test_single_header_single_entry() {
+        let mut test_headers = HashMap::new();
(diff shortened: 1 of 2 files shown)


SergioBenitez:
Thanks so much for finding, investigating, and proposing a solution for this!

The code looks good, but, can you add a test case to `lib/tests` named something like `multiple-headers-issue-223.rs` that verifies the behavior? See the other tests in the directory for examples.

scottschroeder:
### What I tried (and failed)
I did the best I could get tests into `lib/tests/` but ran into a few issues. It seems like most of the tests there use the `MockRequest`, which is unusable for this purpose, because the `add_header` method totally bypasses the actual work done in [from_hyp](https://github.com/SergioBenitez/Rocket/blob/5434f467a9ca77c85f821ac6eb54997fb7ccd9fa/lib/src/request/request.rs#L438). The function `from_hyp` is also private to the crate, so the tests I was running in `lib/tests/` didn't have the visibility to run it directly.

I experimented with trying to use the `hyper::Handler` trait which is what calls [this code path](https://github.com/SergioBenitez/Rocket/blob/4f704e95f2987cf365ef798c03c6a47d67c086a2/lib/src/rocket.rs#L51). That looked like I was going to have to go _really_ low level, and feed hyper raw HTTP just to get the header parsing and extraction logic exercised.

### What I actually did
I opted to create the `request::tests` module, and place my test cases there. I'm open to move things around if you have any better ideas, but this seemed like a reasonable compromise. I also think it would be a good place to exercise other parts of the hyper extraction if needed.

### Test output
As far as the actual tests go, I used whats currently in master and got expected failures:
```
test request::tests::test_multiple_headers_multiple_entries ... FAILED
test request::tests::test_multiple_headers_single_entry ... ok
test request::tests::test_single_header_comma_entry ... FAILED
test request::tests::test_single_header_multiple_entries ... FAILED
test request::tests::test_single_header_single_entry ... ok

failures:

---- request::tests::test_multiple_headers_multiple_entries stdout ----
        thread 'request::tests::test_multiple_headers_multiple_entries' panicked at 'assertion failed: `(left == right)` (left: `"alice"`, right: `"alice, bob"`)', /home/scott/Development/Rocket/lib/src/request/mod.rs:52
note: Run with `RUST_BACKTRACE=1` for a backtrace.

---- request::tests::test_single_header_comma_entry stdout ----
        thread 'request::tests::test_single_header_comma_entry' panicked at 'assertion failed: `(left == right)` (left: `"alice"`, right: `"alice, bob, carol"`)', /home/scott/Development/Rocket/lib/src/request/mod.rs:52

---- request::tests::test_single_header_multiple_entries stdout ----
        thread 'request::tests::test_single_header_multiple_entries' panicked at 'assertion failed: `(left == right)` (left: `"alice"`, right: `"alice, bob"`)', /home/scott/Development/Rocket/lib/src/request/mod.rs:52
```
After applying my patch, its all ✔️.  I _may_ have gone overboard on the tests. The issue I brought up could be tested with just `test_single_header_multiple_entries`, but I felt that these other tests might catch some common issues if this code ever gets refactored.

SergioBenitez:
There were quite a few changes that I wanted made, but unfortunately I didn't have time to give a thorough review, so I went ahead and merged and then made the changes myself.

Merged in https://github.com/SergioBenitez/Rocket/commit/47fe659ebe64e4079105798aa8850dc952c1bdd9. Thanks so much for this! You did an awesome job reporting and fixing this. :)

Issue #22, opened 2016-12-23
JSON is not idiomatic


steveklabnik:
This is your project, so feel free to close.

According to [RFC 430](https://github.com/rust-lang/rfcs/blob/master/text/0430-finalizing-naming-conventions.md#fine-points), it should be `Json`, not `JSON`.

🚀

SkylerLipthay:
Same goes for `APIKey` -> `ApiKey` (non-breaking, because it's just documentation-level).

SergioBenitez:
Indeed, this is by design. There are several other names that don't conform, but I prefer to keep upper-case names for acronyms. If enough people disagree, happy to change it.

LLBlumire:
I'll chime in and say Rocket should conform to what of the rust style guide is defined. So, `Json` and `ApiKey`. This is more intuitive for library users as it's what people are (or should be) used to for rust. 

Same goes for code layout, a cargo fmt goes a long way towards readability (except when it doesn't and screw up royally)

ereichert:
I prefer upper-case names but understand why it would bother someone wanting to contribute.

ghazkarim:
I agree with LLBlumire. Stick to the Rust guidelines for the sake of consistency in the Rust ecosystem.

mgattozzi:
Especially since this is just such a slick framework. It really had potential and I think if people want to use it then making it more idiomatic will be beneficial

kerberosmansour:
@SergioBenitez bravo and building this framework in rust, and looking forward to getting Rocket on to  Rust asynchronous I/O libraries once they stabilize.

reillysiemens:
I second @kerberosmansour's kudos and @LLBlumire's recommendation. Consistency goes a long way (especially for new folks).

cbruun:
Quite a few people have expressed support for following the RFC guideline. Is more feedback needed?

impowski:
I agree, `Json` and `ApiKey`.

tshepang:
It will be more hard to change the names as more people use the project. I imagine there are more pressing things, but this doesn't seem like something that'll take much effort.

SergioBenitez:
Okay, let's do this! 🚀 

Slating for 0.4. In all likelihood, we'll make this happen by 0.3.

drusellers:
whoot

ZelphirKaltstahl:
[...]

PR #226, opened 2017-03-12
Rename APIKey to ApiKey

Rukenshia:
As discussed in #22.


SergioBenitez:
There was brief experiment with rewriting all of history that has now concluded. GitHub seems to have barfed here. Can you please open a new PR with your changes if you'd still like them reviewed? Thanks!

Issue #222, opened 2017-03-10
Error handlers are hard to blend in with the rest of the site


LaylConway:
From the docs:
> Unlike request handlers, error handlers can only take 0, 1, or 2 parameters of types Request and/or Error. 

This is problematic if your site includes for example a navigation bar with a signed in section. These kinds of sections in the site's layout shared between all pages still often need to touch the DB (or cached data) and cookies. Currently the error handler doesn't allow this being done easily. I feel like it would be best to treat error handlers as any other handler, except that they can also take Request and Error as parameters.

An alternative solution is to use a redirect, but that just works around the underlying issue.

SergioBenitez:
Error catchers cannot fail. They are meant to _catch_, not _handle_ errors. That is, they are a sink for errors that aren't handled elsewhere. If you need custom error handling, you can use forwarding and/or `Result` return types to handle this.

echochamber:
@SergioBenitez I think this confusion could be avoided in the future by a docs improvement. The fact that the section on Error Catchers in the docs does not link to the section on Errors (and vice versa) is what I think lead to this confusion.

[From the docs on Error Catchers](https://rocket.rs/guide/requests/#error-catchers)

> When Rocket wants to return an error page to the client, Rocket invokes the catcher for that error

Is not entirely clear on what "wants to return an error page to the client" means from this page alone (the Errors doc page clarifies this). I think a link the Errors section or a short sentence elaborating on what "wants to return an error page to the client" means would make the docs easier to understand and tie the related concepts together.

Thoughts? If :+1:  I can open a PR for it later today.

LaylConway:
I see the reasoning behind this now, but what should I do in the case of 404 errors? Forwarding feels a bit like a hack and I don't know if it's going to have an effect on SEO/accessibility/etc.

SergioBenitez:
@echochamber You're totally right. The docs aren't in any repository yet, but feel free to write something up and submit it as a PR here; it'll force me to finally put the docs up somewhere.

@LaylConway Most 404 pages don't make use of anything in the request except maybe the URL. As a result, error catchers should work well there. Nonetheless, you can invoke request guards directly from the error handler if needed. See, for instance, [how to invoke the `State` request guard](https://rocket.rs/guide/state/#retrieving-state).

[...]

PR #229, opened 2017-03-16
Update error and error catcher docs.

echochamber:
From comment here: https://github.com/SergioBenitez/Rocket/issues/222#issuecomment-285824274

I did two commits since the file didn't previously exists so you could still have a diff view of before and after. The markdown is my best guess at what the actual source for the docs looks like.

### DIFF
--- a/docs/error_catchers.md
+++ b/docs/error_catchers.md
@@ -1,7 +1,11 @@
 # Error Catchers
 
-When Rocket wants to return an error page to the client, Rocket invokes the
-catcher for that error. A catcher is like a route, except it only handles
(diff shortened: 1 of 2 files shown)


theduke:
This is great info, should be merged. (@SergioBenitez)

SergioBenitez:
The guide source is now [in the repository](https://github.com/SergioBenitez/Rocket/tree/master/site/guide). Can you make your changes directly to the guide? I'll close this in the meantime. Thanks so much. :)

echochamber:
@SergioBenitez Sounds good. Will do.

Issue #242, opened 2017-03-26
Add support for non-strict FromForm decoding (ignore extra fields)


naufraghi:
I was transitioning from a `json` api to a form encoded one.

The first thing I did was to replace `JSON` with `FromForm` in the route signature, but I discovered `FromForm` is strict and fails if the request has more field than the struct.

I found this a "surprise" element. In a real life example, developing a Slack Outgoing Webhook endpoint, [Slack documents](https://api.slack.com/custom-integrations/outgoing-webhooks) a set of fields, but in fact emits the `service_id` too.

flosse:
How does your struct look like?

naufraghi:
@flosse here it is:

```rust
#[derive(FromForm)]
struct SlackOutgoing {
    token: String,
    service_id: String,
    team_id: String,
    team_domain: String,
    channel_id: String,
    channel_name: String,
    timestamp: String,
    user_id: String,
    user_name: String,
    text: String,
    trigger_word: Option<String>,
}
```
I currently only need `token`, `user_id` and `text`. If I could I preferred parsing only those fields and be resistant to any API change not effecting the subset I'm using.

flosse:
you could all optional fields ...well wrap into `Option` ;-)

```rust
#[derive(FromForm)]
struct SlackOutgoing {
    token: String,
    service_id:  Option<String>,
    team_id:  Option<String>,
    team_domain:  Option<String>,
    channel_id:  Option<String>,
    channel_name:  Option<String>,
    timestamp:  Option<String>,
    user_id: String,
    user_name:  Option<String>,
    text: String,
    trigger_word: Option<String>,
}
```

naufraghi:
Yes, it can be an _option_ :)

There are a few drawbacks I can see:

- the route will fail if Slack will add a new form field I have no way to `Option`,
- I lose some 1-1 mapping with JSON, where an absent field and a missing field are different (syntactically at least, semantically it depends),
- I need to write a lot :)

SergioBenitez:
Copying my comment from #247 to get some wider feedback:

I see three possibilities here.

  1. The standard behavior could be to allow extra fields. This was actually the initial implementation.

  2. We can allow an attribute to specify the behavior. This would look like:

    #[derive(FromForm)]
    #[form(lenient)]
    struct MyStruct { ... }

  3. We can open an RFC with the Rust team to allow parameters on derives. Something like:

    #[derive(FromForm(lenient))]
    struct MyStruct { ... }

[...]

PR #246, opened 2017-04-01
Relax FormItems iterator slicing rules a little.

crlf0710:
Hi, i find that the current matching rules for query strings are a little too strict. It seems that it rejects everything that's not in `key=[value]` form. However this is essentially not mandatory.

So i relaxed this rule a little, allowing the `value` form. It will turn query strings like "?a&b&c" into [("", "a"), ("", "b"), ("", "c")]), which gives user the chance to accept, ignore or reject these data in the FromForm trait.

This may work together with the solution of #242 for user to ignore rubbish or uninteresting data received without failing the request.

### DIFF
--- a/lib/src/request/form/form_items.rs
+++ b/lib/src/request/form/form_items.rs
@@ -212,23 +212,43 @@ impl<'f> Iterator for FormItems<'f> {
 
     fn next(&mut self) -> Option<Self::Item> {
         let s = &self.string[self.next_index..];
-        let (key, rest) = match memchr2(b'=', b'&', s.as_bytes()) {
-            Some(i) if s.as_bytes()[i] == b'=' => (&s[..i], &s[(i + 1)..]),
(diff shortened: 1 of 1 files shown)


SergioBenitez:
I think we should generally allow query parameters without keys, but I don't like the implementation here.  Returning an empty `key` doesn't feel right. We should probably return an (Option<&str>, &str) instead. Or maybe a `QueryParam` enum with two variants, one for keys and values, and another for just values. Or something else.

Also, there's a lot of duplicate code in the current implementation. We'd need to reduce that to get this in.

crlf0710:
@SergioBenitez yeah i agree. Both of the two approaches are great. Glad to see that you also think that query parameters without keys should be supported, because it does happen at times. I'm looking forward to see such support coming in Rocket. Maybe i'll try PR again if i can find time to implement either of them, but i'm closing this now. Thanks!

PR #247, opened 2017-04-01
[POC] Allow ignoring unused items in query string.

### LINKED ISSUE #242: Add support for non-strict FromForm decoding (ignore extra fields)

crlf0710:
Hi, i'd like something like this to resolve #242, so that the actual struct that implements FromForm trait doesn't have to include everything the request might have. It also works together with #246 to ensure that something without a key doesn't fail the whole request.

The current actual names and identitiers used in this patch is awful. Just put it here to hear what you think.

### DIFF
--- a/codegen/src/decorators/derive_form.rs
+++ b/codegen/src/decorators/derive_form.rs
@@ -49,8 +49,37 @@ fn get_struct_lifetime(ecx: &mut ExtCtxt, item: &Annotatable, span: Span)
     }
 }
 
-// TODO: Use proper logging to emit the error messages.
+trait IgnoreExtraFieldStrategy {
+    fn shall_ignore() -> bool;
+}
+
+struct IgnoreExtraField;
+impl IgnoreExtraFieldStrategy for IgnoreExtraField {
+    fn shall_ignore() -> bool {
+        true
+    }
+}
+
+struct ProhibitExtraField;
+impl IgnoreExtraFieldStrategy for ProhibitExtraField {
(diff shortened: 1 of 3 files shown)


SergioBenitez:
The biggest issue I see with this approach is that it deviates from the convention that `#[derive(Trait)]` derives the trait named `Trait`. With your changes, `#[derive(FromFormIgnorable)]` derives `FromForm` instead of `FromFormIgnorable`. I think this is confusing enough to not integrate this change as it is.

I see three possibilities here.

  1. The standard behavior could be to allow extra fields. This was actually the first implementation.

  2. We can allow an attribute to specify the behavior. This would look like:

    #[derive(FromForm)]
    #[form(lenient)]
    struct MyStruct { ... }

  3. We can open an RFC with the Rust team to allow parameters on derives. Something like:

    #[derive(FromForm(lenient))]
    struct MyStruct { ... }

crlf0710:
I think i like 2 > 1 > 3. As this is just a proof of concept, I'm closing this PR. But I'm looking forward to see such support coming in Rocket. Thanks!

PR #255, opened 2017-04-13
Actually enable TLS in the hello_tls example

fabricedesre:


SergioBenitez:
Good catch! Disabled it to test the error message and forgot to reenable it. Merged in da1f6264e49bdffeff0dd910082cb68cad1eb24c.

PR #256, opened 2017-04-13
Inform about minimum nightly version

rofrol:


SergioBenitez:
I'm not sure what we gain by doing this. The build script will emit an error if it detects you're on a version of `rustc` that is incompatible. And, having the version in the `README` means we have to continuously update it.

rofrol:
1. The version in README is as an example to illustrate what Rocket checks - rustc version and not a nightly. 2. Where to get that information about minimum rustc version prior to compiling - it may take a while and can also be checked programmatically.
3. Show that nightly-2017-03-31 will have rustc of different day: 2017-03-30. This may be confusing.

SergioBenitez:
I don't think this is necessary and can clear to confusion if someone believes we support an older version than we actually do. We should simply say that should always be using the latest nightly. In fact, we do say that in the [getting started section of the guide](https://rocket.rs/guide/getting-started/#nightly-version).

PR #257, opened 2017-04-13
Pin travis to specific nightly

rofrol:
Because of this https://github.com/rust-lang-nursery/rustup.rs/issues/1062#issuecomment-293930677

and future failures.


SergioBenitez:
The latest version of Rocket should always build on the latest working nightly; so far, this has been true with at most a few hours of delay. We purposefully don't pin travis to a specific version because we _want_ to see issues caused by the latest nightly.

Issue #267, opened 2017-04-18
Impl<T> From<T> for Stream<T> where T: Read


dgriffen:
# Feature Request

`Stream` currently has a method called from that takes a `Read` object and returns a Stream. It would be nice if `Stream` implemented `From` so that we could call `.into()` on our `Read` types.

This would be useful in reducing the amount of code written for returning a `Stream` object. Currently in order to return a `Stream` one must type `Stream::from(reader)`, if this impl was added you could simply write `reader.into()`.

This feature needs to exist in Rocket since the Stream type is defined there and impls of traits for types must be done in the same crate.

SergioBenitez:
I'd happily accept a PR for this!

PR #268, opened 2017-04-18
Implements #267

belst:
This implements `std::convert::From` for `Stream<T>`

implements #267

### DIFF
--- a/examples/stream/src/main.rs
+++ b/examples/stream/src/main.rs
@@ -12,14 +12,14 @@ type LimitedRepeat = Take<Repeat>;
 
 #[get("/")]
 fn root() -> content::Plain<Stream<LimitedRepeat>> {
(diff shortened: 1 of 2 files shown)


SergioBenitez:
`From` doesn't need to be imported; it's in the prelude. Also, we should get rid of the old `from` method. Finally, the new `from` method should have a docstring, likely just a copy of the old one.

Oh, and can we keep using 'Stream::from` in the library? I like the explicitness.

belst:
yeah. I derped a little there :P should be good now I guess.

SergioBenitez:
Should the comment be on the `impl` or on the `fn`? I forget how rustdoc shows the two. Can you see what the difference is?

SergioBenitez:
Merged in 0e759edf78a696f8fc3ffcd7f181720dace2382d.

PR #272, opened 2017-04-20
updated serde to v1

skariel:
otherwise it doesn't work with projects using serde 0.9


SergioBenitez:
Unfortunately we need both Tera and Handlebars to upgrade to v1.0 before we can.

skariel:
well, I'll do PRs for them too then

skariel:
Done: upgrading serde v1 PRs for Rocket, Tera and Handlebars-Rust

sunng87:
Handlebars 0.26.1 released for serde 1.0

PR #281, opened 2017-04-29
Fix typo in the guides

tomekw:
Just a minor thing...


SergioBenitez:
Definitely want to catch all of these, no matter how minor. Thanks so much for your fix! :)

Merged in 82c52c903a57d7f5cff0d6cce1b845848ebd2b7d.

PR #282, opened 2017-04-30
site/state.md typos for parameter names

flurdy:
Updated parameter names for count to hit_count as code inside functions expect hit_count so code examples would not even have compiled. This also mirrors the example project code names.

---
This change is [Reviewable](https://reviewable.io/reviews/sergiobenitez/rocket/282)

### DIFF
--- a/site/guide/state.md
+++ b/site/guide/state.md
@@ -54,7 +54,7 @@ current `HitCount` in a `count` route as follows:
 
 ```rust
 #[get("/count")]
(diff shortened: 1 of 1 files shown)


SergioBenitez:
Ah, yeah. This is an awesome catch. Thank you for fixing it! :D

Merged in 18f8a9dd5c51ce3689bd89cccd0a9a9ad71e9d6e.

PR #289, opened 2017-05-13
Add test to check for working condition. Expose fn to get rocket and …

StefanoChiodino:
…run rust fmt

Partially solves #62

### DIFF
--- a/examples/form_kitchen_sink/src/main.rs
+++ b/examples/form_kitchen_sink/src/main.rs
@@ -19,6 +19,8 @@ enum FormOption {
     C,
 }
 
+const INVALID_UTF8_ERROR_MESSAGE : &str = "Form input was invalid UTF8.";
+
 impl<'v> FromFormValue<'v> for FormOption {
     type Error = &'v RawStr;
 
@@ -51,7 +53,7 @@ fn sink(sink: Result<Form<FormInput>, Option<String>>) -> String {
     match sink {
         Ok(form) => format!("{:?}", form.get()),
         Err(Some(f)) => format!("Invalid form input: {}", f),
-        Err(None) => format!("Form input was invalid UTF8."),
+        Err(None) => INVALID_UTF8_ERROR_MESSAGE.to_string(),
     }
 }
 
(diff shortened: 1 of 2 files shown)


StefanoChiodino:
Hi, @SergioBenitez have you had a chance to look at this?

SergioBenitez:
Thanks so much for this, and sorry for the delay in getting back to you. I'm excited to see you contribute these tests in! I've left some feedback. Please let me know if you have any questions. :)

StefanoChiodino:
That's great @SergioBenitez, thanks for the feedback. The review points are all fine (though I personally would call something that just returns rocket `get_rocket` but I saw it being called just `rocket` in other places and I get the point for consistency!).

I expected the lack of tests for failure to come up, but wanted to show what I've got so far. Shouldn't be too difficult to add a few more!

Cheers

StefanoChiodino:
I think I've addressed all you points. I tried passing a string that doesn't serialise to a form struct and  an invalid utf8, then check for the expected error.

SergioBenitez:
Sorry, I may not have been clear. We're not too concerned about getting invalid UTF8 at the application side; I think it's okay to assume that, though it's also fine to test. But we _do_ want to test for validation. We want to make sure that, for example, a form field 'radio' with value 'D' doesn't parse. In other words, we want to make sure that the validation we _think_ is happening is indeed happening.

We should test each of the form fields. So, we'd have 6 classes of tests, one for each field. We want to isolate each field, so each class of tests would use valid values for all fields but the one being tested. For instance, to test the `radio` field, we'd use valid values for all other fields, then perhaps have three tests:

   * One with a valid value for `radio`: should pass.
   * One with `radio` missing: should fail.
   * One with an invalid value for `radio`: should fail.

This gives us thorough testing of the form example.

SergioBenitez:
@StefanoChiodino Do you think you'll be able to take care of this? :)

SergioBenitez:
@StefanoChiodino I wanted to close out #62, so I went ahead and implemented the tests for the 'form_kitchen_sink' example myself. Please do contribute again in the future! It's more than appreciated it!

PR #291, opened 2017-05-16
Tera templates example

icode-live:
Hello Sergio, please accept my gratitude for this wonderful project.

Here is the tera version of the handlebars_templates example.
Hope to see more of your talks on the Web :-)
------
Ran the test so it should be fine
   Compiling tera_templates v0.0.0 (file:///home/tac/Work/myContribs/RUN/Rocket/examples/tera_templates)
    Finished dev [unoptimized + debuginfo] target(s) in 67.33 secs
     Running /home/tac/Work/myContribs/RUN/Rocket/target/debug/deps/tera_templates-2943d598afb2ef4a

running 3 tests
test tests::test_404 ... ok
test tests::test_name ... ok
test tests::test_root ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured


SergioBenitez:
Hi @icode-live! Thanks so much for the pull request!

I think have a Tera templating example is a great idea! Do you think we could do something different, perhaps simpler, though? Maybe just a single route that renders a Tera template.

Also, due to recent changes in the templating library, the `to_string()` method on `Template` is no longer implemented. See the updated handlebars template example for how to use render the template in an example.

icode-live:
ok I'm on it

icode-live:
Hi Sergio, about making the example, are you sure you want is  simpler. Coming from Django it was that handlebars example that got me to dig a little more into Rocket :-)

jaroslaw-weber:
Hey, how about few HTML comments in the template file?

I think the example is simple enough but I know handlebars already so :)

SergioBenitez:
Merged in 7bda1b527b67be1da8bae68ccb3a2fdcc208f19e! Hoorah!

PR #316, opened 2017-06-13
add login and logout tests to session example

flosse:
Do you have any ideas why the logout response doesn't contain a `Set-Cookie` header that clears the cookie value of `user_id`?

### DIFF
--- a/examples/session/src/main.rs
+++ b/examples/session/src/main.rs
@@ -4,6 +4,9 @@
 extern crate rocket_contrib;
 extern crate rocket;
 
+#[cfg(test)]
+mod tests;
+
 use std::collections::HashMap;
 
 use rocket::Outcome;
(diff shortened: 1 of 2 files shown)


SergioBenitez:
You need to pass the cookie back in to the next request, just like a browser would.

flosse:
> You need to pass the cookie back in to the next request, just like a browser would.

This is what I did:
https://github.com/SergioBenitez/Rocket/pull/316/files#diff-a0c764ce77b04fc3701903892cab9774R69

SergioBenitez:
Hi @flosse. So sorry to have closed this! I didn't mean to dismiss your contribution here. I started working on #334, realized the session example didn't have any tests, and starting writing them, forgetting about your PR here. Apologies!

PR #319, opened 2017-06-17
Route.path was renamed to Route.uri

vegai:
Benchmark was broken, simple fix.


SergioBenitez:
Thanks for finding and fixing this! It's appreciated! :)

Merged in 3ef14f5392226db2d3939f4779d6d1b0204488ae.

PR #337, opened 2017-07-06
support dots in form fields

bbigras:
For OpenID I got fields like `openid.ns`.

I have no idea if it's OK to just add `c == '.' ||` but it seems to work.


bbigras:
Thanks!!

PR #347, opened 2017-07-14
update Fairing example to reflect doc's content

stuarth:
Small tweak to align the example code with doc content that it "is both a launch and response fairing" on https://github.com/SergioBenitez/Rocket/blob/cdacda08964f25efb6b7c3d89eaa2cb08de68c24/lib/src/fairing/mod.rs#L275


SergioBenitez:
Awesome, thanks for this! Merged in b3f41f248cf6cdaab12fd00900ef748ee26795c8.

PR #361, opened 2017-07-25
Fix Tera error reporting

lnicola:
Don't skip over the first error line as returned by Tera.

### DIFF
--- a/contrib/src/templates/tera_templates.rs
+++ b/contrib/src/templates/tera_templates.rs
@@ -41,8 +41,8 @@ impl Engine for Tera {
             Ok(string) => Some(string),
             Err(e) => {
                 error_!("Error rendering Tera template '{}'.", name);
-                for error in e.iter().skip(1) {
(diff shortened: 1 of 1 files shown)


lnicola:
Sample output:

```
before:
    => Error rendering Tera template 'index'.
    => Template 'index' failed to render.

after:
    => Error rendering Tera template 'index'.
    => Failed to render 'index': context isn't a JSON object. The value passed needs to be a key-value object: context, struct, hashmap for example..
    => Template 'index' failed to render.
```

lnicola:
Another example (only after):

```
    => Error rendering Tera template 'index'.
    => Failed to render 'index'.
    => Field `foo` not found in context while rendering 'index'.
    => Template 'index' failed to render.
```

SergioBenitez:
Thanks for this. :) Merged in 5f0583eb049892f9795be1fc66134842c37ff88a.

Issue #64, opened 2016-12-27
Allow access to native template features


philwhineray:
At present the Handlebars and Tera template functionality is fully encapsulated.

It is likely that users will choose one based on features unique to that engine. Rocket should be able to expose the appropriate objects to make those features available.

See PR #60 for initial discussion.

kardeiz:
I'm feeling this limitation too.

I think an acceptable resolution would be to just make `rocket_contrib::templates::Context` public. Then users could write their own fairing, which could modify e.g. `Tera` through `context.engines.tera` after calling `Context::initialize`, and then add the `Context` to `Rocket` state.

`Context` doesn't seem to contain any information that *must* be kept private. It's not ideal to make users set the templates directory manually (as in my scenario above), but is not that difficult, and that part of `Template::fairing` could be abstracted out into a `Template::root` function.

seanlinsley:
I've submitted a new PR if anyone wants to try it out: #500.

PR #364, opened 2017-07-27
Allow access to template containers

kardeiz:
and make it possible to interact with them before adding them to managed state.

Also reexports `tera` and `handlebars`.

See #64.

### DIFF
--- a/contrib/src/templates/mod.rs
+++ b/contrib/src/templates/mod.rs
@@ -166,6 +166,7 @@ impl Template {
         })
     }
 
+    /// Returns the path to the template root directory
     pub fn root(config: &Config) -> PathBuf {
         let mut template_root = config.root().join(DEFAULT_TEMPLATE_DIR);
         match config.get_str("template_dir") {
--- a/contrib/src/templates/mod.rs
+++ b/contrib/src/templates/mod.rs
@@ -8,7 +8,6 @@ mod engine;
 mod context;
 
 use self::engine::{Engine, Engines};
-use self::context::Context;
 use self::serde::Serialize;
 use self::serde_json::{Value, to_value};
 use self::glob::glob;
@@ -16,11 +15,13 @@ use self::glob::glob;
 use std::borrow::Cow;
 use std::path::{Path, PathBuf};
 
-use rocket::State;
+use rocket::{Config, State};
 use rocket::request::Request;
 use rocket::fairing::{Fairing, AdHoc};
 use rocket::response::{self, Content, Responder};
 use rocket::http::{ContentType, Status};
+
+pub use self::context::Context;
 
 const DEFAULT_TEMPLATE_DIR: &'static str = "templates";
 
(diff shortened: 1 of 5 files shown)


contrib/src/templates/mod.rs
```diff
@@ -3,25 +3,37 @@ extern crate serde_json;
...
-use rocket::State;
+use rocket::{Config, State};
 use rocket::request::Request;
 use rocket::fairing::{Fairing, AdHoc};
 use rocket::response::{self, Content, Responder};
 use rocket::http::{ContentType, Status};
 
+pub use self::context::Context;
+
+
```
SergioBenitez: Why is there so much empty space here? Just one line is fine.

SergioBenitez:
If we're going to accept this, it's going to need to be thoroughly documented. How do you intend this to be used?

milyin:
It's definitely necessary to give user access to handlebars object to add custom helpers. Without custom helpers handlebars engine is too limited.
May be it's possible to allow to register some callback and call it instead of "let hb = Handlebars::new()" here https://github.com/SergioBenitez/Rocket/blob/master/contrib/src/templates/handlebars_templates.rs#L12 ?

SergioBenitez:
@milyin I was hoping to expose an engine-agnostic mechanism for adding template helpers so that you can add one helper that works for both Handlebars and Tera (and any engines in the future). As far as I can see, being able to add helpers are the primary and perhaps only benefit of this PR. Is that the case? If so, I'd like to implement the nicer solution even though it means delaying being able to do this further.

maddin77:
What about [filters](https://tera.netlify.com/docs/templates/#filters) and [tests](https://tera.netlify.com/docs/templates/#tests) in tera?

SergioBenitez:
Good point, @maddin77.

@kardeiz I'm still in need of documentation for this change. I don't know how you expect these changes will be used; nothing looks obvious to me.

kardeiz:
I'm not sure if this is the best way to do this; having some kind of generator arg (e.g., `Fn() -> Handlebars`) to `Template::fairing` might be better, or some kind of callback for the user (e.g., `Fn(&mut Context)` or `Fn(&mut Handlebars)`. I just put this together as the simplest way to get access to the extra features of `Tera`.

In any case, I was using this PR like:

```rust
rocket::ignite()
    .mount("/", routes![])
    .attach(AdHoc::on_attach(|rocket| {
        match Context::initialize(Template::root(rocket.config())) {
            Some(mut ctxt) => {
                // you can add filters, tests, etc. here
                configure(&mut ctxt.engines.tera);
                Ok(rocket.manage(ctxt))
            },
            None => Err(rocket)
        }
    }))
```

SergioBenitez:
I like the idea of a callback! What if we add another method through which the fairing can be created, say `custom`, that take an `F: FnMut(&mut Engines)` through which the user can do whatever they want with the enabled templating engines before Rocket launches?

SergioBenitez:
Closing this out in hopes of a better solution.

vishusandy:
Are there any plans to allow custom helpers?  I would really like to be able to use them without having to create a local copy of rocket and modify the handlebars code (which I did to get it to work but it creates problems of its own).

PR #366, opened 2017-07-29
Protect against paths with spaces in scripts.

lucaskolstad:
These changes enable the scripts to handle paths that have spaces. The errors are mostly to do with missing double-quoting letting accidental word-splitting occur. You can see the errors before this patch with the following:
```
mkdir ~/dir\ with\ spaces && cd ~/dir\ with\ spaces
git clone https://github.com/SergioBenitez/Rocket
cd Rocket/scripts
./test.sh # errors out
./publish.sh # errors out
./mk-docs.sh # errors out
```

After merging these changes, the above should run successfully without error. The only weird change is inside the `bootstrap_examples()` Bash function in `test.sh`; it has a for loop written a strange way. This was needed so that the environment variables set by the call to `eval $env_vars` remain in scope for the call to `cargo test` later when they are needed. The alternative loops that can also handle paths with spaces require a pipe (`find "${EXAMPLES_DIR}" -maxdepth 1 -type d -name "*" | while read -r file do; .... done`) which causes a subshell, losing the exported environment variables in a different scope. The loop as implemented stores the `find ...` output in a file and just redirects it instead of piping.

### DIFF
--- a/scripts/mk-docs.sh
+++ b/scripts/mk-docs.sh
@@ -7,12 +7,12 @@ set -e
 
 # Brings in: ROOT_DIR, EXAMPLES_DIR, LIB_DIR, CODEGEN_DIR, CONTRIB_DIR, DOC_DIR
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-source $SCRIPT_DIR/config.sh
+source "$SCRIPT_DIR/config.sh"
 
 function mk_doc() {
   local dir=$1
   local flag=$2
-  pushd $dir > /dev/null 2>&1
+  pushd "$dir" > /dev/null 2>&1
     echo ":: Documenting '${dir}'..."
     cargo doc --no-deps --all-features
   popd > /dev/null 2>&1
@@ -21,9 +21,9 @@ function mk_doc() {
 # We need to clean-up beforehand so we don't get all of the dependencies.
 cargo clean
 
-mk_doc $LIB_DIR
-mk_doc $CODEGEN_DIR
-mk_doc $CONTRIB_DIR
+mk_doc "$LIB_DIR"
+mk_doc "$CODEGEN_DIR"
+mk_doc "$CONTRIB_DIR"
 
 # Blank index, for redirection.
-touch ${DOC_DIR}/index.html
+touch "${DOC_DIR}/index.html"
--- a/scripts/mk-docs.sh
+++ b/scripts/mk-docs.sh
@@ -7,12 +7,12 @@ set -e
 
 # Brings in: ROOT_DIR, EXAMPLES_DIR, LIB_DIR, CODEGEN_DIR, CONTRIB_DIR, DOC_DIR
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-source "$SCRIPT_DIR/config.sh"
+source "${SCRIPT_DIR}/config.sh"
 
 function mk_doc() {
   local dir=$1
   local flag=$2
-  pushd "$dir" > /dev/null 2>&1
+  pushd "${dir}" > /dev/null 2>&1
     echo ":: Documenting '${dir}'..."
     cargo doc --no-deps --all-features
   popd > /dev/null 2>&1
@@ -21,9 +21,9 @@ function mk_doc() {
 # We need to clean-up beforehand so we don't get all of the dependencies.
 cargo clean
 
-mk_doc "$LIB_DIR"
-mk_doc "$CODEGEN_DIR"
-mk_doc "$CONTRIB_DIR"
+mk_doc "${LIB_DIR}"
+mk_doc "${CODEGEN_DIR}"
+mk_doc "${CONTRIB_DIR}"
 
 # Blank index, for redirection.
 touch "${DOC_DIR}/index.html"
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -7,24 +7,24 @@ set -e
 
 # Brings in: ROOT_DIR, EXAMPLES_DIR, LIB_DIR, CODEGEN_DIR, CONTRIB_DIR, DOC_DIR
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-source $SCRIPT_DIR/config.sh
+source "$SCRIPT_DIR/config.sh"
 
-if ! [ -z "$(git status --porcelain)" ]; then
-  echo "There are uncommited changes! Aborting."
-  exit 1
-fi
+#if ! [ -z "$(git status --porcelain)" ]; then
+#  echo "There are uncommited changes! Aborting."
+#  exit 1
+#fi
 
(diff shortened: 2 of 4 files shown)


scripts/test.sh
```diff
@@ -50,23 +50,22 @@ function ensure_trailing_whitespace_free() {
...
+      bootstrap_result=$?
+      if [ $bootstrap_result -ne 0 ]; then
+        echo "    Running bootstrap script (${bootstrap_script}) failed!"
+        exit 1
+      else
+        eval $env_vars
       fi
     fi
-  done
+  done < <(find "${EXAMPLES_DIR}" -maxdepth 1 -type d -name "*")
```
SergioBenitez: The `-name "*"` parameter is unnecessary.

scripts/test.sh
```diff
@@ -50,23 +50,22 @@ function ensure_trailing_whitespace_free() {
 }
 
 function bootstrap_examples() {
-  for file in ${EXAMPLES_DIR}/*; do
-    if [ -d "${file}" ]; then
-      bootstrap_script="${file}/bootstrap.sh"
-      if [ -x "${bootstrap_script}" ]; then
-        echo "    Bootstrapping ${file}..."
+  while read -r file;
```
SergioBenitez: Merge this line and the next so that `do` is on the same line as the semicolon.

scripts/test.sh
```diff
@@ -50,23 +50,22 @@ function ensure_trailing_whitespace_free() {
...
 
-        env_vars=$(${bootstrap_script})
-        bootstrap_result=$?
-        if [ $bootstrap_result -ne 0 ]; then
-          echo "    Running bootstrap script (${bootstrap_script}) failed!"
-          exit 1
-        else
-          eval $env_vars
-        fi
+      env_vars=$(bash "${bootstrap_script}")
```
SergioBenitez: Can we do this another way? It's unfortunate to have to call `bash` directly here.

scripts/publish.sh
```diff
@@ -7,29 +7,29 @@ set -e
...
 # Brings in: ROOT_DIR, EXAMPLES_DIR, LIB_DIR, CODEGEN_DIR, CONTRIB_DIR, DOC_DIR
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-source $SCRIPT_DIR/config.sh
+source "$SCRIPT_DIR/config.sh"
 
-if ! [ -z "$(git status --porcelain)" ]; then
-  echo "There are uncommited changes! Aborting."
-  exit 1
-fi
+#if ! [ -z "$(git status --porcelain)" ]; then
```
SergioBenitez: Why has this been commented out?

scripts/publish.sh
```diff
@@ -7,29 +7,29 @@ set -e
...
+#  echo "There are uncommited changes! Aborting."
+#  exit 1
+#fi
 
 # Ensure everything passes before trying to publish.
 echo ":::: Running test suite..."
 cargo clean
-${SCRIPT_DIR}/test.sh
+bash "${SCRIPT_DIR}/test.sh"
```
SergioBenitez: Can we avoid calling `bash` directly?

scripts/publish.sh
lucaskolstad: I had commented this and the `cargo publish` line out while I was working on top of uncomitted changes so I could test that it all ran and then forgot to properly uncomment them both back... woops how embarrassing!

scripts/test.sh
lucaskolstad: I'm actually unsure how to address this. I agree it is unfortunate. I'm thinking about solutions and have left it for now, but if you have a suggestion I'll make the change and squash it into the code review commit.

scripts/test.sh
lucaskolstad: Good catch!

lucaskolstad:
Thanks for reviewing this!

I've added a new commit addressing most of your requested changes. Still haven't fixed the direct call to Bash yet so feel free to wait until there is a smart solution for that which still executes a script potentially having spaces before merging if you want.

scripts/mk-docs.sh
```diff
@@ -7,12 +7,12 @@ set -e
...
 # Brings in: ROOT_DIR, EXAMPLES_DIR, LIB_DIR, CODEGEN_DIR, CONTRIB_DIR, DOC_DIR
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-source $SCRIPT_DIR/config.sh
+source "${SCRIPT_DIR}/config.sh"
 
 function mk_doc() {
   local dir=$1
   local flag=$2
-  pushd $dir > /dev/null 2>&1
+  pushd "$dir" > /dev/null 2>&1
```
SergioBenitez: Use `${dir}` here.

scripts/mk-docs.sh
```diff
@@ -21,9 +21,9 @@ function mk_doc() {
 # We need to clean-up beforehand so we don't get all of the dependencies.
 cargo clean
 
-mk_doc $LIB_DIR
-mk_doc $CODEGEN_DIR
-mk_doc $CONTRIB_DIR
+mk_doc "$LIB_DIR"
```
SergioBenitez: Use `${LIB_DIR}` here. Same thing with the brackets in the two lines below.

scripts/test.sh
```diff
@@ -3,22 +3,22 @@ set -e
...
+source "${SCRIPT_DIR}/config.sh"
 
 # Add Cargo to PATH.
 export PATH=${HOME}/.cargo/bin:${PATH}
 
 # Checks that the versions for Cargo projects $@ all match
 function check_versions_match() {
   local last_version=""
-  for dir in $@; do
+  for dir in "$@"; do
```
SergioBenitez: Use `${@}`.

Actually, do we even need to use quotes here? The inputs should be properly separated already. Or will they be re-separated here?

lucaskolstad: I believe we do need them, I did a quick check by creating a directory with spaces in the path like in the opening pull request description and without the double quotes it errors out looking for the wrong truncated path and doesn't error with the quotes.

scripts/mk-docs.sh
lucaskolstad: fixed in the amended code review commit

lucaskolstad:
I amended the code review commit with those last spots that needed braces that you pointed out.

scripts/test.sh
SergioBenitez: How about `eval` instead of `bash`?

SergioBenitez:
Can you remind me on where we settled with not using `bash`? I believe we discussed using `eval` or `exec` on IRC. What was the conclusion of that discussion?

SergioBenitez:
@lucaskolstad Ping!

SergioBenitez:
Okay. We'll depend on `bash`.

SergioBenitez:
Thanks so much for this! Merged in 2d72928ba193da38f2597fdcdffb2c02c8ded23a. You rock, @lucaskolstad! :)

PR #370, opened 2017-08-01
Reference the Birthday problem (ID collision in Pastebin example)

roSievers:
I think that the probability of some collision is more important than the chance of two particular IDs colliding. However the birthday paradox is too complicated to explain in a short comment, so I referred to Wikipedia instead.

I don't believe that it is necessary to refer to the birthday problem at all, only if you believe that talking about the probability of collision is important.

### DIFF
--- a/site/guide/pastebin.md
+++ b/site/guide/pastebin.md
@@ -144,10 +144,9 @@ pub struct PasteID<'a>(Cow<'a, str>);
 
 impl<'a> PasteID<'a> {
     /// Generate a _probably_ unique ID with `size` characters. For readability,
-    /// the characters used are from the sets [0-9], [A-Z], [a-z]. The
-    /// probability of a collision depends on the value of `size` and
-    /// the amount of IDs generated. If you want to learn more, this is known as
-    /// the birthday problem: https://en.wikipedia.org/wiki/Birthday_problem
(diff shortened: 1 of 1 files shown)


site/guide/pastebin.md
```diff
@@ -145,8 +145,9 @@ pub struct PasteID<'a>(Cow<'a, str>);
 impl<'a> PasteID<'a> {
     /// Generate a _probably_ unique ID with `size` characters. For readability,
     /// the characters used are from the sets [0-9], [A-Z], [a-z]. The
-    /// probability of a collision depends on the value of `size`. In
-    /// particular, the probability of a collision is 1/62^(size).
+    /// probability of a collision depends on the value of `size` and
+    /// the amount of IDs generated. If you want to learn more, this is known as
```
SergioBenitez: Change "amount" to "number". Let's say something like:

"The probability of a collision depends on the value of `size` and the number of IDs generated thus far." I think it would be fine if this is all we said about the probability, i.e, end the doc comment there. What do you think?

roSievers: I think that your version is fine and I copied it in verbatim.

SergioBenitez:
Thanks so much! Merged in 3a76fe03db9f5da7f9304f6ba731e2a6cce71d1d.

PR #374, opened 2017-08-03
Print description of error when tera fails to init

bjorn3:


SergioBenitez:
Can you show a diff of the output with this change?

bjorn3:
Never mind description is the same as the first item of the iter.

bjorn3:
Never mind description is the same as the first item of the iter.

PR #379, opened 2017-08-09
Fix DbConn guard documentation

bahlo:
You have to use `&*conn` to get the `&SqliteConnection` from the `DbConn` guard (was `&conn`).


SergioBenitez:
Merged in 6279039361d786cad621fe4b395f8d1735730d9a. Thanks so much!

PR #397, opened 2017-08-23
router: avoid sorting routes on add

lheckemann:
Resolves the TODO in the comment :)

It would probably be nicer to do the whole thing with a data structure suited for a priority queue like a BinaryHeap, which should improve insertion efficiency as well; I've started doing this in 7424729dddbaa8bda0551530e239cb82a1f050c1 but the tests aren't passing and I have yet to work out why. It also requires a fair bit of extra code.


SergioBenitez:
Merged in c63bb67fb97fec5373df0aaaadd7cdc13858959a. Thanks for this! :)

PR #410, opened 2017-09-03
Improve Debug impl for Config by using `debug_struct()`

LukasKalbertodt:
Hi there :crab: 

This makes it possible to debug-output `Config` with `"{:#?}"`. Please see the commit message for a full explanation.

While looking through all manual `Debug` impls in Rocket, I've found something questionable in the impl for `Outcome`: it doesn't output the values at all -- it only outputs the variant name. Is there a reason for that? If not, a `#[derive(Debug)]` should be fine. I could change this, too, if you want.

### DIFF
--- a/lib/src/config/config.rs
+++ b/lib/src/config/config.rs
@@ -791,15 +791,18 @@ impl Config {
 
 impl fmt::Debug for Config {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "Config[{}] {{ address: {}, port: {}, workers: {}, log: {:?}",
-               self.environment, self.address, self.port, self.workers,
-               self.log_level)?;
+        let mut s = f.debug_struct("Config");
+        s.field("environment", &self.environment);
+        s.field("address", &self.address);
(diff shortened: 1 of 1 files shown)


lib/src/config/config.rs
```diff
@@ -791,15 +791,18 @@ impl Config {
 
 impl fmt::Debug for Config {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "Config[{}] {{ address: {}, port: {}, workers: {}, log: {:?}",
-               self.environment, self.address, self.port, self.workers,
-               self.log_level)?;
+        let struct_name = format!("Config[{}]", self.environment);
```
SergioBenitez: Let's make `environment` a field of the `debug_struct`.

LukasKalbertodt:
@SergioBenitez Should be fixed :)

SergioBenitez:
Thanks so much, @LukasKalbertodt! I've merged this in 329711db3be8ce3b367edbdbee9b27945ba77614. :)

PR #413, opened 2017-09-04
impl Responder for Arc<R>

zoumi:


zoumi:
fn respond_to(self, req: &Request) -> Result<Response<'r>, Status>

Why respond_to() take the ownership?

PR #414, opened 2017-09-04
Clarifies on_request Fairing documentation

belst:
After a bit of confusion on irc/matrix about the Fairing documentation @LukasKalbertodt and me thought that this would be more clear.


belst:
Updated the message

SergioBenitez:
Awesome. Merged in e656ec87d7752176e776e28c94d5402429ca6342.

PR #415, opened 2017-09-05
indent comments to prevent being commonmark code blocks

QuietMisdreavus:
When running rustdoc with `--enable-commonmark`, it found a few paragraphs that were indented off-margin from the numbered list they were within, and treated them as code blocks instead. This PR shifts them all over by one space to align with the text in the list item, which turns them back into regular paragraphs.


SergioBenitez:
Thanks for this! Merged in 62f96331e2ab983bb5c7b3aa33c166e2b0f1f814. :)

PR #420, opened 2017-09-07
remove reliance on lookup_host

dimbleby:
`lookup_host` doesn't look like being stabilised.  

`ToSocketAddrs` is kinda hokey - we have to provide a dummy port number to use it - but it's stable and does exactly the job that's wanted.

### DIFF
--- a/lib/src/config/config.rs
+++ b/lib/src/config/config.rs
@@ -1,5 +1,5 @@
 use std::collections::HashMap;
-use std::net::{IpAddr, lookup_host};
+use std::net::ToSocketAddrs;
 use std::path::{Path, PathBuf};
 use std::convert::AsRef;
 use std::fmt;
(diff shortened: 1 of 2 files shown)


SergioBenitez:
Is it possible that the port matters when resolving the address? IE, `1.2.3.4:A` might resolve while `1.2.3.4:B` won't?

dimbleby:
That's certainly not an issue with the current implementation, which just passes the socket through, and deals with the string by first [trying to parse it into an IP address](https://github.com/rust-lang/rust/blob/a59a6d8a5687dd6aee871de0e7c9f734709ac544/src/libstd/net/addr.rs#L860-L868) and then [calling `lookup_host()`](https://github.com/rust-lang/rust/blob/a59a6d8a5687dd6aee871de0e7c9f734709ac544/src/libstd/net/addr.rs#L848-L852).

And I think that the [docs](https://doc.rust-lang.org/std/net/trait.ToSocketAddrs.html) justify the expectation that it'll always be this way:

> `(&str,u16)`: the string should be either a string representation of an `IpAddr` address as expected by `FromStr` implementation or a host name.

(I'm of the opinion that this having this `impl` be the standard library's only stable way to DNS lookup was a mis-step: but we are where we are).

SergioBenitez:
Excellent! Merged in bd08d50f5974b01b13f2170df698e1bd496f461c. Thanks so much for this. This is a great contribution!

PR #427, opened 2017-09-13
Fix errors![...] macro

teiesti:
In most applications, Rocket "lives" in the parent module. But sometimes, it
might be useful to encapsulate all Rocket code in a distict module like
`my_crate::http`.

In that case, rocket needs to be addressed with a absolute path instead of a
relative path. While the routes![...] macro does this correct, the errors![...]
macro did not.

This commit fixes the errors![...] macro.

### DIFF
--- a/codegen/src/macros/mod.rs
+++ b/codegen/src/macros/mod.rs
@@ -56,6 +56,6 @@ pub fn routes(ecx: &mut ExtCtxt, sp: Span, args: &[TokenTree])
 pub fn errors(ecx: &mut ExtCtxt, sp: Span, args: &[TokenTree])
         -> Box<MacResult + 'static> {
     prefixing_vec_macro(CATCH_STRUCT_PREFIX, |ecx, path| {
-        quote_expr!(ecx, rocket::Catcher::from(&$path))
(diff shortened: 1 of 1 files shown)


teiesti:
In case someone wants to try:

```rust
#![feature(plugin)]
#![plugin(rocket_codegen)]

extern crate rocket;

pub mod http {
    pub fn serve() {
        ::rocket::ignite()
            .catch(errors![
                not_found,
            ])
            .launch();
    }

    #[error(404)]
    fn not_found() {
        unimplemented!()
    }
}

fn main() {
    http::serve();
}
```

SergioBenitez:
This is a really great catch; thanks so much for your contribution. Apologies for taking so long to get to this! It's been merged in af58f72018f4779ce7f2ed61a15120f68c66be72.

zoumi:
errors!() macro is missing in master version?

```
error: cannot find macro `errors!` in this scope
   --> src\main.rs:183:16
    |
183 |         .catch(errors![not_found])
    |                ^^^^^^

```

SergioBenitez:
@zoumi It was renamed to `catchers!`. See the [errors example](https://github.com/SergioBenitez/Rocket/blob/master/examples/errors/src/main.rs).

Issue #368, opened 2017-07-31
How to add a private cookie to a LocalRequest?


golddranks:
Using Rocket 0.3.

I tried to look for a way to add private cookie to a LocalRequest for testing, but it seems that the only way to get private cookies for tests is to generate them first on the server (simulate a session-initiating request). It would be helpful if adding private cookies would be directly available for testing.

I think it would be sufficient to be documented in the API docs, maybe as a method `cookie_private` on `LocalRequest`.

SergioBenitez:
This should definitely be possible and shouldn't be difficult to implement. If anyone wants to tackle this, I'd be more than happy to review.

PR #430, opened 2017-09-14
add private_cookie to LocalRequest

henriiik:
Hello!

I implemented the feature request in #368 since it had the "help wanted" label. I am still new to rust and any feedback is appreciated. The new method is very much like the `cookie` method.

### DIFF
--- a/lib/src/local/request.rs
+++ b/lib/src/local/request.rs
@@ -189,6 +189,28 @@ impl<'c> LocalRequest<'c> {
         self
     }
 
+    /// Add a private cookie to this request.
+    ///
+    /// # Examples
+    ///
+    /// Add private `user_id` cookie:
+    ///
+    /// ```rust
+    /// use rocket::local::Client;
+    /// use rocket::http::Cookie;
+    ///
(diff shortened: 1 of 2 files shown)


turboladen:
Looks good to me. I actually forked earlier to make these same changes. Looking forward to seeing them merged.

SergioBenitez:
I'm sorry; I haven't had any time to review changes lately. But I'll note that this isn't correct. The way the code's written in this PR means that the added cookie will be marked as "added" to the request, not as part of the original request. As a result, Rocket will generate a `Set-Cookie` header incorrectly. If you look at the code for `Cookies`, you'll see that Rocket calls `add_original`, not simply `add`, for regular cookies. This prevents the cookie from being marked as "added" and thus the header being emitted. We need to do something similar for private cookies.

There's a correct but inelegant solution that doesn't require changes to the `cookie` crate, and then there's the better solution that does. Unfortunately I don't have time to go into this in detail now, but I'll leave this here as a note.

SergioBenitez:
Closing in favor of #487.

PR #431, opened 2017-09-15
Added closure to configure template engines

sunng87:
Currently there is no way to access template engine, for example, adding custom helper to handlebars.

This patch added a closure to `Template::fairing` which allows a closure to configure those engines.

### DIFF
--- a/contrib/src/lib.rs
+++ b/contrib/src/lib.rs
@@ -48,6 +48,12 @@ extern crate serde;
 #[cfg_attr(feature = "json", macro_reexport(json_internal))]
 extern crate serde_json;
 
+#[cfg(feature = "handlebars_templates")]
+pub extern crate handlebars;
+
+#[cfg(feature = "tera_templates")]
+pub extern crate tera;
+
 #[cfg(feature = "json")]
 #[cfg_attr(feature = "json", macro_use)]
 #[doc(hidden)]
@@ -74,9 +80,3 @@ mod uuid;
 
 #[cfg(feature = "uuid")]
 pub use uuid::{UUID, UuidParseError};
-
-#[cfg(feature = "handlebars_templates")]
-pub extern crate handlebars;
-
-#[cfg(feature = "tera_templates")]
-pub extern crate tera;
--- a/contrib/src/lib.rs
+++ b/contrib/src/lib.rs
@@ -74,3 +74,9 @@ mod uuid;
 
 #[cfg(feature = "uuid")]
 pub use uuid::{UUID, UuidParseError};
+
+#[cfg(feature = "handlebars_templates")]
+pub extern crate handlebars;
+
+#[cfg(feature = "tera_templates")]
+pub extern crate tera;
--- a/contrib/src/templates/context.rs
+++ b/contrib/src/templates/context.rs
@@ -49,10 +49,6 @@ impl Context {
             Context { root, templates, engines }
         })
     }
-
-    pub fn engines_mut(&mut self) -> &mut Engines {
-        &mut self.engines
-    }
 }
 
 /// Removes the file path's extension or does nothing if there is none.
--- a/contrib/src/templates/context.rs
+++ b/contrib/src/templates/context.rs
@@ -49,6 +49,10 @@ impl Context {
             Context { root, templates, engines }
         })
     }
+
+    pub fn engines_mut(&mut self) -> &mut Engines {
+        &mut self.engines
+    }
 }
 
 /// Removes the file path's extension or does nothing if there is none.
--- a/contrib/src/templates/engine.rs
+++ b/contrib/src/templates/engine.rs
@@ -15,9 +15,9 @@ pub trait Engine: Send + Sync + 'static {
 
 pub struct Engines {
     #[cfg(feature = "tera_templates")]
-    tera: Tera,
+    pub tera: Tera,
     #[cfg(feature = "handlebars_templates")]
-    handlebars: Handlebars,
+    pub handlebars: Handlebars,
 }
 
 impl Engines {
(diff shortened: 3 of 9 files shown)


SergioBenitez:
Can you remove all of the style changes? We don't use 'rustfmt' since it doesn't quite work well enough yet.

I can review the PR once only functional changes are present.

sunng87:
@SergioBenitez disabled my `format-on-save` and commit again.

contrib/src/templates/context.rs
```diff
@@ -49,6 +49,10 @@ impl Context {
             Context { root, templates, engines }
         })
     }
+
+    pub fn engines_mut(&mut self) -> &mut Engines {
```
SergioBenitez: Remove this method.

contrib/src/templates/engine.rs
```diff
@@ -69,4 +69,14 @@ impl Engines {
 
         None
     }
+
+    #[cfg(feature = "handlebars_templates")]
+    pub fn handlebars(&mut self) -> &mut Handlebars {
```
SergioBenitez: Remove this method.

contrib/src/templates/engine.rs
```diff
@@ -69,4 +69,14 @@ impl Engines {
...
+        &mut self.handlebars
+    }
+
+    #[cfg(feature = "tera_templates")]
+    pub fn tera(&mut self) -> &mut Tera {
```
SergioBenitez: Remove this method.

contrib/src/templates/mod.rs
```diff
@@ -158,9 +158,37 @@ impl Template {
...
+    ///     rocket::ignite()
+    ///         // ...
+    ///         .attach(Template::config_and_fairing(|ctxt| {
+    ///             // ctxt.engines_mut().handlebars().register_helper ...
+    ///         }))
+    ///         // ...
+    ///     # ;
+    /// }
+    /// ```
+    pub fn config_and_fairing<F>(f: F) -> impl Fairing
```
SergioBenitez: Name this method `custom`.

contrib/src/templates/mod.rs
```diff
@@ -158,9 +158,37 @@ impl Template {
...
+        where F: Fn(&mut Context) + Send + Sync + 'static
```
SergioBenitez: Pass in an `&mut Engines` instead. Make the fields of the `Engines` struct public.

examples/handlebars_templates/src/main.rs
```diff
@@ -46,6 +47,27 @@ fn rocket() -> rocket::Rocket {
...
 
+// when you want to register handlebars helpers
+#[allow(dead_code)]
+fn rocket_with_hbs_helper() -> rocket::Rocket {
+    rocket::ignite()
+        .mount("/", routes![index, get])
+        .attach(Template::config_and_fairing(|ctxt| {
+            ctxt.engines_mut()
+                .handlebars()
+                .register_helper("test",
```
SergioBenitez: We need a real test here. Create a helper than does something, and use it in one of the templates. Ensure that the returned value is what's expected.

examples/handlebars_templates/src/main.rs
```diff
@@ -46,6 +47,27 @@ fn rocket() -> rocket::Rocket {
         .catch(errors![not_found])
 }
 
+// when you want to register handlebars helpers
+#[allow(dead_code)]
+fn rocket_with_hbs_helper() -> rocket::Rocket {
```
SergioBenitez: Don't create a function that's not used. Instead, perhaps you can name this function something like:

```rust
fn install_helper(handlebars: &mut Handlebars) { .. }
```

 and then call it from `rocket()`:

```rust
.attach(Template::custom(|engines| install_helper(&mut engines.handlebars)));
```

contrib/src/lib.rs
```diff
@@ -74,3 +74,9 @@ mod uuid;
 
 #[cfg(feature = "uuid")]
 pub use uuid::{UUID, UuidParseError};
+
+#[cfg(feature = "handlebars_templates")]
+pub extern crate handlebars;
```
SergioBenitez: Move these `extern crate`s to the top of the file below the existing `extern crate`s.

contrib/src/templates/mod.rs
SergioBenitez: Also, why does `F` have to be `Send + Sync + 'static`?

sunng87: I think it `Adhoc::attach` requested. It won't compile if F is not marked `Send + Sync + 'static`

examples/handlebars_templates/src/main.rs
```diff
@@ -42,7 +43,21 @@ fn not_found(req: &Request) -> Template {
...
+                                     if let Some(p0) = h.param(0) {
+                                         rc.writer.write(p0.value()
+                                                         .render()
+                                                         .into_bytes()
+                                                         .as_ref())?;
+                                     }
+                                     Ok(())
+                                 }));
+
+        }))
```
sunng87: I just changed default `rocket()` to use `custom`

SergioBenitez:
Please see the build failure.

sunng87:
@SergioBenitez it was `Template::show` that has no helper config. Will a API break change on `Template::show` work for you? I'm going to add a closure argument to it. Of course we can a `Template::custom_and_show` but I think it's a little verbose for a test method. WDYT?

SergioBenitez:
So sorry for the delay here! As we spoke about briefly in person, I think the API for `Template::show` needs to change a bit. The upside is that I think it'll actually be cleaner.

Here's the signature I'm imagining for `Template::show()`:

```rust
    pub fn show<S, C>(rocket: &Rocket, name: S, context: C) -> Option<String>
        where S: Into<Cow<'static, str>>, C: Serialize
    {
```

The big change is that `show()` now takes in a `Rocket` instance explicitly. This means that we can use the cached versions of templates (yay!) along with their configuration (yay!) and carrying around a root path is no longer required (yay!). An instance of `Rocket` should be available at all points necessary. In tests, it can be obtained via `client.rocket()`.

What do you think?

SergioBenitez:
Just a heads up that #500 also seeks to implement this.

PR #434, opened 2017-09-17
Make query string parsing non-strict by default

carlosdp:
I'd like to suggest this change which makes query string parsing non-strict by default. Query strings often contain garbage data and having erroneous parameters we don't care about in it should not factor in on whether the route matches or not.

For example, if you are implementing a GET webhook, the webhook developer might add a new parameter to the payload, which they can do without notice since it isn't a breaking change to the API, and this would break our application because the route would no longer match, even though we can still parse the data just fine.

Or, a Facebook share button might attach a `share=facebook_newsfeed` parameter to the URL to indicate where the user came from. In the current default behavior, they would get a 404 if the route took query parameters.

Finally, no other web framework to my knowledge rejects a request for having a super set of the expected query parameters. This behavior is very confusing for developers coming over from Rails or Node, etc. If someone has a good use case for making query parameters strict on the route, we should provide the option to do so, but the default should be non-strict to match more in-line with what web developers expect.

### DIFF
--- a/codegen/src/decorators/route.rs
+++ b/codegen/src/decorators/route.rs
@@ -68,7 +68,7 @@ impl RouteParams {
             #[allow(non_snake_case)]
             let $name: $ty = {
                 let mut items = ::rocket::request::FormItems::from($form_string);
-                let form = ::rocket::request::FromForm::from_form(items.by_ref(), true);
+                let form = ::rocket::request::FromForm::from_form(items.by_ref(), false);
                 #[allow(unreachable_patterns)]
                 let obj = match form {
                     Ok(v) => v,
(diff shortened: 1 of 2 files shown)


Korvox:
Just to add some diversity to this conversation I mocked up what adding another route parameter to make this a togglable thing would look like over [here](https://github.com/SergioBenitez/Rocket/pull/449).

One thing to worry about in both cases is how route rank works for lenient queries. Under either implementations right now the first lenient route to match will take the request unless you specify ranks by hand. It would be quite a bit of work to try to interpret all possible routes and then figure out the one that parses the _most_ query parameters to use (and I don't believe Rocket currently is even designed to support that kind of route retracing) but its something to think about.

SergioBenitez:
Closing this in favor of #608.

carlosdp:
@SergioBenitez that is a reference issue that mentions this PR, is there another PR open that does this same thing?

SergioBenitez:
@carlosdp No, not yet, but we want a holistic solution that envelops this PR instead of a one-off solution.

PR #442, opened 2017-09-28
Fix docs: fairings.md "these event" -> "these events"

colelawrence:
Small grammar error, remove my incorrect grammar from the previous PR.


SergioBenitez:
Good catch! Merged in edb494c6c3f679bda1fa2990a7e3b58edaf6b394.

PR #447, opened 2017-10-03
Add a default catcher for Gateway Timeout.

golddranks:
Gateway Timeout is an useful response, when the server fails to receive a timely response from a backend server (for example, an LDAP server for user authentication). At the moment Rocket doesn't provide a default catcher for it, although it could.

### DIFF
--- a/lib/src/catcher.rs
+++ b/lib/src/catcher.rs
@@ -239,9 +239,8 @@ pub mod defaults {
                 method, or it lacks the ability to fulfill the request.", handle_501,
             503, "Service Unavailable", "The server is currently unavailable.",
                 handle_503,
-            504, "Gateway Timeout", "The server didn't receive a timely response
-                from another server.",
-                handle_504,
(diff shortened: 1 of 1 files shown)


lib/src/catcher.rs
```diff
@@ -239,6 +239,9 @@ pub mod defaults {
                 method, or it lacks the ability to fulfill the request.", handle_501,
             503, "Service Unavailable", "The server is currently unavailable.",
                 handle_503,
+            504, "Gateway Timeout", "The server didn't receive a timely response
```
SergioBenitez: Let's change the following: "didn't" to "did not", and "another" to "an upstream". Let's also merge the last two lines together.

golddranks: Aye! Pushing a new commit.

SergioBenitez:
Merged in b0c1a0a07fc426571901e12ac1df3ec195d48d3a.

PR #449, opened 2017-10-05
Lenient Query Statement Support

Korvox:
This is a proof of concept, but I did want to solicit feedback on the idea before I continue (because there are several paths to fixing this and they are fairly divergent).

The impl in this commit adds another route parameter that can only validly hold the form `query = "lenient"` such as `#[get("foo/bar?<baz>", query = "lenient")]`. It was the most straightforward way to go about it given the current way routes are tokenized (hardcoded url first, then key values after). That being said, there are several alternate implementations that are more ergonomic that I've been thinking about:

- Have some special trailing characters following the query statement, for example `&..` to indicate leniency. This should feel really intuitive for complete novices since it mimics the "grab all" syntax of slices. It would look like `#[get("foo/bar?<baz>&..")]`. Could also leave off the & or be any general pattern at the end of a url (? or ! etc).
- Have a boolean flag rather than a key-value pair. Figuring out how to name it properly feels a bit clunky to me since only `lenient_query` would really be descriptive enough.
- Try wrapping the route decorators in another decorator just to apply leniency. I feel this would be the most ergonomically cumbersome and a pain in the butt to implement (at least for me, who spent all last weekend trying to read into the finer details about libsyntax and how different it is from syn and quote...).
- Do what form data did. Since to get the final form you `from_form` on `from_data` you have a struct like `Form<ImplsFromForm>` which adds a point of boilerplate for `LenientForm` to drop in at. To do this though queries would have to be directed through some wrapper trait that specializes on form strictness like `Query<Foo>` vs `LenientQuery<Foo>`. This adds a boilerplate wrapper type for no practical reason, but makes Queries and Forms behave like each other again.

I'm up for implementing any of these, just wondering what the preferred course of action would be. Or I could tidy up this mockup, add some tests, and we could go with this for now.

### DIFF
--- a/codegen/src/decorators/route.rs
+++ b/codegen/src/decorators/route.rs
@@ -51,7 +51,8 @@ impl RouteParams {
     fn gen_form(&self,
                 ecx: &ExtCtxt,
                 param: Option<&Spanned<Ident>>,
-                form_string: P<Expr>)
+                form_string: P<Expr>,
+                strict: bool)
                 -> Option<Stmt> {
         let arg = param.and_then(|p| self.annotated_fn.find_input(&p.node.name));
         if param.is_none() {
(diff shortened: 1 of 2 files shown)


SergioBenitez:
I don't like the idea of adding special syntax just for this. What I'd like, and what should be eventually implemented, is to remove the special casing of query strings as forms and instead have a new trait, `FromQuery`, which we implement for `Form` and `LenientForm`. Then, if you'd like a strict form, you'd use `Form`. Otherwise, you'd use `LenientForm`. If you want something totally different, such as to parse things that _arent't_ forms, you can implement `FromQuery` for some type yourself.

Closing out the PR for this reason.

Issue #317, opened 2017-06-14
Default binding to port 80 for `production` env


benwilber:
Is it appropriate for Rocket to try to bind to port 80 by default in production mode?  A normal user can't test their app configuration via `ROCKET_ENV=prod cargo run` because they can't bind to the privileged port.  I understand it's trivial to change the port via `ROCKET_PORT`, but why should this be necessary when running Rocket on port 80 is uncommon (except maybe in a Docker container).  I think most deployments would be behind something like nginx/varnish which themselves would be bound on port 80.

```
$ ROCKET_ENV=prod rustup run nightly cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `/Users/benw/Projects/rocket/target/debug/hello_world`
Warning: environment is 'production', but no `secret_key` is configured
Error: Rocket failed to launch due to an I/O error.
thread 'main' panicked at 'Permission denied (os error 13)', /Users/benw/Projects/rocket/lib/src/error.rs:204
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

It's also unfriendly and ambiguous why Rocket `thread 'main' panicked at 'Permission denied (os error 13)',` unless you know that it's because you can't bind to port 80.  Wouldn't it be more appropriate to just have standard rocket port (`8000`) for all the envs?  And then change to the uncommon port 80 as needed.

mehcode:
A normal user _shouldn't_ be able to run the app in production mode. That's what staging is for. Why does the normal user / developer have production database access?

I'd say that the docker container environment is proving to be the common environment for new deployments.

benwilber:
@mehcode The staging env mode also tries to bind to port 80.  I think this is incorrect assumption that Rocket is making.  Configuring binding to a privileged port is the sysadmin's job, which they would prefer to be explicit with `ROCKET_PORT=80`, if that's appropriate for their stack.

I think that rocket should just have standard port `8000` regardless of env.  Also the whole built-in notion of "production", "staging", "development" environments is flawed.  12factor via environment is the correct way to configure an app.

shssoichiro:
In addition to the comment above, I would argue that even if docker is the common environment now, we shouldn't be using defaults that only work for a docker environment. Port 8000 works both inside and outside of docker.

mehcode:
I can see the argument here. I'm all for it.

---

@benwilber With #318 you could use [config-rs](https://github.com/mehcode/config-rs) to easily implement config like that.

[...]

PR #450, opened 2017-10-05
Replacing port 80 with port 8000 for staging and production

darArch:
PR to fix the issue raised here: https://github.com/SergioBenitez/Rocket/issues/317

- replaced all usages of port 80 in code with port 54321
- replaced all usages of port 80 in comments and documentation with port 54321
- updated configuration guide to remove the line around need for `sudo` as with port number > `1024` this is no longer the case
- wasn't sure if there is a preferred port number for this, so I picked `54321` in honour of a Rocket's countdown
- manually triggered a Travis build as I forgot to enable my fork before pushing, all tests passed
- set up a simple `Hello, World!` test app using the fork and with no custom configuration launched using port `54321` when running with `ROCKET_ENV=stage` and `ROCKET_ENV=prod`

### DIFF
--- a/lib/src/config/mod.rs
+++ b/lib/src/config/mod.rs
@@ -35,7 +35,7 @@
 //!     listen on
 //!     * examples: `"localhost"`, `"0.0.0.0"`, `"1.2.3.4"`
 //!   * **port**: _[integer]_ a port number to listen on
-//!     * examples: `"8000"`, `"80"`, `"4242"`
+//!     * examples: `"8000"`, `"54321"`, `"4242"`
 //!   * **workers**: _[integer]_ the number of concurrent workers to use
 //!     * examples: `12`, `1`, `4`
 //!   * **log**: _[string]_ how much information to log; one of `"normal"`,
@@ -78,7 +78,7 @@
 //!
 //! [staging]
 //! address = "0.0.0.0"
-//! port = 80
+//! port = 54321
 //! workers = [number_of_cpus * 2]
 //! log = "normal"
 //! secret_key = [randomly generated at launch]
@@ -86,7 +86,7 @@
 //!
 //! [production]
 //! address = "0.0.0.0"
-//! port = 80
+//! port = 54321
 //! workers = [number_of_cpus * 2]
 //! log = "critical"
 //! secret_key = [randomly generated at launch]
--- a/lib/src/config/mod.rs
+++ b/lib/src/config/mod.rs
@@ -35,7 +35,7 @@
 //!     listen on
 //!     * examples: `"localhost"`, `"0.0.0.0"`, `"1.2.3.4"`
 //!   * **port**: _[integer]_ a port number to listen on
-//!     * examples: `"54321"`, `"80"`, `"4242"`
+//!     * examples: `"8000"`, `"80"`, `"4242"`
 //!   * **workers**: _[integer]_ the number of concurrent workers to use
 //!     * examples: `12`, `1`, `4`
 //!   * **log**: _[string]_ how much information to log; one of `"normal"`,
@@ -70,7 +70,7 @@
 //! ```toml
 //! [development]
 //! address = "localhost"
-//! port = 54321
+//! port = 8000
 //! workers = [number_of_cpus * 2]
 //! log = "normal"
 //! secret_key = [randomly generated at launch]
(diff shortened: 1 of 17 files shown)


SergioBenitez:
Awesome! Thanks for taking this on. Looking forward to merging this.

There a couple of minor issues; I'll leave line comments for those. But the bigger question is: is 54321 the right port number for Rocket? This is a bit of a silly thing, but I'd love for Rocket to have a fun and unique port number! 54321 is pretty good, actually; do you know if anyone else using it?

Another number I had considered was 31680. This number is derived from the Rocket emoji code point 0x1F680, with 0x1F converted to its decimal form (31) and the rest left in hex form. 🚀 A little contest would be fun for this, too!

lib/src/config/mod.rs
```diff
@@ -35,7 +35,7 @@
 //!     listen on
 //!     * examples: `"localhost"`, `"0.0.0.0"`, `"1.2.3.4"`
 //!   * **port**: _[integer]_ a port number to listen on
-//!     * examples: `"8000"`, `"80"`, `"4242"`
+//!     * examples: `"8000"`, `"54321"`, `"4242"`
```
SergioBenitez: You can leave the `80` here and change the `8000`.

benwilber:
What about `64711` in honor of Apollo 11?

```
>>> "".join(map(str, [sum(map(ord, ['a', 'p', 'o', 'l', 'l', 'o'])), 11]))
'64711'
```

darArch:
@SergioBenitez I've held off on making the changes to the other (non staging/production) defaults until a port number is agreed, but I'll do that as soon as one has been chosen.

Neither `54321` nor `31680` nor `64711` appear to be used by default by any other web framework that I looked at. I also did some Googling, and the only possibly disqualifying information I found was that port `54321` seems to have been used [by some trojans in the past](https://www.speedguide.net/port.php?port=54321).

Either way, as soon as there is agreement around which number to use I'll update the PR with the requested changes.

giarve:
`54321` seems easier to remember (yes, launch countdown).
Rocket can wipe the past of this port and give it a better life.

lib/src/config/mod.rs
darArch: Done

darArch:
Updated PR code to use `54321` for _all_ default ports and updated docs per inline comment. 

I was going to wait, but `54321` is so distinctive that a simple Find & Replace All will allow this to be updated if a different port number is ultimately decided upon.

theduke:
How about `1206`.

With **a lot** of imagination, this could be short for "rock":

`12` => R
`0` => O
`6` => K

SergioBenitez:
I was about to merge this, and I had to type 54321 for some reason, and I realized: it is not a particularly easy number to type. This sounds silly, and it might be, but I feel like the default port number should be an easy one to type. `8000` is nice and easy due to the duplicate 0s, but 54321 isn't, unless you swipe the keys from 5 to 1.

Happy to field thoughts on this 😆 .

karuna:
The only bad thing aside from typing problem is that this port used to be used by trojans. So some security software might block it.

darArch:
I'm more than happy to change the port number for this PR, is there a consensus on what port number Rocket should use. As a reminder, the candidates (excluding 54321) proposed so far are: 31680, 64711, 12014 or 8000?

Let me know and I'll get this updated. For reasons of typing ease I'd favour 8000 but I don't have strong feelings about any of the options so I'm happy to go with whatever people like most :)

darArch:
Given the lack of any real consensus, I've updated the default port to `8000` replacing `54321`. Again, if any kind of favoured alternative option emerges it can be updated with a simple find & replace, but for now I thought it best to update to `8000` to allow for merging.

lib/src/local/request.rs
```diff
@@ -152,13 +152,13 @@ impl<'c> LocalRequest<'c> {
     ///
     /// # Examples
     ///
-    /// Set the remote address to "8.8.8.8:80":
+    /// Set the remote address to "8.8.8.8:8000":
```
SergioBenitez: Let's keep this at `80`.

SergioBenitez:
Merged in 19f59b1f9b587c6fcf60bf8b0ac81f813de625f6.

PR #454, opened 2017-10-13
unprovided "?" query should act like empty query

### LINKED ISSUE #376: Route queries without ? for optional parameters
`Version: 0.3.0`

The route with queries can only match with request with queries. But in cases where the parameters are options, requesting with a brother-less `?` is kind of awkward.

    #[get("/login?<redirect_url>")]
    fn login(redirect_url: Option<RedirectUrl>) -> Redirect {
        ...
    }

If there is no `RedirectUrl`, then this resource could be requested with a simple `/login` rather than a `/login?`. But in order to have such behaviour, I must duplicate the code with a new `#[get("/login")]` function.

It would be nice to have an additional rule of having `?` omitted if the all parameters are optional and no query-less route exists.

Also, another request, is route regex in the timeline of rocket?

lucasem:
Resolves #376.
Doesn't change the default ranking system.

API change:
- method `URI::query` now returns `&str` instead of `Option<&str>`
- added method `URI::query_explicit` returning a `bool` of whether a query was explicitly specified.

Impact:
- Routes for `/a` and `/a?<foo>` can _both_ match the request `/a`, unless the empty string fails to parse into the type indicated for the `foo` param. Previously, a path would necessarily start with `/a?` to match the latter route (and again, the match would fail if the empty string fails to parse into `foo`'s type). **Queries are irrelevant for collision with a route.** Rank and failed form parsing should be relied on instead.

Tests have been added and modified to reflect these changes.

### DIFF
--- a/codegen/src/decorators/route.rs
+++ b/codegen/src/decorators/route.rs
@@ -114,12 +114,7 @@ impl RouteParams {
 
     fn generate_query_statement(&self, ecx: &ExtCtxt) -> Option<Stmt> {
         let param = self.query_param.as_ref();
-        let expr = quote_expr!(ecx,
-           match __req.uri().query() {
-               Some(query) => query,
-               None => return ::rocket::Outcome::Forward(__data)
-           }
-        );
+        let expr = quote_expr!(ecx, __req.uri().query());
 
         self.gen_form(ecx, param, expr)
     }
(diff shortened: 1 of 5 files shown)


SergioBenitez:
Closing this in favor of #608.

PR #455, opened 2017-10-13
fix uri query parse

lucasem:
URIs like `/#?`, where the first question mark occurs in the frament, weren't correctly handled.


SergioBenitez:
Partially merged in 372d85ddb39a9296ca7cd43708d58ab1d6b2bcc5.

PR #456, opened 2017-10-13
Fix language type of code block

sinkuu:
Changes langauge type of the toml snippet to remove unnecessary "it's not tested" warning in the documentation.


SergioBenitez:
Does this still result in the blog getting syntax highlighted? This is why I didn't use this in the first place, but maybe that's been fixed.

sinkuu:
I see, sorry for the noise.

PR #458, opened 2017-10-14
Added MIME types for audio and video to prevent compiler warnings.

nickdonnelly:
Added MIME types:
* `video/webm`
* `video/ogg`
* `audio/wav`


nickdonnelly:
`audio` and `video` are official type prefixes as per the [RFC spec](https://tools.ietf.org/html/rfc2045#section-5.2). As for the specific subtypes, they are used commonly across the web and are [supported by major vendors](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Complete_list_of_MIME_types). 

The official [IANA list](https://www.iana.org/assignments/media-types/media-types.xhtml) uses `video/VP8` instead of `video/webm`, but the two are synonymous. I do not, however, believe that `audio/wav` (sometimes `audio/x-wav`) is on the official list, though it is widely supported.

SergioBenitez:
Okay! Merged in 9bc6502fa7ce68300eb86edf286c96bc61e0297e.

PR #482, opened 2017-11-22
Add support for setting a custom Content-Length

hamhut1066:
Currently it is not possible to set a custom Content-Length header. This can be needed when conforming to a spec for example. The default behavior is definitely recommended for 95% of use-cases.

Thanks for a great project, I'm really enjoying working with Rocket.

### DIFF
--- a/lib/src/rocket.rs
+++ b/lib/src/rocket.rs
@@ -132,6 +132,10 @@ impl Rocket {
             hyp_res.headers_mut().append_raw(name, value);
         }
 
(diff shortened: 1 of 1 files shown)


SergioBenitez:
I'm closing this due to the reasons I explain in https://github.com/SergioBenitez/Rocket/issues/514#issuecomment-358284445. I do hope to get something in with the same end-purpose as this PR, however.

hamhut1066:
Thanks, reading through your reply now

PR #483, opened 2017-11-27
add MVC example

madeindjs:
Hi,

I am Rust newbie and I recently try your framework to build a simple application. I think it will be a good idea to integrate it in *example* frameworks because when I try to learn Rocket I didn't find enough information (Template system, Diesel integration, etc..).

This sample just an application who allow you to edit recipes according to [REST](https://en.wikipedia.org/wiki/Representational_state_transfer) conventions. Recipes have just a simple `id` & `name` field. So I create theses routes: 

- **List** recipes: GET _/recipes_ 
- **Create** recipes: POST _/recipes_
- **Show** recipe: GET _/recipes/<recipe_id>_
- **New** recipe: GET _/recipes/new_
- **Edit** recipe: GET _/recipes/<recipe_id>/edit_
- **Update** recipe: PUT _/recipes/<recipe_id>_
- **Delete** recipe: DELETE _/recipes/<recipe_id>_

I didn't find any convention to build an MVC application (Model, View, Controller) so I do as best I can.

 & I open for any suggestion.

### DIFF
--- a/examples/mvc/LICENSE
+++ b/examples/mvc/LICENSE
@@ -1,7 +1,0 @@
-Copyright 2017 ROUSSEAU Alexandre
-
-Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
-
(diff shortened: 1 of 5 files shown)


SergioBenitez:
Thanks for the pull request! The code looks fairly good, but I'm afraid it's a bit out of scope for the examples in the repository. I think this would be better served as a tutorial or blog post.

I'm hoping to add a "vetted external documentation" section to the documentation on the official Rocket website. This would be a great for that! I'll ping you once that happens.

Thanks again!

Issue #234, opened 2017-03-17
[Feature Idea] Register templates from strings, rust-embed, or equivalent


kestred:
This feature would be required to allow strings embedded in the program (specifically like in https://github.com/SergioBenitez/Rocket/issues/227) to be used as templates.
Currently rocket_contrib::Template only supports templates which exist on the real file system.

This feature could exist outside of rocket, by just using the template engine directly and passing in the template string and context.  However, this feature may still make sense in `rocket_contrib` to remove boilerplate from the common desire to embed static files in the executable.

This feature may also be solved by or related to implementing https://github.com/SergioBenitez/Rocket/issues/64.

Issue was labelled as idea instead of request, mostly because I'm looking for input and less certain about what shape the ideal solution will be in.

flosse:
What about using https://maud.lambda.xyz?

Qqwy:
I think this is a good idea. Right now, e.g. Maud can already work with Rocket ([see the maud documentation](https://maud.lambda.xyz/web_frameworks.html#rocket)), as when Maud's `rocket` feature is enabled, an impl of the Responder trait is generated for maud HTML. Of course, that means skipping the template layer, and calling Maud's functionality in-line.

theduke:
This would be very nice.

Distributing the templates when deploying the binary is quite annoying.

Diesel has an `embed_migrations!("the/path")` macro to load migrations from the file system at compile time and embed them into the code.

That would be pretty awesome.

seanlinsley:
#500 might make this possible.

SergioBenitez:
https://github.com/SergioBenitez/Rocket/pull/521 will land shortly and make this possible!

PR #500, opened 2017-12-10
Provide a template registration callback

seanlinsley:
This makes it possible to register helper functions with Handlebars and Tera, resolving #64.

I'm using this locally to update Crates.io to use server-side rendering: https://github.com/rust-lang/crates.io/pull/1173.

That branch is out of date currently; here's how I'm using this feature:

```rust
// src/lib.rs

pub fn route(app: App) {
    rocket::ignite()
        .mount("/", routes![html::index])
        .attach(Template::fairing_with(|context| {
            html::register_helpers(&mut context.engines.handlebars);
        }))
        .launch();
}
```
```rust
// src/html/mod.rs

#[get("/")]
pub fn index() -> Template {
    let mut json = json!({});
    json["current_user"] = json!({"id": 1, "name": "Sean Linsley"});

    Template::render("index", &json)
}

pub fn register_helpers(handlebars: &mut Handlebars) {
    handlebars.register_helper("format_num", Box::new(format_num));
    handlebars.register_helper("embed_svg", Box::new(embed_svg));
}
```

There have been two previous PRs to try adding this feature: #60 and #364. @SergioBenitez has wanted the ability to register the same helper functions on both rendering engines simultaneously, but I'm not sure that's possible because of the library-specific code inside of the helper functions themselves.

Instead I opted to make the underlying template objects accessible via a callback passed to `Template::fairing_with`. Is there a better name for this function?

Once we're happy with this change, I'll update the documentation and write a test.

### DIFF
--- a/contrib/src/templates/mod.rs
+++ b/contrib/src/templates/mod.rs
@@ -15,6 +15,7 @@ use self::glob::glob;
 
 use std::borrow::Cow;
 use std::path::{Path, PathBuf};
+use std::marker::{Send, Sync};
 
 use rocket::State;
 use rocket::request::Request;
@@ -158,7 +159,11 @@ impl Template {
     /// }
     /// ```
     pub fn fairing() -> impl Fairing {
-        AdHoc::on_attach(|rocket| {
+        Template::fairing_with(|_context| ())
+    }
+
+    pub fn fairing_with<F>(callback: F) -> impl Fairing where F: Fn(&mut Context) + Send + Sync + 'static {
+        AdHoc::on_attach(move |rocket| {
             let mut template_root = rocket.config()
                 .root_relative(DEFAULT_TEMPLATE_DIR);
 
@@ -172,7 +177,10 @@ impl Template {
             };
 
             match Context::initialize(template_root) {
-                Some(ctxt) => Ok(rocket.manage(ctxt)),
+                Some(mut context) => {
+                    callback(&mut context);
+                    Ok(rocket.manage(context))
+                },
                 None => Err(rocket)
             }
         })
--- a/contrib/src/templates/engine.rs
+++ b/contrib/src/templates/engine.rs
@@ -15,9 +15,9 @@ pub trait Engine: Send + Sync + 'static {
 
 pub struct Engines {
     #[cfg(feature = "tera_templates")]
-    tera: Tera,
+    pub tera: Tera,
     #[cfg(feature = "handlebars_templates")]
-    handlebars: Handlebars,
+    pub handlebars: Handlebars,
 }
 
 impl Engines {


seanlinsley:
Because the callback returns `Context` instead of `Engines`, it would also allow application developers to register templates from strings: #234. Though it's possible that other changes are needed for that to work.

contrib/src/templates/mod.rs
```diff
@@ -15,6 +15,7 @@ use self::glob::glob;
 
 use std::borrow::Cow;
 use std::path::{Path, PathBuf};
+use std::marker::{Send, Sync};
```
seanlinsley: Oops, this isn't alphabetical

SergioBenitez:
This looks like the original implementation in #431. Perhaps you can finish off the contribution started there?

seanlinsley:
Oh, I didn't see that PR because it wasn't linked to the previous issues & PRs.

I haven't used `Template::show` before, so I'm not sure I understand why it needs to accept a closure.

SergioBenitez:
[`Template::show()`](https://api.rocket.rs/rocket_contrib/struct.Template.html#method.show) is used to render a template manually, usually for testing. The current implementation for `Template::show()` loads the requested template and template engine each time the function is called. As a result, any helpers or additional items attached to an engine via the `Template` fairing won't be present and templates rendered with `Template::show()` will fail to render if they make use of an expected helper.

In #431, I describe a solution to that issue: https://github.com/SergioBenitez/Rocket/pull/431#issuecomment-345361698. We can't accept any PR for this unless testing works as well. We also can't accept a PR for this unless there is documentation and new tests for the feature.

seanlinsley:
Refactoring `Template::show` probably deserves to be a separate PR. I'd be happy to open that.

One question: how do you get a `Context` from a `Rocket` object? Perhaps via `fairings`?

SergioBenitez:
It needs to be in one PR, otherwise, you can't possible test the new addition. You don't need to be able to retrieve anything additional from a `Rocket` instance. The `Template` code already does what is needed.

seanlinsley:
Your [previous comment](https://github.com/SergioBenitez/Rocket/pull/431#issuecomment-345361698) said to update `Template::show` to accept a `Rocket` instance, and to retrieve the already-loaded templates from it.

Being very unfamiliar with the code, I was thinking that something like this might work:
```rust
    pub fn show<S, C>(rocket: &Rocket, name: S, context: C) -> Option<String>
        where S: Into<Cow<'static, str>>, C: Serialize
    {
        let ctxt: Context = rocket.fairings.find(|fairing| fairing.info().name == "")?;

        Template::render(name, context).finalize(&ctxt).ok().map(|v| v.0)
    }
```

Though this appears to handle controller renders:

https://github.com/SergioBenitez/Rocket/blob/a9c66c9426bec57ee958480ae4aa3d789f20488f/contrib/src/templates/mod.rs#L268-L277

seanlinsley:
`rocket.state` should be the same object behind `request.guard()` if I understand correctly, but for some reason `state` is treated as private:
```rust
error[E0616]: field `state` of struct `rocket::Rocket` is private
   --> contrib/src/templates/mod.rs:243:20
    |
243 |         let ctxt = rocket.state.get::<Context>();
    |                    ^^^^^^^^^^^^

```

even though it should be accessible in the crate, as it's defined with `pub(crate)`:
```rust
pub struct Rocket {
    pub(crate) config: Config,
    router: Router,
    default_catchers: HashMap<u16, Catcher>,
    catchers: HashMap<u16, Catcher>,
    pub(crate) state: Container,
    fairings: Fairings,
}
```

SergioBenitez:
@seanlinsley The implementation details of managed state should _never_ be exposed to any outside crate. The `contrib` crate is _not_ part of the core `rocket` crate and should be treated (and is) exactly as if it were a third-party crate.

I think it's fair (and necessary, here) to expose a method that allows retrieving managed state from a `Rocket` instance. The signature (and implementation) should be:

```rust
/// Retrieve the managed state of type `T`, if it exists.
/// 
/// # Example (...)
#[inline(always)]
pub fn state<T: Send + Sync + 'static>(&self) -> Option<&'r T> {
    self.state.try_get()
}
```

You can then call `rocket.state::<Content>()?` to retrieve the managed state from `show()`.

seanlinsley:
Great to hear that someone managed to finish this feature 😄 Thanks @jebrosen!

PR #506, opened 2017-12-17
fix: Make form items parsing less strict

sebasmagri:
Currently Rocket is pretty strict on form items parsing, which affects how parameters without a value in query strings are handled, for example for `?flag` or `?f&foo=bar`. The [RFC 3986](https://tools.ietf.org/html/rfc3986#page-23) do not impose any constraints to the usage of empty query parameters and it seems like the way most framewors and servers handle this is by assigning an empty value to the parameter ([ref](https://stackoverflow.com/questions/4557387/is-a-url-query-parameter-valid-if-it-has-no-value)).

This change modifies the `FormItems` iterator to handle empty or unvalued parameters.

This affects the limits handling as well, since a `POST` request with a form content type and a body `valu`, as it results from limiting `value=Hello+world` to 4 bytes, will produce a valid form items collection like `&[("valu", "")]`. Then when trying to parse it agains a given type without a `valu` field this will result in a `422 Unprocessable Entity` error instead of a `400 Bad Request`. See the change to the `limits_tests::much_too_small` test.

Sorry for the whitespace changes, it's all `rustfmt` work.

### DIFF
--- a/lib/src/request/form/form_items.rs
+++ b/lib/src/request/form/form_items.rs
@@ -74,7 +74,7 @@ use http::RawStr;
 /// ```
 pub struct FormItems<'f> {
     string: &'f RawStr,
-    next_index: usize
+    next_index: usize,
 }
 
 impl<'f> FormItems<'f> {
@@ -145,7 +145,7 @@ impl<'f> FormItems<'f> {
     /// ```
     #[inline]
     pub fn exhaust(&mut self) -> bool {
-        while let Some(_) = self.next() {  }
+        while let Some(_) = self.next() {}
         self.completed()
     }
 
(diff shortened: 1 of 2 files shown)


samueltardieu:
> Sorry for the whitespace changes, it's all rustfmt work.

Note that you can choose which chunks to include in a commit using, for example, `git add -p`. This would let you exclude formatting-only changes.

SergioBenitez:
I think this was handled inadvertently by https://github.com/SergioBenitez/Rocket/commit/de8e1978c5da6efc0a95fa0f4340990df7ab21b9. Closing as a result; let me know if this isn't the case.

PR #520, opened 2017-12-27
Respond with status code 400 when the HTTP method is unknown

### LINKED ISSUE #465: Return 405 Method Not Allowed for unsupported HTTP methods instead of just 500ing
Makes it easier to build security alerts.

Helps distinguish attack traffic from run of the mill bugs in the rocket handlers. Having Rocket report incoming bad methods with 405 makes it easy to separate security alerts from developer alerts (500 code from logic errors, missing template files, missing db connection, ..etc). 

Rocket is already parsing the method, so why do the work again in another system? One could read the code of Rocket and generate a whitelist of methods in the config language of their upstream proxy, but then they have to keep that config up to date with Rocket source code changes. This kind of silent dependency imposes an extra cost on organizations using Rocket. The extra work that would be independently required for every organization is why this should not exist outside of Rocket.

The below diff shows the general idea of distinguishing between request construction failures. Note: it fails to set the Allow header which is required by the spec, not sure the best way of generating that header value (should the registered routes be inspected at runtime to see which methods are allowed? ).
```diff
diff --git a/lib/src/request/request.rs b/lib/src/request/request.rs
index f0492de..d3e78e5 100644
--- a/lib/src/request/request.rs
+++ b/lib/src/request/request.rs
@@ -568,17 +568,17 @@ impl<'r> Request<'r> {
                            h_headers: hyper::header::Headers,
                            h_uri: hyper::RequestUri,
                            h_addr: SocketAddr,
-                           ) -> Result<Request<'r>, String> {
+                           ) -> Result<Request<'r>, (Error, String)> {
         // Get a copy of the URI for later use.
         let uri = match h_uri {
             hyper::RequestUri::AbsolutePath(s) => s,
-            _ => return Err(format!("Bad URI: {}", h_uri)),
+            _ => return Err((Error::Internal, format!("Bad URI: {}", h_uri))),
         };
 
         // Ensure that the method is known. TODO: Allow made-up methods?
         let method = match Method::from_hyp(&h_method) {
             Some(method) => method,
-            None => return Err(format!("Invalid method: {}", h_method))
+            None => return Err((Error::BadMethod, format!("Invalid method: {}", h_method)))
         };
 
         // Construct the request object.
diff --git a/lib/src/rocket.rs b/lib/src/rocket.rs
index 52f276a..bdb6f96 100644
--- a/lib/src/rocket.rs
+++ b/lib/src/rocket.rs
@@ -55,10 +55,13 @@ impl hyper::Handler for Rocket {
         let req_res = Request::from_hyp(self, h_method, h_headers, h_uri, h_addr);
         let mut req = match req_res {
             Ok(req) => req,
-            Err(e) => {
-                error!("Bad incoming request: {}", e);
+            Err((reason, log_message)) => {
+                error!("Bad incoming request: {}", log_message);
                 let dummy = Request::new(self, Method::Get, Uri::new("<unknown>"));
-                let r = self.handle_error(Status::InternalServerError, &dummy);
+                let r = match reason {
+                    Error::BadMethod => self.handle_error(Status::MethodNotAllowed, &dummy),
+                    _ => self.handle_error(Status::InternalServerError, &dummy),
+                };
                 return self.issue_response(r, res);
             }
         };
```

maxfurman:
Fixes #465

### DIFF
--- a/lib/src/request/request.rs
+++ b/lib/src/request/request.rs
@@ -568,17 +568,17 @@ impl<'r> Request<'r> {
                            h_headers: hyper::header::Headers,
                            h_uri: hyper::RequestUri,
                            h_addr: SocketAddr,
-                           ) -> Result<Request<'r>, String> {
+                           ) -> Result<Request<'r>, (Error, String)> {
         // Get a copy of the URI for later use.
         let uri = match h_uri {
             hyper::RequestUri::AbsolutePath(s) => s,
-            _ => return Err(format!("Bad URI: {}", h_uri)),
+            _ => return Err((Error::Internal, format!("Bad URI: {}", h_uri))),
         };
 
         // Ensure that the method is known. TODO: Allow made-up methods?
         let method = match Method::from_hyp(&h_method) {
             Some(method) => method,
-            None => return Err(format!("Invalid method: {}", h_method))
+            None => return Err((Error::BadMethod, format!("Invalid method: {}", h_method)))
         };
 
         // Construct the request object.
--- a/lib/src/request/request.rs
+++ b/lib/src/request/request.rs
@@ -568,17 +568,17 @@ impl<'r> Request<'r> {
                            h_headers: hyper::header::Headers,
                            h_uri: hyper::RequestUri,
                            h_addr: SocketAddr,
-                           ) -> Result<Request<'r>, (Error, String)> {
+                           ) -> Result<Request<'r>, String> {
         // Get a copy of the URI for later use.
         let uri = match h_uri {
             hyper::RequestUri::AbsolutePath(s) => s,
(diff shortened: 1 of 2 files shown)


SergioBenitez:
Awesome! Thanks for taking this on!

After taking a look at this, I noticed that we should also return a `BadRequest` if the URI was invalid. And then I realized that that's all of the error cases in `from_hyp`, so we can simply return a `BadRequest` if we get an error at all! So this is actually as simple as changing `Status::InternalServerError` to `Status::BadRequest` on line 61 of `rocket.rs`!

maxfurman:
That certainly is a lot simpler!

SergioBenitez:
Merged in 4be75accf7b47ba270964dce89003c65cd157f2a. Thanks so much!

PR #521, opened 2017-12-28
Provide template registration callback for native template engine features

### LINKED ISSUE #64: Allow access to native template features

jebrosen:
Closes #64. Supersedes #431, #500.

This includes most of the commits from #431 and is based on further feedback on #431 and #500, in particular changing the signature of `Template::show`. I considered including #500 instead and still can if that is desired, but #431 already added to the tests for handlebars

### DIFF
--- a/contrib/src/templates/mod.rs
+++ b/contrib/src/templates/mod.rs
@@ -187,7 +187,6 @@ impl Template {
     pub fn custom<F>(f: F) -> impl Fairing where F: Fn(&mut Engines) + Send + Sync + 'static {
         AdHoc::on_attach(move |rocket| {
             let mut template_root = rocket.config().root_relative(DEFAULT_TEMPLATE_DIR);
-
             match rocket.config().get_str("template_dir") {
                 Ok(dir) => template_root = rocket.config().root_relative(dir),
                 Err(ConfigError::NotFound) => { /* ignore missing configs */ }
@@ -243,6 +242,7 @@ impl Template {
     ///
     /// Returns `Some` if the template could be rendered. Otherwise, returns
     /// `None`. If rendering fails, error output is printed to the console.
+    /// `None` is also returned if a Template fairing has not been attached.
     ///
     /// # Example
     ///
@@ -255,21 +255,32 @@ impl Template {
     /// use rocket_contrib::Template;
     ///
     /// fn main() {
-    ///     let rocket = rocket::ignite().attach(Template::fairing());
+    ///     let client =
+    ///         rocket::local::Client::new(rocket::ignite()
+    ///             .attach(Template::fairing())
+    ///         ).expect("valid rocket");
     ///
     ///     // Create a `context`. Here, just an empty `HashMap`.
     ///     let mut context = HashMap::new();
     ///
     ///     # context.insert("test", "test");
     ///     # #[allow(unused_variables)]
-    ///     let template = Template::show(&rocket, "index", context);
+    ///     let template = Template::show(client.rocket(), "index", context);
     /// }
     /// ```
     #[inline]
     pub fn show<S, C>(rocket: &Rocket, name: S, context: C) -> Option<String>
         where S: Into<Cow<'static, str>>, C: Serialize
     {
-        let ctxt = rocket.state::<Context>().unwrap();
+        let ctxt = match rocket.state::<Context>() {
+            Some(ctxt) => ctxt,
+            None => {
+                warn_!("Uninitialized template context: missing fairing.");
+                info_!("To use templates, you must attach `Template::fairing()`.");
+                info_!("See the `Template` documentation for more information.");
+                return None;
+            }
+        };
         Template::render(name, context).finalize(&ctxt).ok().map(|v| v.0)
     }
 
--- a/contrib/src/templates/mod.rs
+++ b/contrib/src/templates/mod.rs
@@ -247,18 +247,17 @@ impl Template {
     /// # Example
     ///
     /// ```rust
+    /// # extern crate rocket;
+    /// # extern crate rocket_contrib;
+    ///
     /// use std::collections::HashMap;
     ///
-    /// extern crate rocket;
-    /// extern crate rocket_contrib;
-    ///
     /// use rocket_contrib::Template;
+    /// use rocket::local::Client;
     ///
     /// fn main() {
-    ///     let client =
-    ///         rocket::local::Client::new(rocket::ignite()
-    ///             .attach(Template::fairing())
-    ///         ).expect("valid rocket");
+    ///     let rocket = rocket::ignite().attach(Template::fairing());
+    ///     let client = Client::new(rocket).expect("valid rocket");
     ///
     ///     // Create a `context`. Here, just an empty `HashMap`.
     ///     let mut context = HashMap::new();
@@ -275,9 +274,9 @@ impl Template {
         let ctxt = match rocket.state::<Context>() {
             Some(ctxt) => ctxt,
             None => {
-                warn_!("Uninitialized template context: missing fairing.");
-                info_!("To use templates, you must attach `Template::fairing()`.");
-                info_!("See the `Template` documentation for more information.");
+                warn!("Uninitialized template context: missing fairing.");
+                info!("To use templates, you must attach `Template::fairing()`.");
+                info!("See the `Template` documentation for more information.");
                 return None;
             }
         };
--- a/contrib/src/templates/mod.rs
+++ b/contrib/src/templates/mod.rs
@@ -230,24 +230,23 @@ impl Template {
     }
 
     /// Render the template named `name` with the context `context` into a
-    /// `String`. This method should
-    /// **not** be used in any running Rocket application. This method should
-    /// only be used during testing to validate `Template` responses. For other
-    /// uses, use [`render`](#method.render) instead.
+    /// `String`. This method should **not** be used in any running Rocket
+    /// application. This method should only be used during testing to
+    /// validate `Template` responses. For other uses, use
+    /// [`render`](#method.render) instead.
     ///
     /// The `context` can be of any type that implements `Serialize`. This is
     /// typically a `HashMap` or a custom `struct`.
     ///
     /// Returns `Some` if the template could be rendered. Otherwise, returns
     /// `None`. If rendering fails, error output is printed to the console.
-    /// `None` is also returned if a Template fairing has not been attached.
+    /// `None` is also returned if a `Template` fairing has not been attached.
     ///
     /// # Example
     ///
     /// ```rust
     /// # extern crate rocket;
     /// # extern crate rocket_contrib;
-    ///
     /// use std::collections::HashMap;
     ///
     /// use rocket_contrib::Template;
--- a/contrib/src/templates/mod.rs
+++ b/contrib/src/templates/mod.rs
@@ -158,9 +158,37 @@ impl Template {
     /// }
     /// ```
     pub fn fairing() -> impl Fairing {
-        AdHoc::on_attach(|rocket| {
-            let mut template_root = rocket.config()
-                .root_relative(DEFAULT_TEMPLATE_DIR);
+        Template::config_and_fairing(|_| {})
+    }
+
+    /// Returns a fairing that intializes and maintains templating state.
+    ///
+    /// This method allows you to configure the template context via the
+    /// closure.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// extern crate rocket;
+    /// extern crate rocket_contrib;
+    ///
+    /// use rocket_contrib::Template;
+    ///
+    /// fn main() {
+    ///     rocket::ignite()
+    ///         // ...
+    ///         .attach(Template::config_and_fairing(|ctxt| {
+    ///             // ctxt.engines_mut().handlebars().register_helper ...
+    ///         }))
+    ///         // ...
+    ///     # ;
+    /// }
+    /// ```
+    pub fn config_and_fairing<F>(f: F) -> impl Fairing
+        where F: Fn(&mut Context) + Send + Sync + 'static
+    {
+        AdHoc::on_attach(move |rocket| {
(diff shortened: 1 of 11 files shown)


contrib/src/templates/mod.rs
```diff
@@ -158,9 +158,35 @@ impl Template {
...
+    ///     rocket::ignite()
+    ///         // ...
+    ///         .attach(Template::custom(|engines| {
+    ///             // engines.handlebars.register_helper ...
+    ///         }))
+    ///         // ...
+    ///     # ;
+    /// }
+    /// ```
+    pub fn custom<F>(f: F) -> impl Fairing where F: Fn(&mut Engines) + Send + Sync + 'static {
```
SergioBenitez: Do we need the `Send + Sync + 'static` bounds here? I can't see what would make that the case. Also, we should be able to take in an `FnOnce` instead of `Fn` for greater flexibility. We should be able to have:

```rust
pub fn custom<F: FnOnce(&mut Engines)>(f: F) -> impl Fairing { ... }
```

contrib/src/templates/mod.rs
```diff
@@ -158,9 +158,35 @@ impl Template {
...
+        AdHoc::on_attach(move |rocket| {
+            let mut template_root = rocket.config().root_relative(DEFAULT_TEMPLATE_DIR);
```
SergioBenitez: Kill this empty line.

contrib/src/templates/mod.rs
```diff
@@ -219,22 +248,29 @@ impl Template {
...
+    ///
+    /// extern crate rocket;
+    /// extern crate rocket_contrib;
+    ///
     /// use rocket_contrib::Template;
     ///
-    /// // Create a `context`. Here, just an empty `HashMap`.
-    /// let mut context = HashMap::new();
+    /// fn main() {
+    ///     let rocket = rocket::ignite().attach(Template::fairing());
```
SergioBenitez: This method will be used almost exclusively during testing. As such, we should probably use a `Client` here, and call `client.rocket()` for illustrative purposes.

contrib/src/templates/mod.rs
```diff
@@ -219,22 +248,29 @@ impl Template {
...
-    pub fn show<P, S, C>(root: P, name: S, context: C) -> Option<String>
-        where P: AsRef<Path>, S: Into<Cow<'static, str>>, C: Serialize
+    pub fn show<S, C>(rocket: &Rocket, name: S, context: C) -> Option<String>
+        where S: Into<Cow<'static, str>>, C: Serialize
     {
-        let root = root.as_ref().to_path_buf();
-        Context::initialize(root).and_then(|ctxt| {
-            Template::render(name, context).finalize(&ctxt).ok().map(|v| v.0)
-        })
+        let ctxt = rocket.state::<Context>().unwrap();
```
SergioBenitez: Don't `unwrap()`. There's no guarantee that `Context` is being managed. Simply do:

```rust
let ctxt = rocket.state::<Context>()?;
```

Better yet, emit a warning if there's no `Context` being managed:

```rust
let ctxt = match rocket.state::<Context>() {
    Some(ctxt) => ctxt,
    None => {
            warn_!("Uninitialized template context: missing fairing.");
            info_!("To use templates, you must attach `Template::fairing()`.");
            info_!("See the `Template` documentation for more information.");
            return None;
    }
}
```

Then, amend this method's documentation to note that `None` can also be returned if a `Template::fairing()` isn't attached.

contrib/tests/templates.rs
```diff
@@ -1,17 +1,26 @@
...
+use rocket::config::{Config, Environment};
+use rocket_contrib::Template;
+
 fn template_root() -> PathBuf {
     let cwd = env::current_dir().expect("current working directory");
     cwd.join("tests").join("templates")
 }
 
+fn rocket() -> Rocket {
+    let config = Config::build(Environment::Development).extra("template_dir", template_root().to_str().unwrap()).unwrap();
```
SergioBenitez: This line is far too long. Split this into multiple lines, method chaining style. Also, use `expect()` instead of `unwrap()`.

examples/handlebars_templates/src/main.rs
```diff
@@ -42,7 +43,21 @@ fn not_found(req: &Request) -> Template {
 fn rocket() -> rocket::Rocket {
     rocket::ignite()
         .mount("/", routes![index, get])
-        .attach(Template::fairing())
+        .attach(Template::custom(|engines| {
```
SergioBenitez: This looks pretty ugly. Can we factor the helper out into a function and then just call `register_helper("echo", Box::new(echo_help))` here?

examples/handlebars_templates/src/tests.rs
```diff
@@ -4,20 +4,18 @@ use rocket::http::Method::*;
...
+        $test_fn(&client, client.req($method, $path).dispatch());
     })
 }
 
 #[test]
 fn test_root() {
     // Check that the redirect works.
     for method in &[Get, Head] {
-        dispatch!(*method, "/", |mut response: LocalResponse| {
+        dispatch!(*method, "/", |_client: &Client, mut response: LocalResponse| {
```
SergioBenitez: Use `_: &Client`.

examples/tera_templates/src/tests.rs
```diff
@@ -4,20 +4,18 @@ use rocket::http::Method::*;
...
+        $test_fn(&client, client.req($method, $path).dispatch());
     })
 }
 
 #[test]
 fn test_root() {
     // Check that the redirect works.
     for method in &[Get, Head] {
-        dispatch!(*method, "/", |mut response: LocalResponse| {
+        dispatch!(*method, "/", |_client: &Client, mut response: LocalResponse| {
```
SergioBenitez: Same here.

lib/src/rocket.rs
```diff
@@ -604,6 +604,19 @@ impl Rocket {
         self
     }
 
+    /// Get the managed state of type T, if it exists.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// let rocket = rocket::ignite().manage("example");
```
SergioBenitez: This isn't a good example because it encourages the bad practice of managing a type not controlled by the library. Do something like this instead:

```rust
struct MyState(&'static str);

let rocket = rocket::ignite().manage(MyState("hello!"));
assert_eq!(rocket.state::<MyState>(), Some(MyState("hello!")));

let client = rocket::local::Client::new(rocket).expect("valid rocket");
assert_eq!(client.rocket().state::<MyState>(), Some(MyState("hello!")));
```

contrib/src/templates/mod.rs
jebrosen: (blocked by #522)

examples/handlebars_templates/src/main.rs
```diff
@@ -39,10 +40,22 @@ fn not_found(req: &Request) -> Template {
     Template::render("error/404", &map)
 }
 
+struct EchoHelper;
```
SergioBenitez: You shouldn't need a struct here. You should be able to pass a function with the following signature to the `register_helper` method directly

```rust
(&Helper, &Registry, &mut RenderContext) -> HelperResult;
```

contrib/src/templates/mod.rs
```diff
@@ -214,27 +242,46 @@ impl Template {
...
+    ///
+    /// extern crate rocket;
+    /// extern crate rocket_contrib;
+    ///
     /// use rocket_contrib::Template;
     ///
-    /// // Create a `context`. Here, just an empty `HashMap`.
-    /// let mut context = HashMap::new();
+    /// fn main() {
+    ///     let client =
```
SergioBenitez: I don't like this style. Initialize the `rocket` instance on one line and a `client` on another.

contrib/src/templates/mod.rs
```diff
@@ -214,27 +242,46 @@ impl Template {
...
     /// `None`. If rendering fails, error output is printed to the console.
+    /// `None` is also returned if a Template fairing has not been attached.
     ///
     /// # Example
     ///
     /// ```rust
     /// use std::collections::HashMap;
+    ///
+    /// extern crate rocket;
+    /// extern crate rocket_contrib;
```
SergioBenitez: We shouldn't need this `extern crate`.

contrib/src/templates/mod.rs
```diff
@@ -214,27 +242,46 @@ impl Template {
...
     /// Returns `Some` if the template could be rendered. Otherwise, returns
     /// `None`. If rendering fails, error output is printed to the console.
+    /// `None` is also returned if a Template fairing has not been attached.
     ///
     /// # Example
     ///
     /// ```rust
     /// use std::collections::HashMap;
+    ///
+    /// extern crate rocket;
```
SergioBenitez: Do we need this one? If so, move it to the top of the example and hide it using `# extern crate rocket`. Import the `Client` under the `use rocket_contrib::Template` line: `use rocket::local::Client`.

contrib/src/templates/mod.rs
```diff
@@ -214,27 +242,46 @@ impl Template {
...
+        where S: Into<Cow<'static, str>>, C: Serialize
     {
-        let root = root.as_ref().to_path_buf();
-        Context::initialize(root).and_then(|ctxt| {
-            Template::render(name, context).finalize(&ctxt).ok().map(|v| v.0)
-        })
+        let ctxt = match rocket.state::<Context>() {
+            Some(ctxt) => ctxt,
+            None => {
+                warn_!("Uninitialized template context: missing fairing.");
```
SergioBenitez: Is this the correct indentation level? I think it should probably be `warn!` instead of `warn_!`, but best to trigger the warning manually and confirm with the log output.

contrib/tests/templates.rs
```diff
@@ -1,17 +1,28 @@
...
 fn template_root() -> PathBuf {
     let cwd = env::current_dir().expect("current working directory");
     cwd.join("tests").join("templates")
 }
 
+fn rocket() -> Rocket {
+    let config = Config::build(Environment::Development)
+        .extra("template_dir", template_root().to_str().expect("template directory"))
+        .expect("valid configuration");
+    rocket::custom(config, true).attach(Template::fairing())
```
SergioBenitez: Add an empty line above here.

contrib/src/templates/mod.rs
jebrosen: Without these, I see a lot of "error[E0432]: unresolved import `rocket_contrib`" or error[E0433]: failed to resolve. Maybe a missing `extern crate rocket;`.

The only success I had was adding both extern crate declarations and wrapping the code in a `fn main()`.

SergioBenitez: Okay, no worries. Let's just move these to the top and hide them with `#`. So the top of the file should look like:

```rust
# extern crate rocket;
# extern crate rocket_contrib;

...
```

contrib/src/templates/mod.rs
jebrosen: `warn_!` indents with `=>`; `warn!` does not indent. The indented versions seem to be more consistent with the rest of the crate and with `impl<Responder> for Template`, but to be honest I don't know which is more appropriate for this message.

SergioBenitez: This should be `warn!` since it is being called from a top-level method.

contrib/src/templates/mod.rs
```diff
@@ -201,40 +229,56 @@ impl Template {
         Template { name: name.into(), value: to_value(context).ok() }
     }
 
-    /// Render the template named `name` located at the path `root` with the
-    /// context `context` into a `String`. This method is _very slow_ and should
+    /// Render the template named `name` with the context `context` into a
+    /// `String`. This method should
     /// **not** be used in any running Rocket application. This method should
```
SergioBenitez: Word wrap at 80 cols.

contrib/src/templates/mod.rs
```diff
@@ -201,40 +229,56 @@ impl Template {
...
     ///
     /// The `context` can be of any type that implements `Serialize`. This is
-    /// typically a `HashMap` or a custom `struct`. The path `root` can be
-    /// relative, in which case it is relative to the current working directory,
-    /// or absolute.
+    /// typically a `HashMap` or a custom `struct`.
     ///
     /// Returns `Some` if the template could be rendered. Otherwise, returns
     /// `None`. If rendering fails, error output is printed to the console.
+    /// `None` is also returned if a Template fairing has not been attached.
```
SergioBenitez: Should be `Template`, in ticks.

contrib/src/templates/mod.rs
```diff
@@ -201,40 +229,56 @@ impl Template {
...
     ///
     /// # Example
     ///
     /// ```rust
+    /// # extern crate rocket;
+    /// # extern crate rocket_contrib;
+    ///
```
SergioBenitez: I think if we don't kill this new line we'll have an empty line at the start of the code block in the rendered docs. Can you confirm? If so, kill it.

lib/src/rocket.rs
```diff
@@ -604,6 +604,25 @@ impl Rocket {
         self
     }
 
+    /// Get the managed state of type T, if it exists.
```
SergioBenitez: Let's say...

"Returns `Some` of the managed state value for the type `T` if it exists. Otherwise, returns `None`."

SergioBenitez:
Awesome, awesome, awesome! Thanks so much for persevering, @jebrosen! Will merge this later today. Great job!

SergioBenitez:
Merged in f9f1ed75cd0cc274eb480037e93e3fea6d4f43ac, 80e7339ebe1dbaa763a823a55bec7da2349e52f6, and d79cb9d8f0b860ac66605860332118e63afccebb. Thanks so much, @jebrosen!

PR #525, opened 2017-12-29
Proposal: Diesel documentation change in state.md

chillcaw:
When I was trying to integrate Rocket with Diesel I got a few errors with the tutorial code. I opened [#1441](https://github.com/diesel-rs/diesel/issues/1441) in the diesel repository and made the changes necessary to fix the error to the documentation. Apparently diesel_codegen is deprecated in the latest version of diesel.

### DIFF
--- a/site/guide/state.md
+++ b/site/guide/state.md
@@ -178,9 +178,9 @@ use the following dependencies:
 ```
 [dependencies]
 rocket = "0.4.0-dev"
-diesel = { version = "1.0.0-rc1", features = ["sqlite"] }
+diesel = { version = "*", features = ["sqlite"] }
 r2d2-diesel = "*"
-r2d2 = "*"
+r2d2 = "0.8.*"
 ```
 
 Your `diesel` dependency information will differ. In particular, you should
--- a/site/guide/state.md
+++ b/site/guide/state.md
@@ -180,7 +180,7 @@ use the following dependencies:
 rocket = "0.4.0-dev"
 diesel = { version = "*", features = ["sqlite"] }
 r2d2-diesel = "*"
-r2d2 = "0.8.*"
+r2d2 = "0.8"
 ```
 
 Your `diesel` dependency information will differ. In particular, you should
--- a/site/guide/state.md
+++ b/site/guide/state.md
@@ -166,7 +166,7 @@ fn handler(conn: DbConn) { ... }
 ```
 
 [`diesel`]: http://diesel.rs/
-[`r2d2`]: https://docs.rs/r2d2/0.7.2/r2d2/
+[`r2d2`]: https://docs.rs/r2d2/
 
 ### Dependencies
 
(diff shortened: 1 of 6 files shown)


site/guide/state.md
```diff
@@ -179,9 +179,8 @@ use the following dependencies:
 [dependencies]
 rocket = "0.4.0-dev"
 diesel = { version = "*", features = ["sqlite"] }
-diesel_codegen = { version = "*", features = ["sqlite"] }
 r2d2-diesel = "*"
-r2d2 = "*"
+r2d2 = "0.8.*"
```
Eijebong: Does that even work ?? :o

chillcaw: It does work on my local dev, is it not supposes to?

Eijebong: Nah I meant that I never saw it written this way, this is equivalent to "0.8" which is more common.

SergioBenitez:
This looks pretty good. Can we also update the `todo` example to use the latest diesel and this code as a POC?

chillcaw:
diesel_codegen is deprecated in 0.99 I think.

I've been trying to get the todo example to run, but I can't get the `DATABASE_URL=db/db.sql diesel migrate run` to work.

My Error:
`near "OR": syntax error`

SergioBenitez:
@chillcaw We should remove every use of `*` and replace it with the correct version. Using `*` is a bad idea because it will self-update and result in code breakage.

chillcaw:
@SergioBenitez The code for the todo example works with both `0.99` and `1.0.0-rc1`, but I've specified 1.0.0-rc1 in the documentation to match diesel's getting started docs.

1) Are there any other examples you want me to update for compatibility with diesel `1.0.0-rc1`?
2) Are there any other places in the docs with * that should be changed to specific versions?

site/guide/state.md
```diff
@@ -178,20 +178,17 @@ use the following dependencies:
...
-r2d2 = "*"
+diesel = { version = "1.0.0-rc1", features = ["sqlite"] }
+r2d2-diesel = "1.0.0-rc1"
+r2d2 = "0.8"
 ```
 
-Your `diesel` dependency information will differ. In particular, you should
-specify the latest versions of these libraries as opposed to using a `*`. The
+Your `diesel` dependency information may differ. The
 crates are imported as well:
```
SergioBenitez: Word wrap this and the above line at 80 cols.

SergioBenitez:
Looks like Diesel 1.0 was released! Might as well update these for 1.0 before we merge.

Eijebong:
Upgrading diesel to 1.0 should be easy as there are no changes between 1.0-rc1 and 1.0. You just have to change the version numbers everywhere.

SergioBenitez:
Great! Merged in 401bedebb2664a5b0aae67f9feb4921efc3905b9. Thanks so much @Eijebong, @chillcaw!

PR #526, opened 2018-01-02
Fix: tera_templates using old catchers expressions

t04glovern:
`examples/tera_templates` example uses the old `catchers` syntax and produces the following error when run

```bash
error: cannot find macro `catchers!` in this scope
  --> src/main.rs:45:16
   |
45 |         .catch(catchers![not_found])
   |                ^^^^^^^^

```

Based on the documentation for [rocket/struct.Catcher](https://api.rocket.rs/rocket/struct.Catcher.html#code-generation) I've resolved the issue. 

One of the side effects of doing this meant I also updated the `serde` dependencies to the latest release and confirmed example still functioned appropriately.

### DIFF
--- a/examples/tera_templates/Cargo.toml
+++ b/examples/tera_templates/Cargo.toml
@@ -6,9 +6,9 @@ workspace = "../../"
 [dependencies]
 rocket = { path = "../../lib" }
 rocket_codegen = { path = "../../codegen" }
-serde = "1.0"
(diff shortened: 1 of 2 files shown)


SergioBenitez:
This is the _new_ syntax. It sounds like you're looking at examples on the master branch while using Rocket v0.3. Examples compatible with v0.3 are on the v0.3 branch.

PR #533, opened 2018-01-09
Update cookie dependency revision

Ameobea:
Swapping to a different revision of `cookie` pulls in an updated version of `ring` which works around an issue in recent versions of nightly.  This allows Rocket to build successfully on the latest version of nightly.


SergioBenitez:
This required further changes since _all_ downstream `ring` dependencies had to be updated. I went ahead and made the changes. Thanks for bringing this up!

Ameobea:
Thanks so much for that!

Issue #522, opened 2017-12-28
Attach and Launch fairings shouldn't need to be `Send + Sync`


SergioBenitez:
These fairings are called _once_ on the main thread, so the `Send + Sync` bounds are unnecessary. Furthermore, attach fairings are never stored anywhere, so they don't need the `'static` bound. This also implies that `AdHoc::on_attach` and `AdHoc::on_launch` should accept an `F: FnOnce(..)` instead of an `Fn`.

In all, the loosest bounds are:

  * `on_attach`: none (`FnOnce`)
  * `on_launch`: 'static (`FnOnce`)
  * `on_request`: Send + Sync + 'static (`Fn`)
  * `on_response`: Send + Sync + 'static (`Fn`)

It would be ideal if Rocket imposed the least restrictions.

PR #537, opened 2018-01-13
Reload templates when they have been changed on disk

### LINKED ISSUE #163: Option to reload or prevent caching of templates
It would be super-helpful for development and debugging if Rocket could reload cached templates when I modify them, or else turn off caching altogether in development mode.

jebrosen:
Resolves #163.

There are a few general aspects I would like to get feedback on:
- The dependency on `notify` in Cargo.toml. it would be nice if this is enabled only when both `debug_assertions` and `feature="templates"` are active, i.e. `target.'cfg(all(debug_assertions, feature="templates"))'.dependencies`. In its current form, `notify` is a dependency for all debug builds but remains unused. I will continue to search for a solution to this, but it seems cargo does not support all the cfg() options that rust does, at least not in some places.
  - Side note: what about making this a separate feature, instead of being based on build type? That would make this bullet point obsolete, and allow consumers to control template reloading independently of the build type.
  - EDIT (2018-04-15): Discussions on IRC have led to keeping notify as a dependency in all debug builds, and also adding a new default feature (perhaps "`template_reload`"). The template reload behavior would now happen in `cfg(all(debug_assertions, feature="template_reload"))`.
- ~~The mutex workaround in `custom()`. The `on_attach` callback is `Fn()`, so it must use a workaround such as interior mutability and `Option.take()` in order to move `f` out of its environment. The interior mutability must also be thread-safe due to the `Send + Sync + 'static` bounds, thus the choice of `Mutex`. #522 would make this workaround unnecessary, but that issue is nowhere near trivial to fix and might necessitate some amount of change to the entire Fairings system.~~
  - EDIT (2018-04-12): Changed approach by using a new `Fairing` type, so this point is no longer relevant.
- ~~The `ContextManager` wrapper. Current consumers of `Context` assume they can access the fields of `Context` (that is, they have a `Deref<Target=Context>`, for example `RwLockReadGuard<Context>`).~~
- ~~This is a difficult feature to write automated tests for, and likely unreliable. Not all filesystems on all platforms are guaranteed to have reliable filesystem notification, and failing a CI build because of something nondeterministic like change notification timing would be unfortunate.~~
    - EDIT (2018-07-10): There is a test now, and it strives to be reliable on CI. Hopefully it actually is.
- User-facing and internal documentation.


Mononofu:
Looking forward to this!

In the meantime, pseudo code for one possible way of testing in the face of unreliable change notification, if you don't want to use a mock:

```
initial = fetch_url()

timeout = now() + 5s
modify_template()
while fetch_url() == initial && now() < timeout:
  modify_template()
  sleep(10ms)

expect_ne(initial, fetch_url())
```

This test should pass as long as change notifications fire eventually.

contrib/src/templates/mod.rs
```diff
@@ -140,6 +143,37 @@ pub struct TemplateInfo {
     data_type: ContentType
 }
 
+struct ManagedContext(
```
SergioBenitez: I think I'd rather see:

```rust
#[cfg(not(debug_assertions))]
mod context {
    type Context = Context;

    impl ManagedContext {
        fn new(ctxt: Context) -> ManagedContext {
            ManagedContext(ctxt)
        }

        fn get(&self) -> impl Deref<Target=Context> {
            &self.0
        }
    }
}

#[cfg(debug_assertions)]
mod context {
    use std::sync::{RwLock, RwLockReadGuard, RwLockWriteGuard};

    struct ManagedContext(RwLock<Context>);

    impl ManagedContext {
        fn new(ctxt: Context) -> ManagedContext {
            ManagedContext(RwLock::new(ctxt))
        }

        fn get(&self) -> impl Deref<Target=Context> {
            self.0.read().unwrap()
        }

        fn get_mut(&self) -> impl DerefMut<Target=Context> {
            self.0.write().unwrap()
        }
    }
}
```

Note the use of `impl Deref`.

contrib/src/templates/mod.rs
```diff
@@ -215,13 +249,18 @@ impl Template {
...
-                }
-                None => Err(rocket)
-            }
+            let ctxt = match Context::initialize(template_root, Box::new(f)) {
+                Some(ctxt) => ctxt,
+                None => return Err(rocket),
+            };
+
+            #[cfg(debug_assertions)]
+            let rocket = rocket.attach(AdHoc::on_request(|req, _| {
```
SergioBenitez: I think we're better off creating a custom `Fairing` implementation at this point as opposed to using `AdHoc`. This is really confusing. It took me a while to realize that this is what's actually going to reload the templates.

I think we want to create a new file, `fairing.rs` or something, that contains `ManagedContext` and some new type, `TemplateFairing`, for which we implement `Fairing` for. This time it will have both `on_attach` and `on_request` handlers. The `custom` method should then get really really small, probably just: `TemplateFairing::new()`.

contrib/src/templates/mod.rs
```diff
@@ -215,13 +249,18 @@ impl Template {
...
+                let mc = req.guard::<State<ManagedContext>>().succeeded().expect("context wrapper");
```
SergioBenitez: No panics here! Log an `internal error` and proceed normally.

contrib/src/templates/mod.rs
```diff
@@ -315,7 +355,12 @@ impl Template {
             Status::InternalServerError
         })?;
 
-        let string = ctxt.engines.render(name, &info, value).ok_or_else(|| {
+        let engines = ctxt.engines.as_ref().ok_or_else(|| {
```
SergioBenitez: See my other comment about this being an `Option`.

contrib/src/templates/mod.rs
```diff
@@ -329,12 +374,13 @@ impl Template {
...
 impl Responder<'static> for Template {
     fn respond_to(self, req: &Request) -> response::Result<'static> {
-        let ctxt = req.guard::<State<Context>>().succeeded().ok_or_else(|| {
+        let mc = req.guard::<State<ManagedContext>>().succeeded().ok_or_else(|| {
             error_!("Uninitialized template context: missing fairing.");
             info_!("To use templates, you must attach `Template::fairing()`.");
             info_!("See the `Template` documentation for more information.");
             Status::InternalServerError
         })?;
+        let ctxt = mc.get();
```
SergioBenitez: Just add the `.get()` to the thing above.

contrib/src/templates/mod.rs
SergioBenitez: I don't like that `get_mut` is only included in the `debug_assertions` case. I suppose it's fine, but it would be nice if these were symmetric.

contrib/src/templates/watch.rs
```diff
@@ -0,0 +1,41 @@
+extern crate notify;
+
+use self::notify::{watcher, DebouncedEvent, RecommendedWatcher, RecursiveMode, Watcher};
+
+use std::path::Path;
+use std::sync::mpsc::{channel, Receiver};
+use std::sync::Mutex;
+use std::time::Duration;
+
+pub struct TemplateWatcher {
```
SergioBenitez: Do we really need this extra layer of indirection?

contrib/src/templates/context.rs
```diff
@@ -11,20 +13,41 @@ pub struct Context {
     pub root: PathBuf,
     /// Mapping from template name to its information.
     pub templates: HashMap<String, TemplateInfo>,
-    /// Mapping from template name to its information.
-    pub engines: Engines
+    /// Loaded template engines, or None if there was a failure during a reload
+    pub engines: Option<Engines>,
```
SergioBenitez: Why is this an `Option`? If there was a failure, we should keep using the old engines.

contrib/src/templates/mod.rs
```diff
@@ -203,7 +237,7 @@ impl Template {
     ///     # ;
     /// }
     /// ```
-    pub fn custom<F>(f: F) -> impl Fairing where F: Fn(&mut Engines) + Send + Sync + 'static {
+    pub fn custom<F>(f: F) -> impl Fairing where F: Fn(&mut Engines) + Copy + Send + Sync + 'static {
```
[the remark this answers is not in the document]
jebrosen: Will revert to the previous Mutex<Option<>> hack

contrib/src/templates/mod.rs
jebrosen: `get_mut` can only even exist because of the `RwLock` in `debug_assertions`; the structure is asymmetric because the release build does not need to write to `Context` and can keep the read-only form

contrib/src/templates/mod.rs
jebrosen: I will see what happens with other refactors and check this again, but I believe this was done because `ctxt` borrows from `mc`, so the expression assigned `mc` can't be a temporary.

contrib/src/templates/watch.rs
jebrosen: By extra layer do you mean the `watch.rs` file, or the entire structure? The motivation for this was to keep the `notify` dependency relatively isolated in the code.

contrib/src/templates/context.rs
```diff
@@ -11,12 +11,15 @@ pub struct Context {
...
-    pub engines: Engines
+    /// Loaded template engines
+    pub engines: Engines,
 }
 
 impl Context {
-    pub fn initialize(root: PathBuf) -> Option<Context> {
+    pub fn initialize(
+        root: PathBuf,
+        customize_callback: &(Fn(&mut Engines) + Send + Sync + 'static),
```
jebrosen: Would be cleaner to remove the Send+Sync+'static bounds here, it's only at the level of the Fairing that these bounds are necessary

contrib/src/templates/fairing.rs
```diff
@@ -0,0 +1,121 @@
...
+            custom_callback: Box::new(f),
+        }
+    }
+}
+
+impl Fairing for TemplateFairing {
+    fn info(&self) -> Info {
+        Info {
+            name: "Templates",
+            kind: Kind::Attach | Kind::Request,
```
jebrosen: There is probably a speed boost to be gained by only specifying Kind::Request in debug builds, where it is actually used.

contrib/src/templates/watch.rs
jebrosen: These fields and/or behavior are small enough they could maybe move into the debug-only version of `mod context` in `fairing.rs`.

contrib/src/templates/fairing.rs
```diff
@@ -0,0 +1,121 @@
...
+
+        pub fn get<'a>(&'a self) -> impl Deref<Target=Context> + 'a {
+            self.context.read().unwrap()
+        }
+
+        fn get_mut<'a>(&'a self) -> impl DerefMut<Target=Context> + 'a {
+            self.context.write().unwrap()
+        }
+
+        pub fn reload_if_needed(&self, custom_callback: &(Fn(&mut Engines) + Send + Sync + 'static)) {
```
jebrosen: Same as above about Send+Sync+'static

contrib/src/templates/context.rs
```diff
@@ -45,7 +48,8 @@ impl Context {
             }
         }
 
-        Engines::init(&templates).map(|engines| {
+        Engines::init(&templates).map(|mut engines| {
+            customize_callback(&mut engines);
```
SergioBenitez: We should call the user's function after initialization. That way, we don't have to pass it in to `initialize` at all, and `Context` can stay ignorant that it even exists.

contrib/src/templates/fairing.rs
```diff
@@ -0,0 +1,125 @@
...
+    use super::{Context, Engines, TemplateWatcher};
+
+    /// Wraps a Context. With `cfg(debug_assertions)` active, this structure
+    /// additionally provides a method to reload the context at runtime.
+    pub struct ContextManager{ context: RwLock<Context>, watcher: Option<TemplateWatcher> }
+
+    impl ContextManager {
+        pub fn new(ctxt: Context) -> ContextManager {
+            let root = ctxt.root.clone();
+            ContextManager {
```
SergioBenitez: I think you can reorder this so that you don't need the `let` binding:

```rust
ContextManager {
    watcher: TemplateWather::new(ctxt.root.clone()),
    context: RwLock::new(ctxt)
}
```

contrib/src/templates/fairing.rs
```diff
@@ -0,0 +1,125 @@
...
+
+        fn get_mut<'a>(&'a self) -> impl DerefMut<Target=Context> + 'a {
+            self.context.write().unwrap()
+        }
+
+        pub fn reload_if_needed<F: Fn(&mut Engines)>(&self, custom_callback: F) {
+            if self.watcher.as_ref().map(TemplateWatcher::needs_reload).unwrap_or(false) {
+                warn!("Change detected, reloading templates");
+                let mut ctxt = self.get_mut();
+                match Context::initialize(ctxt.root.clone(), custom_callback) {
```
SergioBenitez: Maybe an `if let` will work better here?

contrib/src/templates/fairing.rs
```diff
@@ -0,0 +1,125 @@
...
+                };
+            }
+        }
+    }
+}
+
+pub use self::context::ContextManager;
+
+pub struct TemplateFairing {
+    custom_callback: Box<Fn(&mut Engines) + Send + Sync + 'static>,
```
SergioBenitez: You should create a type alias for the function type instead of using the long type everywhere.

contrib/src/templates/fairing.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+impl TemplateFairing {
+    pub fn new<F>(f: F) -> TemplateFairing
+        where F: Fn(&mut Engines) + Send + Sync + 'static
+    {
+        TemplateFairing {
```
SergioBenitez: If you name the argument the same as the field, then you can do: `TemplateFairing { custom_callback }`.

contrib/src/templates/fairing.rs
```diff
@@ -0,0 +1,125 @@
...
+            custom_callback: Box::new(f),
+        }
+    }
+}
+
+impl Fairing for TemplateFairing {
+    fn info(&self) -> Info {
+        Info {
+            name: "Templates",
+            kind: if cfg!(debug_assertions) {
```
SergioBenitez: Better to use a `#[cfg(debug_assertions)]` on each field.

contrib/src/templates/fairing.rs
```diff
@@ -0,0 +1,125 @@
...
+            Ok(dir) => template_root = rocket.config().root_relative(dir),
+            Err(ConfigError::NotFound) => { /* ignore missing configs */ }
+            Err(e) => {
+                e.pretty_print();
+                warn_!("Using default templates directory '{:?}'", template_root);
+            }
+        };
+
+        match Context::initialize(template_root.clone(), &*self.custom_callback) {
+            Some(ctxt) => { Ok(rocket.manage(ContextManager::new(ctxt))) }
```
SergioBenitez: As in the previous comment, we don't need to pass the function to the `Context` initialization at all. Just call it after the context returns. Better code, and nicer separation of concerns to boot.

contrib/src/templates/fairing.rs
```diff
@@ -0,0 +1,125 @@
...
+            None => Err(rocket),
+        }
+    }
+
+    fn on_request(&self, _req: &mut Request, _data: &Data) {
+        #[cfg(debug_assertions)]
```
SergioBenitez: Use this on the `fn`. Not need to create a new block here.

contrib/src/templates/fairing.rs
```diff
@@ -0,0 +1,125 @@
...
+        {
+            use rocket::State;
+            let cm = _req.guard::<State<ContextManager>>().unwrap();
```
SergioBenitez: Rename the argument to `req` since it's being used.

contrib/src/templates/mod.rs
```diff
@@ -5,11 +5,14 @@ extern crate glob;
 #[cfg(feature = "tera_templates")] mod tera_templates;
 #[cfg(feature = "handlebars_templates")] mod handlebars_templates;
 mod engine;
+mod fairing;
 mod context;
+#[cfg(debug_assertions)] mod watch;
```
SergioBenitez: Move this to line 7, after the rest of the `cfg`-dependent imports.

contrib/src/templates/watch.rs
SergioBenitez: Yeah, that's what I'm thinking as well. Play around with it a bit. If we don't need the indirection, and it doesn't make code significantly cleaner, let's get rid of it.

contrib/src/templates/fairing.rs
jebrosen: What should this be named? TemplateCustomizationCallback? Or maybe just CustomizationCallback?

jebrosen:
Thanks! Addressed most of those comments.

I will check again on the cfg situation -- the problem I encountered before was that `Cargo.toml` didn't seem to support the same cfg flags that `rustc` does.

As for the `TemplateWatcher`, I will try moving that into `ContextManager` and see how it changes things.

contrib/src/templates/fairing.rs
SergioBenitez: Maybe just `CustomEngineFn`? Something ending in `Fn` would be good.

jebrosen: This is all that changes, is it still worth it?
```diff
diff --git a/contrib/src/templates/fairing.rs b/contrib/src/templates/fairing.rs
index ab15dd3..74c6e12 100644
--- a/contrib/src/templates/fairing.rs
+++ b/contrib/src/templates/fairing.rs
@@ -104,12 +104,14 @@ mod context {
 
 pub use self::context::ContextManager;
 
+pub type CustomEngineFn = Fn(&mut Engines) + Send + Sync + 'static;
+
 pub struct TemplateFairing {
-    custom_callback: Box<Fn(&mut Engines) + Send + Sync + 'static>,
+    custom_callback: Box<CustomEngineFn>,
 }
 
 impl TemplateFairing {
-    pub fn new(custom_callback: Box<Fn(&mut Engines) + Send + Sync + 'static>) -> TemplateFairing
+    pub fn new(custom_callback: Box<CustomEngineFn>) -> TemplateFairing
     {
         TemplateFairing { custom_callback }
     }
```

The only other place all the bounds show up is in a trait bound, where the type alias can't be used.

contrib/src/templates/fairing.rs
```diff
@@ -0,0 +1,155 @@
...
+        _watcher: Option<RecommendedWatcher>,
+        /// Receive end of the message queue for events from `_watcher`
+        recv_queue: Mutex<Receiver<DebouncedEvent>>,
+    }
+
+    impl ContextManager {
+        pub fn new(ctxt: Context) -> ContextManager {
+            let (tx, rx) = channel();
+
+            let _watcher = if let Ok(mut watcher) = watcher(tx, Duration::from_secs(1)) {
```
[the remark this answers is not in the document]
jebrosen: They can be used (somewhat awkwardly), but they don't appear to clean it up. I'll take another stab at it later though.

jebrosen:
The trait alias workaround didn't work as easily as I expected, with some errors about missing the `Output` associated type of `FnOnce`, which I attempted to solve a few ways demonstrated here:

https://play.rust-lang.org/?gist=cd5cdbf4680729afb0c8a8feb724bd37&version=nightly

jebrosen:
(adapted from IRC) I tried adding `template_reload` as a feature that is on by default, but I realized most of the docs use `default_features = false` on rocket_contrib and I suspect that isn't an uncommon situation. Alternatively, an opt-*in* feature called no_template_reload would permit the desired behavior. Thoughts?

jebrosen:
Sorry about going off the radar for a while. In summary, I think this is what's still unaddressed as far as design:

* [ ] Trying to not repeat `Fn(&mut Engines) + Send + Sync + 'static` by using an alias. A trait alias ends up being repetitive in other ways (`<Output=()>`) because of rust-lang/rust#24010 AKA rust-lang/rust#23856 .
* [ ] Gating by a feature, e.g. "template_reload". The main question I have about this is whether it makes sense to be a default-on feature, when several docs for `rocket_contrib` and therefore many users are likely to use `default_features=false` anyway.
* [ ] Tests. These should, at a minimum, verify that writing to a file in the templates directory triggers a template reload in the appropriate profile/feature combinations. I haven't thought much about testing the different error conditions.

I would like feedback/a decision on the first two points, and I was hoping to write some tests within the next few days.

SergioBenitez:
@jebrosen Looks like the test failed in debug mode.

jebrosen:
Local test output:
```
GET /:
Warning: Change detected, reloading templates
    => Error: No matching routes for GET /.
    => Warning: Responding with 404 Not Found catcher.
```

Looks like the change is not picked up on the CI. I will switch from using notify's `DebouncedEvent` to `RawEvent` in case it is a timing issue.

jebrosen:
So far, that's one CI build that failed with raw_watcher, and one CI build that succeeded with raw_watcher. I haven't had the test fail locally (yet).

SergioBenitez:
Looks like a race with the test. Perhaps try adding a `flush()` or `sink()` before checking that the new template was picked up. Additionally, or instead, you may want to add a short, say, 100ms delay between writing and checking for a reload.

jebrosen:
Ah. I've looked over the different methods available on `File` and their implementations and I will try `sync_all` for this. I had assumed that `Drop` would do the right thing, but I forgot that an immediate sync isn't guaranteed. In the meantime, do you have any thoughts on my [earlier comments](https://github.com/SergioBenitez/Rocket/pull/537#issuecomment-385215782) about the trait alias and the feature declaration?

contrib/lib/tests/templates.rs
```diff
@@ -116,5 +116,61 @@ mod templates_tests {
...
+            let client = Client::new(rocket()).unwrap();
+
+            // verify that the initial content is correct
+            let initial_rendered = Template::show(client.rocket(), RELOAD_TEMPLATE, ());
+            assert_eq!(initial_rendered, Some(INITIAL_TEXT.into()));
+
+            // write a change to the file
+            write_file(&reload_path, NEW_TEXT);
+
+            let mut attempts = 0;
```
SergioBenitez: I'd prefer if it was written the following way:

```rust
for i in 0..6 {
    // dispatch any request to trigger a template reload
    client.get("/").dispatch();

    // if the new content is correct, we are done
    let new_rendered = Template::show(client.rocket(), RELOAD_TEMPLATE, ());
    if new_rendered == Some(NEW_TEXT.into()) {
        remove_file(reload_path).expect("delete template file");
        return;
    }

    thread::sleep(Duration::from_millis(250));
}

panic!("failed to reload template in 1.5s");
```

contrib/lib/src/templates/fairing.rs
```diff
@@ -0,0 +1,154 @@
...
+                None
+            };
+
+            ContextManager {
+                watcher,
+                context: RwLock::new(ctxt),
+            }
+        }
+
+        pub fn get<'a>(&'a self) -> impl Deref<Target=Context> + 'a {
```
SergioBenitez: We should rename `get()` and `get_mut()` to `context()` and `context_mut()`.

contrib/lib/src/templates/fairing.rs
```diff
@@ -0,0 +1,154 @@
...
+}
+
+pub use self::context::ContextManager;
+
+pub struct TemplateFairing {
+    custom_callback: Box<Fn(&mut Engines) + Send + Sync + 'static>,
+}
+
+impl TemplateFairing {
+    pub fn new(custom_callback: Box<Fn(&mut Engines) + Send + Sync + 'static>) -> TemplateFairing {
```
SergioBenitez: It looks like this is called only once. Why not write `TemplateFairing { custom_callback }` where this is called instead of having this method? That would be just as clear (if not clearer due to the heavy type signature here) while also being fewer lines of code.

contrib/lib/src/templates/fairing.rs
```diff
@@ -0,0 +1,154 @@
...
+        TemplateFairing { custom_callback }
+    }
+}
+
+impl Fairing for TemplateFairing {
+    fn info(&self) -> Info {
+        Info {
+            name: "Templates",
+            #[cfg(debug_assertions)]
+            kind: Kind::Attach | Kind::Request,
```
SergioBenitez: A short comment explaining why there's a difference between debug and not debug would be nice.

contrib/lib/src/templates/fairing.rs
```diff
@@ -0,0 +1,154 @@
...
+            #[cfg(not(debug_assertions))]
+            kind: Kind::Attach,
+        }
+    }
+
+    fn on_attach(&self, rocket: Rocket) -> Result<Rocket, Rocket> {
```
SergioBenitez: A short docstring here would be nice. Something like:

```rust
/// Initializes the template context. Templates will be searched for in the `template_dir` config variable
/// or the default `DEFAULT_TEMPLATE_DIR`. The user's callback, if any was supplies, is called to customize
/// the template engines. When debug is enabled, the `ContextManager::new` method initializes a directory
/// watcher for auto-reloading of templates. 
```

contrib/lib/src/templates/fairing.rs
```diff
@@ -0,0 +1,154 @@
...
+                (self.custom_callback)(&mut ctxt.engines);
+                Ok(rocket.manage(ContextManager::new(ctxt)))
+            }
+            None => Err(rocket),
+        }
+    }
+
+    #[cfg(debug_assertions)]
+    fn on_request(&self, req: &mut ::rocket::Request, _data: &::rocket::Data) {
+        let cm = req.guard::<::rocket::State<ContextManager>>().unwrap();
```
SergioBenitez: Let's use an `expect()` with a short explanation of why this can't fail.

contrib/lib/src/templates/mod.rs
```diff
@@ -205,26 +207,10 @@ impl Template {
...
-                    f(&mut ctxt.engines);
-                    Ok(rocket.manage(ctxt))
-                }
-                None => Err(rocket)
-            }
-        })
+    pub fn custom<F>(f: F) -> impl Fairing
+        where F: Fn(&mut Engines) + Send + Sync + 'static
+    {
+        TemplateFairing::new(Box::new(f))
```
SergioBenitez: We should do `TemplateFairing { custom_callback: f }`.

contrib/lib/src/templates/mod.rs
```diff
@@ -329,12 +315,13 @@ impl Template {
...
-        let ctxt = req.guard::<State<Context>>().succeeded().ok_or_else(|| {
+        let cm = req.guard::<State<ContextManager>>().succeeded().ok_or_else(|| {
             error_!("Uninitialized template context: missing fairing.");
             info_!("To use templates, you must attach `Template::fairing()`.");
             info_!("See the `Template` documentation for more information.");
             Status::InternalServerError
         })?;
+        let ctxt = cm.get();
 
         let (render, content_type) = self.finalize(&ctxt)?;
```
SergioBenitez: Presumably we can't call `cm.get()` here directly because of some borrow checking issues?

contrib/lib/tests/templates.rs
```diff
@@ -116,5 +116,57 @@ mod templates_tests {
             let response = client.get("/tera/test").dispatch();
             assert_eq!(response.status(), Status::NotFound);
         }
+
+        #[cfg(debug_assertions)]
+        #[test]
```
SergioBenitez: Can we flip these two attributes so that longer one is at the bottom?

contrib/lib/tests/templates.rs
```diff
@@ -116,5 +116,57 @@ mod templates_tests {
...
+        fn test_template_reload() {
+            use rocket::local::Client;
```
SergioBenitez: Do the `std` imports before the `rocket` imports.

contrib/lib/tests/templates.rs
```diff
@@ -116,5 +116,57 @@ mod templates_tests {
...
+            use std::thread;
+            use std::time::Duration;
+
+            fn write_file(path: &Path, text: &str) {
+                let mut file = File::create(path).expect("open file");
+                file.write_all(text.as_bytes()).expect("write file");
+                file.sync_all().expect("sync file");
+            }
+
+            const RELOAD_TEMPLATE: &'static str = "hbs/reload";
```
SergioBenitez: `const`s should come before everything else. Note that we don't need the `'static` since that's implied.

contrib/lib/tests/templates.rs
```diff
@@ -116,5 +116,57 @@ mod templates_tests {
...
+            const INITIAL_TEXT: &'static str = "initial";
+            const NEW_TEXT: &'static str = "reload";
+
+            let reload_path = Path::join(
+                Path::new(env!("CARGO_MANIFEST_DIR")),
+                "tests/templates/hbs/reload.txt.hbs"
+            );
+
+            write_file(&reload_path, INITIAL_TEXT);
```
SergioBenitez: Add a comment above this. I suppose it's `// set up the template with the initial text _before_ initializing the Rocket instance so that it's picked up on the initial load`.

contrib/lib/src/templates/mod.rs
jebrosen: Yes, and I think I may have finally figured out why. The `State` type implements `Deref`, whose lifetime is tied to the `State` object itself. Using `.inner()` returns a longer lifetime and "escapes" out of the temporary `State` value.

contrib/lib/src/templates/fairing.rs
```diff
@@ -0,0 +1,154 @@
...
+        pub fn reload_if_needed<F: Fn(&mut Engines)>(&self, custom_callback: F) {
+            self.watcher.as_ref().map(|w| {
+                let rx = w.1.lock().expect("receive queue");
+                let mut changed = false;
+                while let Ok(_) = rx.try_recv() {
+                    changed = true;
+                }
+
+                if changed {
+                    warn!("Change detected, reloading templates");
```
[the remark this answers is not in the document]
jebrosen: Should this maybe be `info_!` as well? I have to admit I'm not sure when to use the indented forms and when not to.

contrib/lib/src/templates/fairing.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    impl ContextManager {
+        pub fn new(ctxt: Context) -> ContextManager {
+            let (tx, rx) = channel();
+
+            let watcher = if let Ok(mut watcher) = raw_watcher(tx) {
+                if watcher.watch(ctxt.root.clone(), RecursiveMode::Recursive).is_ok() {
+                    Some((watcher, Mutex::new(rx)))
+                } else {
+                    warn_!("Could not monitor the templates directory for changes.");
```
SergioBenitez: Ah, so, the `_` just adds a level of indentation to allow for grouping messages. Here, we'd want the one on the top _not_ to use the `_` so we get this:

```rust
Warning: Could not monitor the templates directory for changes.
    => Live template reload will be unavailable.
```

In other words, the `_` is for cosmetic/grouping purposes.

contrib/lib/src/templates/fairing.rs
```diff
@@ -0,0 +1,164 @@
...
+
+        pub fn context<'a>(&'a self) -> impl Deref<Target=Context> + 'a {
+            self.context.read().unwrap()
+        }
+
+        fn context_mut<'a>(&'a self) -> impl DerefMut<Target=Context> + 'a {
+            self.context.write().unwrap()
+        }
+
+        /// Checks whether any template files have changed on disk.  If
```
SergioBenitez: Word wrap at 80 columns.

contrib/lib/src/templates/fairing.rs
```diff
@@ -0,0 +1,164 @@
...
+                    };
+                }
+            });
+        }
+    }
+}
+
+pub use self::context::ContextManager;
+
+/// The TemplateFairing initializes the template system on attach,
```
SergioBenitez: Word wrap at 80 cols.

contrib/lib/src/templates/fairing.rs
```diff
@@ -0,0 +1,164 @@
...
+/// running custom_callback after templates have been loaded.
+/// In debug mode, the fairing checks for modifications to templates
+/// before every request and reloads them if necessary.
+pub struct TemplateFairing {
+    pub(crate) custom_callback: Box<Fn(&mut Engines) + Send + Sync + 'static>,
```
SergioBenitez: Can we get a short docstring on `custom_callback` as well?

contrib/lib/src/templates/mod.rs
```diff
@@ -76,6 +78,11 @@ const DEFAULT_TEMPLATE_DIR: &'static str = "templates";
 /// [Serde](https://github.com/serde-rs/json) and would serialize to an `Object`
 /// value.
 ///
+/// In debug mode (without the `--release` flag passed to `cargo`), templates
+/// will be automatically reloaded from disk if any changes have been made to
+/// the templates directory since the previous request. In release builds,
+/// template reloading is unavailable for better performance.
```
jebrosen: If you like this wording, I can add something similar to the guide.

SergioBenitez: I do like this wording, though I'd say "disabled" instead of "unavailable". I'd also make it clear that there's no way to enable it.

SergioBenitez:
Merged in 491b04cf5abed3d23bb9cefa7d94a4777cd90288. Whew! So happy to finally land this. Excellent job as always, @jebrosen!

PR #544, opened 2018-01-18
Content length handling

### LINKED ISSUE #514: Allow for setting a custom Content-Length

This is a follow on from a PR (https://github.com/SergioBenitez/Rocket/pull/482) I made against this repository last month. It was not discussed, and the use-case may be a little unclear. I am creating an issue as I think this should be considered in the Rocket codebase as it is not possible to implement around it.

There are several specs that require setting a custom _Content-Length_, especially in the case of a HEAD request. 

The current behaviour does not comply with the HTTP Spec.

Excerpt from the spec (highlighting mine):

> The Content-Length entity-header field indicates the size of the entity-body, in decimal number of OCTETs, sent to the recipient or, __in the case of the HEAD method, the size of the entity-body that would have been sent had the request been a GET__.
https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html

 The following function [write_response](https://github.com/hamhut1066/Rocket/blob/dfc3ddebf162c36342aca1604f2b17997ecdb343/lib/src/rocket.rs#L135) overwrites the _Content-Length_ regardless of input, so any library/application that utilises rocket must have their Content-Length forcibly changed.

# General Comments

I welcome a discussion on this, (perhaps on reasons why this may _not_ be a good idea). I am relying on this behaviour, so hope we can work out how to implement it.

Thanks!

amouat:
Closes #514.

I moved all logic for setting Content-Length out of `write_response` and into `dispatch`.

I expanded the tests in lib/tests/head_handling.rs.

### DIFF
--- a/lib/src/rocket.rs
+++ b/lib/src/rocket.rs
@@ -131,11 +131,6 @@ impl Rocket {
             hyp_res.headers_mut().append_raw(name, value);
         }
 
-        if response.body().is_none() {
-            hyp_res.headers_mut().set(header::ContentLength(0));
-            return hyp_res.start()?.end();
-        }
-
         match response.body() {
             None => {
                 hyp_res.headers_mut().set(header::ContentLength(0));
--- a/lib/src/rocket.rs
+++ b/lib/src/rocket.rs
@@ -21,7 +21,7 @@ use error::{Error, LaunchError, LaunchErrorKind};
 use fairing::{Fairing, Fairings};
 
 use http::{Method, Status, Header};
-use http::hyper::{self, header};
+use http::hyper;
 use http::uri::Uri;
 
 /// The main `Rocket` type: used to mount routes and catchers and launch the
@@ -133,11 +133,9 @@ impl Rocket {
 
         match response.body() {
             None => {
-                hyp_res.headers_mut().set(header::ContentLength(0));
                 hyp_res.start()?.end()
             }
-            Some(Body::Sized(body, size)) => {
-                hyp_res.headers_mut().set(header::ContentLength(size));
+            Some(Body::Sized(body, _)) => {
                 let mut stream = hyp_res.start()?;
                 io::copy(body, &mut stream)?;
                 stream.end()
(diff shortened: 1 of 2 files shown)


amouat:
Test failure is due to out-of-sync libraries. The fix I used is to bump `rusqlite` to "0.13" in `examples/raw_sqlite/Cargo.toml`

SergioBenitez:
It's failing because of https://github.com/rust-lang/cargo/issues/4902.

SergioBenitez:
This changeset is much more complicated than it needs to be. All that needs to happen is that the body needs to be stripped using `Response.strip_body()` if the user's handler responded successfully to a `HEAD` request. The code already does the rest. This should be a line or two of changes.

amouat:
I'm new to Rust and know very little of Rocket internals, so there may well be a better way of doing this. I'm afraid I can't see it though, and I did play with a few approaches.

I'll take you through my logic. In order to correctly meet the HTTP spec, we need to:

 1) Set the content-length on HEAD requests to the same as the GET equivalent.
 2) Set the content-length on empty body requests to 0 (I'm not sure if this is essential)
 3) Do not set the content-length for streaming responses
 4) Strip the body from HEAD requests

In the original code, the body is stripped before the content-length is set. To correctly set the length, we either need to move the code that sets the content-length, or save the value somewhere before stripping the body. I felt it was simpler to move the content-length code to dispatch.

The final match branch can probably be simplified to `_` at the loss of some explicitness. It could also be changed to an if/else, but I think the match is cleaner.

I wonder if it would be possible to automatically generate HEAD routes at start-up rather than handle them in dispatch if they aren't set. This would simplify dispatch and get rid of the need to cache the method type for this patch.

SergioBenitez:
Unfortunately I don't have a lot of time at the moment, so I went ahead and took care of it, @amouat. Apologies for being a bit short on my previous response.

amouat:
No worries, I had assumed strip_body set the size to 0, which was wrong.

PR #547, opened 2018-01-22
Use from_reader_eager instead of from_reader when deserialize Json

messense:
`serde_json::from_reader` is considerably slow for some reason.

https://github.com/serde-rs/json/issues/160

### DIFF
--- a/contrib/src/json.rs
+++ b/contrib/src/json.rs
@@ -14,6 +14,23 @@ use serde_json;
 
 pub use serde_json::error::Error as SerdeError;
 
+/// Like [`from_reader`] but eagerly reads the content of the reader to a string
+/// and delegates to `from_str`.
+///
+/// [`from_reader`]: https://docs.serde.rs/serde_json/fn.from_reader.html
+fn from_reader_eager<R, T>(mut reader: R) -> serde_json::Result<T>
+where
+    R: Read,
+    T: DeserializeOwned,
+{
+    let mut s = String::new();
+    if let Err(io_err) = reader.read_to_string(&mut s) {
+        // Error::io is private to serde_json. Do not use.
+        return Err(SerdeError::io(io_err));
+    }
+    serde_json::from_str(&s)
+}
+
 /// The JSON type: implements `FromData` and `Responder`, allowing you to easily
 /// consume and respond with JSON.
 ///
@@ -97,7 +114,7 @@ impl<T: DeserializeOwned> FromData for Json<T> {
         }
 
         let size_limit = request.limits().get("json").unwrap_or(LIMIT);
-        serde_json::from_reader(data.open().take(size_limit))
+        from_reader_eager(data.open().take(size_limit))
             .map(|val| Json(val))
             .map_err(|e| { error_!("Couldn't parse JSON body: {:?}", e); e })
             .into_outcome(Status::BadRequest)


contrib/src/json.rs
```diff
@@ -97,7 +97,13 @@ impl<T: DeserializeOwned> FromData for Json<T> {
         }
 
         let size_limit = request.limits().get("json").unwrap_or(LIMIT);
-        serde_json::from_reader(data.open().take(size_limit))
+        let mut buf = String::new();
+        data.open()
+            .take(size_limit)
+            .read_to_string(&mut buf)
+            .map_err(|e| { error_!("IO Error: {:?}", e); e })
+            .unwrap();
```
messense: I am not sure how to deal with this `unwrap()`.

messense:
Before this patch:

image

After this patch:

image

I was using Rocket v0.3.6 when profiling this.

Dowwie:
@messense  what library are you using to profile?

messense:
@Dowwie `Instruments.app` bundled within Xcode

messense:
Just did a profile on using `BufReader`:

```patch
diff --git a/contrib/src/json.rs b/contrib/src/json.rs
index a461127..ff2e441 100644
--- a/contrib/src/json.rs
+++ b/contrib/src/json.rs
@@ -1,5 +1,5 @@
 use std::ops::{Deref, DerefMut};
-use std::io::Read;
+use std::io::{Read, BufReader};

 use rocket::outcome::{Outcome, IntoOutcome};
 use rocket::request::Request;
@@ -98,7 +98,8 @@ impl<T: DeserializeOwned> FromData for Json<T> {
         }

         let size_limit = request.limits().get("json").unwrap_or(LIMIT);
-        serde_json::from_reader(data.open().take(size_limit))
+        let reader = BufReader::new(data.open().take(size_limit));
+        serde_json::from_reader(reader)
             .map(|val| Json(val))
             .map_err(|e| { error_!("Couldn't parse JSON body: {:?}", e); e })
             .into_outcome(Status::BadRequest)
```

image

Much slower than `serde_json::from_str`

SergioBenitez:
That is...surprising. Have you investigated why `serde` performs so differently? The internal returned `DataStream` is already buffered, so I wouldn't expect any gains there. But it's surprising to see such a difference between `from_str` and `from_reader`.

dtolnay:
We optimize `from_str` in lots of ways that it is not possible to optimize `from_reader`. For example `from_reader` needs to track line and column position as it goes, in case a parsing error happens later and we need to report the position; `from_str` can figure out line and column lazily only in the case that an error happened.

contrib/src/json.rs
SergioBenitez: Unfortunately it doesn't look like Serde's error type can be created from an `io:Error`, so we'll have to change the error type here so that it returns an `io::Error` instead. We can then use [`impl From<serde::Error> for io::Error`](https://docs.serde.rs/serde_json/error/struct.Error.html#impl-From%3CError%3E). Alternatively, @dtolnay: would it be possible to add a `From<io::Error> for serde_json::error::Error` implementation?

dtolnay: I would prefer not to have `From<io::Error> for serde_json::Error`. JSON errors are for Serde to create.

messense: Switched to `io::Error`

dtolnay:
I filed https://github.com/serde-rs/json/issues/407 to support this in serde_json. Let me know if anyone has API suggestions.

contrib/src/json.rs
```diff
@@ -88,18 +88,23 @@ impl<T> Json<T> {
...
-    fn from_data(request: &Request, data: Data) -> data::Outcome<Self, SerdeError> {
+    fn from_data(request: &Request, data: Data) -> data::Outcome<Self, io::Error> {
         if !request.content_type().map_or(false, |ct| ct.is_json()) {
             error_!("Content-Type is not JSON.");
             return Outcome::Forward(data);
         }
 
         let size_limit = request.limits().get("json").unwrap_or(LIMIT);
-        serde_json::from_reader(data.open().take(size_limit))
+        let mut buf = String::new();
```
gsquire: Could you use `with_capacity` here with `size_limit` to optimize this a little more?

SergioBenitez: We don't want to over-allocate memory. Perhaps we could choose a safe lower-bound: something like 1k or 4k.

SergioBenitez:
@messense Lets copy the code from https://github.com/serde-rs/json/issues/407 (https://docs.rs/eager_json/0.1.0/src/eager_json/lib.rs.html#13-24) into Rocket. Then we could do this correctly. :)

SergioBenitez:
Merged in ae8e9025d21c8ff512d8a0917494e21b9eaafc2e. Thanks so much for finding, reporting, and fixing this @messense! Awesome work. :)

messense:
I will measure the performance implications with different lower bounds when I have the time.

Would be great to have this patch backported to 0.3.x version!

messense:
Regarding the buffer capacity question, I just did a simple instrument with the `json` example crate.

I am using the following Python script to send some requests the server.

```python
# -*- coding: utf-8 -*-
from concurrent.futures import ThreadPoolExecutor

from requests import Session

session = Session()


def post_json(index):
    return session.post(
        'http://localhost:8000/message/{}'.format(index),
        json={
            'contents': 'Hello World!' * 1000
        }
    )


def bench():
    with ThreadPoolExecutor() as executor:
        fs = [executor.submit(post_json, i) for i in range(10000)]
        _rs = [r.result() for r in fs]

if __name__ == '__main__':
    bench()

```

#### String::new()

image

#### String::with_capacity(1024)

image

#### String::with_capacity(4096)

image

Though I am not sure this is the correct way to measure its performance. To me it seems that use `1K` buffer is reasonable.

Issue #528, opened 2018-01-06
Support brotli compression of responses


Martin1887:
Hello.

I think that brotli support will be great. A fast (though not as fast as original C Brotli) implementation in Rust is available in https://github.com/dropbox/rust-brotli. All modern browsers support Brotli compression, so it could speed up many applications!

This feature request is related to #195.

Maybe it could be implemented with custom responders as it is suggested in #195, but I think that this a feature that can be good for all requests, and therefore wrapping all responses in a custom Brotli responder could be boilerplate in many applications. 

Thanks and regards.

SergioBenitez:
This can be implemented as both a `Responder` and as a `Fairing`, if you want all responses compressed. This can be done entirely outside of Rocket. I don't think something like this belongs in the core library, but perhaps a nice implementation can make its way into `contrib`. Because of this, I'm closing out this issue as declined.

If someone implements this as a `Fairing` and `Responder`, I'm happy to review it for inclusion in `contrib`!

PR #550, opened 2018-01-27
Compression Fairing in contrib crate

Martin1887:
Hello.

This pull request provides compression support to all responses using the optional feature 'Compression', as requested in #528.

This feature provides brotli and gzip compression with the following requirements:

- Gzip and Brotli are used only when the request has a 'Accept-Encoding' header with gzip or brotli respectively.
- When Gzip and Brotli are supported, Brotli is used.
- If the response has a 'Content-Encoding' header gzip, then gzip is used instead brotli.
- If the response has no 'Content-Encoding' headers or it has br Content-Encoding, brotli is used.
- Responses with 'Content-Type' 'image/*' are never compressed.
- Responses with 'Content-Type' 'font/*' and brotli support are compressed in font brotli mode.
- Responses with 'Content-Type' 'text/*' and brotli support are compressed in text brotli mode.
- For the rest of content types, generic brotli mode is used.
- Brotli quality is set to 2 for fast compression with compression ratio similar to gzip.
- Tests have been created to test the funcionalities and all of them pass.

Regards.


contrib/src/compression.rs
```diff
@@ -0,0 +1,166 @@
...
+
+impl Compression {
+    fn remove_content_header<'a, 'b: 'a>(to_delete: &'a str,
+                                         headers: HeaderMap<'b>,
+                                         response: &mut Response<'a>) {
+        response.remove_header("Content-Encoding");
+        for enc in headers.into_iter()
+            .filter(|x| x.name() == "Content-Encoding" && x.value() != to_delete) {
+            let header = enc.clone();
+            println!("{:?}", header);
```
messense: Remove this debugging print?

Martin1887:
Sorry for the println!, I have removed it.

I think that the build failed is because a DATABASE_URL environment variable is needed to compile the contrib crate for a example that uses it.

Thanks for the finding!

Regards.

Martin1887:
The error in the build seems to be caused by an error compiling TestOpts for Travis, it is not related to the last commit or the pull request:

```
error[E0560]: struct `test::TestOpts` has no field named `quiet`
  --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/compiletest_rs-0.3.5/src/lib.rs:97:9
   |
97 |         quiet: config.quiet,
   |         ^^^^^^ `test::TestOpts` does not have this field
   |
   = note: available fields are: `list`, `filter`, `filter_exact`, `run_ignored`, `run_tests` ... and 8 others
```

Regards.

Martin1887:
The build has been fixed.

Regards.

jebrosen:
First, I love the idea of a fairing for compression support and would probably use one if it was merged.

I haven't looked at the whole PR comprehensively, but there is at least one very serious bug: the fairing uses a single `AtomicBool` each to track whether or not brotli and/or gzip are supported. If two requests are sent (e.g. one accepting gzip and one accepting only identity), the behavior will depend on the timing of the requests and responses -- a response might be sent with gzip encoding when only identity was supported by that client.

Reading the `Accept-Encoding` header in `on_response` and eliminating `on_request` entirely should be both simpler and more correct.

On a separate note, some projects might want a choice of whether or not to support brotli, gzip, or other compression methods, either at compile time with feature selection or at runtime with different configs.

Martin1887:
Hello @jebrosen  and thanks for your comment.

Firstly, I don't understand why I didn't view that the request is available on the on_response function, my first idea was using only on_response but then I found that the Accept-Encoding is only in the request and then I added the on_request function without noticing that the request is available on the on_response function 😖😖😖😖... my fault. I expect fixing it today.

I also found useful adding features to choose the compression method, but then I found that brotli is not supported in all browsers and that it is only supported in encrypted (https) connections, so gzip must be always supported. Then, currently the unique useful option would be ignoring brotli compression method because you want to use always gzip. As the compression method chosen is in cascade (if brotli is supported use brotli, else if gzip is supported use gzip, etc.) this adds too much complexity to the fairing, so I decided it does not worth. As a workaround, you can force gzip in your responses adding the Content-Encoding header, but then you are downloading and including in your executable brotli fairing without using it. However, I will study the features option while fixing the on_response function.

Thanks and regards.

Martin1887:
Hello.

The accept headers are check in the on_response function now.

I have found a easy manner of separated brotli_compression and gzip_compression features and they have been implemented too.


Thanks and regards.

contrib/src/compression.rs
```diff
@@ -0,0 +1,169 @@
...
+impl Fairing for Compression {
+    fn info(&self) -> Info {
+        Info {
+            name: "Brotli and gzip compressors for responses",
+            kind: Kind::Response,
+        }
+    }
+
+    fn on_response(&self, request: &Request, response: &mut Response) {
+        let accept_headers: Vec<&str> = request.headers().get("Accept-Encoding").collect();
```
SergioBenitez: This line and the series of booleans below are unnecessary. 

The first thing this should do is check whether this is an image. The `ContentType` header is cached in the request, so you can do this really easy and correctly with:

```rust
// We don't compress images, so bail early.
if request.content_type().top() == "image" {
    return;
}
```

Then we simply want to check whether `br` or `gzip` are accepted content-encoding and perform GZIP or BR compression if so, preferring to perform brotli compression if it's accepted. This is can be done simpler as well:

```rust
let accepted_encodings = request.headers()
    .get("Accept-Encoding")
    .flat_map(|e| e.split(","));

#[cfg(feature = "brotli_compression")]
if accepted_encodings.any(|e| e == "br") {
    // Do brotli compression and return.
}

#[cfg(feature = "brotli_compression")]
if accepted_encodings.any(|e| e == "gzip") {
    // Do GZIP compression and return.
}
```

contrib/src/compression.rs
```diff
@@ -0,0 +1,169 @@
...
+            {
+                brotli_compressed = true;
+                response.adjoin_header(Header::new("Content-Encoding", "br"));
+                let body = response.body_bytes();
+                if let Some(body) = body {
+                    let mut plain = Cursor::new(body);
+                    let mut compressed = Cursor::new(Vec::<u8>::new());
+                    let mut params = brotli::enc::BrotliEncoderInitParams();
+                    params.quality = 2;
+                    let content_type = headers
```
SergioBenitez: What is this trying to do? You can get the raw content type string with `request.headers().get_one("Content-Type")`. But I don't think you need it. Just use `request.content_type()`.

contrib/src/compression.rs
```diff
@@ -0,0 +1,169 @@
...
+                        .get("Content-Type")
+                        .collect::<Vec<&str>>()
+                        .join(", ");
+                    if content_type.contains("text/") {
```
SergioBenitez: Use `top()`.

Martin1887:
Hello @SergioBenitez.

My idea was that the user could disable compression in specific requests (due to performance usually) forcing a `Content-Encoding`, but it is true that this is a weird case that could not worth it and even produce misunderstandings. Therefore, I will remove it.

contrib/src/compression.rs
Martin1887: Here I collect and join in case the request had several `Content-Type` headers and some of them were `text/*`. Once more, this corner case maybe could be omited.

Martin1887:
Hello.

The code has been simplified and the compression algorithm choose via `Content-Encoding` header removed.

Nevertheless, some of your suggestions have not be able to be implemented as proposed because the `content_type()` method returns an Option (and `if let` or `match`is needed to check the `Some` value) and cfg attributes are not allowed on if expressions (`if cfg!`plus attr code block is needed).


Regards.

Martin1887:
I have not used pull requests changes requested before, but I think that @SergioBenitez should approve the changes before merging the pull request when he has reviewed them. Or should I do anything more?

I think that all requested changes have been implemented and the pull request is ready to be merged.


Regards.

contrib/src/compression.rs
```diff
@@ -0,0 +1,129 @@
...
+            }
+        }
+
+        // The compression is done if the request supports brotli or gzip and
+        // the corresponding feature is enabled
+        if cfg!(feature = "brotli_compression")
+            && request
+                .headers()
+                .get("Accept-Encoding")
+                .flat_map(|e| e.trim().split(","))
```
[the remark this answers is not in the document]
Martin1887: I use trim in case any browser set the header with spaces (for instance: "Accept-Encoding: deflate, gzip, br").

In Firefox and Chrome this does not happen, but I think that this makes the fairing more robust.

contrib/src/compression.rs
```diff
@@ -0,0 +1,129 @@
...
+        } else if cfg!(feature = "gzip_compression")
+            && request
+                .headers()
+                .get("Accept-Encoding")
+                .flat_map(|e| e.trim().split(","))
+                .any(|e| e.trim() == "gzip")
+        {
+            #[cfg(feature = "gzip_compression")]
+            {
+                let body = response.body_bytes();
```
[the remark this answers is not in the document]
Martin1887: I are doing changes here, but I don't sure if they are fine.

I'm trying to use `response.body()` function to get streamed body, but the only manner of get the `Read` inside is using `body.into_bytes()` and then passing it as parameter to the corresponded compression function.

Is this ok?

Doing it has also forced to me to use an auxiliar boolean in order to use response mut reference only once in the same block:

            let mut success = false;
            let mut compressed = Cursor::new(Vec::<u8>::new());
            {
                let plain = response.body();
                if let Some(plain) = plain {
                    let body_bytes = &plain.into_bytes();
                    if let Some(body_bytes) = body_bytes {
                        success = GzEncoder::new(&mut compressed, flate2::Compression::default())
                            .write_all(&body_bytes)
                            .is_ok();
                    }
                }
            }
            if success {
                Compression::set_body_and_header(response, compressed, "gzip");
            }

Is this acceptable?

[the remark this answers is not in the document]
Martin1887: I don't understand how the streaming interface works.

If I set a sized body to the response, then everything works:

    fn set_body_and_header<'h, B>(response: &mut Response<'h>, body: B, header: &'h str)
    where
        B: Read + Seek + 'h,
    {
        response.remove_header("Content-Encoding");
        response.adjoin_header(Header::new("Content-Encoding", header));
        response.set_sized_body(body);
    }

However, replacing in this function the `set_sized_body`call with `set_streaming_body` makes the body empty into the response, in spite of passing the same bytes as `body`to the function.

Is there a special function to call when the streaming has finished?

Has the `chunk_size` parameter of the `set_chunked_body` special requirements?

What am I doing wrong?

Sorry if my question is too much stupid.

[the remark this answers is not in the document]
SergioBenitez: You can get the actual body with `response.body().into_inner()` or `response.take_body().into_inner()`. The returned value from these types implement `Read`. The compression libraries should have some kind of wrapper over any `Read` type that returns another `Read` type that, when read, compresses on the fly. This is what we need. Once you have that type, use `response.set_streamed_body(compression_wrapper)` to set the new body on the response.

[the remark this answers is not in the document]
Martin1887: Yes, I was doing that: in my `Fairing` the compression libraries read the body as a `Cursor<Vec<u8>>` and write the compressed bytes into another `Cursor<Vec<u8>>`.

I have done some tests and I have found that `response.set_streamed_body` works fine using a `repeat().take()` as in the documentation example. Nevertheless, when a `Cursor<Vec<u8>>` is passed to this function, the response body is empty.

This is the code and the results that I have used to test it:

Set a `Cursor`fails:

        response.remove_header("Content-Encoding");
        response.adjoin_header(Header::new("Content-Encoding", header));
        response.set_streamed_body(body);
        println!("{:?}", repeat(87).take(5));
        // println!("{:?}", body);
        println!("{:?}", response.body().unwrap().into_bytes());

Output:

    Take { inner: Repeat { .. }, limit: 5 }
    Some([])


Set a `repeat().take()`works, the `Cursor` has 16 bytes.

        response.remove_header("Content-Encoding");
        response.adjoin_header(Header::new("Content-Encoding", header));
        response.set_streamed_body(repeat(87).take(5));
        println!("{:?}", repeat(87).take(5));
        println!("{:?}", body);
        println!("{:?}", response.body().unwrap().into_bytes());

Output:

       Take { inner: Repeat { .. }, limit: 5 }
       Cursor { inner: [139, 5, 128, 72, 101, 108, 108, 111, 32, 119, 111, 114, 108, 100, 33, 3], pos: 16 }
       Some([87, 87, 87, 87, 87])


As you can see, the response body has bytes when a `repeat().take()` is passed to `set_streamed_body` but when a `Cursor<Vec<u8>>` is passed then the body is empty. Using `set_sized_body` when the same `Cursor<Vec<u8>>` works fine.

I don't know if you can figure out where is the problem.


Thanks.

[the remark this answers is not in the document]
Martin1887: I have found the error: my misunderstanding about the `Cursor` struct. Seeking `body` to the start before calling `set_streamed_body` everything works.

All tests pass and I have tested it in a own project with JS and CSS files getting response times between 0 and 2 ms 😊.

I think that now the pull request is ready to be merged, I wait your comments.

Martin1887:
@SergioBenitez, have you had any chance to review the last changes?

contrib/src/compression.rs
```diff
@@ -0,0 +1,149 @@
...
+                    let mut params = brotli::enc::BrotliEncoderInitParams();
+                    params.quality = 2;
+                    if let Some(ref content_type) = content_type {
+                        if content_type.top() == "text" {
+                            params.mode = BrotliEncoderMode::BROTLI_MODE_TEXT;
+                        } else if content_type.top() == "font" {
+                            params.mode = BrotliEncoderMode::BROTLI_MODE_FONT;
+                        }
+                    }
+                    if plain.into_inner().read_to_end(&mut plain_buffer).is_ok() {
```
jebrosen: `read_to_end` is the problematic part of this: what if the response was streaming a 3GB file from disk? Without compression, it will be streamed to the client piece by piece keeping only a bit in memory at a time. With compression, it now has to read all 3 GB into memory, compress it (keeping both uncompressed and compressed versions in memory for a short time), then send the compressed version to the client.

A safe approach would involve something like this (modulo parameters and error handling):
```
let reader = response.take_body().into_inner();
let new_body = brotli::CompressorReader::new(reader, buffer_size, q, lgwin);
Compression::set_body_and_header(response, new_body, "br");
```

(The same reasoning applies to gzip below)

Martin1887: I will try your suggestion when #646 is resolved and I can compile Rocket.

However, `flate2` API only supports to write `&[u8]`, so I don't find any manner of converting the `Read` into a `&[u8]` without `read_to_end` or similar (I guess that `take` has the same problem).

Thanks for your help.

jebrosen: `flate2::read::GzEncoder` looks like it would work for this purpose  (instead of `flate2::write::GzEncoder`)

Martin1887: I understand that then I could use `read` to write in a streamed manner the bytes compressed into a buffer of my desired size, no?

It seems weird to me that `read` writes compressed data xD, that is why I didn't found this API before.

Thanks again.

SergioBenitez:
@jebrosen Could you check this out?

SergioBenitez:
@Martin1887 Please rebase.

contrib/src/compression.rs
```diff
@@ -0,0 +1,133 @@
...
+        request
+            .headers()
+            .get("Accept-Encoding")
+            .flat_map(|e| e.split(","))
+            .any(|e| encodings.contains(&e.trim()))
+    }
+
+    fn set_body_and_header<'h, B>(response: &mut Response<'h>, mut body: B, header: &'h str)
+    where
+        B: Read + Seek + 'h,
```
jebrosen: This function should not use `Seek`, to align with the changes below.

contrib/src/compression.rs
```diff
@@ -0,0 +1,133 @@
...
+                    let mut params = brotli::enc::BrotliEncoderInitParams();
+                    params.quality = 2;
+                    if let Some(ref content_type) = content_type {
+                        if content_type.top() == "text" {
+                            params.mode = BrotliEncoderMode::BROTLI_MODE_TEXT;
+                        } else if content_type.top() == "font" {
+                            params.mode = BrotliEncoderMode::BROTLI_MODE_FONT;
+                        }
+                    }
+                    brotli::BrotliCompress(&mut plain.into_inner(), &mut compressed, &params)
```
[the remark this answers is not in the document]
Martin1887: `q` is the quality, I am using 2 in this crate to have really fast compressions with compression ratio similar to gzip.

`lgwin` is the size of the window, whose default in the crate is 20. Recommended values are 20 and 22.

contrib/src/lib.rs
```diff
@@ -53,13 +55,18 @@ pub extern crate handlebars;
...
+extern crate flate2;
+
 #[cfg(feature = "json")]
 #[cfg_attr(feature = "json", macro_use)]
 #[doc(hidden)]
 pub mod json;
 
 #[cfg(feature = "json")]
-pub use json::{Json, SerdeError, JsonValue};
+pub use json::{Json, JsonValue, SerdeError};
```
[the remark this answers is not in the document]
Martin1887: Yes, RLS did it. I will undone that changes.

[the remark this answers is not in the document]
Martin1887: A commit undoing the changes would be enough or does exist any git command to discard the changes done in that file on a specified commit (I have not found it)?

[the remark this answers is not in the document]
jebrosen: For now, a commit to undo them would be enough; the history can be cleaned up later.

Martin1887:
I have implemented the changes, but I am waiting that the pull request https://github.com/dropbox/rust-brotli/pull/23 to be merged and released in rust-brotli repository, since currently `CompressionReader` does not accept compression mode.

Regards.

SergioBenitez:
@Martin1887 Please ping us when that PR (or something that offers similar functionality) has been accepted.

contrib/lib/src/compression.rs
```diff
@@ -0,0 +1,133 @@
+//! This module provides brotli ang gzip compression for all non-image
```
jens1o: `and`?

Martin1887: Sorry, I don't understand your comment.

contrib/lib/src/compression.rs
```diff
@@ -0,0 +1,133 @@
...
+#[cfg(feature = "brotli_compression")]
+use brotli::enc::backward_references::BrotliEncoderMode;
+
+#[cfg(feature = "gzip_compression")]
+use flate2;
+#[cfg(feature = "gzip_compression")]
+use flate2::write::GzEncoder;
+
+#[derive(Debug, Default)]
+pub struct Compression;
```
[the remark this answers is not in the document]
Martin1887: My previous comments about streaming still stand -- were you waiting for the changes in brotli to update that code?

I have not pushed the last commit because it will not compile until the pull request in rust-brotli is merged.

There are few brotli crates in crates.io, and rust-brotli is the fastest one. A post in the Dropbox blog describes the algorithm with more details: https://blogs.dropbox.com/tech/2016/06/lossless-compression-with-brotli/.

`Default` is used in the `fairing` function, but it is not strictly necessary. `Debug` derive can also be dropped. I add the empty tuple as field in the struct like you suggest and I also remove the `derive`s. It will be pushed as soon as the pull request is merged in rust-brotli repository.

contrib/lib/src/compression.rs
jens1o: I think that you made a typo here. Shouldn't it read `This module provides brotli and gzip compression for all non-image`?

Martin1887: Right, thanks. These silly typos are the most difficult to see. I have changed it and it will be fixed in the next commit.

SergioBenitez:
@Martin1887 We're waiting on you. I'll close this PR, to clear the queue, in a week if we don't hear back.

Martin1887:
My Pull Request has been merged and the new crates.io release will be launched on the next week.

Then, I will commit and push my changes after test them.

Martin1887:
I have committed the last changes and rebased `master`.

All `Compression` tests pass and I have also tested the library in another independent project using my working copy as crate path.

I hope that this pull request is finally ready to be merged 🙂.

jebrosen:
Please rebase or reorder commits again against the lastest `master` if you can; many unrelated commits are showing up in the diff and it's hard to filter through the relevant changes.

I only noticed one problem with the latest implementation: any response that already has a `Content-Encoding` header will be re-compressed. Compression should only happen if there is not already a `Content-Encoding` header, or if the header's value is `identity`.

Everything else looks reasonable to me, just nitpicks and stylistic changes.

jebrosen:
> Sorry, but I am not very skillful in git. All commits in this pull request contain Brotli (the firsts ones) or [CompressionFairing]. I think that this can easy the merge work. Anyway, if you know a good tutorial about reordering commits, rebase, cherry pick, squash, etc. I will appreciate it.

I'm not sure exactly what happened, but I did try rebasing it myself and things got pretty hairy. Would you be able to make a new branch off of the upstream `master` branch, copy and paste in all of your changes as a single commit, and make a new pull request from that branch? My guess is that adding these commits directly to your `master` branch instead of a separate feature branch made the rebasing and merging harder later.

PR #556, opened 2018-02-01
Make errors with binding more understandable

asonix:
Specifically add `Bind` as a `LaunchErrorKind` discriminant to allow for errors binding (like trying to bind to port 80 without root) to be more understandable.

### DIFF
--- a/lib/src/error.rs
+++ b/lib/src/error.rs
@@ -33,6 +33,7 @@ pub enum Error {
 /// other kinds of launch errors.
 #[derive(Debug)]
 pub enum LaunchErrorKind {
+    Bind(hyper::Error),
     Io(io::Error),
     Collision(Vec<(Route, Route)>),
     FailedFairing,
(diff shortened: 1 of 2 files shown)


SergioBenitez:
Is this really correct? Are all errors returned from Hyper due to bind failures?

asonix:
The call to `Server::http` is handled in hyper [here](https://github.com/hyperium/hyper/blob/c529d77c786d32a3b6265176239034980687c2ae/src/server/mod.rs#L205)
with the line
```rust
HttpListener::new(addr).map(Server::new)
```
So we look at `HttpListener::new()` defined [here](https://github.com/hyperium/hyper/blob/c529d77c786d32a3b6265176239034980687c2ae/src/net.rs#L244) with the line
```rust
Ok(HttpListener::from(try!(TcpListener::bind(addr))))
```
So it just wraps a call to `TcpListener::bind(addr)`, which returns an io::Error on a failed bind. This is the only error source in the codepath.

asonix:
I previously brought up a change to the Hyper folks here: https://github.com/hyperium/hyper/pull/1430

SergioBenitez:
I really dislike working around Hyper. Oh well. Let's do it.

SergioBenitez:
Merged in aab74ced2327fc95c7af505002f10a7373c19b8b.

PR #557, opened 2018-02-04
HEAD response bodies are stripped before fairings are run

### LINKED ISSUE #546: HEAD response bodies are stripped before fairings are run
I believe response fairings should run before the body is stripped. Currently it is not possible for a fairing to access the body, and one can also add a body to a HEAD response because of this.

I was attempting to implement a fairing for automatic ETag generation / conditional get when I discovered this. ETags are intended to be generated and checked for HEAD requests as well, but this is impossible in an `on_response` fairing because the body is stripped before the fairing is run.

teuron:
fixed #546 

What needs to be implemented

- [x]  Test: Fairings should only be called once per request/response pair.
- [x]  Test: Response fairings see a `GET` request for auto-handled `HEAD` requests.
- [x]  Request fairings should be called with the original request.
- [x]  Update the docs to clarify the new constraints.

### DIFF
--- a/lib/src/request/request.rs
+++ b/lib/src/request/request.rs
@@ -618,7 +618,7 @@ impl<'r> Request<'r> {
 
     /// Set the method of `self`, even if self is a shared reference
     #[inline(always)]
-    pub(crate) fn set_method_interior(&self, method: Method) {
+    pub(crate) fn _set_method(&self, method: Method) {
         self.method.set(method);
     }
 
--- a/lib/src/request/request.rs
+++ b/lib/src/request/request.rs
@@ -37,7 +37,7 @@ struct RequestState<'r> {
 /// data. This includes the HTTP method, URI, cookies, headers, and more.
 #[derive(Clone)]
 pub struct Request<'r> {
-    method: Method,
+    method: Cell<Method>,
     uri: Uri<'r>,
     headers: HeaderMap<'r>,
     remote: Option<SocketAddr>,
@@ -53,7 +53,7 @@ impl<'r> Request<'r> {
                                                 method: Method,
                                                 uri: U) -> Request<'r> {
         Request {
-            method: method,
+            method: Cell::new(method),
             uri: uri.into(),
             headers: HeaderMap::new(),
             remote: None,
@@ -91,7 +91,7 @@ impl<'r> Request<'r> {
     /// ```
     #[inline(always)]
     pub fn method(&self) -> Method {
-        self.method
+        self.method.get()
     }
 
     /// Set the method of `self`.
@@ -111,7 +111,7 @@ impl<'r> Request<'r> {
     /// ```
     #[inline(always)]
     pub fn set_method(&mut self, method: Method) {
-        self.method = method;
+        self.method.set(method);
     }
 
     /// Borrow the URI from `self`, which is guaranteed to be an absolute URI.
@@ -446,7 +446,7 @@ impl<'r> Request<'r> {
     /// ```
     pub fn format(&self) -> Option<&MediaType> {
         static ANY: MediaType = MediaType::Any;
-        if self.method.supports_payload() {
+        if self.method.get().supports_payload() {
             self.content_type().map(|ct| ct.media_type())
         } else {
             // FIXME: Should we be using `accept_first` or `preferred`? Or
@@ -616,6 +616,12 @@ impl<'r> Request<'r> {
         *self.state.params.borrow_mut() = route.get_param_indexes(self.uri());
     }
 
+    /// Set the method of `self`, even if self is a shared reference
+    #[inline(always)]
+    pub(crate) fn set_method_interior(&self, method: Method) {
+        self.method.set(method);
+    }
+
     /// Replace all of the cookies in `self` with those in `jar`.
     #[inline]
     pub(crate) fn set_cookies(&mut self, jar: CookieJar) {
@@ -701,7 +707,7 @@ impl<'r> fmt::Display for Request<'r> {
     /// Pretty prints a Request. This is primarily used by Rocket's logging
     /// infrastructure.
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{} {}", Paint::green(&self.method), Paint::blue(&self.uri))?;
+        write!(f, "{} {}", Paint::green(&self.method.get()), Paint::blue(&self.uri))?;
 
         // Print the requests media type when the route specifies a format.
         if let Some(media_type) = self.format() {
--- a/lib/src/request/request.rs
+++ b/lib/src/request/request.rs
@@ -446,7 +446,7 @@ impl<'r> Request<'r> {
     /// ```
     pub fn format(&self) -> Option<&MediaType> {
         static ANY: MediaType = MediaType::Any;
-        if self.method.get().supports_payload() {
+        if self.method().supports_payload() {
             self.content_type().map(|ct| ct.media_type())
         } else {
             // FIXME: Should we be using `accept_first` or `preferred`? Or
@@ -707,7 +707,7 @@ impl<'r> fmt::Display for Request<'r> {
     /// Pretty prints a Request. This is primarily used by Rocket's logging
     /// infrastructure.
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{} {}", Paint::green(&self.method.get()), Paint::blue(&self.uri))?;
+        write!(f, "{} {}", Paint::green(self.method()), Paint::blue(&self.uri))?;
 
         // Print the requests media type when the route specifies a format.
         if let Some(media_type) = self.format() {
--- a/lib/src/rocket.rs
+++ b/lib/src/rocket.rs
@@ -203,6 +203,8 @@ impl Rocket {
         self.preprocess_request(request, &data);
         self.fairings.handle_request(request, &data);
 
+	let mut auto_handeled = false;
+
         // Route the request to get a response.
         let mut response = match self.route(request, data) {
             Outcome::Success(mut response) => {
@@ -226,8 +228,14 @@ impl Rocket {
                 if request.method() == Method::Head {
                     info_!("Autohandling {} request.", Paint::white("HEAD"));
                     request.set_method(Method::Get);
-                    let mut response = self.dispatch(request, data);
-                    response.strip_body();
+                    let response = self.dispatch(request, data);
+
+		    //TODO: Remove auto_handeled and change it to
+		    //request.set_method(Method::Head); 
+		    //Rust thinks that `request` is still borrowed here, but it's
+		    //not, so wait until NLL has shipped. 
+		    auto_handeled = true;
+
                     response
                 } else {
                     self.handle_error(Status::NotFound, request)
@@ -242,7 +250,7 @@ impl Rocket {
         self.fairings.handle_response(request, &mut response);
 
         // Strip the body if this is a `HEAD` request.
-        if request.method() == Method::Head {
+        if request.method() == Method::Head || auto_handeled {
             response.strip_body();
         }
 
--- a/lib/src/rocket.rs
+++ b/lib/src/rocket.rs
@@ -236,15 +236,15 @@ impl Rocket {
             Outcome::Failure(status) => self.handle_error(status, request),
         };
 
-        // Strip the body if this is a `HEAD` request.
-        if request.method() == Method::Head {
-            response.strip_body();
-        }
-
         // Add the 'rocket' server header to the response and run fairings.
         // TODO: If removing Hyper, write out `Date` header too.
         response.set_header(Header::new("Server", "Rocket"));
         self.fairings.handle_response(request, &mut response);
+
+        // Strip the body if this is a `HEAD` request.
+        if request.method() == Method::Head {
+            response.strip_body();
+        }
 
         response
     }
--- a/lib/src/rocket.rs
+++ b/lib/src/rocket.rs
@@ -211,7 +211,6 @@ impl Rocket {
         // Route the request and run the user's handlers.
         let mut response = self.route_and_process(request, data);
 
-
         // Add the 'rocket' server header to the response and run fairings.
         // TODO: If removing Hyper, write out `Date` header too.
         response.set_header(Header::new("Server", "Rocket"));
@@ -226,10 +225,9 @@ impl Rocket {
         response
     }
 
-
     fn route_and_process<'s, 'r>(
         &'s self,
-        request: &'r Request<'s>,
+        request: &'r mut Request<'s>,
         data: Data
     ) -> Response<'r> {
         // Route the request to get a response.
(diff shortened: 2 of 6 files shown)


jebrosen:
This will still strip response bodies for autohandled routes. I think that aspect could be fixed by changing line 230 from `request.strip_body()`  to `request.set_method(Method::Head);

teuron:
Thanks, but Rust thinks here that request is still borrowed. I choose this solution but I could also make an unsafe block here and create a new reference. What do you think is better?

teuron:
And is it actually wanted, that the on_response fairings are called twice on autohandeled HEAD request?

jebrosen:
Without seeing the error message, I'm guessing that request is actually borrowed by the response. A perhaps cleaner solution would be to stash something like `let was_head_request = (request.method() == Method::Head);` early on and then check against `was_head_request` to strip the body or not -- that reduces the special-casing of auto-handled HEAD.

That would help with strip_body, but there is another issue I just noticed: when a HEAD request is auto-handled, response fairings are told the method was GET. IMO This behavior should either be an explicit design choice or a bug that gets fixed.

jebrosen:
I posted the previous before seeing your other comment about fairings being called twice. The recursion for auto-handled HEAD certainly makes everything harder to reason about... now I'm unsure of the right way to address all this.

If I'm reading right, the following happens for auto-handled requests:
1. GET route runs
2. fairings run against the GET request
3. body is stripped
4. fairings run again, this time against the stripped GET request <-- oops
5. response is returned
In this case, (4.) should not happen.

And for "real" head requests:
1. HEAD route runs
2. body is stripped
3. fairings run against the stripped HEAD request
4. response is returned
In this case, (2) and (3) should probably be switched, if only so that fairings cannot add bodies to head request.

teuron:
My proposed solution would be this workflow: 

Auto-handeled HEAD Requests: 

1. Head request fairings run
2. GET route runs without any fairings
3. Head response fairings run
4. Response body is stripped
5. Return response

Normal HEAD Requests: 

1. HEAD route runs
2. fairings run against HEAD request
3. Response body is stripped
4. Return response

teuron:
@jebrosen I implemented a working solution with my proposed workflow.

SergioBenitez:
Looks like we've identified issues with the existing implementation: awesome!

Here's what _should_ happen:
  * Fairings should be called _once_ for a given request/response pair.
  * Fairings should be called with the original request (post-[`preprocess`](https://github.com/SergioBenitez/Rocket/blob/master/lib/src/rocket.rs#L172)).
    - This means requests fairings should see a `HEAD` regardless of what Rocket does thereafter.
    - Bodies should be stripped on incoming `HEAD` requests  _even if_ a fairing changes the method.
  * Response fairings should see the `Request` the passed in `Response` was generated for.
    - This is consistent with response fairings seeing changes made by other request fairings.
    - This means response fairings see a `GET` request for auto-handled `HEAD` requests.

This PR and #546 add the following constraint:
  * Fairings should be called with the final response _before_ body stripping.

I am in favor of this change, and I'm certainly in favor of fixing the existing implementation so that it always does what should have been happening. @teuron I feel that your current implementation is a bit hard to follow. We need to ensure that the `dispatch` method is easy to understand.

Note that this PR will also need to fix the documentation on fairings and automatic head handling where appropriate to clarify these semantics.

SergioBenitez:
@jebrosen Thank you for the commentary on this PR. It's very helpful and appreciated!

jebrosen:
@SergioBenitez that analysis/solution sounds good to me, and I also agree that this implementation is a bit hard to follow as-is. The latest version also adds a `transmute` that appears to create two `&mut` references to the same request, which I am very uncomfortable with. Hopefully there is a simpler way to achieve all the goals outlined above.

I will mention that I have changed my own code so that this and #546 are no longer blocking me personally, but what I did only works on autohandled requests and only for fairings that can tell from the response they have been run once already.

teuron:
@SergioBenitez @jebrosen I've implemented everything based on your suggestion. 

I've added some notes in the docs, and I think that would be enough. 


Thanks for your support!

SergioBenitez:
@jebrosen Would you be able to review the updated code?

jebrosen:
I have read the latest code and it "looks right" to me, for whatever that's worth. I would like to test it more thoroughly, but unfortunately will not have the time for a while.

lib/src/rocket.rs
```diff
@@ -191,17 +191,33 @@ impl Rocket {
...
-        self.fairings.handle_request(request, &data);
+
+        if run_fairings {
+            self.fairings.handle_request(request, &data);
+        }
+
+        let was_head_request = request.method() == Method::Head;
 
         // Route the request to get a response.
         let mut response = match self.route(request, data) {
```
SergioBenitez: I think the better refactor is to split this big `match` out into another method, put everything before and after it in `dispatch`, and call the new method in the autohandled case instead of dispatch. In psuedocode:

```rust
fn dispatch(request, data) -> Response<'r> {
    // Remember whether this was originally a HEAD request or not.
    let head_request = request.method().is_head();

    // Preprocess the request and run the request fairings.
    self.preprocess_request(request, &data);
    self.fairings.handle_request(request, &data);

    // Route the request and run the user's handlers.
    let mut response = self.route_and_process(request, &data);

    // Run response fairings and strip the body if this was a HEAD request.
    self.fairings.handle_response(request, &mut response);
    if head_request { response.strip_body(); }

    response
}

fn route_and_process(request, data) -> Response<'r> {
    match self.route(request, data) {
        ...
        Outcome::Forward(data) => {
            if request.method() == Method::Head {
                request.set_method(Method::Get);
                self.route_and_process(request, data)
            } else {
                self.handle_error(Status::NotFound, request)
            }
        }
        ...
    }
}
```

This removes all of this `if` statements, clarifies the logic, and separates concerns nicely.

SergioBenitez: Both this PR and the existing implementation were calling `self.preprocess_request()` twice as well. The refactor I propose makes it clear to see that this doesn't happen.

teuron: @SergioBenitez thanks thats the best way we can do it. I will implement it next week and update the PR.

lib/src/rocket.rs
```diff
@@ -191,20 +191,49 @@ impl Rocket {
...
+            response.strip_body();
+        }
+
+        response
+    }
+
+
+    fn route_and_process<'s, 'r>(
+        &'s self,
+        request: &'r Request<'s>,
```
jebrosen: This should be `&'r mut` instead of just `&'r`, since this function might modify `request`

lib/src/rocket.rs
```diff
@@ -214,20 +243,24 @@ impl Rocket {
...
-                    unsafe { (&mut *(request as *const _ as *mut _)) };
 
                 // There was no matching route. Autohandle `HEAD` requests.
                 if request.method() == Method::Head {
                     info_!("Autohandling {} request.", Paint::white("HEAD"));
+
+                    // Rust thinks `request` is still borrowed here, but it's
+                    // obviously not (data has nothing to do with it), so we
+                    // convince it to give us another mutable reference.
+                    // TODO: Use something that is well defined, like UnsafeCell.
```
jebrosen: What is the status of this TODO -- is there a better solution than the `unsafe` block? NLL is available now.

lib/src/rocket.rs
SergioBenitez: Absolutely! Great catch! This mistake leads to highly undefined behavior when the `&mut Request` is created later from it.

lib/src/rocket.rs
SergioBenitez: I'm not sure if NLL resolves this particular issue, but I'd be curious to know if it does. That being said, there are far too many soundness problems with the NLL implementation as it stands for me to be comfortable enabling it in Rocket.

teuron: I tried to use it with NLL, but it does not solve the problem. Do you have any suggestions?

SergioBenitez: You can leave this as is.

lib/src/rocket.rs
```diff
@@ -191,20 +191,47 @@ impl Rocket {
...
+        // Check if the request is a `HEAD` request.
+        let was_head_request = request.method() == Method::Head;
+
+        // Route the request and run the user's handlers.
+        let mut response = self.route_and_process(request, data);
+
+        // Add the 'rocket' server header to the response and run fairings.
+        // TODO: If removing Hyper, write out `Date` header too.
+        response.set_header(Header::new("Server", "Rocket"));
+
```
SergioBenitez: Remove this empty line.

lib/src/request/request.rs
```diff
@@ -616,6 +616,12 @@ impl<'r> Request<'r> {
         *self.state.params.borrow_mut() = route.get_param_indexes(self.uri());
     }
 
+    /// Set the method of `self`, even if self is a shared reference
+    #[inline(always)]
+    pub(crate) fn set_method_interior(&self, method: Method) {
```
SergioBenitez: Let's name this `_set_method()`. We should use `_method_name()` when we have an internal-only method that conflicts with a public method.

lib/src/request/request.rs
```diff
@@ -446,7 +446,7 @@ impl<'r> Request<'r> {
     /// ```
     pub fn format(&self) -> Option<&MediaType> {
         static ANY: MediaType = MediaType::Any;
-        if self.method.supports_payload() {
+        if self.method.get().supports_payload() {
```
SergioBenitez: Don't use `self.method.get()`. Instead, use `self.method()`. Make this change everywhere else `self.method.get()` is used.

lib/src/rocket.rs
```diff
@@ -214,20 +241,16 @@ impl Rocket {
...
-                    unsafe { (&mut *(request as *const _ as *mut _)) };
 
                 // There was no matching route. Autohandle `HEAD` requests.
                 if request.method() == Method::Head {
                     info_!("Autohandling {} request.", Paint::white("HEAD"));
-                    request.set_method(Method::Get);
-                    let mut response = self.dispatch(request, data);
-                    response.strip_body();
+
+                    request.set_method_interior(Method::Get);
```
SergioBenitez: Move this line just after the next comment with no empty line before or after it.

lib/tests/fairing_before_head_strip-issue-546.rs
```diff
@@ -0,0 +1,74 @@
+#![feature(plugin, decl_macro)]
+#![plugin(rocket_codegen)]
+
+extern crate rocket;
+
+use rocket::response::content;
+
+
```
SergioBenitez: Remove this empty line.

lib/tests/fairing_before_head_strip-issue-546.rs
```diff
@@ -0,0 +1,74 @@
...
+    struct Counter {
+        get: AtomicUsize,
+    }
+
+    #[test]
+    fn not_empty_before_fairing() {
+        let rocket = rocket::ignite()
+            .mount("/", routes![index])
+            .attach(AdHoc::on_response(|req, res| {
+                if req.method() != Method::Head || res.body_string() != Some("{ 'test': 'dont strip before fairing' }".into()) {
```
SergioBenitez: Can we simply `assert!` here and cause the test to fail by panicking?

lib/tests/fairing_before_head_strip-issue-546.rs
```diff
@@ -0,0 +1,74 @@
...
+    fn not_empty_before_fairing_autohandeled() {
+        let counter = Counter::default();
+        let rocket = rocket::ignite()
+            .mount("/", routes![auto])
+            .manage(counter)
+            .attach(AdHoc::on_request(|req, _| {
+                 let c = req.guard::<State<Counter>>().unwrap();
+
+                 if c.get.fetch_add(1, Ordering::Release) != 0 {
+                     req.set_method(Method::Put); //On Error set bad Method
```
SergioBenitez: Comment should be `// This is the comment.`

lib/tests/fairing_before_head_strip-issue-546.rs
```diff
@@ -0,0 +1,74 @@
...
+                 }
+            }))
+            .attach(AdHoc::on_response(|req, res| {
+                if req.method() != Method::Get || res.body_string() != Some("{ 'test': 'dont strip before fairing' }".into()) {
```
SergioBenitez: Same with the `assert!` here.

site/guide/fairings.md
```diff
@@ -96,7 +96,8 @@ events is described below:
     a response fairing can be used to provide a response when the greater
     applications fails to by rewriting **404** responses as desired. As another
     example, response fairings can also be used to inject headers into all
-    outgoing responses.
+    outgoing responses. Beware that the body of a `HEAD` request will be
```
SergioBenitez: This seems very out of context.

If you'd like, we can focus on the code for this PR and I can take care of the documentation changes afterwards.

lib/src/rocket.rs
```diff
@@ -214,20 +240,15 @@ impl Rocket {
...
                 // There was no matching route. Autohandle `HEAD` requests.
                 if request.method() == Method::Head {
                     info_!("Autohandling {} request.", Paint::white("HEAD"));
-                    request.set_method(Method::Get);
-                    let mut response = self.dispatch(request, data);
-                    response.strip_body();
+
+                    // Dispatch the request again with Method `GET`.
+                    request._set_method(Method::Get);
+                    let response = self.route_and_process(request, data);
```
SergioBenitez: Remove the `let` and just return `self.route_and_process(request, data)` directly from this branch.

lib/tests/fairing_before_head_strip-issue-546.rs
```diff
@@ -0,0 +1,69 @@
+#![feature(plugin, decl_macro)]
+#![plugin(rocket_codegen)]
+
+extern crate rocket;
+
+use rocket::response::content;
+
+#[head("/")]
+fn index() -> content::Json<&'static str> {
+    content::Json("{ 'test': 'dont strip before fairing' }")
```
SergioBenitez: Can we move this string into a `const` so that we don't keep repeating it?

SergioBenitez: Also, is there a reason you're returning `Json` instead of just a `'static str`? It's better to keep it simple and focused on what we're trying to test.

lib/tests/fairing_before_head_strip-issue-546.rs
```diff
@@ -0,0 +1,69 @@
...
+        let rocket = rocket::ignite()
+            .mount("/", routes![index])
+            .attach(AdHoc::on_response(|req, res| {
+                assert_eq!(req.method(), Method::Head);
+                assert_eq!(res.body_string(), Some("{ 'test': 'dont strip before fairing' }".into()));
+            }));
+
+        let client = Client::new(rocket).unwrap();
+        let response = client.head("/").dispatch();
+
```
SergioBenitez: Remove this empty line. Same in the test below.

SergioBenitez:
Merged in 3a1d6718946a0a3830ffff3a7b15faf9d48182e5. Super awesome work, @teuron!

Issue #21, opened 2016-12-21
Improve Logging: Migrate to `tracing`


SergioBenitez:
It's currently not possible to tap into Rocket's logging infrastructure in any easy way. Of course, logging is important, and so it's necessary to have a nice, clean API for Rocket applications to log messages. Libraries like [slog](https://github.com/slog-rs/slog) and [log](https://github.com/rust-lang-nursery/log), the latter currently used by Rocket, provide nice, structured ways to do this. Look into restructuring Rocket's logging to expose their functionalities.

The current alternative is to simply `println` to log.

divoxx:
Sorry, I missed this one. 

Glad you're already thinking about it. Any thoughts around supporting structured logging? The main blocker I see is that sometimes for the same "event" (i.e. an error) you generate a warn and an error message instead of just one.

Is this something you'd be looking for help with? I'm happy to propose something but seems like you're already thinking about it.

mehcode:
Heh. Guess I wasn't the only one who missed this. Apologies.

@SergioBenitez I'm itching to jump in and do something here. I can implement slog-rs in Rocket in a PR if that's the direction you'd like to go.

SergioBenitez:
@mehcode I'd love to see how this would look! I have a few requirements for Rocket's "next-gen" logging:

  1. Every log statement should have a `Request` associated with it for easy filtering.
  2. You should be able to log without using a macro.
  3. Log messages should look pretty by default. I _really_ like the way they look now.
  4. You should be able to switch to a "compact" mode.
  5. Logging should not look garbled amidst concurrent requests.
  6. Logging should have zero overhead in the presence of no errors after initialization in `release`.
  7. Users should be able to log at any level trivially.
  8. Rocket should be able to colorize output as it wishes.

Whatever log solution we go with should meet these requirements.

marcusball:
I pretty much agree with all of those. Though, I'm curious, what's your motivation behind 2? 

I really like 5, but I'm concerned how that would be achieved. The most aesthetically pleasing would be to have logs associated with each request grouped together into a block; The simplest implementation of that—off the top of my head—would be to buffer the logs and dump them at the end of each request. Unfortunately, that wouldn't work well for debugging, as presumably you'd want to see logs leading to, for example, an incomplete request due to a panic. 

[...]

PR #558, opened 2018-02-06
Add interface to configure log output location

dqsully:
In Rust applications that use packages like [indicatif](https://crates.io/crates/indicatif), when Rocket logs to stdout it may get overwritten by a running progress bar or otherwise mess up the tty output. This change adds a `log_output` property to Rocket's config which lets the programmer to configure Rocket to log to stdout, stderr, or to a custom logging implementation of his/her own.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1,0 +1,1483 @@
+[[package]]
+name = "aho-corasick"
+version = "0.6.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "memchr 2.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "antidote"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "atty"
+version = "0.2.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "libc 0.2.42 (registry+https://github.com/rust-lang/crates.io-index)",
+ "termion 1.5.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "winapi 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "backtrace"
+version = "0.3.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "backtrace-sys 0.1.23 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cfg-if 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.42 (registry+https://github.com/rust-lang/crates.io-index)",
+ "rustc-demangle 0.1.8 (registry+https://github.com/rust-lang/crates.io-index)",
+ "winapi 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "backtrace-sys"
+version = "0.1.23"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "cc 1.0.17 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.42 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "base64"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "byteorder 1.2.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "safemem 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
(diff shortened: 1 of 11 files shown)


SergioBenitez:
I believe you can already effectively do this by registering a [`log`](https://github.com/rust-lang-nursery/log) logger and using `rocket::custom(_, false)` to create a `Rocket` instance. Is there something else this PR adds?

dqsully:
That's a good idea. In that case I'd like to add log output location/coloring/emojii configuration into the `Rocket.toml` config file, some documentation on how to intercept Rocket's logging, as well as change the log formatting to be more compatible with other implementations. At least that way developers can hook into Rocket's existing logging infrastructure until `slog` is implemented.

dqsully:
Alright, I've implemented those changes.

lib/src/fairing/fairings.rs
```diff
@@ -8,12 +8,17 @@ pub struct Fairings {
     launch: Vec<&'static Fairing>,
     request: Vec<&'static Fairing>,
     response: Vec<&'static Fairing>,
+    use_emoji: bool,
```
jebrosen: This is strange at first glance -- ideally the Fairings struct would not need to know itself whether to print emojis or not, and it's another value to pass around and keep track of.

lib/src/fairing/fairings.rs
```diff
@@ -110,7 +115,11 @@ impl Fairings {
         }
 
         if !self.all_fairings.is_empty() {
-            info!("{}{}:", Paint::masked("📡  "), Paint::purple("Fairings"));
+            if self.use_emoji {
```
jebrosen: These `self.use_emoji` checks repeated everywhere add a lot of redundancy.

Can Paint be extended to have this "use-emoji-or-not-aware" capability, *in addition to* Paint::masked? (@SergioBenitez) I see the value in treating emoji use and color orthogonally, but I don't like it duplicated at each log invocation.

lib/src/logger.rs
```diff
@@ -57,12 +57,80 @@ impl fmt::Display for LoggingLevel {
...
 #[doc(hidden)] #[macro_export]
-macro_rules! log_ { ($name:ident: $($args:tt)*) => { $name!(target: "_", $($args)*) }; }
+macro_rules! log_ {
+    ($name:ident: $fmt_str:expr) => {
+        $name!(
+            concat!("    {rocket_indent} ", $fmt_str),
+            rocket_indent=$crate::yansi::Paint::white("=>")
+        )
+    };
+    ($name:ident: $fmt_str:expr, $($args:tt)*) => {
```
jebrosen: Does this need to be two cases? `$($args:tt)*` or a similar pattern should cover the no-arguments case also.

lib/src/logger.rs
```diff
@@ -74,10 +142,41 @@ macro_rules! debug_ { ($($args:expr),+) => { log_!(debug: $($args),+); }; }
 #[doc(hidden)] #[macro_export]
 macro_rules! warn_ { ($($args:expr),+) => { log_!(warn: $($args),+); }; }
 
+pub fn format_log(record: &log::Record) -> String {
```
jebrosen: Instead of returning a String, this could format directly into the destination to avoid extra allocations, especially in the `Level::Debug` case.

This would mean at minimum taking a `Write` argument and using `write!(writer, ...)` instead of `format!(...)` everywhere.

lib/src/rocket.rs
```diff
@@ -344,7 +345,10 @@ impl Rocket {
     /// Creates a new `Rocket` application using the supplied custom
     /// configuration information. The `Rocket.toml` file, if present, is
     /// ignored. Any environment variables setting config parameters are
-    /// ignored. If `log` is `true`, logging is enabled.
+    /// ignored. If `log` is `false` or the config has `log_output` set to
+    /// `LoggingOutput::Disabled`, the default log implementation is disabled,
+    /// allowing you to use/write one of your own. (`log` is kept around
```
jebrosen: I think having both a parameter and a config option that can disable logging is confusing. I would rather it be just the config option, but that definitely is a breaking change.

Cargo.lock
```diff
@@ -0,0 +1,1483 @@
+[[package]]
```
jebrosen: This file is in the .gitignore, please remove it.

lib/src/rocket.rs
```diff
@@ -531,7 +548,11 @@ impl Rocket {
     /// ```
     #[inline]
     pub fn catch(mut self, catchers: Vec<Catcher>) -> Self {
-        info!("👾  {}:", Paint::purple("Catchers"));
+        if self.config.use_emoji {
+            info!("👾  {}:", Paint::purple("Catchers"));
```
jebrosen: This alien monster should use `Paint::masked`; I believe it not being masked before was just an oversight.

SergioBenitez:
I think between the new ability to disable coloring and logging entirely in 0.4, and already being able to register a logger by using `log` directly, this PR no longer adds too much new. I'd like to see a general rework of logging in Rocket (#21) rather than piecemeal features.

PR #569, opened 2018-02-21
Update state guide to reflect Diesel reexport

ELD:
Diesel now reexports r2d2 functionality, so rather than including that
library explicitly, it's better to leverage the reexport.


SergioBenitez:
Merged in c48905f4835f701784fa23ef72a1f92888c790d8. Thanks so much!

PR #561, opened 2018-02-09
Fix typo in guide

pixelistik:

PR #579, opened 2018-03-01
Correct various spellings and typos in the guide.

DJMcNab:
Does not include the change suggested in SergioBenitez/Rocket#561
as there is already that PR open for it.


SergioBenitez:
Merged in 96bb608e3af034771f9731e47758e45794346068. Thanks so much!

PR #584, opened 2018-03-15
failed parsing in the case where `#` precedes `?`

takayahilton:
fragment    = *( pchar / "/" / "?" )
https://www.ietf.org/rfc/rfc3986.txt


SergioBenitez:
Fragments aren't sent along to the server, so we should never see these. In any case, the `URI` parsing has been completely rewritten and will be merged soon. As such, I'm closing out this PR.

SergioBenitez:
Eh, why not merge this? Merged in 0b0d0998649600c5342549cc6cedd026f4937ad2.

PR #591, opened 2018-03-22
Fix minor typo

mtn:
avaiable -> available.
It also propagates to the docs, but I haven't looked into how those are generated.

### DIFF
--- a/lib/src/request/from_request.rs
+++ b/lib/src/request/from_request.rs
@@ -114,7 +114,7 @@ impl<S, E> IntoOutcome<S, (Status, E), ()> for Result<S, E> {
 ///     Extracts the [Route](/rocket/struct.Route.html) from the request if one
 ///     is available. If a route is not available, the request is forwarded.
 ///
-///     For information of when a route is avaiable, see the
+///     For information of when a route is available, see the
 ///     [`Request::route`](/rocket/struct.Request.html#method.route)
 ///     documentation.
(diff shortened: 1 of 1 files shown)


ELD:
@mtn The docs are generated from that comment (i.e. any comment that starts with `///`). Unless you mean the Rocket guides on the rocket.rs site.

So, if it's in the API documentation, this change will propagate to them.

mtn:
Right, I was referring to the API documentation. I mentioned it because sometimes maintainers want the documentation regenerated as well.

SergioBenitez:
@mtn The guide is at https://github.com/SergioBenitez/Rocket/tree/master/site. They're regenerated every new release, but only from the relevant branch.

lib/src/request/from_request.rs
```diff
@@ -114,7 +114,7 @@ impl<S, E> IntoOutcome<S, (Status, E), ()> for Result<S, E> {
 ///     Extracts the [Route](/rocket/struct.Route.html) from the request if one
 ///     is available. If a route is not available, the request is forwarded.
 ///
-///     For information of when a route is avaiable, see the
```
SergioBenitez: `of` should also be `on`

SergioBenitez:
Merged in e9ec5b2d1303ca7c874fba04e64243364dd3e109. Thanks for finding this!

PR #592, opened 2018-03-24
feat(example): update handlebars template ex to include partials and layouts

ashleygwilliams:


examples/handlebars_templates/src/main.rs
```diff
@@ -15,7 +15,9 @@ use handlebars::{Helper, Handlebars, RenderContext, RenderError, JsonRender};
 #[derive(Serialize)]
 struct TemplateContext {
     name: String,
-    items: Vec<String>
+    items: Vec<String>,
+    title: String,
+    parent: String,
```
SergioBenitez: Can we make this `&'static str`? My guess is that's what you'd want in a real application anyway.

examples/handlebars_templates/src/main.rs
```diff
@@ -24,13 +26,31 @@ fn index() -> Redirect {
 }
 
 #[get("/hello/<name>")]
-fn get(name: String) -> Template {
+fn hello(name: String) -> Template {
+    let page = "index".to_string();
+    let title = format!("Rocket Example - {}", page).to_string();
```
SergioBenitez: Why not pass in `"index"` as the template context and then render it as `"Rocket Example - {}"` in the template?

examples/handlebars_templates/src/tests.rs
```diff
@@ -40,15 +40,16 @@ fn test_root() {
...
-            items: vec!["One".into(), "Two".into(), "Three".into()]
+            items: vec!["One".into(), "Two".into(), "Three".into()],
+            title: "hello".to_string(),
+            parent: "layout".to_string(),
         };
 
-        let expected = Template::show(client.rocket(), "index", &context).unwrap();
         assert_eq!(response.status(), Status::Ok);
-        assert_eq!(response.body_string(), Some(expected));
+        assert!(response.body_string().unwrap().contains("Jack"));
```
SergioBenitez: The template isn't actually being checked any more. We need to add that back in. Nothing should be ignored here.

examples/handlebars_templates/templates/about.hbs
```diff
@@ -0,0 +1,8 @@
+{{#*inline "page"}}
+
+<section id="about">
+  <h1>Here's anoter page!</h1>
```
SergioBenitez: another

SergioBenitez:
@ashleygwilliams Ping

ashleygwilliams:
hey whoops sorry, was in europe and forgot about this. will send over changes, all seem very reasonable and good :)

SergioBenitez:
Merged in f00c68252c8d735fe5bb190d0a22511c00060e4a and 94cfba8b496bd51feb22d11a34518b16999e43a4.

PR #604, opened 2018-04-14
refactor: rename UUID to Uuid

jkcclemens:
I did a case-sensitive find and replace.

It compiles and `cargo test` passes, so good?

### DIFF
--- a/examples/uuid/src/main.rs
+++ b/examples/uuid/src/main.rs
@@ -8,8 +8,8 @@ extern crate rocket;
 extern crate rocket_contrib;
 
 use std::collections::HashMap;
-use uuid::Uuid;
-use rocket_contrib::UUID;
+use uuid::Uuid as RealUuid;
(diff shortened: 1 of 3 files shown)


jkcclemens:
Alright, now `scripts/test.sh` passes.

examples/uuid/src/main.rs
```diff
@@ -8,21 +8,21 @@ extern crate rocket;
 extern crate rocket_contrib;
 
 use std::collections::HashMap;
-use uuid::Uuid;
-use rocket_contrib::UUID;
+use uuid::Uuid as RealUuid;
```
SergioBenitez: Don't import this as `RealUuid`. Instead use `uuid::Uuid` as needed in the example.

SergioBenitez:
Hooray! Merged in 105137a46d2ee3c366fb8ac4b000d329fe7e8c8e.

PR #605, opened 2018-04-14
Fix documentation with old key length

course-correction:
Seems like there was a small oversight in 722ee93f8ba8924c9700c23a24a74569c8b1ab2f to document the change from 192 bit to 256 bit long secret keys. If there is no objection I request to merge these changes.


SergioBenitez:
Good catch! Merged in 4157c573e8d218d84a4eb9a068daf6ccff9b294a.

PR #612, opened 2018-04-26
Add doc string to silence missing_docs lint

Henning-K:
Basically the same as https://github.com/SergioBenitez/Rocket/commit/0d18faf91ee473b29e91bb26e94908c2976777b9 but for catch instead of route.


SergioBenitez:
@Henning-K Apologies that it's taken so long for me to get to this. Nonetheless, it's now merged in 2f2e0b83fc801738437ec69e4bd16281e6a446fd. Thank you!

PR #617, opened 2018-05-01
Added audio/flac as a known MIME type

imron:
Added the `audio/flac` MIME type so that routes can specify this as a format without generating a compiler warning.

Although the [official IANA list of MIME types](https://www.iana.org/assignments/media-types/media-types.xhtml#audio) does not include `audio/flac` as a MIME type, both [Wikipedia](https://en.wikipedia.org/wiki/FLAC) and [xiph.org](https://wiki.xiph.org/MIME_Types_and_File_Extensions) list this as the MIME type for FLAC files.


SergioBenitez:
Merged in a383d49ab0c224394353bae76a2b231287483cfe. Thanks for this!

PR #619, opened 2018-05-02
Custom request methods

### LINKED ISSUE #232: [Feature Request] Allow route decorator to define a custom HTTP method.
Allow the generic `route` decorator to accept _any_ method. I'd be okay with requiring quotes for non-standard methods.

```rust
#[route(FOO, path = "...")]
```

Or

```rust
#[route("FOO", path = "...")]
```

---

Key motivation for this is I have a colleague who uses https://tools.ietf.org/html/draft-snell-link-method-12 in his API definitions. 

There are also a dozen more defined methods that aren't "standard HTTP". It's not reasonable for rocket to have first-class support for any method. I think this is a good middle ground.

tcmal:
Core: 
  - Add `Extension(String)` to Method enum
  - Remove `Copy` trait from Method enum (breaking)
  - `Method.from_str()` makes an `Extension` if the string given doesn't match, meaning it can never fail. (still returns a result for compatibility).
  - `Method.as_str()` now returns a string with the lifetime of the method object, not static.
  - Make `method` in `StaticRequestInfo` a `&str`

Codegen: 
  - `request` decorator can now take a string literal as the method argument.

Examples:
  - Add example for custom methods

Fixes #232

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ members = [
   "codegen_next/",
   "contrib/",
   "examples/cookies",
+  "examples/custom_method",
   "examples/errors",
   "examples/form_validation",
   "examples/hello_person",
--- a/codegen/src/decorators/route.rs
+++ b/codegen/src/decorators/route.rs
@@ -8,20 +8,13 @@ use utils::*;
 
 use syntax::codemap::{Span, Spanned, dummy_spanned};
 use syntax::tokenstream::TokenTree;
-use syntax::ast::{Arg, Ident, Item, Stmt, Expr, MetaItem, Path};
+use syntax::ast::{Arg, Ident, Item, Stmt, Expr, MetaItem};
 use syntax::ext::base::{Annotatable, ExtCtxt};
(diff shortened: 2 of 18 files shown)


tcmal:
I'm not too sure why some of your commits are included, sorry about that.

tcmal:
Rebased & Squashed.

SergioBenitez:
Did you have a need for this? In any case, any custom HTTP methods are likely to be known a priori, so it would suffice to have `Extension(String)` instead of `Extension(&'static str)`. This also lets us keep `Method: Copy`.

tcmal:
I don't think a static lifetime would work because of `.from_hyp()`, `&'a str` with `'a` being from hyper's method could work which I think would last long enough at least for the handler, although that is still breaking, albeit to a lesser extent.

tcmal:
As said in #232, it's common in most frameworks & non-standard methods are used relatively often, though I can see the problems with implementing it.

Either way, I'll close this PR & either use `&str` or just not do it altogether, please do tell me which you'd prefer.

PR #655, opened 2018-06-06
Implement request-state local cache

### LINKED ISSUE #654: Request-Local State Cache
Rocket doesn't currently support "per request cache". Usually this is done with middlewares (#55) on other frameworks (and Go does this with `Context`). Such functionality is usually used to store (as an example) loaded `User` object to the context of current request. This way, if the user object is needed in multiple places in the request handling path, the same object can be used.

Here's an simple example of such use case: https://gist.github.com/vhakulinen/5f72f380d5959b7f00435fc473b2c1e3. In the `cargo run` file, you can see that the user object is being loaded multiple times, and this cannot currently be avoided in any non-hacky way.

That said, I'm interested in implementing this feature, as I said earlier in Matrix. I don't yet have any plan yet other that to extend the `Request` or `RequestState` to have `context` property which would act as a cache. This is so that both fairings and request guards have access to the per request cache.

Example usage (for request guard) would be following:
```rust
struct User {
    uid: usize
}

struct PermissionViewContent{}

impl<'a, 'r> FromRequest<'a, 'r> for PermissionViewContent {
    type Error = ();

    fn from_request(request: &'a Request<'r>) -> request::Outcome<PermissionViewContent, ()> {
        let u = request.guard::<User>()?;
        if u.uid != 1 {
            return Outcome::Forward(())
        }

        Outcome::Success(PermissionViewContent{})
    }
}

impl<'a, 'r> FromRequest<'a, 'r> for User {
    type Error = ();

    fn from_request(request: &'a Request<'r>) -> request::Outcome<User, ()> {
        if let Some(user) = request.context.get("my-unique-user-key") {
            return user;
        }

        let u = ...database access and so on... 

        request.context.set("my-unique-user-key", u);

        return u;
    }
}
```

I'm quite new to rust, so there probably is more 'rust' way to do this. I was wondering if something like the `request.guard::<User>()` could be implemented for the per request cache, but that would be limiting in some ways: for example, you'd be limited to have one cached `User` object per request.

vhakulinen:
Opening this PR for feedback. Based completely on @SergioBenitez's design.

I guess the next step is to extend the `state` crate to allow more relaxed bounds for `T` in the `local_cache`?

The `cache` needs to be wrapped in `Rc` so it gets `Clone`, but it was mentioned that in future `Request` might be send to another thread to be processed, so should we already use `Arc` instead?

Simple example usage https://gist.github.com/vhakulinen/fb84746fa102761c0683f694523a47cd.

Closes #654.

### DIFF
--- a/core/lib/src/request/request.rs
+++ b/core/lib/src/request/request.rs
@@ -91,10 +91,10 @@ impl<'r> Request<'r> {
     /// ```rust
     /// # use rocket::http::Method;
     /// # use rocket::Request;
-    /// # struct User();
-    /// fn current_user(_: &Request) -> User {
+    /// # struct User;
+    /// fn current_user() -> User {
     ///     // Load user...
-    ///     # User()
+    ///     # User
     /// }
     ///
     /// # Request::example(Method::Get, "/uri", |request| {
@@ -103,12 +103,12 @@ impl<'r> Request<'r> {
     /// ```
     pub fn local_cache<T, F>(&self, f: F) -> &T
         where T: Send + Sync + 'static,
-              F: FnOnce(&Request) -> T {
+              F: FnOnce() -> T {
 
         match self.state.cache.try_get() {
             Some(cached) => cached,
             None => {
-                self.state.cache.set(f(self));
+                self.state.cache.set(f());
                 self.state.cache.get()
             }
         }
--- a/core/lib/src/request/request.rs
+++ b/core/lib/src/request/request.rs
@@ -1,3 +1,4 @@
+use std::rc::Rc;
 use std::cell::{Cell, RefCell};
 use std::net::{IpAddr, SocketAddr};
 use std::fmt;
@@ -26,6 +27,7 @@ struct RequestState<'r> {
     cookies: RefCell<CookieJar>,
     accept: Storage<Option<Accept>>,
     content_type: Storage<Option<ContentType>>,
+    cache: Rc<Container>,
 }
 
 /// The type of an incoming web request.
@@ -41,7 +43,7 @@ pub struct Request<'r> {
     uri: Uri<'r>,
     headers: HeaderMap<'r>,
     remote: Option<SocketAddr>,
-    state: RequestState<'r>
+    state: RequestState<'r>,
 }
 
 impl<'r> Request<'r> {
@@ -67,6 +69,7 @@ impl<'r> Request<'r> {
                 cookies: RefCell::new(CookieJar::new()),
                 accept: Storage::new(),
                 content_type: Storage::new(),
+                cache: Rc::new(Container::new()),
             }
         }
     }
@@ -76,6 +79,23 @@ impl<'r> Request<'r> {
         let rocket = Rocket::custom(Config::development().unwrap(), true);
         let mut request = Request::new(&rocket, method, uri);
         f(&mut request);
+    }
+
+    /// Retrieves the cached value for type `T` from the request-local cached
+    /// state of `self`. If no such value has previously been cached for
+    /// this request, `f` is called to produce the value which is subsequently
+    /// returned.
+    pub fn local_cache<T, F>(&self, f: F) -> &T
+        where T: Send + Sync + 'static,
+              F: FnOnce(&Request) -> T {
+
+        match self.state.cache.try_get() {
+            Some(cached) => cached,
+            None => {
+                self.state.cache.set(f(self));
+                self.state.cache.get()
+            }
+        }
     }
 
     /// Retrieve the method from `self`.
--- a/core/lib/src/request/request.rs
+++ b/core/lib/src/request/request.rs
@@ -85,6 +85,22 @@ impl<'r> Request<'r> {
     /// state of `self`. If no such value has previously been cached for
     /// this request, `f` is called to produce the value which is subsequently
     /// returned.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # use rocket::http::Method;
+    /// # use rocket::Request;
+    /// # struct User();
+    /// fn current_user(_: &Request) -> User {
+    ///     // Load user...
+    ///     # User()
+    /// }
+    ///
+    /// # Request::example(Method::Get, "/uri", |request| {
+    /// let user = request.local_cache(current_user);
+    /// # });
+    /// ```
     pub fn local_cache<T, F>(&self, f: F) -> &T
         where T: Send + Sync + 'static,
               F: FnOnce(&Request) -> T {
--- a/examples/request_state_local_cache/src/main.rs
+++ b/examples/request_state_local_cache/src/main.rs
@@ -2,70 +2,59 @@
 #![plugin(rocket_codegen)]
 extern crate rocket;
 
-use std::thread;
-use std::time::Duration;
+use std::sync::atomic::{AtomicUsize, Ordering};
 
-use rocket::request::{self, Request, FromRequest};
+use rocket::request::{self, Request, FromRequest, State};
 use rocket::outcome::Outcome::*;
 
 #[cfg(test)] mod tests;
 
-#[derive(Clone, Copy)]
-struct Sleep();
-struct SleepCached(Sleep);
-struct ForwardGuard();
+struct Atomics {
+    first: AtomicUsize,
+    second: AtomicUsize,
+}
 
-impl<'a, 'r> FromRequest<'a, 'r> for ForwardGuard {
+struct Guard1();
+struct Guard2();
+
+impl<'a, 'r> FromRequest<'a, 'r> for Guard1 {
     type Error = ();
 
-    fn from_request(_: &'a Request<'r>) -> request::Outcome<Self, ()> {
-        Forward(())
+    fn from_request(req: &'a Request<'r>) -> request::Outcome<Self, ()> {
+        req.guard::<State<Atomics>>()?.first.fetch_add(1, Ordering::Relaxed);
+        req.local_cache(|req| {
+            req.guard::<State<Atomics>>().unwrap().second.fetch_add(1, Ordering::Relaxed);
+        });
+
+        Success(Guard1())
     }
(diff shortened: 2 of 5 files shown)


SergioBenitez:
Please add some doc-tests, unit tests, and integration tests.

vhakulinen:
@SergioBenitez wouldn't it make more sense to have the `F` in `local_cache` be a trait?

vhakulinen:
Sorry, didn't mean to close this PR.

SergioBenitez:
Depending on "wall time" for the test is error prone. Instead, lets use something like two request guards each which access the same pair of `AtomicUsize`s. One of the atomics should be incremented in the closure that initializes the cached value, and the other should be incremented outside of the closure. Thus, after calling a handler that makes use of both request guards, the atomic accessed inside the initialization closure should have a value of one while the other should have a value of two.

In pseudocode, this is:

```rust
#[get("/")]
fn handler(Guard1, Guard2) { .. }

impl FromRequest for {Guard1, Guard2} {
    req.guard::<State<Atomics>>()?.first.increment();
    req.local_cache(|req| req.guard::<State<Atomics>>()?.second.increment());
}

fn main() {
    rocket::ignite().manage(Atomics);
}

#[test]
fn test() {
    client.get("/").dispatch();
    assert_eq!(client.rocket().state::<Atomics>().unwrap().first.value(), 2);
    assert_eq!(client.rocket().state::<Atomics>().unwrap().second.value(), 1);
}
```

examples/request_state_local_cache/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+
+        Success(Guard1())
+    }
+}
+
+impl<'a, 'r> FromRequest<'a, 'r> for Guard2 {
+    type Error = ();
+
+    fn from_request(req: &'a Request<'r>) -> request::Outcome<Self, ()> {
+        req.guard::<State<Atomics>>()?.first.fetch_add(1, Ordering::Relaxed);
```
SergioBenitez: Just call the `Guard1` request guard here.

examples/request_state_local_cache/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+use rocket::outcome::Outcome::*;
+
+#[cfg(test)] mod tests;
+
+struct Atomics {
+    first: AtomicUsize,
+    second: AtomicUsize,
+}
+
+struct Guard1();
```
SergioBenitez: No need for `()` here or when it's constructed.

examples/request_state_local_cache/src/tests.rs
```diff
@@ -0,0 +1,14 @@
...
+use ::Atomics;
+use super::rocket;
+use rocket::local::Client;
+
+#[test]
+fn test() {
+    let client = Client::new(rocket()).unwrap();
+    client.get("/").dispatch();
+
+    assert!(client.rocket().state::<Atomics>().unwrap().first.load(Ordering::Relaxed) == 2, "First is two");
```
SergioBenitez: Store the reference to `Atomics` in a variable to clean these `assert!`s up:

```rust
let atomics = client.rocket().state::<Atomics>().unwrap();
assert_eq!(atomics.uncached.load(Ordering::Relaxed), 2);
assert_eq!(atomics.cached.load(Ordering::Relaxed), 2);
```

examples/request_state_local_cache/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+use rocket::request::{self, Request, FromRequest, State};
+use rocket::outcome::Outcome::*;
+
+#[cfg(test)] mod tests;
+
+struct Atomics {
+    first: AtomicUsize,
```
SergioBenitez: Let's rename these `cached` and `uncached`.

examples/request_state_local_cache/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+
+#[get("/")]
+fn index(_g1: Guard1, _g2: Guard2) {
+}
+
+fn rocket() -> rocket::Rocket {
+    rocket::ignite()
+        .manage(Atomics{
```
SergioBenitez: Instead, derive `Default` for `Atomics` and call `Atomics::default()` here.

core/lib/src/request/request.rs
```diff
@@ -78,6 +81,39 @@ impl<'r> Request<'r> {
...
+    ///     # User()
+    /// }
+    ///
+    /// # Request::example(Method::Get, "/uri", |request| {
+    /// let user = request.local_cache(current_user);
+    /// # });
+    /// ```
+    pub fn local_cache<T, F>(&self, f: F) -> &T
+        where T: Send + Sync + 'static,
+              F: FnOnce(&Request) -> T {
```
SergioBenitez: On second thought: why are we passing `self` to `F`? The caller already has a handle to `request`, so they can use it if they need to. I think this should probably be `F: FnOnce() -> T`.

examples/request_state_local_cache/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+            req.guard::<State<Atomics>>().unwrap().second.fetch_add(1, Ordering::Relaxed);
+        });
+
+        Success(Guard2())
+    }
+}
+
+
+#[get("/")]
+fn index(_g1: Guard1, _g2: Guard2) {
```
SergioBenitez: Leave a comment saying something like:

```rust
// This is exists only to run the request guards.
```

examples/request_state_local_cache/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+struct Guard1();
+struct Guard2();
+
+impl<'a, 'r> FromRequest<'a, 'r> for Guard1 {
+    type Error = ();
+
+    fn from_request(req: &'a Request<'r>) -> request::Outcome<Self, ()> {
+        req.guard::<State<Atomics>>()?.first.fetch_add(1, Ordering::Relaxed);
```
SergioBenitez: Pull out the `atomics` state loading:

```rust
let atomics = req.guard::<State<Atomics>>()?;

atomics.uncached.fetch_add(1, Ordering::Relaxed);
req.local_cache(|| atomics.cached.fetch_add(1, Ordering::Relaxed));

Guard1
```

examples/request_state_local_cache/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+        req.local_cache(|req| {
```
SergioBenitez: This is saving a value of type `()` in the cache. I suppose that's fine, but let's at least leave a comment indicating this.

vhakulinen:
Updated the PR based on feedback.

I've been wondering whether or not the use of trait would be better for `local_cache` instead of `FnOnce`. I've yet to come up with compelling example, where such situation would be favorable - I guess it would only make it a bit more explicit that the cache is per type.

examples/request_state_local_cache/src/main.rs
SergioBenitez: This change wasn't made.

examples/request_state_local_cache/src/tests.rs
```diff
@@ -0,0 +1,15 @@
...
+use super::rocket;
+use rocket::local::Client;
+
+#[test]
+fn test() {
+    let client = Client::new(rocket()).unwrap();
+    client.get("/").dispatch();
+
+    let atomics = &client.rocket().state::<Atomics>().unwrap();
+    assert!(atomics.uncached.load(Ordering::Relaxed) == 2, "First is two");
```
SergioBenitez: No need for the string comment. Also, use `assert_eq`. Same for the one below.

examples/request_state_local_cache/src/tests.rs
```diff
@@ -0,0 +1,15 @@
...
+use ::Atomics;
+use super::rocket;
+use rocket::local::Client;
+
+#[test]
+fn test() {
+    let client = Client::new(rocket()).unwrap();
+    client.get("/").dispatch();
+
+    let atomics = &client.rocket().state::<Atomics>().unwrap();
```
SergioBenitez: I don't think you need the `&` there.

examples/request_state_local_cache/src/main.rs
```diff
@@ -0,0 +1,59 @@
...
+
+impl<'a, 'r> FromRequest<'a, 'r> for Guard2 {
+    type Error = ();
+
+    fn from_request(req: &'a Request<'r>) -> request::Outcome<Self, ()> {
+        req.guard::<Guard1>()?;
+        Success(Guard2)
+    }
+}
+
```
SergioBenitez: Kill this extra empty line.

examples/request_state_local_cache/src/main.rs
```diff
@@ -0,0 +1,59 @@
...
+struct Guard1;
+struct Guard2;
+
+impl<'a, 'r> FromRequest<'a, 'r> for Guard1 {
+    type Error = ();
+
+    fn from_request(req: &'a Request<'r>) -> request::Outcome<Self, ()> {
+        req.local_cache(|| {
```
SergioBenitez: Please make the change in the previous comment.

examples/request_state_local_cache/src/tests.rs
vhakulinen: Hmh, when I tried without `&`, I had to de-reference (`(*atomics).cached...`) the value when I wanted to access it for some reason. Seems to work without one now.

examples/request_state_local_cache/src/main.rs
vhakulinen: I see I commited my experiments.

SergioBenitez:
Merged in 97c6b3ace842747e5ba5e4d323f09e1be0278c5d. Hoorah! Thanks so much for this!

PR #666, opened 2018-06-19
Update codegen for 2018-06-22 nightly.

### LINKED ISSUE #660: Compile fails with latest nightly (as of 2018-06-11)
Hi,

I was playing around with Rocket and Rust using the helloworld example from https://rocket.rs/guide/getting-started. I first tried on my private laptop a few days ago and was very satisfied with the results. Today (2018-06-11) i tried to get everything running on my laptop at work and compilation fails with the following message:

```
Compiling rocket_codegen v0.3.12
error[E0432]: unresolved import `syntax::parse::common`
 --> /home/sja/.cargo/registry/src/github.com-1ecc6299db9ec823/rocket_codegen-0.3.12/src/utils/parser_ext.rs:5:20
  |
5 | use syntax::parse::common::SeqSep;
  |                    ^^^^^^ Could not find `common` in `parse`

error[E0624]: method `parse_seq_to_end` is private
  --> /home/sja/.cargo/registry/src/github.com-1ecc6299db9ec823/rocket_codegen-0.3.12/src/utils/parser_ext.rs:13:14
   |
13 |         self.parse_seq_to_end(&Eof,
   |              ^^^^^^^^^^^^^^^^

error: aborting due to 2 previous errors

Some errors occurred: E0432, E0624.
```

I am using the latest nightly as of today. I am running everthing on my Linux(`Linux 4.14.48-2-MANJARO #1 SMP PREEMPT Fri Jun 8 20:41:40 UTC 2018 x86_64 GNU/Linux`) Laptop and used rustup to install nightly versions.

I have switched over to the nighly `nightly-2018-06-09-x86_64-unknown-linux-gnu (default)` and everything compiles fine. So i assume it has something to do with Rust itself affecting Rocket.

Investigating further and reading the suggested error messages did not help for me as i am still not that familiar with Rust.

Regards

Sebastian

jebrosen:
Fixes #660 in the `master` branch. ~~Depends on rust-lang/rust#51664.~~

`parse_seq` is identical to the method removed from libsyntax.

Also includes fix for next breakage, introduced around 2018-06-22.

- ~~`parse_optional_str` and `mk_mac_expr` (which are private) are now re-implemented by Rocket in `parser_ext.rs` by a copy/paste and fixing up the paths.~~

- ~~`parse_seq` was deleted. This reimplementation is not *quite* identical: it uses `parse_seq_to_end`, which we already asked be made public, instead of `parse_seq_to_before_end`, which is still private. *I think the span associated with the result is one character longer than it was in the original implementation.* I would like to fix that but I am not sure how to correct it (or even inspect or verify it).~~

~~If a "quick fix" that makes no changes to libsyntax is useful or helpful, this appears to work. It does have a likelihood of bitrot, but Rocket will eventually move away from libsyntax anyway.~~

~~If requesting changes to libsyntax is preferred, I think we can get by with `parse_optional_str` and `mk_mac_expr` public, and either reintroduce `parse_seq` or re-implement it (correctly, not necessarily this implementation) in Rocket. Having `parse_seq_to_before_end` would help if we were to re-implement `parse_seq` in Rocket.~~

~~EDIT: I have not (yet) verified that the nightly bound (2018-06-12) is actually sufficient, but I believe it should be.~~


SergioBenitez:
@jebrosen The implementations for these functions: are they copied straight from the rustc source code, or did you reimplement them yourself?

jebrosen:
These are all from `rustc` with only path/`use` directive changes, and the difference mentioned for `parse_seq`. From the standpoint of correctness, they should be fine now but would need to be kept up to date with any future changes in libsyntax. From the standpoint of attribution and maintenance, I would definitely ask the rust project for their opinion before considering merging this or any similar code. I probably jumped the gun by posting this as a PR.

Rereading at my original comment, it was not very clear that my intent here is to demonstrate that it is *possible* for this code to be duplicated and in the case of of `parse_seq` resurrected in rocket, not to argue that it is a desirable approach.

From my point of view, these are the options I see in order of desirability:

1. Move away from using `libsyntax`, which I understand Rocket is not quite ready to do yet.
2. Request the methods to be made public and/or defined again, which will take at least one or two days, and they may not want to bring back `parse_seq`.
3. This solution, which I consider a stopgap in case `parse_seq` will not be brought back in `libsyntax`.

SergioBenitez:
We absolutely want to move away from `libsyntax` and compiler-plugins in general in favor of proc-macros, but proc-macros appear to be more unstable that compiler-plugins themselves. What's more, they're ill-defined at the moment, quite buggy, and not powerful enough. I plan to aid in ameliorating this situation soon.

I think we should proceed with a combination of 2) and 3) as follows: if the method still exists in `libsyntax`, lets submit a PR to make them public again. If not, let's reimplement them in `rocket` for now. We want to reduce our exposure to breakage as much as possible.

jebrosen:
Sounds good.

I am preparing a PR for rust-lang/rust that makes the these methods public, and will remove them from this PR. `parse_seq` can then be identical to the implementation that was removed from rustc.

jebrosen:
Rebased, rewrote, and also updated for compatibility with changes introduced by rust-lang/rust#48149.

core/codegen/src/decorators/derive_form.rs
```diff
@@ -27,11 +27,11 @@ fn struct_lifetime(ecx: &mut ExtCtxt, item: &Annotatable, sp: Span) -> Option<St
             ItemKind::Struct(_, ref generics) => {
                 let mut lifetimes = generics.params.iter()
                     .filter_map(|p| match *p {
-                        GenericParam::Lifetime(ref def) => Some(def),
+                        GenericParam { ref ident, ref kind, .. } if *kind == GenericParamKind::Lifetime => Some(ident),
```
jebrosen: This is the only part I'm not quite certain of. Seems to work, but there might be a better way of doing this.

core/codegen/src/utils/parser_ext.rs
```diff
@@ -14,6 +15,10 @@ pub trait ParserExt<'a> {
 
     // Like `parse_ident` but also looks for an `ident` in a `Pat`.
     fn parse_ident_inc_pat(&mut self) -> PResult<'a, Ident>;
+
+    // Duplicates previously removed method in libsyntax
```
jebrosen: Should this link back to the rust repository more concretely? I wasn't sure if I should refer to a PR, commit, URL, or just leave a generic note like this.

core/codegen/src/decorators/derive_form.rs
SergioBenitez: Does `ident` include the initial tick `'`? You should take a look at the generated code and ensure that it looks correct, even if it's accepted by the compiler otherwise. Aside from this, this looks good, though it would seem that `kind` doesn't need to be a `ref kind`.

core/codegen/src/utils/parser_ext.rs
SergioBenitez: Let's link to the PR that removed the method. Something like, "See https://github.com/rust-lang/rust/foo/bar/baz for more details.".

core/codegen/src/decorators/derive_form.rs
SergioBenitez: I'm pushing a slightly different fix for this in 0.3. Let's use the same one for master.

SergioBenitez:
Merged in 5b8f8eebc6a4aeb405e4cb5e6c27ce5f6d4d9d22. Thanks for the excellent work, @jebrosen!

PR #672, opened 2018-06-24
fix uri! macro for params with initial underscore

### LINKED ISSUE #603: uri! can't be used on routes with params whose names begin with _
Rocket version: `master`
Operating system: Arch Linux; rustc 1.27.0-nightly (0b72d48f8 2018-04-10)

When a uri!() macro is used on a route with parameters named with an initial _, there is an error in the macro expansion:

```
error: invalid format string: expected `'}'`, found `'_'`
 --> examples/uri_fail/src/main.rs:1:1
  |
1 | #![feature(plugin, decl_macro)]
  | ^
...
9 |     uri!(fail: "test");
  |     ------------------- in this macro invocation
```

Code to reproduce:
```
#![feature(plugin, decl_macro)]
#![plugin(rocket_codegen)]

extern crate rocket;

#[get("/")]
fn hello() -> &'static str {
    uri!(works: "test");
    uri!(fail: "test");
    "whatever"
}

#[get("/works/<used>")]
fn works(used: String) -> String {
    // can uri! to this one
    used
}

#[get("/fail/<_unused>")]
fn fail(_unused: String) -> &'static str {
    //can't uri! to this one
    "text"
}

fn main() {
    rocket::ignite().mount("/", routes![hello, works, fail]).launch();
}
```

`ROCKET_CODEGEN_DEBUG=1` reveals the cause:
```
--> codegen/src/macros/uri.rs:172
Emitting URI expression: expr(4294967295: {
    let _unused =
        <String as
            ::rocket::http::uri::FromUriParam<_>>::from_uri_param("test");
    ::rocket::http::uri::Uri::from(format!("/fail/{_unused}" , _unused = &
                                           _unused as &
                                           ::rocket::http::uri::UriDisplay ,))
})
```

Apparently the format! macro can't be used with leading underscores in names, which is what the compiler suggests to do with unused variables.

I believe Rocket should either work around this or print a better diagnostic.

jebrosen:
Fixes #603.

### DIFF
--- a/core/codegen/src/macros/uri.rs
+++ b/core/codegen/src/macros/uri.rs
@@ -129,6 +129,7 @@ pub fn uri_internal(
     for (i, &(mut ident, ref ty)) in internal.fn_args.iter().enumerate() {
         let (span, mut expr) = (exprs[i].span, exprs[i].clone());
         ident.span = span;
+        let f_ident = Ident::from_str(&("f_".to_string() + &ident.to_string()));
 
         // path for call: <T as FromUriParam<_>>::from_uri_param
         let idents = split_idents("rocket::http::uri::FromUriParam");
@@ -160,7 +161,7 @@ pub fn uri_internal(
 
         // generating: arg assignment tokens for format string
         let uri_display = quote_path!(ecx, ::rocket::http::uri::UriDisplay);
-        let mut tokens = quote_tokens!(ecx, $ident = &$ident as &$uri_display,);
+        let mut tokens = quote_tokens!(ecx, $f_ident = &$ident as &$uri_display,);
         tokens.iter_mut().for_each(|tree| tree.set_span(span));
         format_assign_tokens.push(tokens);
     }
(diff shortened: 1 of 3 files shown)


core/codegen/src/macros/uri.rs
```diff
@@ -129,6 +129,7 @@ pub fn uri_internal(
     for (i, &(mut ident, ref ty)) in internal.fn_args.iter().enumerate() {
         let (span, mut expr) = (exprs[i].span, exprs[i].clone());
         ident.span = span;
+        let f_ident = Ident::from_str(&("f_".to_string() + &ident.to_string()));
```
SergioBenitez: Can we simply rebind `ident` here so that we don't have two `ident`-like variables? Also, can you add a comment above the redefinition explaining what/why this is happening?

SergioBenitez:
Perfect. Can you also add a test case to `core/codegen/tests/typed-uris.rs` as a regression test?

PR #674, opened 2018-06-24
Improve session example.

incon:
Set dev secret_key.
Remove success message because it’s never displayed.

### DIFF
--- a/examples/session/Rocket.toml
+++ b/examples/session/Rocket.toml
@@ -1,3 +1,6 @@
+[dev]
+secret_key = "itlYmFR2vYKrOmFhupMIn/hyB6lYCCTXz4yaQX89XVg="
+
 [staging]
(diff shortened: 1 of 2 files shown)


examples/session/Rocket.toml
```diff
@@ -1,3 +1,6 @@
+[dev]
+secret_key = "itlYmFR2vYKrOmFhupMIn/hyB6lYCCTXz4yaQX89XVg="
```
SergioBenitez: Let's set this globally instead:

```toml
[global]
secret_key = "itlYmFR2vYKrOmFhupMIn/hyB6lYCCTXz4yaQX89XVg="
```

incon: I've updated the configuration. Made secret_key global, removed unneeded staging settings.

SergioBenitez:
Merged in 76bbcb5de8d2fae6f13daa08d44e4ec729a9e11c.

PR #675, opened 2018-06-26
Implementing FromRequest for &'a Request<'r>

mettke:
This PR allows using `&Request` in request handlers created via the codegen module

### Example

```rust
#[get("/example/<test>")]
pub fn example(request: &Request, test: &RawStr) -> String {
    "test".into()
}
```

### DIFF
--- a/core/lib/src/request/request.rs
+++ b/core/lib/src/request/request.rs
@@ -16,6 +16,7 @@ use error::Error;
 use http::{Method, Header, HeaderMap, Cookies, CookieJar};
 use http::{RawStr, ContentType, Accept, MediaType};
 use http::hyper;
+use outcome;
 
 #[derive(Clone)]
 struct RequestState<'r> {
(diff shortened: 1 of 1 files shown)


SergioBenitez:
This is very purposefully _not_ implemented so that you are required to use a request guard to validate and parse any information you might need from a request.

mettke:
Sry, but even thou I can understand the purpose behind it, it still sounds kinda odd to me.

According to the [docs](https://api.rocket.rs/rocket/struct.Route.html#example-1), I can get the raw request object in a request handler by using:
```rust
use rocket::{Request, Route, Data};
use rocket::handler::Outcome;
use rocket::http::Method;

fn handler<'r>(request: &'r Request, _data: Data) -> Outcome<'r> {
    Outcome::from(request, "Hello, world!")
}

// this is a rank 1 route matching requests to `GET /`
let index = Route::ranked(1, Method::Get, "/", handler);
```

Why is it allowed to use the raw object when I'm manually specifying the routes and not allowed when I'm using codegen. Especially as it would render my second PR useless 'cause no one would be able to get the request object to check if a template is available except by defining all routes manually.

SergioBenitez:
The only sanctioned API is the codegen-based one, which compiles down to use the raw API you're alluding to. You can use the raw API if necessary, of course, but the use-cases should be near zero. It exists because it is necessary, but the vast majority of users should have no need to use. The codegen API should be safer, more secure, and easier to use.

Issue #404, opened 2017-08-31
Request: Access to managed state in `from_param`


rbalicki2:
* I love the pattern shown in https://rocket.rs/guide/state/#usage where the `DbConn` is created from the managed state pool and then exposed as a Request Guard.
* I'd like to do be able to do the same thing in a `FromParam` implementation. Namely, I want to enable

```
#[get("/todos/<todo>")]
pub fn get_todo(
  todo: Todo
) -> Json<Todo> {
  Json(todo)
}
```

And have Todo's `FromParam` implementation fetch a todo with a particular ID.

* Right now, I use a separate `db::establish_connection` function which returns a `ConnectionResult<PgConnection>`. This is not ideal, as it doesn't share the same pool.
* Alternatively, I could implement `FromRequest` for `Todo`, and call `request.route().unwrap().uri`. However, then Rocket complains that `x is declared as an argument but isn't in the function signature`. Likewise, I wouldn't know which path parameter to match (e.g. it is ambiguous in the case of `/user/<user>/photos/<photo>` where we want to do this for both users and photos)
* I think ideally, one should be able to do this in other places, for example when deserializing a struct from data, although I haven't thought very deeply about how that would actually happen.
* One possible solution is to make the request accessible from the `RawStr`
* Or, one could pass an object `PathParamWrapper` which dereferences or `Into`'s to the `RawStr` (honestly, I'm not sure if what I just said makes sense).

inferiorhumanorgans:
> In my opinion this isn't needed & should be in your routing function, letting you handle errors correctly. Users aren't passing you a Todo; they're passing you a reference to one they'd like you to resolve (if you can).

Being able to access state (e.g. the database pool) from a parameter guard would be useful here for turning the reference into a todo object.

Currently you can shove db requests into a request guard but then you'll have no real way of figuring out what the reference actually is.  In my use case I've got a path with a variety of segments representing objects in the database.  Some classes are used multiple times in the path, and some classes will want to do a bit of inspection higher up the hierarchy.  My current solution is lots of boilerplate code in each route which I'd love to get rid of.

Jayshua:
I'll add my use case: Ensuring the user has access to the resource they are requesting. For example, imagine a todo:

```rust
#[get("/todo/<id>")]
fn get_todo(id: TodoId, authorization: ApiToken) -> Json<Todo> {
   ...
}
```

[...]

PR #676, opened 2018-06-26
Allow users to check whether a template is available

mettke:
This PR allows users to check whether a template is available and thus allowing them to provide a fallback

### Example

```rust
#[get("/example/<test>")]
pub fn example(request: &Request, test: &RawStr) -> Result<Template, Status> {
    Template::contains(test, request)
        .and_then(|available| {
            if available {
                return Ok(Template::render(test.to_string(), json!({})));
            }
            return Ok(Template::render("alternative", json!({})));
        })
}
```


SergioBenitez:
This is very purposefully _not_ implemented so that you are required to use a request guard to validate and parse any information you might need from a request.

SergioBenitez:
Apologies! Commented on the wrong PR!

jebrosen:
I don't like this specific implementation, mostly because of the interaction with #675. I do think it has merit, though. Would it possible to instead expose this functionality in the form of a request guard? I imagine the example you gave could then look something like this:

```rust
#[get("/example/<template>")]
pub fn example(template: Option<ValidTemplate>) -> Template {
    match template {
        Some(ValidTemplate(name)) => Template::render(name, json!({})),
        None => Template::render("alternative", json!({})),
    }
}
```

mettke:
@jebrosen Implementing it as request guard certainly is an awesome idea. That way #675 wouldn't be necessary at all. 

The only problem I see is that it kinda limits interactions. Being able to actively check whether a template exists or not allows people to be a lot more creative what template to return in which case. It allows appending a prefix or a postfix and multiple choice instead of a binary one.

Another option would be to create a new struct which contains the request (private) and exposes a function like `template_exists()`. That way we are able to add new functionality which requires direct access to the template engine without exposing too much.

```rust
#[get("/example/<template>")]
pub fn example(test: &RawStr, engine: TemplateEngine) -> Template {
    if engine.template_exists(test) {
         Template::render(name, json!({}))
    }
    Template::render("alternative", json!({}))
}
```

jebrosen:
I realize now the example I wrote above would need "ValidTemplate" to be `FromParam`, but `FromParam` can't currently access managed state (#404).

I agree that your second example with the hypothetical `TemplateEngine` type is more flexible, and also more readily implemented.

SergioBenitez:
I would happily review a PR for the second option outlined above.

contrib/src/templates/template_engine.rs
```diff
@@ -0,0 +1,55 @@
...
+    ///         return Template::render("specfic", json!({}));
+    ///     }
+    ///     Template::render("fallback", json!({}))
+    /// }
+    /// ```
+    pub fn template_exists(&self, name: &str) -> bool {
+        self.ctxt.templates.contains_key(name)
+    }
+
+    fn get_context(req: &'a Request) -> Result<State<'a, Context>, Status> {
```
mettke: We have code similar to this one in template [mods.rs](https://github.com/SergioBenitez/Rocket/blob/master/contrib/lib/src/templates/mod.rs#L332). In fact that file contains it at 2 different locations. 

Maybe we want to create a global function to get the context in these 3 cases?

contrib/lib/src/templates/template_engine.rs
```diff
@@ -0,0 +1,55 @@
...
+    ///         return Template::render("specfic", json!({}));
+    ///     }
+    ///     Template::render("fallback", json!({}))
+    /// }
+    /// ```
+    pub fn template_exists(&self, name: &str) -> bool {
+        self.ctxt.templates.contains_key(name)
+    }
+
+    fn get_context(req: &'a Request) -> Result<State<'a, Context>, Status> {
```
mettke: We have code similar to this one in template [mods.rs](https://github.com/SergioBenitez/Rocket/blob/master/contrib/lib/src/templates/mod.rs#L332). In fact that file contains it at 2 different locations. 

Maybe we want to create a global function to get the context in these 3 cases?

SergioBenitez:
We also need some (checked) tests/examples and a bit more thorough documentation. You should include a complete example in the docstring for `TemplateEngines` as well as some unit tests in `contrib/lib/tests`.

mettke:
I have added two tests (for tera and handlebars) and an example in the docstring. I'm, however, not sure how to create good checked examples, 'cause it requires a request handler which cannot be generated in an example (as far as I know)

contrib/lib/tests/templates.rs
```diff
@@ -48,12 +51,26 @@ mod templates_tests {
             let template = Template::show(&rocket, "tera/html_test", &map);
             assert_eq!(template, Some(ESCAPED_EXPECTED.into()));
         }
+
+        #[test]
+        fn test_template_engine_with_tera() {
+            let rocket = rocket();
+            rocket.attach(AdHoc::on_request(|req, _| {
```
SergioBenitez: This is never actually run, so this test isn't checking anything.

contrib/lib/tests/templates.rs
```diff
@@ -69,5 +86,15 @@ mod templates_tests {
             let template = Template::show(&rocket, "hbs/test", &map);
             assert_eq!(template, Some(EXPECTED.into()));
         }
+
+        #[test]
+        fn test_template_engine_with_handlebars() {
+            let rocket = rocket();
+            rocket.attach(AdHoc::on_request(|req, _| {
```
SergioBenitez: Same here.

SergioBenitez:
We still need unit tests that do something. I believe that simply switching the request fairing into an attach fairing in the unit tests you've modified suffices. Please give this a try and ensure you have both positive and negative tests.

contrib/lib/src/templates/mod.rs
```diff
@@ -6,6 +6,7 @@ extern crate glob;
 #[cfg(feature = "handlebars_templates")] mod handlebars_templates;
 mod engine;
 mod context;
+pub mod template_engine;
```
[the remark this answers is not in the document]
mettke: Unfortunately, it has to be. Otherwise, compilation fails with

```
error[E0603]: module `template_engine` is private
  --> contrib/lib/src/lib.rs:77:40
   |
77 | pub use templates::{Template, Engines, template_engine::TemplateEngine};```

File: contrib/lib/src/templates/template_engine.rs [hunk 6]
```diff
@@ -0,0 +1,88 @@
...
+/// ```rust,ignore
+/// #[get("/")]
+/// fn homepage(engine: TemplateEngine) -> Template {
+///     if engine.template_exists("specific") {
+///         return Template::render("specfic", json!({}));
+///     }
+///     Template::render("fallback", json!({}))
+/// }
+/// ```
+pub struct TemplateEngine<'a> {
```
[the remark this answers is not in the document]
mettke: My suggestion would be `TemplateMetadata`. Guess `EngineMetadata` wouldn't be verbose enough to the user, even thou the name is used internally

contrib/lib/src/templates/template_engine.rs
```diff
@@ -0,0 +1,88 @@
...
+    ctxt: State<'a, Context>,
```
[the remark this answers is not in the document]
mettke: I don't know how. I can't use `.guard::<State<&'a Context>>()` in combination with `*ctxt`. That is an unmanaged state

I can't use `.guard::<State<Context>>()` in combination with `&*ctxt` 'cause the local variable wouldn't live long enough

contrib/lib/src/templates/mod.rs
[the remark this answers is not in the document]
SergioBenitez: You want `mod metadata` here (assuming you rename the `template_engine.rs` file to `metedata.rs`) and a `pub use self::metadata::TemateMetadata'` below.

contrib/lib/src/templates/template_engine.rs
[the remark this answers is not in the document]
SergioBenitez: You can call `.inner()` on the `State` type.

contrib/lib/src/templates/template_engine.rs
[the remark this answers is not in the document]
SergioBenitez: Yeah, that sounds good to me!

contrib/lib/tests/templates.rs
```diff
@@ -48,12 +53,30 @@ mod templates_tests {
             let template = Template::show(&rocket, "tera/html_test", &map);
             assert_eq!(template, Some(ESCAPED_EXPECTED.into()));
         }
+
+        #[test]
+        fn test_template_engine_with_tera() {
+            let rocket = rocket().attach(AdHoc::on_request(|req, _| {
+                let engine = TemplateMetadata::from_request(req).unwrap();
+                assert_eq!(engine.contains_template("tera/txt_test"), true);
```
SergioBenitez: Use `assert!(engine.contains_template("tera/txt_test"))`.

contrib/lib/tests/templates.rs
```diff
@@ -48,12 +53,30 @@ mod templates_tests {
...
+                assert_eq!(engine.contains_template("tera/html_test"), true);
+                assert_eq!(engine.contains_template("tera/not_existing"), false);
```
SergioBenitez: Use `assert!(!engine.contains_template("tera/not_existing"))`.

contrib/lib/tests/templates.rs
```diff
@@ -69,5 +92,17 @@ mod templates_tests {
             let template = Template::show(&rocket, "hbs/test", &map);
             assert_eq!(template, Some(EXPECTED.into()));
         }
+
+        #[test]
+        fn test_template_engine_with_handlebars() {
+            let rocket = rocket().attach(AdHoc::on_request(|req, _| {
+                let engine = TemplateMetadata::from_request(req).unwrap();
+                assert_eq!(engine.contains_template("hbs/test"), true);
```
SergioBenitez: Same here with the `assert_eq!`.

PR #680, opened 2018-06-28
Reimplement routes! and catchers! as proc_macros

jebrosen:
What the title says. Also adds compiletest to the codegen_next crate and updates all examples and tests to properly import the macros.

Unfortunately, at present `#![feature(proc_macro_non_items)]` is required on downstream crates. They are also imported with `use` like items are instead of with `#[macro_use]`, which makes for a noisy commit.

This branch does not currently move any of the associated documentation.

### DIFF
--- a/core/codegen_next/src/prefixing_vec.rs
+++ b/core/codegen_next/src/prefixing_vec.rs
@@ -1,0 +1,44 @@
+use proc_macro::TokenStream;
+use proc_macro2::TokenStream as TokenStream2;
+
+use syn::{Ident, Path};
+
+use derive_utils::parser::{Parser, Seperator, Result as PResult};
+
+#[inline]
+fn prefix_path(prefix: &str, path: &mut Path) {
+    let mut last_seg = path.segments.last_mut().expect("last path segment");
+    let last_value = last_seg.value_mut();
+    last_value.ident = Ident::new(&format!("{}{}", prefix, last_value.ident), last_value.ident.span());
+}
+
+pub fn prefixing_vec_macro_internal<F>(prefix: &str, to_expr: F, args: TokenStream) -> PResult<TokenStream>
+        where F: FnMut(Path) -> TokenStream2
+{
+    let mut parser = Parser::new(args);
+    let mut paths = parser.parse_sep(Seperator::Comma, |p| {
+        p.parse::<Path>()
+    })?;
+    parser.eof().map_err(|_| {
+        parser.current_span()
+            .error("expected `,` or `::` or end of macro invocation")
+    })?;
+
+    for ref mut p in &mut paths {
+        prefix_path(prefix, p);
+    }
+    let path_exprs: Vec<_> = paths.into_iter().map(to_expr).collect();
+
+    let tokens = quote! { vec![#(#path_exprs),*] };
+    Ok(tokens.into())
+}
+
+pub fn prefixing_vec_macro<F>(prefix: &str, to_expr: F, args: TokenStream) -> TokenStream
+    where F: FnMut(Path) -> TokenStream2
+{
+    prefixing_vec_macro_internal(prefix, to_expr, args)
+        .unwrap_or_else(|diag| {
+            diag.emit();
+            (quote! { vec![] }).into()
+        })
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,4 +41,5 @@ members = [
   "examples/raw_sqlite",
   "examples/tls",
   "examples/fairings",
+  "examples/hello_2018",
 ]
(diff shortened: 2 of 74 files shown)


core/codegen_next/tests/compiletest.rs
```diff
@@ -0,0 +1,104 @@
...
+    let mut config = compiletest::Config::default();
+    config.mode = mode.parse().expect("Invalid mode");
+    config.src_base = format!("tests/{}", mode).into();
+    config.clean_rmeta();
+
+    config.target_rustcflags = Some([
+                                    link_flag("-L", "crate", &[]),
+                                    link_flag("-L", "dependency", &["deps"]),
+                                    extern_dep("rocket_codegen", Kind::Dynamic).expect("find codegen dep"),
+                                    extern_dep("rocket", Kind::Static).expect("find core dep")
```
jebrosen: Should `rocket_codegen_next` go in this list as well?

core/codegen_next/src/prefixing_vec.rs
```diff
@@ -0,0 +1,32 @@
...
+}
+
+pub fn prefixing_vec_macro<F>(prefix: &str, to_expr: F, args: TokenStream) -> PResult<TokenStream>
+        where F: FnMut(Path) -> TokenStream2
+{
+    let mut parser = Parser::new(args);
+    let mut paths = parser.parse_sep(Seperator::Comma, |p| {
+        p.parse::<Path>()
+    })?;
+    parser.eof()?;
```
jebrosen: Testing for `eof()` is necessary, otherwise something like `routes![a, b c]` will be accepted and parse as only `a,b`. The compiler plugin macro says it expected `,`, `::`, or `<eof>`. This proc_macro currently just says `expected eof`.

jebrosen: I will try replacing the error message with a better one, like "expected paths separated by commas".

jebrosen:
As discussed in IRC, it would be nice if users didn't need to import these macros just like today. 

Current situation:
`routes!`, `catchers!`, `uri!`, and `rocket_internal_uri!` appear to be "magic" globally available macros because they're globally registered via the compiler plugin. They don't need to be (and probably can't) be imported via any means other than the plugin, that I know of.

proc_macro situation:
`routes!` and `catchers!` (so far) are proc_macros. These are only available via `use_extern_macros`, whether by a use statement or a full path at the call site. I thought that a `macro_rules` macro could get around that, but I always seem to end up needing `use_extern_macros` in the consumer crate anyway. It does look like it will be possible to use a `macro_rules` macro in conjunction with some `pub use` and `$crate` in order to allow `#[macro_use]` instead of `use rocket::routes;`, but the feature gate may have to stay.

2018 crates:
Crates targeting 2018 can drop 'extern crate' declarations, and importing individual macros via `use` is posed as an alternative for `#[macro_use]`. The question is, what path should Rocket's macros be available at for 2018 crates. I will do some more testing to figure out how modules, paths, and imports interact with this.

Proposed design, assuming this all works:
* `proc_macro`s (routes, catchers, etc.) are defined in a `proc_macro` crate
* The `rocket` crate will import the `proc_macro`s into <MODULE A>
* The `rocket` crate will define `macro_rules` wrapper macros in <MODULE B> that call `$crate::<MODULE A>::macro!`
* 2015 crates will now add `#[macro_use]` to `extern crate rocket;` and have access to the `macro_rules` macros
* 2018 crates do not need `#[macro_use]`. They should be able to `use` macros from either <MODULE A> or <MODULE B> (I will verify this) but we should recommend one or the other, probably the one with the direct proc_macro reexports.

One of <MODULE A> or <MODULE B> could be the crate root, depending on how we want the paths to look to the 2018 crates that import macros.

Note that attribute proc_macros and proc_macro derives are not covered above. They are in a similar situation as far as imports go, but a macro_rules workaround can't be used in this situation. Derives appear to be covered by `#[macro_use]`, I'm not sure about attributes though.

jebrosen:
Now implements the following variant of the above design:

* `routes` and `catchers` are reimplemented as `proc_macros`: `rocket_routes_internal` and `rocket_catchers_internal`, in `rocket_codegen_next`
* `rocket` already glob-reexports the macros from `rocket_codegen_next`
* `rocket` now defines a macro_rules! macro called `routes!` that simply forwards its arguments to `$crate::rocket_routes_internal`, and the same with `catchers!`
  * *this requires use_extern_macros, which will hopefully be stable soon. It is already active in Rocket's examples and tests via the decl_macro feature.*
* All consumer crates need to add `#[feature(proc_macro_path_invoc)]`, and import the macros:
  * Import Option A, idiomatic for Rust 2015: add a `#[macro_use] extern crate rocket;`
  * Import Option B, idiomatic for Rust 2018: add a `use rocket::routes;` or `use rocket::catchers;` and omit `extern crate rocket;` completely.

My plan is to investigate the use of `proc_macro_hack` next. Hopefully it will make `#[feature(proc_macro_path_invoc)]` and maybe even the macro_rules wrappers unnecessary. `mashup` is another possibility, but the error messages can be very confusing when `routes!` is used incorrectly.

SergioBenitez:
Merged in 8e779610c499b645edcd0448c062b2654dce29a3. Woo!

Issue #419, opened 2017-09-06
The [#error]  attribute conflict with log crate's "error!"


zoumi:
compile error:
```
error: macro `error` may not be used in attributes
  --> src\main.rs:95:1
   |
95 | #[error(404)]
   | ^^^^^^^^^^^^^

error: aborting due to previous error

error: Could not compile ....
```

SergioBenitez:
See #21 and https://github.com/SergioBenitez/Rocket/issues/109#issuecomment-300326193.

PR #681, opened 2018-06-28
deprecate '#[error]' and 'errors!

pjenvey:
with their use now emitting a deprecation warning

(as discussed on #rocket-dev as an alternative fix for #419)


codegen/src/decorators/catch.rs
```diff
@@ -98,3 +98,15 @@ pub fn catch_decorator(
...
     output
 }
+
+pub fn error_decorator(
+    ecx: &mut ExtCtxt,
+    sp: Span,
+    meta_item: &MetaItem,
+    annotated: Annotatable
+) -> Vec<Annotatable> {
+    ecx.struct_span_warn(sp, "the `error` decorator has been deprecated as of Rocket v0.3.15")
```
SergioBenitez: We should use the word "attribute" in the warning instead of decorator. I just used the word decorator in the source code because...arbitrary decision. :)

SergioBenitez: Also, let's align this closer to the real deprecated message. Let's say:

```
warning: use of deprecated Rocket attribute `error` (deprecated since v0.3.15)
help: the `error` attribute was replaced by the `catch` attribute:`#[catch(..)]`
```

Same goes for the `errors` bang macro, of course.

SergioBenitez:
Fantastic! And last but not least, can you include a `compile-fail` test that checks that the warning is in fact emitted? You can use `codegen/tests/compile-fail/unknown-media-type.rs` as a template.

SergioBenitez:
Excellent work! Merged in 9ac9cd462d52564db3a7c22ef3809fd07c107150. This will be part of the next 0.3.x release.

PR #685, opened 2018-07-03
Adds option to disable logging completely

### LINKED ISSUE #496: RFC: ability to disable logging
After following the various bug reports/feature requests on the topic of logging (https://github.com/SergioBenitez/Rocket/issues/21, https://github.com/SergioBenitez/Rocket/issues/152, etc), I decided to look into whether it would be possible to implement logging as a Fairing. I have a proof-of-concept using `slog`, along with a fork of `rocket.rs` that adds the ability to disable Rocket's built-in logging. The user would put something like `log = "off"` in their config to achieve this. I also have an implementation of a Fairing that allows the user to pass their own slog `Logger` into the application, giving the user more flexibility of where the logs go, their format, and how they are handled.

Before going any further, I wanted to find out whether something like this was worth pursuing.

Thanks for any & all comments

brizental:
Fixes #496 

Let me know what you guys think about this or if this is a feature you don't want we can just close it up :)

Basically I added an `off` option for `log` here. What it does is log nothing. As I mentioned in my comment this is pretty useful when you want to implement your own logging mechanism and don't want rocket's default logs to get in the way (as cool as they are).

r? @SergioBenitez


core/lib/src/logger.rs
```diff
@@ -77,7 +82,7 @@ macro_rules! warn_ { ($($args:expr),+) => { log_!(warn: $($args),+); }; }
 impl log::Log for RocketLogger {
     #[inline(always)]
     fn enabled(&self, record: &log::Metadata) -> bool {
-        record.target().starts_with("launch") || record.level() <= self.0.max_log_level()
+        record.target().starts_with("launch") || self.0.max_log_level().is_some() && self.0.max_log_level().unwrap() >= record.level()
```
DJMcNab: Could this also be something like:
```rust
record.target().starts_with("launch") || if let Some(level) = self.0.max_log_level() { level >= record.level() } else { false }
```

It's probably better to avoid the unwrap to improve maintainability, but it does seem a shame to require the explicit `false`.

brizental: I'm actually going to change this to

```
record.target().starts_with("launch") && self.0.max_log_level().is_some() || 
if let Some(level) = self.0.max_log_level() { 
   level >= record.level() 
} else { 
   false 
}
```

Because this also gets rid of the launch logs so we actually have no rocket logs whatsoever.

vladikoff:
@pwoolcoc @jrozner ^

SergioBenitez:
I'm in favor of an "off" option for logging levels, but this is not the correct way to implement such a feature. The `RocketLogger` should never be aware that logging is disabled. In general, if `RocketLogger` is being used, then logging is happening; that's its job. Thus, disabling logging needs to prevent `RocketLogger` from ever existing at all.

As I alluded to in #496, logging can already be disabled by passing `false` to `custom`, thus we can use the path that the code takes when `false` is passed in as an example of how a correct implementation of this feature should look. In particular, we can see that in `core/lib/src/rocket.rs`, the `Rocket::configured` method only enables logging if `log` is `true`. With an implementation of this feature, the `log` parameter to `Rocket::custom` should be removed, and the check should instead observe that the `log` configuration parameter has been set to `off`.

In short, the changes needed are:

1. Add `LoggingLevel::Off` that parses from `"off"`. (this is already done)
2. Remove the `log` parameter from `Rocket::custom()` and `Rocket::configured()`.
a. Additionally, fix up the documentation to note this change.
3. Only enable logging (in `core/lib/src/rocket.rs`) if `config.log_level != Off`. This can be checked in either `try_init` or in `core/lib/src/rocket.rs`. I'd change `try_init` to return a `bool` that is `true` only if logging was actually enabled. The log enabling code in `rocket.rs` then becomes:

    ```rust
    if logger:try_init(config.log_level) {
        logger::push_max_level(logger::LoggingLevel::Normal);
    }
    ``` 

Finally, the implementation of the `enabled()` method is odd. I'd recommend do something like the following instead:

```rust
match self.0.max_log_level() {
    Some(max) => record.level() <= max || record.target().starts_with("launch"),
    None => false
}
```

brizental:
@SergioBenitez thanks for such a detailed review!
I tried to apply all the stuff you reviewed, let me know if it's getting better :)

core/lib/src/logger.rs
```diff
@@ -17,15 +17,27 @@ pub enum LoggingLevel {
...
+            LoggingLevel::Debug => Some(log::Level::Trace),
+            LoggingLevel::Off => None
+        }
+    }
+
+    fn to_level_filter(&self) -> log::LevelFilter {
+        match *self {
+            LoggingLevel::Critical => log::LevelFilter::Warn,
+            LoggingLevel::Normal => log::LevelFilter::Info,
+            LoggingLevel::Debug => log::LevelFilter::Debug,
```
SergioBenitez: This should be `log::LevelFilter::Trace`.

SergioBenitez: This should be `log::LevelFilter::Trace`.

core/lib/src/logger.rs
```diff
@@ -17,15 +17,27 @@ pub enum LoggingLevel {
...
     /// Shows everything.
     Debug,
+    /// Shows nothing.
+    Off,
 }
 
 impl LoggingLevel {
     #[inline(always)]
-    fn max_log_level(&self) -> log::Level {
+    fn max_log_level(&self) -> Option<log::Level> {
```
SergioBenitez: This can now simply be: `self.to_level_filter().to_level()` (with a correct implementation of `to_level_filter()` below).

SergioBenitez: This wasn't implemented.

coffeebe4code:
@SergioBenitez I have read a few of these "issues" PR's, is it possible you would allow a different logging framework to be "injected" in. ie, we have all the necessary trait implementations you need, and pass ownership to rocket?

We use Dynatrace at our company, and are big in Open Telemetry, I would like to log our requests to fit into the OTEL model, so it is easily digestible in cloud native infrastructures.

Issue #560, opened 2018-02-08
Overly strict FromSegments implementation for PathBuf


Noughmad:
In the interest of safety, the implementation of `FromSegments` for `PathBuf` ([code](
https://github.com/SergioBenitez/Rocket/blob/8b1aaed0cecb0a26dde8f1aa4ce89a2a4003f957/lib/src/request/param.rs#L339)) rejects all segments that start with a dot (`.`). I believe this is overly strict, since

1. The dot-dot case (a filename of exactly `..`) is already handled.
2. A single dot segment (exactly `.`) could also be valid, and just skipped.
3. Lots of hidden files start with `.`, such as `.gitignore` and other config files.

In my use case I only require the third case, as I am trying to create a file browser, where the path is specified as `<path..>` and `path: PathBuf`. However, such a route cannot browse a file with a name starting with a dot, and I get a `Failed to parse 'path': BadStart('.')` error message. 

I can reproduce this behavior on version `0.3.6` on ArchLinux, but looking at the code I can see it is present in `master` as well.

zxvfxwing:
Hi,

I have also encountered this behavior when I wanted to accept dotfiles (as you say) in my routes.
@SergioBenitez answered me on matrix channel / IRC bridge after exposing my problem.
You can read the entire conversation [here](https://riot.im/app/#/room/!NkeCEOxYAAanlPMmEf:matrix.org/$15106551905wxJcZ:spokonline.net).

So I made a [workaround](https://github.com/zxvfxwing/fancyndex/blob/v1.0/src/filesystem/unsafepath.rs) (used in [one of my personal](https://github.com/zxvfxwing/fancyndex/blob/96e4cae57dbedc732a3064ede1b880cf1f8fbf0f/src/api/home.rs#L53) project) which is certainly not perfect but works.

Use case:

```rust
#[get("/<unsafe_p..>")]
pub fn dotfiles(unsafe_p: UnsafePBuf) -> ... {
    let dotpath = unsafe_p.path();
    /* Do wathever you want with your path */
}
```
Hopefully, my answer met your expectations.

SergioBenitez:
The danger is that users will unknowingly leak hidden files if this change is made. I'd like to keep the safe-by-default behavior while also allowing users to disable this on a per-case basis.

[...]

Issue #239, opened 2017-03-22
Static file serving


luser:
First off: let me just say that Rocket is very nice and has been a pleasure to use, kudos!

I looked around the docs and couldn't find any support for serving a set of static files from a directory. This is something that I've seen supported in every web framework I've ever used. As an arbitrary example, [here's how it works in Flask](http://flask.pocoo.org/docs/0.12/quickstart/#static-files). This is a pretty common pattern to serve things like CSS files and JS scripts, so I think it would be useful to have. Obviously one can serve individual files by just having a handler that returns a `File`, but this seems like generally useful functionality. Ideally one would simply be able to do something like `rocket::ignite().mount("/static", StaticFiles(some_path)`.

haheute:
I am not sure if this helps.. Have you seen the the `static_files` example ?

SergioBenitez:
Indeed, this is covered in the [static_files example](https://github.com/SergioBenitez/Rocket/tree/master/examples/static_files).

luser:
I had not seen that, thanks! I did eventually find [the bit about segments parameters](https://rocket.rs/guide/requests/#dynamic-segments) which has that same code as an example, but I still think a helper that simplified this extremely common use case would be nice. It took me a while to find that part of the documentation because I was looking around for "static file serving" and I did not expect to find it under "Dynamic Segments". :)

SergioBenitez:
@luser Perhaps something in `contrib` that you can simply mount. So it would look something like:

```rust
use rocket_contrib::StaticFiles;

rocket::ignite().mount("/path_to_serve_on", StaticFiles::from("/path_to_serve_from"));
```

This would presumably serve files from the `/path_to_serve_from` directory at the path `/path_to_serve_on`.

I think something like this would be nice, though I'm not sure if it's actually possible to do this in a clean way at the moment. I'll look into it.

luser:
That sounds great!

theduke:
I think the way to do it right now is sufficiently simple, just showing how  in the guide under "Serving Static Files" or something alike would suffice.

lnicola:
I think it would be best to offer something that supports HTTP caching.

marti1125:
I have a question I am using bootstrap only (/css/bootstrap.min.css, /css/bootstrap-theme.min.css)
It shows errors in console  (GET /css/bootstrap-theme.min.css.map:) ??
any reason about this =?
[...]

PR #696, opened 2018-07-09
Allow "." but disallow ".." in segments

soyuka:
Hello,

You may want to allow `.` in your paths (for example `.hidden`) and the `starts_with('.')` prevents this.

I'm not sure that this is the best fix, maybe that we should check that the segment equals `.` instead?

### DIFF
--- a/core/lib/src/request/param.rs
+++ b/core/lib/src/request/param.rs
@@ -336,7 +336,7 @@ impl<'a> FromSegments<'a> for PathBuf {
 
             if decoded == ".." {
                 buf.pop();
-            } else if decoded.starts_with('.') {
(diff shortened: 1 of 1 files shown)


DJMcNab:
For prior discussion see https://github.com/SergioBenitez/Rocket/issues/560#issuecomment-364695939

soyuka:
Oh nice thanks @DJMcNab I should've looked it up before :). 

Anyway I'm not using this for Static, I just find the default guard a bit too much restrictive in term of paths. Going to clause and follow #239 especially that [this workaround](https://github.com/SergioBenitez/Rocket/issues/560#issuecomment-364173013) is easy to set up.

PR #705, opened 2018-07-13
future-proof FromForm against scope resolution changes

jebrosen:
See rust-lang/rust#50504, rust-lang/rust#51952.

There is uncertainty regarding `proc_macro` hygiene and whether this actually future-proofs, and those concerns would need to be addressed before taking this approach.

### DIFF
--- a/core/codegen_next/src/lib.rs
+++ b/core/codegen_next/src/lib.rs
@@ -74,7 +74,7 @@ fn real_derive_from_form_value(input: TokenStream) -> PResult<TokenStream> {
         impl<'v> ::rocket::request::FromFormValue<'v> for #name {
             type Error = &'v ::rocket::http::RawStr;
 
-            fn from_form_value(v: &'v ::rocket::http::RawStr) -> Result<Self, Self::Error> {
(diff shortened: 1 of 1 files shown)


core/codegen_next/src/lib.rs
```diff
@@ -71,24 +71,15 @@ fn real_derive_from_form_value(input: TokenStream) -> PResult<TokenStream> {
...
-        mod scope {
-            extern crate std;
-            extern crate rocket;
+        impl<'v> ::rocket::request::FromFormValue<'v> for #name {
+            type Error = &'v ::rocket::http::RawStr;
 
-            use self::std::prelude::v1::*;
-            use self::rocket::request::FromFormValue;
-            use self::rocket::http::RawStr;
+            fn from_form_value(v: &'v ::rocket::http::RawStr) -> Result<Self, Self::Error> {
```
SergioBenitez: Should this also be `::std::result::Result`?

jebrosen:
Instead of this significant change, `use super::*;`, `use super::#name;`, or typing out `super::#name` as necessary should be no better or worse than the current code wrt hygiene, but will silence the warning in consuming crates.

SergioBenitez:
Merged in 706cd320539c79aea0f7c8b51726dad088fcf7f4.

PR #713, opened 2018-07-19
spelling and grammar fixes in core and contrib crates

jebrosen:
Covers core and contrib only -- examples, tests, and codegen not included in this pass. Most changes should be non-controversial. A few more opinionated changes are in there somewhere, though.

Regarding functions that "set" or "add" a value, some of them are written in the style of "[this function] Sets the ..." or "[this function] Adds the..." and others in the imperative: "Set the..." or "Add the...". I noticed and changed a few of these to stay consistent with nearby functions, but it would be nice to have a consistent style at some point.


jebrosen:
Merged in 0c80217.

PR #716, opened 2018-07-19
Send response when there is unread body data

Yarn:


SergioBenitez:
Thanks so much for investigating this and proposing a fix! I've merged the fix itself, without the tests, in 995603666. The tests were very useful in determining that this is a correct fix, however!

PR #720, opened 2018-07-28
Finish applying Clippy suggestions.

sstangl:
This finishes applying Clippy's basic code cleanup suggestions. It turns out that there were not many!

I've only applied Clippy suggestions that should be obviously-unobjectionable in review, based off discussion from the previous patch.

### DIFF
--- a/core/codegen/src/decorators/route.rs
+++ b/core/codegen/src/decorators/route.rs
@@ -87,12 +87,10 @@ impl RouteParams {
     }
 
     fn generate_data_statement(&self, ecx: &ExtCtxt) -> Option<Stmt> {
-        let param = self.data_param.as_ref().map(|p| &p.value);
-        let arg = param.and_then(|p| self.annotated_fn.find_input(&p.node.name));
-        if param.is_none() {
-            return None;
-        } else if arg.is_none() {
-            self.missing_declared_err(ecx, param.unwrap());
+        let param = self.data_param.as_ref().map(|p| &p.value)?;
+        let arg = self.annotated_fn.find_input(&param.node.name);
+        if arg.is_none() {
+            self.missing_declared_err(ecx, param);
             return None;
         }
 
--- a/core/lib/src/config/custom_values.rs
+++ b/core/lib/src/config/custom_values.rs
@@ -129,7 +129,7 @@ impl Limits {
             "forms" => self.forms = limit,
             _ => {
                 let mut found = false;
-                for tuple in self.extra.iter_mut() {
+                for tuple in &mut self.extra {
                     if tuple.0 == name {
                         tuple.1 = limit;
                         found = true;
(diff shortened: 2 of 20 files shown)


core/lib/src/router/route.rs
```diff
@@ -77,17 +77,16 @@ impl Route {
...
     /// ```
-    pub fn new<S>(m: Method, path: S, handler: Handler) -> Route
+    pub fn new<S>(method: Method, path: S, handler: Handler) -> Route
         where S: AsRef<str>
     {
         let uri = Uri::from(path.as_ref().to_string());
+        let rank = default_rank(&uri);
+
         Route {
+            method, handler, uri, rank,
```
jebrosen: This line (and again in `ranked`) now contains more than one of the fields, and they are in a different order. I'm not entirely sold on that.

sstangl: Sergio asked me to do that on the last patch, so I applied it again here.

core/codegen/src/decorators/route.rs
```diff
@@ -87,12 +87,10 @@ impl RouteParams {
     }
 
     fn generate_data_statement(&self, ecx: &ExtCtxt) -> Option<Stmt> {
-        let param = self.data_param.as_ref().map(|p| &p.value);
-        let arg = param.and_then(|p| self.annotated_fn.find_input(&p.node.name));
-        if param.is_none() {
-            return None;
-        } else if arg.is_none() {
-            self.missing_declared_err(ecx, param.unwrap());
+        let param = self.data_param.as_ref().map(|p| &p.value)?;
```
SergioBenitez: Clippy was able to find this change? That's awesome.

core/lib/src/config/custom_values.rs
```diff
@@ -211,7 +211,7 @@ pub fn u64(conf: &Config, name: &str, value: &Value) -> Result<u64> {
 
 pub fn u16(conf: &Config, name: &str, value: &Value) -> Result<u16> {
     match value.as_integer() {
-        Some(x) if x >= 0 && x <= (u16::max_value() as i64) => Ok(x as u16),
+        Some(x) if x >= 0 && x <= (i64::from(u16::max_value())) => Ok(x as u16),
```
SergioBenitez: Is there a reason to prefer this over the `as` coercion? The former seems clearer to me.

svmnotn: I had the same question, this seems to be the answer as to why clippy does this lint:
> Rust’s `as` keyword will perform many kinds of conversions, including silently lossy conversions. Conversion functions such as `i32::from` will only perform lossless conversions. Using the conversion functions prevents conversions from turning into silent lossy conversions if the types of the input expressions ever change, and make it easier for people reading the code to know that the conversion is lossless.

core/lib/src/router/route.rs
SergioBenitez: Yeah. This one seems messier, though. Oh well; let's just do this for consistency.

SergioBenitez:
@sstangl Would you be able to rebase?

SergioBenitez:
Awesome! Kept the vast majority of this (minus the `X::from` instead of `as`) and merged in 9bf585496. Thanks again, @sstangl! :)

PR #721, opened 2018-07-29
Adding and_then methods for outcome

mettke:
Currently, it is quite tedious to transform between two Outcomes. There is the `map` function, but unfortunately, it requires the result to be a success. 

```rust
impl<'a, 'r> FromRequest<'a, 'r> for Anything {
    type Error = ();

    fn from_request(request: &'a Request) -> Outcome<Self, ()> {
        let outcome = Something::from_request(request);
        match outcome {
            outcome::Outcome::Success(value) => {
                if value == "test" {
                    outcome::Outcome::Success(value)
                }
                outcome::Outcome::Forward(value)
            },
            outcome::Outcome::Forward(value) => outcome::Outcome::Forward(value),
            outcome::Outcome::Failure(value) => outcome::Outcome::Failure(value),
        }
    }
}
```

This PR comes with a functionality like [enum.Option.html#method.and_then](https://doc.rust-lang.org/std/option/enum.Option.html#method.and_then) allowing a mapping to an alternative outcome value. 

```rust
impl<'a, 'r> FromRequest<'a, 'r> for Anything {
    type Error = ();

    fn from_request(request: &'a Request) -> Outcome<Self, ()> {
        Something::from_request(request).and_then(|value| {
            if value == "test" {
                outcome::Outcome::Success(value)
            }
            outcome::Outcome::Forward(value)
        })
    }
}
```

I'm only unsure about the naming. Currently it is `and_then`, `and_then_forward` and `and_then_failure`. Maybe something like `success_then`, `forward_then` and `failure_then` would be more self explaining

### DIFF
--- a/core/lib/src/outcome.rs
+++ b/core/lib/src/outcome.rs
@@ -484,6 +484,93 @@ impl<S, E, F> Outcome<S, E, F> {
         }
     }
 
+    /// Maps an `Outcome<S, E, F>` to an `Outcome<T, E, F>` by applying the
+    /// function `f` to the value of type `S` in `self` if `self` is an
+    /// `Outcome::Success`.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use rocket::outcome::Outcome;
+    /// # use rocket::outcome::Outcome::*;
+    /// #
+    /// let x: Outcome<i32, &str, bool> = Success(10);
+    ///
+    /// let mapped = x.and_then(|v| match v {
+    ///    10 => Success("10"),
+    ///    1 => Forward(false),
+    ///    _ => Failure("30")
+    /// });
+    ///
+    /// assert_eq!(mapped, Success("10"));
+    /// ```
+    #[inline]
+    pub fn and_then<T, M: FnOnce(S) -> Outcome<T, E, F>>(self, f: M) -> Outcome<T, E, F> {
+        match self {
+            Success(val) => f(val),
+            Failure(val) => Failure(val),
+            Forward(val) => Forward(val),
+        }
+    }
+
+    /// Maps an `Outcome<S, E, F>` to an `Outcome<S, T, F>` by applying the
+    /// function `f` to the value of type `E` in `self` if `self` is an
+    /// `Outcome::Failure`.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use rocket::outcome::Outcome;
+    /// # use rocket::outcome::Outcome::*;
+    /// #
+    /// let x: Outcome<i32, &str, bool> = Failure("hi");
+    ///
+    /// let mapped = x.failure_then(|v| match v {
+    ///    "hi" => Failure(10),
+    ///    "test" => Forward(false),
+    ///    _ => Success(10)
+    /// });
+    ///
+    /// assert_eq!(mapped, Failure(10));
+    /// ```
+    #[inline]
+    pub fn failure_then<T, M: FnOnce(E) -> Outcome<S, T, F>>(self, f: M) -> Outcome<S, T, F> {
+        match self {
+            Success(val) => Success(val),
+            Failure(val) => f(val),
+            Forward(val) => Forward(val),
+        }
+    }
+
+    /// Maps an `Outcome<S, E, F>` to an `Outcome<S, E, T>` by applying the
+    /// function `f` to the value of type `F` in `self` if `self` is an
+    /// `Outcome::Forward`.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use rocket::outcome::Outcome;
+    /// # use rocket::outcome::Outcome::*;
+    /// #
+    /// let x: Outcome<i32, &str, Option<bool>> = Forward(Some(false));
+    ///
+    /// let mapped = x.forward_then(|v| match v {
+    ///    Some(true) => Success(10),
+    ///    Some(false) => Forward(20),
+    ///    None => Failure("10")
+    /// });
+    ///
+    /// assert_eq!(mapped, Forward(20));
+    /// ```
+    #[inline]
+    pub fn forward_then<T, M: FnOnce(F) -> Outcome<S, E, T>>(self, f: M) -> Outcome<S, E, T> {
+        match self {
+            Success(val) => Success(val),
+            Failure(val) => Failure(val),
+            Forward(val) => f(val),
+        }
+    }
+
     /// Converts from `Outcome<S, E, F>` to `Outcome<&mut S, &mut E, &mut F>`.
     ///
     /// ```rust


SergioBenitez:
The build has failed:

```rust
failures:
---- src/outcome.rs - outcome::Outcome<S, E, F>::and_then_failure (line 521) stdout ----
error[E0308]: mismatched types
  --> src/outcome.rs:533:15
   |
15 |       Success("10")
   |               ^^^^ expected i32, found reference
   |
   = note: expected type `i32`
              found type `&'static str`
error[E0308]: mismatched types
  --> src/outcome.rs:536:1
   |
18 | assert_eq!(mapped, Failure(10));
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected &str, found integral variable
   |
   = note: expected type `rocket::Outcome<i32, &str, usize>`
              found type `rocket::Outcome<_, {integer}, _>`
   = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)
```

mettke:
@SergioBenitez fixed

SergioBenitez:
The new functionality here looks useful and well motivated to me. I'm for this change.

@jebrosen Can you review this and ping me when it's ready to merge?

core/lib/src/outcome.rs
```diff
@@ -484,6 +484,96 @@ impl<S, E, F> Outcome<S, E, F> {
         }
     }
 
+    /// Maps an `Outcome<S, E, F>` to an `Outcome<T, E, F>` by applying the
+    /// function `f` to the value of type `S` in `self` if `self` is an
+    /// `Outcome::Success`.
```
jebrosen: I think the shorter wording of [Result::and_then](https://doc.rust-lang.org/nightly/std/result/enum.Result.html#method.and_then) is clearer. For example: "Calls `f` if `self` is a [Success], otherwise returns the [Failure]/[Forward] value of `self`"

The same goes for the other two methods (compare to `Result::or_else`).

mettke:
> `and_then_forward` doesn't actually seem that useful to me.

I kindly have to disagree. Imagine the following situation. A route should be accessible with either a valid token or a valid API key. We can require one of them being available by creating a `struct` `Authenticated` implementing `FromRequest`. In that function, we check for the `Token` and if it is not available we check for an `APIKey`. 

```rust
pub struct Authenticated {
    pub token: Option<Token>,
    pub api_key: Option<APIKey>,
}

impl<'a, 'r> FromRequest<'a, 'r> for Authenticated {
    type Error = ();

    fn from_request(request: &'a Request<'r>) -> Outcome<Self, Self::Error> {
        Token::from_request(request)
            .map(|token| Authenticated {
                token: Some(token),
                api_key: None,
            })
            .and_then_forward(|_| {
                APIKey::from_request(request).map(|api_key| Authenticated {
                    token: None,
                    api_key: Some(api_key),
                })
            })
    }

    fn from_request_current_implementation(request: &'a Request<'r>) -> Outcome<Self, Self::Error> {
        let token = Token::from_request(request)
            .map(|token| Authenticated {
                token: Some(token),
                api_key: None,
            });

        match token {
            outcome::Outcome::Success(value) => outcome::Outcome::Success(value),
            outcome::Outcome::Forward(_) => {
                APIKey::from_request(request).map(|api_key| Authenticated {
                    token: None,
                    api_key: Some(api_key),
                })
            }
            outcome::Outcome::Failure(value) => outcome::Outcome::Failure(value),
        }
    }
}

pub struct Token;

impl<'a, 'r> FromRequest<'a, 'r> for Token {
    type Error = ();

    fn from_request(request: &'a Request<'r>) -> Outcome<Self, Self::Error> {
        unimplemented!()
    }
}

pub struct APIKey;

impl<'a, 'r> FromRequest<'a, 'r> for APIKey {
    type Error = ();

    fn from_request(request: &'a Request<'r>) -> Outcome<Self, Self::Error> {
        unimplemented!()
    }
}
```

`and_then_forward` allows us to react to a missing `Token` without having to check for an `APIKey` when a token exists. 

Coming up with a good example for `and_then_error` is however quite tough. I'm not sure whether there is a situation where I want to transform an `Error` to either a `Forward` or a `Success`. Except maybe if I disagree with the author whether an error should be fatal or a simple forward.

jebrosen:
After seeing the example I agree that `and_then_forward` is useful. I like `and_then` for Success because it parallels Option and Result's `and_then`. I'd be fine with using something like `forward_and_then`, `failure_and_then`, `forward_then`, `failure_then`.

core/lib/src/outcome.rs
mettke: Done

jebrosen:
See the build failures -- the inline code examples weren't updated to match the new method names.

I see now that you were following the documentation style of the existing Outcome methods. I still prefer the shorter style myself, but consistency is more important here; they should be changed back so they match. Sorry about the back and forth.

mettke:
@jebrosen No problem. Sry about the build failure - hope I got it all right this time. Sometimes building rocket is quite difficult due to the nature of the nightly compiler...

core/lib/src/outcome.rs
```diff
@@ -484,6 +484,102 @@ impl<S, E, F> Outcome<S, E, F> {
...
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use rocket::outcome::Outcome;
+    /// # use rocket::outcome::Outcome::*;
+    /// #
+    /// let x: Outcome<i32, &str, usize> = Success(10);
```
SergioBenitez: Can we make this an `Outcome<i32, &str, bool>` to use three very different, distinct types? I think that will make the example clearer.

core/lib/src/outcome.rs
```diff
@@ -484,6 +484,102 @@ impl<S, E, F> Outcome<S, E, F> {
...
+    ///
+    /// let mapped = x.and_then(|v| {
```
SergioBenitez: Use a match to make this example (and the other two) clearer:

```rust
let mapped = x.and_then(|v| match v {
    10 => Success("10"),
    1 => Forward(20),
    _ => Failure(false)
});
```

core/lib/src/outcome.rs
```diff
@@ -484,6 +484,102 @@ impl<S, E, F> Outcome<S, E, F> {
...
+    ///    if v == 10 {
+    ///       Success("10")
+    ///    } else if v == 1 {
+    ///       Forward(20)
+    ///    } else {
+    ///       Failure("30")
+    ///    }
+    /// });
+    /// assert_eq!(mapped, Success("10"));
```
SergioBenitez: Add an empty line above here.

core/lib/src/outcome.rs
mettke: I guess you mean 
```rust
let mapped = x.and_then(|v| match v {
   10 => Success("10"),
   1 => Forward(false),
   _ => Failure("30")
});
```
The third generic is the type of `Forward` which is now a bool ;)

jebrosen:
Merged in cd776d5. Thanks, @mettke!

Issue #687, opened 2018-07-05
Update published ring dependency to 0.12


Perseus101:
Rocket currently uses the slightly outdated 0.11 version of Ring, which is conflict with another crate I am trying to use which requires 0.12.

It appears Rocket's dependency for version 0.11 comes from using version 0.9 of cookies. Upgrading cookies to 0.10 should solve the problem.

#615 also references the outdated version of ring as a possible problem causing compilation issues.

Perseus101:
It looks like the most up to date version of rocket supports ring 0.10, but isn't published. What is the estimate for when a version with this updated dependency be published?

SergioBenitez:
See #627, #516, #492, and https://github.com/SergioBenitez/Rocket/issues/492#issuecomment-351699847.

The hope is to release Rocket v0.4 by mid-August.

jens1o:
@SergioBenitez Could you give an update on when you want to release `0.4`?

SergioBenitez:
@jens1o As soon as humanly possible. A lot of things have come up since I last gave an estimate that have consumed my time in the last month, many of which are directly relevant to Rocket. I'm super excited to share these with the community. Until then, all I can realistically say is: as soon as possible.

PR #722, opened 2018-07-29
update `cookie` to v0.11 to pull in latest `ring`

rushmorem:


jebrosen:
This cannot be done in `0.3`, and the commit that `master` points to is identical to cookie 0.11 except for the stated version number.

See the links in https://github.com/SergioBenitez/Rocket/issues/687#issuecomment-402843320.

rushmorem:
> This cannot be done in 0.3

This pull request does not make any assumption about Rocket's version. `Cargo.toml` is already set to `0.4.0-dev`.

> the commit that master points to is identical to cookie 0.11 except for the stated version number

Since they are equivalent, why not just use the version number, now that it's up on `crates.io`?

jebrosen:
> This pull request does not make any assumption about Rocket's version. Cargo.toml is already set to 0.4.0-dev.

Sorry, thinking you actually wanted this for 0.3 was a hasty assumption on my part given the number of past requests to do it.

> Since they are equivalent, why not just use the version number, now that it's up on crates.io?

I see no problem with doing this now, assuming cookie 0.11 has all of the features Rocket 0.4 will need.

SergioBenitez:
We're already depending on a `cookie` that depends on `0.13` (since e25a82b14b). This change would have no affect on which version of `ring` is used.

PR #724, opened 2018-08-01
Add documentation and examples for request-local state.

### LINKED ISSUE #654: Request-Local State Cache

jebrosen:
Closes #654.

### DIFF
--- a/core/lib/src/fairing/mod.rs
+++ b/core/lib/src/fairing/mod.rs
@@ -260,6 +260,67 @@ pub use self::info_kind::{Info, Kind};
 ///     }
 /// }
 /// ```
+///
+/// ## Request-Local Cache
+///
+/// Fairings can use the *request-local cache* to persist data between the
+/// request and the response, or to pass data to a request guard.
+///
+/// ```
+/// # use std::time::{Duration, SystemTime};
+/// # use rocket::Outcome;
+/// # use rocket::{Request, Data, Response};
+/// # use rocket::fairing::{Fairing, Info, Kind};
+/// # use rocket::http::Status;
+/// # use rocket::request::{self, FromRequest};
+/// #
+/// struct RequestTimer;
+/// #[derive(Copy, Clone)]
+/// struct StartTime(SystemTime);
+///
+/// impl Fairing for RequestTimer {
+///     fn info(&self) -> Info {
+///         Info {
+///             name: "Request Timer",
+///             kind: Kind::Request | Kind::Response
+///         }
+///     }
+///
+///     /// Stores the start time of the request
+///     fn on_request(&self, request: &mut Request, _: &Data) {
+///         request.local_cache(|| {
+///             // Note the use of a newtype to disambiguate `StartTime` from
+///             // other SystemTime values that might be stored in request-local
+///             // cache for other purposes
+///             Some(StartTime(SystemTime::now()))
+///         });
+///     }
+///
+///     // Adds a header to the response indicating how long the server took to
+///     // process the request
+///     fn on_response(&self, request: &Request, response: &mut Response) {
+///         let start_time: &Option<StartTime> = request.local_cache(|| None);
+///         if let Some(duration) = start_time.and_then(|s| s.0.elapsed().ok()) {
+///             response.set_raw_header("X-Response-Time", format!("{} ms",
+///                 duration.as_secs() * 1000 + duration.subsec_millis() as u64));
+///         }
+///     }
+/// }
+///
+/// // Allows a route to access the time the request was initiated
+/// impl<'a, 'r> FromRequest<'a, 'r> for StartTime {
+///     type Error = ();
+///
+///     fn from_request(request: &'a Request<'r>) -> request::Outcome<StartTime, ()> {
+///         let start_time: &Option<StartTime> = request.local_cache(|| None);
+///         match *start_time {
+///             Some(time) => Outcome::Success(time),
+///             None => Outcome::Failure((Status::InternalServerError, ())),
+///         }
+///     }
+/// }
+/// ```
+
 pub trait Fairing: Send + Sync + 'static {
     /// Returns an [`Info`](/rocket/fairing/struct.Info.html) structure
     /// containing the `name` and [`Kind`](/rocket/fairing/struct.Kind.html) of
--- a/core/lib/src/fairing/mod.rs
+++ b/core/lib/src/fairing/mod.rs
@@ -276,7 +276,7 @@ pub use self::info_kind::{Info, Kind};
 /// #
 /// struct RequestTimer;
 /// #[derive(Copy, Clone)]
-/// struct StartTime(SystemTime);
+/// struct StartTime(pub Option<SystemTime>);
 ///
 /// impl Fairing for RequestTimer {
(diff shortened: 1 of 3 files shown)


core/lib/src/request/from_request.rs
```diff
@@ -207,6 +207,118 @@ impl<S, E> IntoOutcome<S, (Status, E), ()> for Result<S, E> {
...
+///     fn from_request(request: &'a Request<'r>) -> request::Outcome<&'a User, ()> {
+///         request.local_cache(|| {
+///             // This closure will run only once per request, and future
+///             // invocations will reuse the result of the first calculation
+///             let db = request.guard::<Database>().succeeded()?;
+///             request.cookies()
+///                 .get_private("user_id")
+///                 .and_then(|cookie| cookie.value().parse().ok())
+///                 .and_then(|id| db.get_user(id).ok())
+///         }).as_ref().or_forward(())
```
jebrosen: The return type and last line of this function aren't very obvious but are necessary given how `local_cache` works. It also means that `impl<'a, 'r> FromRequest<'a, 'r> for Admin` must be changed to something with a lifetime, such as `for Admin<'a>`.

site/guide/state.md
```diff
@@ -85,6 +85,19 @@ fn from_request(req: &'a Request<'r>) -> request::Outcome<T, ()> {
 
 [`Request::guard()`]: https://api.rocket.rs/rocket/struct.Request.html#method.guard
 
+### Request-Local State
```
jebrosen: What other parts of the guide should Request-Local State / Request-Local Cache be called out in?

core/lib/src/request/from_request.rs
SergioBenitez: Yeah, that's pretty interesting.

I would put the comment outside of the call (for style/readability purposes). And then, I _might_ do:

```rust
let user_result = request.local_cache(|| {
    // everything
});

user_result.as_ref().or_forward(())
```

I'd also add a postlude to this example with information on what's going on with the references here, which is pretty interesting. Also, it would help to complete the example by including the `Admin` guard as well.

site/guide/state.md
```diff
@@ -85,6 +85,19 @@ fn from_request(req: &'a Request<'r>) -> request::Outcome<T, ()> {
...
+
+In addition to application-wide state, each `Request` has its own state container
```
SergioBenitez: I would start this by saying: "While managed state is _global_ and available application-wide, request-local state is _local_ to a given request, carried along with the request, and dropped once the request is completed." And then continue by describing why these semantics are useful.

site/guide/state.md
```diff
@@ -85,6 +85,19 @@ fn from_request(req: &'a Request<'r>) -> request::Outcome<T, ()> {
...
+that can be used by request guards and fairings. This container can be used to
+cache expensive computations and lookups, such as to a database or external
+server, that might be used by multiple request guards during routing. Fairings
```
SergioBenitez: Not just guards, but anything with access to a `Request`. Think of a way to word this more generally.

site/guide/state.md
```diff
@@ -85,6 +85,19 @@ fn from_request(req: &'a Request<'r>) -> request::Outcome<T, ()> {
...
+can provide request-specific data for consumption by request guards in the same
+manner. See the documentation for the [`FromRequest`] and [`Fairing`] traits for
```
SergioBenitez: I would be more specific about what kind of things you would want to use request-local state for and what problem they're solving. We want users to read this and have a pretty good idea about when to reach for request-local state. I don't think this accomplishes this quite yet.

Don't be afraid to include some example code here too to drive the code home.

SergioBenitez:
These two examples are really nice, @jebrosen!

core/lib/src/fairing/mod.rs
```diff
@@ -260,6 +260,67 @@ pub use self::info_kind::{Info, Kind};
...
+///             // other SystemTime values that might be stored in request-local
+///             // cache for other purposes
+///             Some(StartTime(SystemTime::now()))
+///         });
+///     }
+///
+///     // Adds a header to the response indicating how long the server took to
+///     // process the request
+///     fn on_response(&self, request: &Request, response: &mut Response) {
+///         let start_time: &Option<StartTime> = request.local_cache(|| None);
```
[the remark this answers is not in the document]
jebrosen: I ended up doing something else, since  `StartTime(Option<SystemTime>)` can't contain a `Duration`. I'm still happy with the fairing, but the request guard becomes a bit awkward when the Option is inside the wrapper.

site/guide/state.md
```diff
@@ -85,6 +85,50 @@ fn from_request(req: &'a Request<'r>) -> request::Outcome<T, ()> {
...
+/// A global counter for arbitrary request IDs
+static request_id_counter: AtomicUsize = AtomicUsize::new(0);
+/// A type that represents request IDs
+struct RequestId(pub usize);
+
+/// Returns the current request's RequestId, assigning one
+/// if the current request does not have one already.
+impl<'a, 'r> FromRequest<'a, 'r> for RequestId {
+    fn from_request(request: &'a Request<'r>) -> request::Outcome {
+        // The closure passed to local_cache will only be run once, the
```
SergioBenitez: Should say, "will be executed at most once per request".

SergioBenitez:
Awesome as always, @jebrosen! Merged in d1cfdbaa8e575c4f0b8a0509b06356cd0ffb1987.

PR #734, opened 2018-08-12
Fix benchmark compilation

messense:


core/lib/benches/format-routing.rs
```diff
@@ -13,7 +13,7 @@ fn post() -> &'static str { "post" }
 
 fn rocket() -> rocket::Rocket {
     let config = Config::new(Environment::Production).unwrap();
-    rocket::custom(config, false).mount("/", routes![get, post])
+    rocket::custom(config).mount("/", routes![get, post])
```
jebrosen: I believe this will enable logging, where it wasn't before. This was changed in 965c90af; something like`config.log = Off` may be appropriate here.

messense: Addressed.

core/lib/benches/format-routing.rs
```diff
@@ -12,8 +12,9 @@ fn get() -> &'static str { "get" }
 fn post() -> &'static str { "post" }
 
 fn rocket() -> rocket::Rocket {
-    let config = Config::new(Environment::Production).unwrap();
-    rocket::custom(config, false).mount("/", routes![get, post])
+    let mut config = Config::new(Environment::Production).unwrap();
```
SergioBenitez: I'd prefer:

```rust
let config = Config::build(Environment::Production).log_level(LoggingLevel::Off);
rocket::custom(config.unwrap()) (and so on..)
```

messense: Updated.

SergioBenitez:
Awesome. Good catch, @jebrosen! Merged in 589743579.

PR #736, opened 2018-08-15
Update codegen for 2018-08-14 nightly

messense:


SergioBenitez:
It looks like `ThinVec` is also being exported from `syntax` at `syntax::ThinVec`. We should grab it from there so we don't need to `extern crate` another crate.

SergioBenitez:
Merged in 135b7d29. Thanks so much!

PR #739, opened 2018-08-16
Minor documentation fixes

EloD10:
It should be good.


SergioBenitez:
I think I prefer `diesel_postgres_pool`, actually. `pg` isn't as clear, in my opinion.

ELD:
@SergioBenitez I was taking cues from the type names. In Diesel, it abbreviates the connection type as `PgConnection` rather than `PostgresConnection`. However, I'm not opposed to renaming the feature.

SergioBenitez:
Went with `diesel_postgres_pool`. Merged the docstring fix in this PR in 46da03c3c3c. Thanks!

PR #746, opened 2018-08-19
Embed the diesel database migrations in the `todo` example.

jebrosen:
With this change, `diesel_cli` will no longer be needed in order to prepare or run the example or tests, making the example easier to work with (and possibly reducing the time taken for CI builds?). It will still be necessary to install `diesel_cli` manually in order to add migrations or regenerate the schema.

### DIFF
--- a/examples/todo/src/main.rs
+++ b/examples/todo/src/main.rs
@@ -20,6 +20,9 @@ use diesel::SqliteConnection;
 
 use task::{Task, Todo};
 
+// This macro from `diesel_migrations` defines an `embedded_migrations` module
+// containing a function named `run`. This allows the example to be run and
+// tested without any outside setup of the database.
 embed_migrations!();
 
 #[database("sqlite_database")]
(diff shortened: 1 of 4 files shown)


examples/todo/src/main.rs
```diff
@@ -19,6 +22,8 @@ use diesel::SqliteConnection;
 
 use task::{Task, Todo};
 
+embed_migrations!();
```
SergioBenitez: What does this do, exactly? Perhaps a small comment above would help.

jebrosen:
Another thought: it might even be possible to eliminate the bootstrap script entirely by running the tests against an in-memory database, by setting the url in the configuration to ":memory:".

jebrosen:
@SergioBenitez With this set of changes, the bootstrap script now only deletes the database file. I planned to try making the bootstrap script entirely unnecessary -- maybe by using an in-memory database, database transactions, or a temporary file in test mode. Would you prefer to wait for those changes here, or for me to open a separate PR later?

jebrosen:
Merged in 30cf753.

PR #753, opened 2018-08-24
Remove uses of the `const_fn` feature.

jebrosen:
This feature is currently only used in the `http` crate to make `Status::new` a const fn. A `Status` instance can already be constructed in a const using `Status { code, reason }` syntax, so this doesn't remove any capabilities but is a breaking change for anyone using `Status::new` in a const.


SergioBenitez:
Merged in 1da506ea5d8cd842380982b475ccf74c63211917.

PR #761, opened 2018-08-31
Added support for Clone to rocket::http::uri::Uri

talg:


core/http/src/parse/indexed.rs
```diff
@@ -31,7 +31,7 @@ impl AsPtr for [u8] {
 }
 
 
-#[derive(PartialEq)]
+#[derive(PartialEq, Clone)]
```
SergioBenitez: This is incorrect as it adds a `Clone` bound to `T`, which is unnecessary. `Indexed` already implements `Clone`; no changes are needed to `indexed.rs`.

core/http/src/parse/indexed.rs
```diff
@@ -175,7 +175,7 @@ impl<'a, T: ?Sized + ToOwned + 'a> Indexed<'a, T>
 }
 
 impl<'a, T: ToOwned + ?Sized + 'a> Clone for Indexed<'a, T> {
-    fn clone(&self) -> Self {
+    default fn clone(&self) -> Self {
```
SergioBenitez: This is unnecessary.

talg:
Good catch, fixed in the latest commit.

SergioBenitez:
Merged in 8172f748b32c79f980020d7dabdff22ee131c46f. Thanks so much! :)

PR #762, opened 2018-09-02
Added JsonResult for lenient routing

dnorman:
When receiving a webhook or other types of API requests, often times the contract is unclear, or the deviation from the contract can be hard to diagnose without the raw JSON body which was actually sent.

I considered various ways of logging the json body in this situation, but this one seemed to be the simplest, most intuitive, and involved no copying of the underlying json body string.

### DIFF
--- a/contrib/lib/src/json.rs
+++ b/contrib/lib/src/json.rs
@@ -99,6 +99,52 @@ impl<T> Json<T> {
     }
 }
 
+/// ## Lenient Routing in the case of a JSON parsing failure
+///
+/// If you're receiving JSON data but you want your handler to fire regardless of
+/// whether the JSON parsing is successful or not, you can specify JsonResult
+/// you can
+/// ```rust,ignore
+/// #[post("/users/", data = "<user>")]
+/// fn new_user(r: JsonResult<User>) {
+///     match r {
+///         JsonResult::Ok(user) => {..}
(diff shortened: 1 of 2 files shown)


SergioBenitez:
Why don't we simply change the existing `FromData` impl for `Json` so that it also returns the `String` in the error case? Something like:

```rust
#[derive(Clone, PartialEq, etc..)]
pub enum JsonError {
    /// An I/O error occurred while reading the incoming request data.
    Io(io::Error),
    /// The client's data was received successfully but failed to parse as valid JSON. The
    /// `String` value in `.0` is the raw data received from the user, while the `SerdeError`
    /// in `.1` is the deserialization error.
    InvalidJson(String, SerdeError),
}

impl<T: DeserializeOwned> FromData for Json<T> {
    type Error = JsonError;
    
    ...
}
```

Then you can simply use `Result` in a handler as today, but with more information:

```rust
#[post("/", data = "<json>")]
fn handler(json: Result<Json<MyValue>, JsonError>) {
    match json {
        Ok(value) => println!("The value: {:?}", value),
        Err(JsonError::Io(e)) => println!("I/O Error: {:?}", e),
        Err(JsonError::InvalidJson(raw, e)) => println!("{:?} is not valid JSON: {:?}", raw, e),
    }
}
```

PR #774, opened 2018-09-17
Guard template change watcher inside `Mutex`

jebrosen:
...so that `ContextManager` will be Sync on all platforms.

This is a apparently a problem on at least FreeBSD, where the `notify` crate's `RecommendedWatcher` equals `PollWatcher`. That type happens to be `!Sync`, so rocket_contrib fails to compile later when the `Sync` bound is not satisfied.


SergioBenitez:
Merged in 95cbc241db6a632b060c58bd505e880db106e967.

PR #780, opened 2018-10-01
Fix the example code in README.md by applying the following compiler …

NAThompson:
…suggestion:

error[E0658]: procedural macros cannot be expanded to expressions (see issue #38356)
  --> src/main.rs:12:38
   |
12 |     rocket::ignite().mount("/hello", routes![hello]).launch();
   |                                      ^^^^^^^^^^^^^^
   |
   = help: add #![feature(proc_macro_non_items)] to the crate attributes to enable

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Rocket is web framework for Rust (nightly) with a focus on ease-of-use,
 expressibility, and speed. Here's an example of a complete Rocket application:
 
 ```rust
-#![feature(plugin, decl_macro)]
+#![feature(plugin, decl_macro, proc_macro_non_items)]
 #![plugin(rocket_codegen)]
 
 #[macro_use] extern crate rocket;


NAThompson:
This appears to only show up on the very latest `rustc`: I reproduced with

```bash
$ rustc --version
rustc 1.31.0-nightly (fc403ad98 2018-09-30)
```

and a few days ago, this was not a compile error.

SergioBenitez:
It's not rustc that's changed, it's Rocket; we're now using procedural macros for more things than before, requiring different feature gates, and we haven't updated the README or guide docs appropriately. There's a final major change I'm working on now that will fix the README as well, but I'm waiting on https://github.com/rust-lang/rust/pull/52121 to land, which will change the feature gates yet again.

NAThompson:
Ok, cools, looks like that PR will render this one irrelevant, so this is a non-issue.

PR #781, opened 2018-10-02
Compression Fairing

Martin1887:
Clean Compression Fairing Pull Request with only a commit with all changes done about the compression fairing in a single commit.


contrib/lib/src/compression.rs
```diff
@@ -0,0 +1,130 @@
...
+            .get("Accept-Encoding")
+            .flat_map(|e| e.split(","))
+            .any(|e| encodings.contains(&e.trim()))
+    }
+
+    fn already_encoded(response: &Response) -> bool {
+        response
+            .headers()
+            .get("Content-Encoding")
+            .any(|e| e != "identity" && e != "chunked")
```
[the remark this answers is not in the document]
Martin1887: I think that you mean that `chunked` is a `Transfer-Encoding` instead `identity`. I have applied my changes guessing it and yes, it was my failure (I saw it in `Encoding` and I didn't think that `Transfer-Encoding` are included in the same `enum`).

jebrosen:
Sorry for the late response time this go around. I did make formatting, style, and documentation fixes (no significant code changes) but have not yet gotten to the "real-world" tests I planned to do. It's on my list for the weekend, though!

Martin1887:
@jebrosen, are your tests running successfully? Any comments or news about this feature?

Thanks for your work!

jebrosen:
Whoops! I have been waiting for a reply to a comment I forgot to post:

Functionality wise, this is good. I have cleaned up the documentation and code style to closely match other things in `contrib`, and can merge whenever.

I only have one real issue, which is that "all responses are compressed except images or with the identity opt-out" compresses large responses and already-compressed data such as zip files. I would not feel bad releasing this with the disclaimer that it performs badly with those types of responses, but I would also like to add those restrictions in the future without feeling too bad about "breaking" anything. I also don't want to go to the other extreme of "compress only text/* responses."

@SergioBenitez, how would you feel about providing this functionality as-is with the possibility of restricting it or making it more configurable in the future?

SergioBenitez:
This has come a long way. Nice job, @jebrosen and @Martin1887! There are quite a few style issues, but I figure you've already fixed those locally, @jebrosen.

I think this needs two things:

**1) Configuration.** At minimum, we should allow the user to say which media types _not_ to compress, and we should provide a default list that works for most folks.

```toml
[global.compress]
exclude = ["application/gzip", "application/zip", "image/*"]
```

I _think_ opt-out is the right approach here. The alternative is to ask the user to white-list instead, but that feels significantly more cumbersome and prone to forgetting something.

**2) A Responder.** We should allow users to compress individual responses without using a fairing. Presumable the responder wouldn't be beholden to the `exclude` list, which means you can selectively override the list as well.

jebrosen:
This is my "version" of the whole PR: https://paste.rs/BDF.diff - the bulk of the difference is renaming and reorganizing code and documentation to match the layout of other features in `rocket_contrib`, but notice a few other minor changes I made. It might not be fully up-to-date, but I'll do that comparison again when we're closer to merging.

I think both configuration and a Responder (as described) are a good idea. Don't hesitate to ping me here or on IRC if you have any questions.

Martin1887:
I'm working on the exceptions read from the Rocket configuration and I have almost finished it.

I have added a field `exceptions` of type `Vec<String>` and a `on_attach` function. The problem that I have found is that the `Compression` struct cannot be modified inside the `on_attach` function because it is not mutable at this point.

I have seen that the `templates` fairing uses `ContextManager` to change the state of the Rocket instance inside the `on_attach` function, but this seems too cumbersome to me to only change one parameter. Is there any other alternatives to change the `exceptions` field of the `Compression` struct?

Another question that has arisen is about the responder: a `GzipResponder` and another `BrotliResponder` should be created or it is better having only one `CompressionResponder`? Maybe both options?


Thanks.

Martin1887:
Regarding the `ContentTypes` I have noticed that WASM should also be excluded for compression, since it is already compressed. Do you think so? Any other `ContentTypes` to be excluded?

Martin1887:
I have implemented the `compress.exclude` configuration parameter in the `Compression` trait. The module has also been reorganized inside a folder.

The `Content-Type` comparison is slightly ad-hoc, I would like receiving your advice to improve it:

```rust
fn skip_encoding(
        &self,
        content_type: &Option<rocket::http::ContentType>,
        content_type_top: &Option<&UncasedStr>,
        cm: &rocket::State<ContextManager>,
    ) -> bool {
        let context = cm.context();
        let exceptions = context.exceptions();
        exceptions
            .iter()
            .filter(|c| match content_type {
                Some(ref orig_content_type) => match MediaType::parse_flexible(c) {
                    Some(exc_media_type) => {
                        if exc_media_type.sub() == "*" {
                            Some(exc_media_type.top()) == *content_type_top
                        } else {
                            exc_media_type == *orig_content_type.media_type()
                        }
                    }
                    None => {
                        if c.contains("/") {
                            let split: Vec<&str> = c.split("/").collect();

                            let exc_media_type =
                                MediaType::new(String::from(split[0]), String::from(split[1]));

                            if split[1] == "*" {
                                Some(exc_media_type.top()) == *content_type_top
                            } else {
                                exc_media_type == *orig_content_type.media_type()
                            }
                        } else {
                            false
                        }
                    }
                },
                None => false,
            })
            .count()
            > 0
    }
```

Meanwhile, I will begin to implement the `responder`.

Martin1887:
I have just implemented the `Compressed` `Responder` and its tests.

The code also has been reorganized inside the `compression` module to share the common functions.

Only documentation and improvements should be missing now.

I expect your comments and reviews.


Regards.

Martin1887:
I have completed the suggested changes and added docstrings.

I have also unified the `exclusions` name instead of a mix of `exclusions` and `exceptions`.

I think that the below function and maybe documentation and style are the only things that could need improvements.

> I have implemented the `compress.exclude` configuration parameter in the `Compression` trait. The module has also been reorganized inside a folder.
> 
> The `Content-Type` comparison is slightly ad-hoc, I would like receiving your advice to improve it:
> 
> ```rust
> fn skip_encoding(
>         &self,
>         content_type: &Option<rocket::http::ContentType>,
>         content_type_top: &Option<&UncasedStr>,
>         cm: &rocket::State<ContextManager>,
>     ) -> bool {
>         let context = cm.context();
>         let exceptions = context.exceptions();
>         exceptions
>             .iter()
>             .filter(|c| match content_type {
>                 Some(ref orig_content_type) => match MediaType::parse_flexible(c) {
>                     Some(exc_media_type) => {
>                         if exc_media_type.sub() == "*" {
>                             Some(exc_media_type.top()) == *content_type_top
>                         } else {
>                             exc_media_type == *orig_content_type.media_type()
>                         }
>                     }
>                     None => {
>                         if c.contains("/") {
>                             let split: Vec<&str> = c.split("/").collect();
> 
>                             let exc_media_type =
>                                 MediaType::new(String::from(split[0]), String::from(split[1]));
> 
>                             if split[1] == "*" {
>                                 Some(exc_media_type.top()) == *content_type_top
>                             } else {
>                                 exc_media_type == *orig_content_type.media_type()
>                             }
>                         } else {
>                             false
>                         }
>                     }
>                 },
>                 None => false,
>             })
>             .count()
>             > 0
>     }
> ```
> 
> Meanwhile, I will begin to implement the `responder`.

contrib/lib/src/compression/fairing.rs
```diff
@@ -0,0 +1,150 @@
...
+///
+/// In the brotli compression mode (using the
+/// [rust-brotli](https://github.com/dropbox/rust-brotli) crate), quality is set
+/// to 2 in order to achieve fast compression with a compression ratio similar
+/// to gzip. When appropriate, brotli's text and font compression modes are
+/// used.
+///
+/// In the gzip compression mode (using the
+/// [flate2](https://github.com/alexcrichton/flate2-rs) crate), quality is set
+/// The Compression type implements brotli and gzip compression for responses in
```
jebrosen: This single extraneous line snuck in from somewhere, maybe during a rebase/merge.

contrib/lib/src/compression/fairing.rs
```diff
@@ -0,0 +1,150 @@
...
+/// to the default (9) in order to have good compression ratio.
+///
+/// This fairing does not compress responses that already have a
+/// `Content-Encoding` header.
+///
+/// This fairing ignores the responses with a `Content-Type` matching any of
+/// the following default types:
+///
+/// - application/gzip
+/// - application/brotli
```
jebrosen: I can't find any source for `application/brotli` being a thing.

contrib/lib/src/compression/fairing.rs
```diff
@@ -0,0 +1,150 @@
...
+/// - application/zip
+/// - image/*
+/// - application/wasm
+/// - application/binary
```
jebrosen: Similarly, I don't think `application/binary` is a thing. `application/octet-stream` is, but I'm not sure if it's worth excluding specifically either.

contrib/lib/src/compression/fairing.rs
```diff
@@ -0,0 +1,150 @@
...
+/// This fairing does not compress responses that already have a
+/// `Content-Encoding` header.
+///
+/// This fairing ignores the responses with a `Content-Type` matching any of
+/// the following default types:
+///
+/// - application/gzip
+/// - application/brotli
+/// - application/zip
+/// - image/*
```
jebrosen: Let's add `video/*` as well.

contrib/lib/src/compression/fairing.rs
```diff
@@ -0,0 +1,150 @@
...
+            kind: Kind::Attach | Kind::Response,
+        }
+    }
+
+    fn on_attach(&self, rocket: Rocket) -> Result<Rocket, Rocket> {
+        let mut ctxt = Context::new();
+        match rocket.config().get_slice("compress.exclude") {
+            Ok(excls) => {
+                let mut error = false;
+                let mut exclusions_vec = Vec::with_capacity(excls.len());
```
jebrosen: This can probably be written a bit more succinctly, along the lines of:

```rust
let exclusions: Result<Vec<String>, ConfigError> = excls.map(|e| match e {
    Value::String(s) => Ok(s.clone()),
    _ => Err(ConfigError::BadType(...)),
}).collect();

// Handle Ok, Err variants for exclusions...
```

It might be relatively easy to combine the error handling as well:
```rust
let exclusions = rocket.config().get_slice(...).and_then(|excls| ...);
let ctxt = match exclusions {
    Ok(e) => Context::with_exclusions(e),
    Err(ConfigError::Missing(_)) => { /* ignore missing */ },
    Err(e) => { e.pretty_print(); warn!(...) },
};
```

These should only be taken as rough ideas, I know they don't work directly as-is and it might be trickier than this.

contrib/lib/src/compression/fairing.rs
```diff
@@ -0,0 +1,150 @@
...
+    fn info(&self) -> Info {
+        Info {
+            name: "Response compression",
+            kind: Kind::Attach | Kind::Response,
+        }
+    }
+
+    fn on_attach(&self, rocket: Rocket) -> Result<Rocket, Rocket> {
+        let mut ctxt = Context::new();
+        match rocket.config().get_slice("compress.exclude") {
```
jebrosen: Does compress.exclude work here? I would expect that you would have to do something like `get_table("compress").and_then(|t| t.get_slice("exclude"))`. I don't see any tests for the configuration syntax, which could probably go in one of the existing top-level examples (maybe `static_files`?)

contrib/lib/src/compression/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+        body: B,
+        encoding: Encoding,
+    ) {
+        response.set_header(ContentEncoding(vec![encoding]));
+        response.set_streamed_body(body);
+    }
+
+    fn skip_encoding(
+        content_type: &Option<rocket::http::ContentType>,
+        content_type_top: &Option<&UncasedStr>,
```
jebrosen: This shouldn't take `content_type_top` separately, since it can be derived from `content_type`.

contrib/lib/src/compression/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+        context: &rocket::State<Context>,
+    ) -> bool {
+        let exceptions = &context.exclusions;
+        exceptions
+            .iter()
+            .filter(|c| match content_type {
+                Some(ref orig_content_type) => match MediaType::parse_flexible(c) {
+                    Some(exc_media_type) => {
```
jebrosen: If there was a `media_type.loose_match(other_media_type)` I would suggest that instead, but this looks okay.

contrib/lib/src/compression/fairing.rs
```diff
@@ -0,0 +1,150 @@
...
+            Ok(excls) => {
+                let mut error = false;
+                let mut exclusions_vec = Vec::with_capacity(excls.len());
+                for e in excls {
+                    match e {
+                        Value::String(s) => exclusions_vec.push(s.clone()),
```
jebrosen: Let's do the `MediaType::parse_flexible` here so it happens once at startup instead of every request. Exclusions would now be a `Vec<MediaType>`.

contrib/lib/src/compression/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+                        if exc_media_type.sub() == "*" {
+                            Some(exc_media_type.top()) == *content_type_top
+                        } else {
+                            exc_media_type == *orig_content_type.media_type()
+                        }
+                    }
+                    None => {
```
jebrosen: This whole `None` branch can go away if media types are parsed at startup. I think it's reasonable to reject exclusions that don't parse as a media type.

contrib/lib/src/compression/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+                                exc_media_type == *orig_content_type.media_type()
+                            }
+                        } else {
+                            false
+                        }
+                    }
+                },
+                None => false,
+            })
+            .count()
```
jebrosen: `.filter(|x| /*condition*/).count() > 0` can be reduced to `.any(|x| /*condition*/)`

contrib/lib/src/lib.rs
```diff
@@ -43,12 +42,31 @@
...
+extern crate rocket_contrib_codegen;
+#[cfg(feature = "databases")]
+#[doc(hidden)]
+pub use rocket_contrib_codegen::*;
+
+#[cfg(any(feature = "brotli_compression", feature = "gzip_compression"))]
+pub mod compression;
+
+#[cfg(feature = "brotli_compression")]
+extern crate brotli;
```
jebrosen: This `extern crate` really should be in `compression.rs` for consistency with the rest of `contrib`. I'll try to get it to work again before this is merged.

contrib/lib/tests/compressed.rs
```diff
@@ -0,0 +1,254 @@
...
+    const HELLO: &str = r"This is a message to hello with more than 100 bytes \
+        in order to have to read more than one buffer when gzipping. こんにちは!";
+
+    #[get("/")]
+    pub fn index() -> Compressed<String> {
+        Compressed::new(String::from(HELLO))
+    }
+
+    #[get("/font")]
+    pub fn font() -> Compressed<Response<'static>> {
```
jebrosen: `Compressed<Content<&'static str>>` and `Compressed::new(Content(ContentType::WOFF, HELLO))` is much shorter and a better example to show anyone reading the code. Similar for `image`.

contrib/lib/src/compression/fairing.rs
Martin1887: I have only tested it putting "compress.exclude" in tests, so I have no idea. @SergioBenitez, how do you think that it should be done?

contrib/lib/src/compression/mod.rs
Martin1887: I know it, but I put it as parameter because the function that call this function has to use `content_type_top` anyway some lines below, and the `content_type_top` extraction has to be done twice if it is not received as parameter. Should I remove the parameter anyway?

contrib/lib/src/compression/fairing.rs
jebrosen: In TOML, these are all equivalent (unless I made a mistake somewhere):
```toml
[global]
compress.exclude = ["text/html"]

[global]
compress = { exclude = ["text/html"] }

[global.compress]
exclude = ["text/html"]
```

For `get_slice("compress.exclude")` to work, this syntax is required:
```toml
[global]
"compress.exclude" = ["text/html"]
```
which I think is surprising to users.

contrib/lib/src/compression/mod.rs
jebrosen: You can leave it for now and I'll look again after the other changes.

contrib/lib/src/compression/fairing.rs
Martin1887: I think that `application/octet-stream` shall be excluded, binary data cannot be compressed.

contrib/lib/src/compression/fairing.rs
Martin1887: I have not been able to do this, mutability error in `excls.map` inside the `match` branch.

Martin1887:
I have pushed the last changes, sorry for the delay, these have been busy weeks.

I have marked as resolved the implemented changes and I have commented still open reviews.

Martin1887:
@jebrosen, @SergioBenitez, are you reviewing and testing the last changes?

For any question or proposal don't hesitate on doing it.

Martin1887:
Any news in this pull request?

jebrosen:
I'd been waiting for @SergioBenitez to come back so I could run a few relatively minor implementation/API changes by the both of you:

* Renaming the `Compressed` Responder to `Compress` -- `Compressed` sounded wrong because the inner type is not already compressed, it's something that will be compressed at response time.
* Invalid exclusions. Rather than use the default exclusions if anything is wrong, I've changed it to warn for each invalid MIME type but continue with whichever ones are valid.

I did reorganize some documentation to keep up with the rest of `contrib` and made a few other minor style changes and fixes, but the actual implementation should be otherwise unchanged.

The only thing missing is an example -- is this worthy of a full example itself or is including it in another example somewhere enough?

derekdreery:
It's worth noting that compression can compromise security in certain circumstances (see [BREACH]). In order to be vulnerable, you need all of these things to be true for a response:

 1. It is compressed
 2. The attacker can control some part of the response (it reflects user input, for instance)
 3. The response contains a secret

I'm generally weary of compressing API data, and only compress static assets. Not sure how this applies here - maybe just a note in the docs, or maybe things like `application/json` should not be compressed by default. Up to you.

[BREACH]: http://breachattack.com/

contrib/lib/src/compression/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+                .expect("Compression Context registered in on_attach");
+
+            if CompressionUtils::skip_encoding(&content_type, &context) {
+                return;
+            }
+        }
+
+        // Compression is done when the request accepts brotli or gzip encoding
+        // and the corresponding feature is enabled
+        if cfg!(feature = "brotli_compression") && CompressionUtils::accepts_encoding(request, "br")
```
meven: Could be nice to check gzip and brotli feature upon the fairing's on_attach and save in a variable which compression to use.
A user might not see what he is missing, or think his responses are compressed but are not.
Plus it would save some instructions on every single requests.

jebrosen: The only checks are 1) "is brotli enabled" and 2) "did this request say it supports brotli". 1 is done at compile time with the `cfg!`, and 2 must be checked again for every request.

meven: Thanks, I forgot cfg! was compile time only,

Great Pr btw

meven: Still half my point was that the cfg! check should also be done in on_attach so that if none of the required features are set the user can get an error instead of discovering at runtime that something is amiss, given the user has not read properly the documentation.

Martin1887: In `lib.rs` there is these lines, so if none of compression features are enabled a compile error is raised:

```
#[cfg(any(feature = "brotli_compression", feature = "gzip_compression"))]
pub mod compression;
```

Regards.

Martin1887:
This pull request have been opened too much time. Any news @jebrosen?

It seems that the project is a bit paralyzed, it's a pity because it is a great project.

contrib/lib/src/compression/mod.rs
meven: Ok didn't know about this.

jebrosen:
> This pull request have been opened too much time. Any news jebrosen?

You're right, and this fell off my todo list and got buried a while back. I'm working on merging it right now.

jebrosen:
This is at the point (and has been for a while, sorry!) where going back and forth with GitHub reviews isn't so productive. So I've merged this in 6a55aa7, with my refactoring in 0a3960b to keep up with some changes made to the way `contrib` was organized and to rename the `Responder`.

There are a few things that are not done, but should be easy to modify or add:

* (docs) Guide-level documentation
* (docs) Mention of the security implications raised by @derekdreery (https://github.com/SergioBenitez/Rocket/pull/781#issuecomment-464331608)
* (docs) Example configuration of exclusions
* (tests) Tests that are run when only one of `brotli` or `gzip` is enabled - currently, the tests are only run with both enabled.
* "Real life" stress testing

I will put these on my personal todo list (or open an issue if I won't get to some of them), but anyone interested is welcome to work on them as well.

jebrosen:
Thanks @Martin1887 for persisting, and sorry this has taken so long!

PR #783, opened 2018-10-06
Add toml support

richo:
I think the known_media_types change is good, it's unclear if this really hits the bar for being a first class content type in content.rs.

Tests are running now, so I'll followup if I broke anything. I couldn't find any tests for this, so let me know if there are some I should have updated.


core/lib/src/response/content.rs
```diff
@@ -89,4 +89,5 @@ ctrs! {
     Plain: Plain, "plain text", "text/plain",
     Css: CSS, "CSS", "text/css",
     JavaScript: JavaScript, "JavaScript", "application/javascript"
+    Toml: Toml, "TOML", "application/toml"
```
jebrosen: I'm pretty sure this should be `Toml: TOML, "TOML", "application/toml"` (second one uppercased to match the definition earlier). The `JavaScript` line above needs a trailing comma, as well.

richo: Good catch. I fixed that up, although I think I'm parsing your above comment as prefering not to include toml?

The underlying problem I have is purely that I need to send some toml in a response, so if there's a way to create content types at runtime I'm happy to do that as well, grepping around it seemed like this was the only way to do it.

SergioBenitez: You can simply `ContentType::new("application", "json")`. If you're trying to set the content-type on a response, use [`Content`](https://api.rocket.rs/rocket/response/struct.Content.html).

SergioBenitez:
Because there is no registered media type, I'm closing this out. If TOML gets a media type registered, we can add built in support. Until then, please manually create the `ContentType`.

Side note: It would be interesting for Rocket to support adding new known content types out of band.

richo:
As a note to future readers, Chrome today recognises toml files as an extension in the save dialog  when presented with a toml extension in a content-disposition: filename.

But when I opened this PR I didn't realise how easy it was to create new ContentTypes from my application so  merging this isn't very important.

PR #785, opened 2018-10-06
Allow wildcards in content-type

### LINKED ISSUE #583: Allow wildcards in content-type
Currently (Rocket version 0.4.0-dev) if you use a wildcard in content-type you will get a warning like this:

```rust
warning: 'image/*' is not a known media type
 --> src/handlers/image.rs:5:33
  |
5 | #[post("/image", format = "image/*", data = "<data>")]
  |                                 ^^^^^^^^^
```

Using wildcard e.g. for images is needed as you do not know what the actual type of the image will be (png, gif, jpg etc.) and you want to handle all of them in the same route.

zacps:
Adds the top level media types (From the [iana](https://www.iana.org/assignments/media-types/media-types.xhtml) list) to the known media types.

This doesn't add any tests, I couldn't figure out where they had to go for compiletest to require no warnings.

Fixes #583

### DIFF
--- a/core/http/src/known_media_types.rs
+++ b/core/http/src/known_media_types.rs
@@ -1,6 +1,15 @@
 macro_rules! known_media_types {
     ($cont:ident) => ($cont! {
         Any (is_any): "any media type", "*", "*",
+        Application (is_application): "any application type", "application", "*",
+        Audio (is_audio): "any audio type", "audio", "*",
+        Font (is_font): "any font type", "font", "*",
+        Image (is_image): "any image type", "image", "*",
+        Message (is_message): "any message type", "message", "*",
+        Model (is_model): "any model type", "model", "*",
+        Multipart (is_multipart): "any multipart type", "multipart", "*",
+        Text (is_text): "any text type", "text", "*",
+        Video (is_video): "any video type", "video", "*",
         Binary (is_binary): "binary data", "application", "octet-stream",
         HTML (is_html): "HTML", "text", "html" ; "charset" => "utf-8",
         Plain (is_plain): "plain text", "text", "plain" ; "charset" => "utf-8",
--- a/examples/form_kitchen_sink/src/main.rs
+++ b/examples/form_kitchen_sink/src/main.rs
@@ -28,7 +28,7 @@ struct FormInput<'r> {
     select: FormOption,
 }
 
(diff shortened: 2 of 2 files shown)


zacps:
The build failure seems to just be nightly breakage.

SergioBenitez:
I don't think this is the right approach. We shouldn't have constructors for media types that aren't specific, the lone exception being `Any` for being special. Instead, we should record known top-level media-types in a different way and use that knowledge specifically to prevent the warning [from being emitted](https://github.com/SergioBenitez/Rocket/blob/master/core/codegen_next/src/http_codegen.rs#L78). In other words, we need to update `is_known` for top-level media types or provide another method that checks for known top-level media-types. I'd prefer the latter.

zacps:
I'm not sure I understand the motivation? Why is having a constructor for the top level types bad?

SergioBenitez:
@zacps Because all existing constructors (minus `Any`) are for media-types registered with IANA. These are not media types in the same sense; these are "top-level" types. As such, this PR conflates the intended meaning of the `MediaType` type.

zacps:
How about adding a `is_concrete` method or similar which returns true for specific media types `text/plain` and false for generic ones ` image/*`.

Then `ContentType` could only accept concrete media types and `Accepts` could accept either.

The earlier would be a breaking change is `Any` was counted as not concrete, but it could probably be introduced through a warning.

SergioBenitez:
That already exists as `specificity()`.

jebrosen:
One critical difference between `Accept` and `Content-Type` is that Rocket applications will not be *creating* `Accept` headers nearly as often because they are part of the request. As far as I'm aware, the wildcard designation is only even meaningful in certain contexts (such as the `Accept` header), so it shouldn't be so easy to get one on accident.

If we are going to explicitly support wildcards in the `format` specifier, then I think it's worth taking a close look at the default route collision and ranking logic as well and the interactions between unspecified, top-level-only specified, and fully specified `format`s.

SergioBenitez:
#819 shows another reason why this is the wrong approach: `is_image` (and the rest) match exactly on "image/*", where it should probably match on "image/$anything". As such, I'd like to stick to my original assertion that we should handle these distinctly and am closing this out.

PR #794, opened 2018-10-14
Call unreachable macro to make clippy happy

mmrath:


core/codegen_next/src/attribute/route.rs
```diff
@@ -181,7 +181,7 @@ fn data_expr(ident: &syn::Ident, ty: &syn::Type) -> TokenStream2 {
             Borrowed(Outcome::Success(ref __v)) => {
                 Borrowed(Outcome::Success(::std::borrow::Borrow::borrow(__v)))
             }
-            Borrowed(__o) => Borrowed(__o.map(|_| loop { /* unreachable */ })),
+            Borrowed(__o) => Borrowed(__o.map(|_| loop { unreachable!("Unreachable code"); })),
```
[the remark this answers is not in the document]
SergioBenitez: Alternatively, find a (clean) way to rewrite this that obviates the need for divergence of some sort.

SergioBenitez:
```suggestion
            Borrowed(__o) => Borrowed(__o.map(unreachable!("case handled in previous match block"))),
```

jebrosen: I don't think this is right -- this looks like it evaluates `unreachable!` before entering `map()`

mmrath:
@SergioBenitez @jebrosen I will close this PR for now? I can raise another one with the lambda or you can fix it later.

PR #795, opened 2018-10-14
Nested form attributes. (Closes #313)

### LINKED ISSUE #313: [Feature Request] Nested form attributes
## Feature Requests

Go, Rails, Express all have some way of parsing nested attributes easily from forms.

I think it would be pretty nice to have the ability to able to specify a struct like:

```Rust
#[derive(FromForm]
struct StarterApplication {
  user: User,
  business: Business,
}

#[post("/apply", data="<form>")]
fn post_apply(context: Context, form: StartedApplication) -> Flash<Redirect> {
  //..
}
```

And based on the dot notation or brackets, allow nested attributes in forms to parse automatically:

```html
<input type="email" name="user[email]" />
<input type="email" name="business[name]" />
```

or 

```html
<input type="email" name="user.email" />
<input type="email" name="business.name" />
```

Instead I find myself doing this:

```Rust
#[derive(FromForm, Serialize)]
struct StarterApplication {
    user_email: String,
    user_full_name: String,
//... user fields ...
    business_name: String,
//... business fields ...
}
```

and manually converting it to the individual structs.

That being said, if there was a way to possibly specify multiple structs on FromForm and get access to the form field name and not just the value, that might also be a viable option? I'm not sure what my options are regardless. Perhaps I'm missing something?

It could possibly be an extension of rocket if there was a way to get the values of the fields that come through. I'm not sure that is possible though.

Thanks!!

tcmal:
Not much to say; Creates a new `FormItems` instance for each child but this should be inconsequential as `FormItem` only stores references. If there are form errors in nested children they won't show the prefix however, so a nested `User` model will only show a missing `name`, not a missing `user_name`.

### DIFF
--- a/core/codegen_next/src/derive/from_form.rs
+++ b/core/codegen_next/src/derive/from_form.rs
@@ -3,7 +3,8 @@ use derive_utils::{*, ext::{TypeExt, Split3}};
 
 #[derive(FromMeta)]
 struct Form {
-    field: FormField,
+    field: Option<FormField>,
+    nested: Option<bool>
 }
 
 struct FormField {
@@ -41,8 +42,8 @@ fn validate_struct(gen: &DeriveGenerator, data: Struct) -> Result<()> {
     let mut names = ::std::collections::HashMap::new();
     for field in data.fields().iter() {
         let id = field.ident.as_ref().expect("named field");
-        let field = match Form::from_attrs("form", &field.attrs) {
-            Some(result) => result?.field,
+        let field = match Form::from_attrs("form", &field.attrs) { // TODO: probably could look nicer
+            Some(result) => result?.field.unwrap_or(FormField { span: Spanned::span(&id), name: id.to_string() }),
             None => FormField { span: Spanned::span(&id), name: id.to_string() }
         };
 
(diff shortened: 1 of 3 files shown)


SergioBenitez:
This is great. I don't have a chance now for a deep review, but I wanted to point out the following higher-level issues:

  * Let's call this "filtering" a `subform` or something similar.
  * Create a subform (fn subform(&self, name: &str) -> Self) given a name, without the separator.
  * The separator should be a `.`, not a `_`, and shouldn't be chosen by the user.
  * We shouldn't allocate to crate a subform.
  * I don't think this particular implementation allows for deeper nesting. It needs to.

@jebrosen Do you have anything to add?

tcmal:
The only way I can think of without allocating would be to have `from_form` accept any `Iterator<FormItem>`. The strings are all references so it shouldn't be allocating much.

I also don't see why this wouldn't allow deeper nesting.

SergioBenitez:
> I also don't see why this wouldn't allow deeper nesting.

Apologies, I believe you're right.

> The only way I can think of without allocating would be to have from_form accept any Iterator<FormItem>. The strings are all references so it shouldn't be allocating much.

Alternatively, simply pass the name of the field to use to filter the subform to `FormItems` itself so that it only returns items that match. Then, no allocations of any sort are required.

tcmal:
Doing so would mean items in between would be skipped entirely though, i.e.

```
   ?time=now            read
   &user_name=John      subform -> start
   &profit=2                        skip
   &user_surname=Smith             end
   &done=True           read
```

Unless the subform tracks how many it goes through and 'rewinds' at the end, but this means the root form will waste time reading over the subform's attributes and is weird for others directly using FormItems.

SergioBenitez:
> Doing so would mean items in between would be skipped entirely though, i.e.

That's what we want.

It's no different than what you're doing now, except your solution allocates several times.

tcmal:
As in things meant to be top-level would be skipped, so if someone makes a request in a weird order they could get an error which is likely to be confusing.

tcmal:
Most browsers don't support >2,000 characters in URL length so it's at most (but not really) 500 parameters that could be passed. Everything being cloned is a reference so the absolute worst case is ~4KiB for 64 bit I believe? This is all theoretical though; I'm not sure why anyone would want 500 URL parameters that are all nested.

PR #803, opened 2018-10-28
Update 'contains_template' documentation.

jebrosen:
With template reloading active, the set of loaded templates can change. Since `contains_template` is "live" with respect to the list of templates, this text is more accurate.


SergioBenitez:
Merged in d5fec97fcb2eb20df3be6a59b529289535d28263.

PR #805, opened 2018-10-28
Support usage of rocket macros in 2018 edition crates.

jebrosen:
This commit consists of 3 changes:

* Replace `::rocket::` and `::std::` with `rocket::` and `std::` in `rocket_internal_uri`. This only became necessary quite recently, and I want to track down which commit in rust caused it and file a bug report if it's unintentional. **This change appears to be necessary for 2018-edition crates to be able to call `uri!`, regardless of import method!**
* Change the `hello_2018` example to not use `extern crate rocket;` and `use rocket::{macros...};` instead. This is idiomatic for the 2018 edition and the original motivation for adding the example.
* Change the `#[route]`-generated macro to call `rocket::rocket_internal_uri` instead of `rocket_internal_uri`. This probably requires 'rocket' to be not-renamed and at the crate root, but this is already required by most of the codegen. Without this change, `use rocket::uri;` without `#[macro_use]` must always be accompanied by `use rocket::rocket_internal_uri;`.

The first change may be due to a regression in rustc, but it might also be intentional and only exposed because we define `decl_macro`s from inside a `proc_macro`. We could hold off on it while I track the cause down, but in the meantime no 2018 edition crate can use the type-safe URI generation.

The second and third change align Rocket's macros better with 2018 usage style and demonstrate that it works.

### DIFF
--- a/core/codegen/src/attribute/route.rs
+++ b/core/codegen/src/attribute/route.rs
@@ -318,6 +318,8 @@ fn generate_internal_uri_macro(route: &Route) -> TokenStream2 {
 
     quote! {
         pub macro #generated_macro_name($($token:tt)*) {{
+            extern crate rocket;
+            extern crate std;
             rocket::rocket_internal_uri!(#route_uri, (#(#dynamic_args),*), $($token)*)
         }}
     }
--- a/core/codegen/src/attribute/route.rs
+++ b/core/codegen/src/attribute/route.rs
@@ -317,9 +317,9 @@ fn generate_internal_uri_macro(route: &Route) -> TokenStream2 {
     let route_uri = route.attribute.path.origin.0.to_string();
 
     quote! {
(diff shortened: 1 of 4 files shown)


SergioBenitez:
> no 2018 edition crate can use the type-safe URI generation

I don't think this is true. I think the correct statement is that you can't use the `uri!` macro unless it is imported via `#[macro_use]`.

jebrosen:
> >     no 2018 edition crate can use the type-safe URI generation

> I don't think this is true. I think the correct statement is that you can't use the uri! macro unless it is imported via #[macro_use].

I will verify th specifics when I make a small test case to file a bug in rust, but it did not matter in which way the macros are imported. `#[macro_use]` was the first thing I tried to "work around" the issue when it first popped up, and it didn't help.

SergioBenitez:
Merged in 328bf4b32e105815bffc4f3eda7ea7d31191f3f1.

PR #810, opened 2018-11-03
Make the inner value of Form public, so users can create their own Forms

rukai:
As discussed with sergio on irc.
I've tested it on my project and it works as expected.


SergioBenitez:
Merged in b4f85daaf248e989f22a78a3ec787759674df9d2.

PR #816, opened 2018-11-05
Put private cookies behind a feature

LinusU:
I'm quite new to both Rocket and Rust, so I've probably made a few mistakes. I'm very open to feedback though!

The motivation behind this is to be able to compile Rocket without `ring`, which currently doesn't ship assembly for the MIPS architecture.

todo:
- [x] The testing script needs to be updated so it tests with the feature and without it.
- [x] The documentation needs to be updated.
- [x] The guide should say that you can disable this functionality. 
- [x] Try and decrease the number of `#[cfg(...)]`s

### DIFF
--- a/core/lib/src/config/config.rs
+++ b/core/lib/src/config/config.rs
@@ -10,7 +10,8 @@ use {num_cpus, base64};
 use config::Environment::*;
 use config::{Result, ConfigBuilder, Environment, ConfigError, LoggingLevel};
 use config::{Table, Value, Array, Datetime};
-use http::Key;
+
+#[cfg(feature = "private-cookies")] use http::Key;
 
 /// Structure for Rocket application configuration.
 ///
@@ -49,6 +50,7 @@ pub struct Config {
     /// How much information to log.
     pub log_level: LoggingLevel,
     /// The secret key.
+    #[cfg(feature = "private-cookies")]
     crate secret_key: SecretKey,
     /// TLS configuration.
     crate tls: Option<TlsConfig>,
@@ -231,6 +233,7 @@ impl Config {
         let default_workers = (num_cpus::get() * 2) as u16;
 
         // Use a generated secret key by default.
+        #[cfg(feature = "private-cookies")]
         let key = SecretKey::Generated(Key::generate());
 
         Ok(match env {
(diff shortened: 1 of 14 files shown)


scripts/test.sh
```diff
@@ -107,6 +107,12 @@ if [ "$1" = "--contrib" ]; then
   done
 
   popd > /dev/null 2>&1
+elif [ "$1" = "--no-default-features" ]; then
```
SergioBenitez: Let's have a `--core` that emulates the existing `--contrib`. We only want to test the core library here; it's too expensive to rebuild and test all of the examples. To be clear, you should have an array of features (tls, private-cookies), test once without default features, test once with the default features, and then test with each feature in the array separately enabled.

LinusU: Does this look good? c272ea1

LinusU:
@SergioBenitez would you mind taking a new look 👀 I've tried to address all the issues ☺️ 

The diff is a bit noisy in 1c50a9e, what I did was to move all the `private_*` functions in to another `impl` block, and put the `cfg[...]` toggle on that `impl` block.

core/lib/src/config/config.rs
```diff
@@ -49,6 +50,7 @@ pub struct Config {
     /// How much information to log.
     pub log_level: LoggingLevel,
     /// The secret key.
+    #[cfg(feature = "private-cookies")]
```
jebrosen: I think the possibility was discussed on IRC of leaving `secret_key` in regardless of `cfg`. I realize that `cookie` only exposes SecretKey when `secure` is enabled, but I think `SecretKey` could become a unit struct in the alternative case if reducing the `cfg` annotations is an important goal. I suppose that might have downsides, though.

SergioBenitez:
I merged a few of the `cfg`s and left `secret_key` as a configuration option (though with a "disabled" message when disabled) regardless of whether the feature was enabled. This was merged in 53758c6dd7453c906f134b12112a1c4702ab0277 with my changes in 4224419e63a99b1f21db447f8c4aedab8daec41a. Thanks so much for this!

LinusU:
Thank you, it was nice to read thru the changes to see what I can do better next time!

PR #818, opened 2018-11-06
Make local client Sync

### LINKED ISSUE #814: `std::cell::RefCell<rocket::http::CookieJar>` cannot be shared between threads safely
```
error[E0277]: `std::cell::RefCell<rocket::http::CookieJar>` cannot be shared between threads safely
   --> src/main.rs:359:13
    |
359 |             thread::spawn(move|| {
    |             ^^^^^^^^^^^^^ `std::cell::RefCell<rocket::http::CookieJar>` cannot be shared between threads safely
    |
    = help: within `rocket::local::Client`, the trait `std::marker::Sync` is not implemented for `std::cell::RefCell<rocket::http::CookieJar>`
    = note: required because it appears within the type `std::option::Option<std::cell::RefCell<rocket::http::CookieJar>>`
    = note: required because it appears within the type `rocket::local::Client`
    = note: required because of the requirements on the impl of `std::marker::Send` for `std::sync::Arc<rocket::local::Client>`
    = note: required because it appears within the type `[closure@src/main.rs:359:27: 364:14 client:std::sync::Arc<rocket::local::Client>, tx:std::sync::mpsc::Sender<rocket::http::Status>]`
    = note: required by `std::thread::spawn`
```

`rocket::local::Client` can't be shared between threads in version 0.4.0-rc.1 while it used can be shared in 0.3.x.

messense:
Fixes #814

### DIFF
--- a/core/lib/src/local/request.rs
+++ b/core/lib/src/local/request.rs
@@ -111,7 +111,8 @@ impl<'c> LocalRequest<'c> {
 
         // Set up any cookies we know about.
         if let Some(ref jar) = client.cookies {
-            for cookie in jar.borrow().iter() {
+            let cookies = jar.read().expect("LocalRequest::new lock");
+            for cookie in cookies.iter() {
                 request.cookies().add_original(cookie.clone().into_owned());
             }
         }
@@ -408,7 +409,7 @@ impl<'c> LocalRequest<'c> {
         // If the client is tracking cookies, updates the internal cookie jar
         // with the changes reflected by `response`.
         if let Some(ref jar) = client.cookies {
-            let mut jar = jar.borrow_mut();
+            let mut jar = jar.write().expect("LocalRequest::_dispatch lock");
             let current_time = ::time::now();
             for cookie in response.cookies() {
                 if let Some(expires) = cookie.expires() {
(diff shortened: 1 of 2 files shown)


core/lib/src/local/request.rs
```diff
@@ -111,7 +111,8 @@ impl<'c> LocalRequest<'c> {
 
         // Set up any cookies we know about.
         if let Some(ref jar) = client.cookies {
-            for cookie in jar.borrow().iter() {
+            let cookies = jar.read().unwrap();
```
messense: Not sure how to deal with these `unwrap`s.

jebrosen:
Should we introduce some kind of warning in the documentation that sharing Client between threads can exhibit non-deterministic behavior? There was no risk before cookie tracking and no risk when Client was !Sync, but Sync + cookie tracking opens up the possibility of misuse.

core/lib/src/local/request.rs
messense: Changed to use `expect`

SergioBenitez:
@jebrosen Great idea.

SergioBenitez:
Merged in 64ba2056ef3678152646483c28a20231ad7ae98d. Note added in b16269a30ef9ae9d51878bfd8c6a84622360640d.

PR #830, opened 2018-11-18
Fix HSTS application logic in SpaceHelmet.

jebrosen:
The reorganization flipped the forcing logic and left out a check, so it now applies the forced HSTS header every time *except* when it should, overwriting any per-request HSTS headers.


SergioBenitez:
Good catch! Merged in af84f60bb8c82c48ad38bbc9dd681449eee1ff16.

PR #832, opened 2018-11-18
Fix notify dependency version.

jebrosen:
We are using `impl From<std::io::Error> for notify::Error` at https://github.com/SergioBenitez/Rocket/blob/76c830a467920f818ca1ff7d9e92798c2755f6f8/contrib/lib/src/templates/fairing.rs#L58 , which was added in `notify` 4.0.6.

This was first reported by **groks** in IRC.


SergioBenitez:
Merged in d14c3393a0c3fec4f8c2f83e68f9abc0975f56d8.

PR #833, opened 2018-11-19
Use the 'dyn' keyword explicitly to be compatible with the 'bare_trait_objects' lint.

jebrosen:
`bare_trait_objects` is part of the `rust_2018_idioms` lint group. I believe this will also allow `cargo fix` to upgrade crates that use the `uri!` macro. Currently, the compiler suggests adding the `dyn` keyword inside the macro invocation somewhere which never works.


SergioBenitez:
Merged in ed4055925df801c22f1b6f9fb95f5a0b3ba00dd3.

PR #851, opened 2018-12-09
Remove uses of FnBox

xfix:
`FnBox` is unlikely to be ever stabilized, as `unsized_locals` fixes `Box<dyn FnOnce()>` to work.

### DIFF
--- a/core/lib/src/fairing/ad_hoc.rs
+++ b/core/lib/src/fairing/ad_hoc.rs
@@ -1,5 +1,4 @@
 use std::sync::Mutex;
-use std::boxed::FnBox;
 
 use {Rocket, Request, Response, Data};
 use fairing::{Fairing, Kind, Info};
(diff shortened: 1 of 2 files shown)


core/lib/src/fairing/ad_hoc.rs
```diff
@@ -146,9 +145,10 @@ impl Fairing for AdHoc {
     fn on_attach(&self, rocket: Rocket) -> Result<Rocket, Rocket> {
         if let AdHocKind::Attach(ref mutex) = self.kind {
             let mut option = mutex.lock().expect("AdHoc::Attach lock");
-            option.take()
```
jebrosen: The extra let binding should be unnecessary -- just replace the line `.call_box((rocket,))` with `(rocket)`.

xfix: It is unnecessary, however it looks like ugly when having a function call in a multi-line chain like this.

xfix:
Also, from what I understand, the plan is to deprecate and remove `FnBox` as soon as possible.

SergioBenitez:
@xfix Where is that happening? We should ideally ask for that to be postponed.

xfix:
@SergioBenitez https://github.com/rust-lang/rust/pull/55431

jebrosen:
The last I checked:

* `FnBox` will be deprecated (probably not immediately removed) only once the replacement works. https://github.com/rust-lang/rust/issues/28796#issuecomment-437310015
* `Box<FnOnce>` working is only a partial replacement for `FnBox`, since `Box<FnOnce>` does not yet implement `FnOnce`. This is what rust-lang/rust#55431 will fix.
* The `impl` of `FnOnce` for `Box<FnOnce>` has some open questions.

So I don't think this will happen too quickly and we can probably wait a bit longer. I think the only reason not to merge this would be that `unsized_locals` might end up not panning out and it would have to be reverted.

xfix:
Even then, if `unsized_locals` ends up not panning out (which I don't see happening, I think it's sound enough), by this point there should be `boxed_closure_impls` feature which will make this work.

jebrosen:
With rust-lang/rust#59500 merged, this should be possible without the feature gate once the next nightly is released. That would require increasing the minimum rustc version, more so than this PR already does.

jebrosen:
Merged in 7ab1c42.

Issue #854, opened 2018-12-10
"chained comparison operators require parentheses" when using #[database]


RalfJung:
I am using rocket 0.4.0 on Debian testing and rustc `1.32.0-nightly (4a45578bc 2018-12-07)`.

I am playing around with the new built-in DB support (following https://rocket.rs/v0.4/guide/state/#databases), but running into compilation issues.  My code is (most things removed to get a minimal example)
```rust
#[macro_use] extern crate rocket_contrib;

use rocket_contrib::{
    databases::{r2d2_diesel, diesel}
};

// DB connection guard type
#[database("postgres")]
pub struct DbConn(pub r2d2_diesel::ConnectionManager<diesel::PgConnection>);

fn main() { }
```
with a `Cargo.toml` saying
```
rocket = "0.4"
rocket_contrib = { version = "0.4", features = ["handlebars_templates", "diesel_postgres_pool"] }
```
and my `Rocket.toml` contains
```
[global.databases]
postgres = { url = "postgres://r@/ff-node-monitor" }
```

This fails to compile, saying
```
error: chained comparison operators require parentheses
  --> src/main.rs:45:1
   |
45 | #[database("postgres")]
   | ^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: use `::<...>` instead of `<...>` if you meant to specify type arguments
   = help: or use `(...)` if you meant to specify fn arguments

error: aborting due to previous error
```

BaptisteGelez:
What if you try with

```rust
pub struct DbConn(pub r2d2_diesel::ConnectionManager::<diesel::PgConnection>);
```
(notice the `::` after `ConnectionManager`)

RalfJung:
No luck:
```
warning: unnecessary path disambiguator
  --> src/main.rs:45:53
   |
45 | pub struct DbConn(pub r2d2_diesel::ConnectionManager::<diesel::PgConnection>);
   |                                                     ^^ try removing `::`

error: chained comparison operators require parentheses
  --> src/main.rs:44:1
   |
44 | #[database("postgres")]
   | ^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: use `::<...>` instead of `<...>` if you meant to specify type arguments
   = help: or use `(...)` if you meant to specify fn arguments

error: aborting due to previous error
```

RalfJung:
However, this gets me one step further:
```rust
[...]

Issue #839, opened 2018-11-30
Macro parameterized State types cause arcane scoping errors


Korvox:
Happens on 0.4.0-rc.1 and master.

[See this example program.](https://gist.github.com/Korvox/521af158c49f1a14ab83155a5448a554)

Normally you can generate routes in macros just fine, but if you make the type parameter of a state guard a macro parameter compilation fails with really arcane error messages about scoping issues on internal types such as " cannot find `__req` in this scope"  and "cannot find `__data` in this scope "

To reproduce simply define any route in a macro and parameterize the type of a State guard:

```
macro_rules! foo {
    ($type:ty) => {
        #[get("/")]
        fn index(thing: rocket::State<$type>) -> String {
            format!("Thing: {:?}", *thing)
        }
    }
}
```
@jebrosen on IRC has a tentative patch and is way more knowledgeable on the internals of how Rocket expands its spanned quotes that cause this.

jebrosen:
This patch fixes the test case and passes the existing test suite. At least these functions in the same file probably need similar modifications: `data_expr` (`__req`, `__data`), `query_exprs` (`__trail`, `__data`). Additionally, tests like @Korvox's should be added to guarantee that all kinds of guards work from macros in this way. Any other macros such as `#[catch]` should probably be checked, too.

```diff
diff --git a/core/codegen/src/attribute/route.rs b/core/codegen/src/attribute/route.rs
index 73bf86e..74213c7 100644
--- a/core/codegen/src/attribute/route.rs
+++ b/core/codegen/src/attribute/route.rs
@@ -300,12 +300,14 @@ fn query_exprs(route: &Route) -> Option<TokenStream2> {
 }
 
 fn request_guard_expr(ident: &syn::Ident, ty: &syn::Type) -> TokenStream2 {
+    let req = quote! { __req };
+    let data = quote! { __data };
     let span = ident.span().unstable().join(ty.span()).unwrap().into();
     quote_spanned! { span =>
         #[allow(non_snake_case, unreachable_patterns, unreachable_code)]
-        let #ident: #ty = match <#ty as ___r::FromRequest>::from_request(__req) {
+        let #ident: #ty = match <#ty as ___r::FromRequest>::from_request(#req) {
             ___Outcome::Success(__v) => __v,
-            ___Outcome::Forward(_) => return ___Outcome::Forward(__data),
+            ___Outcome::Forward(_) => return ___Outcome::Forward(#data),
[...]

PR #860, opened 2018-12-11
Fix All The Databases™

ELD:
Fixes the highlighted issues in #854 and #862.

Things to do:

- [x] Add `<>` around the connection pooling type in `contrib/codegen`
- [x] Add UI/regression test for #854
- [x] implement DerefMut
- [x] Test DerefMut trait using Sqlite and `:memory:`
- [x] bump versions of Rusqlite and r2d2-sqlite in docs

### Original text below:

This commit trades the previous error message in #854 for the following:

Code:

```toml
[package]
name = "rocket-db-repro"
version = "0.1.0"
authors = ["Eric Dattore <eric.dattore@gmail.com>"]
edition = "2018"

[dependencies]
rocket = { path = "../rocket/core/lib", version = "0.4" }
rocket_contrib = { path = "../rocket/contrib/lib", version = "0.4", features = ["databases", "diesel_sqlite_pool"] }
```
```rust
use rocket_contrib::{
    database,
    databases::{diesel, r2d2_diesel}
};

#[database("sqlite")]
pub struct DbConn(pub r2d2_diesel::ConnectionManager<diesel::PgConnection>);

fn main() {}

```

Error:
```
error[E0432]: unresolved import `rocket_contrib::databases::r2d2_diesel`
 --> src/main.rs:3:25
  |
3 |     databases::{diesel, r2d2_diesel}
  |                         ^^^^^^^^^^^ no `r2d2_diesel` in `databases`

error[E0412]: cannot find type `PgConnection` in module `diesel`
 --> src/main.rs:7:62
  |
7 | pub struct DbConn(pub r2d2_diesel::ConnectionManager<diesel::PgConnection>);
  |                                                              ^^^^^^^^^^^^ did you mean `Connection`?

error: aborting due to 2 previous errors

Some errors occurred: E0412, E0432.
For more information about an error, try `rustc --explain E0412`.
error: Could not compile `rocket-db-repro`.
```

### DIFF
--- a/contrib/codegen/src/database.rs
+++ b/contrib/codegen/src/database.rs
@@ -20,8 +20,8 @@ const EXAMPLE: &str = "example: `struct MyDatabase(diesel::SqliteConnection);`";
 const ONLY_ON_STRUCTS_MSG: &str = "`database` attribute can only be used on structs";
 const ONLY_UNNAMED_FIELDS: &str = "`database` attribute can only be applied to \
     structs with exactly one unnamed field";
-const NO_GENERIC_STRUCTS: &str = "`database` attribute cannot be applied to a struct \
-    with a generic type";
+const NO_GENERIC_STRUCTS: &str = "`database` attribute cannot be applied to structs \
+    with generics";
 
 fn parse_invocation(attr: TokenStream, input: TokenStream) -> Result<DatabaseInvocation> {
     let attr_stream2 = ::proc_macro2::TokenStream::from(attr);
@@ -31,7 +31,7 @@ fn parse_invocation(attr: TokenStream, input: TokenStream) -> Result<DatabaseInv
 
     let input = ::syn::parse::<DeriveInput>(input).unwrap();
     if !input.generics.params.is_empty() {
-        return Err(input.span().error(NO_GENERIC_STRUCTS));
+        return Err(input.generics.span().error(NO_GENERIC_STRUCTS));
     }
 
     let structure = match input.data {
@@ -56,34 +56,37 @@ fn parse_invocation(attr: TokenStream, input: TokenStream) -> Result<DatabaseInv
     })
 }
 
+#[allow(non_snake_case)]
 pub fn database_attr(attr: TokenStream, input: TokenStream) -> Result<TokenStream> {
     let invocation = parse_invocation(attr, input)?;
 
     // Store everything we're going to need to generate code.
-    let connection_type = &invocation.connection_type;
+    let conn_type = &invocation.connection_type;
     let name = &invocation.db_name;
-    let request_guard_type = &invocation.type_name;
+    let guard_type = &invocation.type_name;
     let vis = &invocation.visibility;
-    let pool_type = Ident::new(&format!("{}Pool", request_guard_type), request_guard_type.span());
+    let pool_type = Ident::new(&format!("{}Pool", guard_type), guard_type.span());
     let fairing_name = format!("'{}' Database Pool", name);
+    let span = conn_type.span().into();
 
     // A few useful paths.
-    let databases = quote!(::rocket_contrib::databases);
-    let r2d2 = quote!(#databases::r2d2);
+    let databases = quote_spanned!(span => ::rocket_contrib::databases);
+    let Poolable = quote_spanned!(span => #databases::Poolable);
+    let r2d2 = quote_spanned!(span => #databases::r2d2);
     let request = quote!(::rocket::request);
 
-    Ok(quote! {
+    let generated_types = quote_spanned! { span =>
         /// The request guard type.
-        #vis struct #request_guard_type(
-            pub #r2d2::PooledConnection<<#connection_type as #databases::Poolable>::Manager>
-        );
+        #vis struct #guard_type(pub #r2d2::PooledConnection<<#conn_type as #Poolable>::Manager>);
 
         /// The pool type.
-        #vis struct #pool_type(
-            #r2d2::Pool<<#connection_type as #databases::Poolable>::Manager>
-        );
+        #vis struct #pool_type(#r2d2::Pool<<#conn_type as #Poolable>::Manager>);
+    };
 
-        impl #request_guard_type {
+    Ok(quote! {
+        #generated_types
+
+        impl #guard_type {
             /// Returns a fairing that initializes the associated database
             /// connection pool.
             pub fn fairing() -> impl ::rocket::fairing::Fairing {
@@ -91,7 +94,7 @@ pub fn database_attr(attr: TokenStream, input: TokenStream) -> Result<TokenStrea
 
                 ::rocket::fairing::AdHoc::on_attach(#fairing_name, |rocket| {
                     let pool = #databases::database_config(#name, rocket.config())
-                        .map(<#connection_type>::pool);
+                        .map(<#conn_type>::pool);
 
                     match pool {
                         Ok(Ok(p)) => Ok(rocket.manage(#pool_type(p))),
@@ -117,12 +120,12 @@ pub fn database_attr(attr: TokenStream, input: TokenStream) -> Result<TokenStrea
             pub fn get_one(rocket: &::rocket::Rocket) -> Option<Self> {
                 rocket.state::<#pool_type>()
                     .and_then(|pool| pool.0.get().ok())
-                    .map(#request_guard_type)
+                    .map(#guard_type)
             }
         }
 
-        impl ::std::ops::Deref for #request_guard_type {
-            type Target = #connection_type;
+        impl ::std::ops::Deref for #guard_type {
+            type Target = #conn_type;
 
             #[inline(always)]
             fn deref(&self) -> &Self::Target {
@@ -130,14 +133,14 @@ pub fn database_attr(attr: TokenStream, input: TokenStream) -> Result<TokenStrea
             }
         }
 
-        impl ::std::ops::DerefMut for #request_guard_type {
+        impl ::std::ops::DerefMut for #guard_type {
             #[inline(always)]
             fn deref_mut(&mut self) -> &mut Self::Target {
                 &mut self.0
             }
         }
 
-        impl<'a, 'r> #request::FromRequest<'a, 'r> for #request_guard_type {
+        impl<'a, 'r> #request::FromRequest<'a, 'r> for #guard_type {
             type Error = ();
 
             fn from_request(request: &'a #request::Request<'r>) -> #request::Outcome<Self, ()> {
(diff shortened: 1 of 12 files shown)


jebrosen:
What happens when the relevant features are enabled? If the error doesn't mention Poolable, we should try to fix that as well.

ELD:
@jebrosen So in the example case, the `Cargo.toml` does enable the relevant features. We don't reexport `r2d2_diesel` because we don't need to.

You're right, though, that we should probably take this further and include an error about the type not implementing Poolable.

jebrosen:
Oh, interesting. I had assumed the test case was with an otherwise valid type.

Instead then, what is the error for an attempted connection type of `Vec<i32>`? That should exhibit the same syntax problem fixed by <>, is a valid type, and does not implement `Poolable`.

SergioBenitez:
We need to add a regression test for this.

ELD:
@jebrosen In that case, it does behave as you'd expect.

Code:
```rust
use rocket_contrib::{
    database,
};

#[database("sqlite")]
pub struct DbConn(pub Vec<i32>);

fn main() {
    println!("Hello, world!");
}
```

Error:
```
error[E0277]: the trait bound `std::vec::Vec<i32>: rocket_contrib::databases::Poolable` is not satisfied
 --> src/main.rs:5:1
  |
5 | #[database("sqlite")]
  | ^^^^^^^^^^^^^^^^^^^^^ the trait `rocket_contrib::databases::Poolable` is not implemented for `std::vec::Vec<i32>`

For more information about this error, try `rustc --explain E0277`.
error: Could not compile `rocket-db-repro`.
```

ELD:
@SergioBenitez Just a reminder that you agreed on IRC to help me write the UI/regression test for this error so we don't get tripped up by it again.

SergioBenitez:
@ELD Added, and fixed some diagnostics issues along the way.

ELD:
@SergioBenitez Okay, great, thanks!

contrib/codegen/src/database.rs
```diff
@@ -56,42 +56,45 @@ fn parse_invocation(attr: TokenStream, input: TokenStream) -> Result<DatabaseInv
...
     let vis = &invocation.visibility;
-    let pool_type = Ident::new(&format!("{}Pool", request_guard_type), request_guard_type.span());
+    let pool_type = Ident::new(&format!("{}Pool", guard_type), guard_type.span());
     let fairing_name = format!("'{}' Database Pool", name);
+    let span = conn_type.span().into();
 
     // A few useful paths.
-    let databases = quote!(::rocket_contrib::databases);
-    let r2d2 = quote!(#databases::r2d2);
+    let databases = quote_spanned!(span => ::rocket_contrib::databases);
```
jebrosen: Should these have the `span` of the `conn_type` or the default? This looks like the same thing that caused #839.

SergioBenitez: Yeah, I considered that. I _think_ the reason why this is okay is because the entire thing now has that `Span`, as opposed to some parts having it and others not. Perhaps you could add a test, like `core/lib/tests/expansion.rs` to ensure this is actually the case?

jebrosen: Ah, I checked again and this looks okay after all. These aren't local variables, they're absolute paths in a use. I'll try to add test tonight or tomorrow to be sure.

contrib/lib/Cargo.toml
```diff
@@ -61,8 +61,8 @@ r2d2 = { version = "0.8", optional = true }
 r2d2_postgres = { version = "0.14", optional = true }
 mysql = { version = "14", optional = true }
 r2d2_mysql = { version = "9", optional = true }
-rusqlite = { version = "0.14.0", optional = true }
-r2d2_sqlite = { version = "0.6", optional = true }
+rusqlite = { version = "0.15.0", optional = true }
```
SergioBenitez: I'm fairly certain that this is a breaking change, which means we won't be able to ship it in `0.4.x`. Can we either 1) allow both versions, or 2) keep the old version?

ELD: I'll investigate option 2, first. I'm sure we can keep the old version. Thanks for calling out the breaking change issue.

contrib/lib/tests/databases.rs
```diff
@@ -11,3 +11,38 @@ mod databases_tests {
...
+            .extra("databases", Value::Table(test_db))
+            .finalize()
+            .unwrap();
+
+        let rocket = rocket::custom(config).attach(SqliteDb::fairing());
+        let connection = SqliteDb::get_one(&rocket);
+
+        assert!(connection.is_some());
+
+        let result: rusqlite::Result<i32> = connection.unwrap().query_row("SELECT 1", &[], |row| row.get(0));
```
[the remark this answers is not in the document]
ELD: Ah, shoot. I didn't look closely enough. Thanks! I'll look for a method I can use that takes `&mut self`.

Never mind, this isn't done yet!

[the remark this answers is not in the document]
ELD: Cool. Referring back to the previous issue, it looks like transactions are the way to go with Sqlite and testing that a `DerefMut` impl exists.

contrib/codegen/src/database.rs
jebrosen: I think a test is of limited utility after all. After trying to write a macro that expanded to a `#[database(...)]` and encountering "must be a string literal", I can't think of much of a reason to *want* to write a database guard inside a macro. Since these are 1) not local variables and 2) the same span, I think it's fine.

SergioBenitez:
@ELD What's the status? Is this ready for a final review?

ELD:
@SergioBenitez yes, I believe it’s ready for final review.

contrib/codegen/tests/ui-fail/database-types.rs
```diff
@@ -0,0 +1,10 @@
+extern crate rocket;
+#[macro_use] extern crate rocket_contrib;
+
+struct Unknown;
+
+#[database("foo")]
+struct A(Unknown);
+//~^ ERROR Unknown: rocket_contrib::databases::Poolable
+
```
jebrosen: This would be a good place to add the regression test for #854, e.g. with `#[database("bar")] struct B(Vec<i32>);` -- that should error because `Vec<i32>` is not `Poolable`.

contrib/lib/tests/databases.rs
```diff
@@ -11,3 +11,59 @@ mod databases_tests {
...
+        test_db.insert("test_db".into(), Value::Table(test_db_opts));
+        let config = Config::build(Environment::Development)
+            .extra("databases", Value::Table(test_db))
+            .finalize()
+            .unwrap();
+
+        let rocket = rocket::custom(config).attach(SqliteDb::fairing());
+        let connection = SqliteDb::get_one(&rocket);
+
+        assert!(connection.is_some());
```
jebrosen: It's `assert-is-some` here, but `.unwrap()` in the test above. I'd like to see these test cases match a bit more closely with each other.

contrib/lib/tests/databases.rs
```diff
@@ -11,3 +11,59 @@ mod databases_tests {
...
+        test_db.insert("test_db".into(), Value::Table(test_db_opts));
+        let config = Config::build(Environment::Development)
+            .extra("databases", Value::Table(test_db))
+            .finalize()
+            .unwrap();
+
+        let rocket = rocket::custom(config).attach(SqliteDb::fairing());
+        let mut conn = SqliteDb::get_one(&rocket).unwrap();
+
+        let tx = conn.transaction().unwrap();
```
jebrosen: Let's add a comment here for posterity like ".transaction() on conn's inner type takes &mut self" so it's clear that this is where the real test is. Similarly with `deref_impl_present`

contrib/lib/tests/databases.rs
```diff
@@ -11,3 +11,65 @@ mod databases_tests {
...
+        test_db.insert("test_db".into(), Value::Table(test_db_opts));
+        let config = Config::build(Environment::Development)
+            .extra("databases", Value::Table(test_db))
+            .finalize()
+            .unwrap();
+
+        let rocket = rocket::custom(config).attach(SqliteDb::fairing());
+        let maybe_conn = SqliteDb::get_one(&rocket);
+
+        assert!(maybe_conn.is_some());
```
SergioBenitez: No need for this `assert` since the next line already panics if it's `None`. Better yet, change the next to an `expect` with a helpful (but short) message.

contrib/lib/tests/databases.rs
```diff
@@ -11,3 +11,65 @@ mod databases_tests {
...
+    use rocket_contrib::database;
+
+    #[database("test_db")]
+    struct SqliteDb(pub rusqlite::Connection);
+
+    #[test]
+    fn deref_mut_impl_present() {
+        let mut test_db: BTreeMap<String, Value> = BTreeMap::new();
+        let mut test_db_opts: BTreeMap<String, Value> = BTreeMap::new();
+        test_db_opts.insert("url".into(), Value::String(":memory:".into()));
```
SergioBenitez: Is this magic `:memory:` URL documented anywhere?

contrib/lib/tests/databases.rs
```diff
@@ -11,3 +11,65 @@ mod databases_tests {
...
+
+        assert!(maybe_conn.is_some());
+        let mut conn = maybe_conn.unwrap();
+
+        // Rusqlite's `transaction()` method takes `&mut self`, hence this is
+        // testing the presence of a `DerefMut` trait on the generated
+        // connection type.
+        let tx = conn.transaction().unwrap();
+
+        let result: rusqlite::Result<i32> = tx.query_row("SELECT 1", &[], |row| row.get(0));
```
SergioBenitez: Remove the `assert` below, change this to:

```suggestion
        let _: i32 = tx.query_row("SELECT 1", &[], |row| row.get(0)).expect("get row");
```

contrib/lib/tests/databases.rs
```diff
@@ -11,3 +11,65 @@ mod databases_tests {
...
+
+        assert!(result.is_ok());
+
+        tx.commit().unwrap();
```
SergioBenitez: Do we need this for some reason?

contrib/lib/tests/databases.rs
```diff
@@ -11,3 +11,65 @@ mod databases_tests {
...
+    }
+
+    #[test]
+    fn deref_impl_present() {
```
SergioBenitez: Same thing in this test with the assets and `expect`s.

contrib/lib/tests/databases.rs
jebrosen: It's originally documented at https://sqlite.org/c3ref/open.html, I believe. Neither `rusqlite` nor `r2d2_sqlite` documents this.

contrib/lib/tests/databases.rs
jebrosen: The transaction is rolled back without it. For these tests, that's probably fine though.

ELD:
@SergioBenitez I addressed all of your nitpicks!

contrib/lib/tests/databases.rs
SergioBenitez: Okay. Would be nice to let users know, but perhaps it's fine to delegate to lower level docs here.

SergioBenitez:
Merged in three commits ending in aba3ad327b862ac05c84a8d86b13566409225c3d. Thanks so much, @ELD!

PR #863, opened 2018-12-12
Eradicate bootstraps

jebrosen:
Removes the only test bootstrap by using temporary files and removes the code that runs test bootstraps. I'm not entirely sold on using `temp_dir()`, but I think I'm otherwise mostly happy with the way I ended up doing this. Hopefully the CI agrees.

### DIFF
--- a/examples/todo/src/main.rs
+++ b/examples/todo/src/main.rs
@@ -78,8 +78,8 @@ fn index(msg: Option<FlashMessage>, conn: DbConn) -> Template {
     })
 }
 
-fn rocket() -> (Rocket, Option<DbConn>) {
-    let rocket = rocket::ignite()
+fn rocket() -> Rocket {
+    rocket::ignite()
         .attach(DbConn::fairing())
         .attach(AdHoc::on_attach("Database Migrations", |rocket| {
             let conn = DbConn::get_one(&rocket).expect("database connection");
@@ -94,16 +94,9 @@ fn rocket() -> (Rocket, Option<DbConn>) {
         .mount("/", StaticFiles::from("static/"))
         .mount("/", routes![index])
         .mount("/todo", routes![new, toggle, delete])
-        .attach(Template::fairing());
-
-    let conn = match cfg!(test) {
-        true => DbConn::get_one(&rocket),
-        false => None,
-    };
-
-    (rocket, conn)
+        .attach(Template::fairing())
 }
 
 fn main() {
-    rocket().0.launch();
+    rocket().launch();
 }
--- a/examples/todo/src/task.rs
+++ b/examples/todo/src/task.rs
@@ -50,4 +50,8 @@ impl Task {
     pub fn delete_with_id(id: i32, conn: &SqliteConnection) -> bool {
         diesel::delete(all_tasks.find(id)).execute(conn).is_ok()
     }
+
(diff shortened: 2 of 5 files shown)


examples/todo/src/main.rs
```diff
@@ -78,8 +78,14 @@ fn index(msg: Option<FlashMessage>, conn: DbConn) -> Template {
     })
 }
 
-fn rocket() -> (Rocket, Option<DbConn>) {
-    let rocket = rocket::ignite()
+fn rocket(override_config: Option<Config>) -> (Rocket, Option<DbConn>) {
+    let rocket = if let Some(config) = override_config {
```
SergioBenitez: Can this be `override_config.map(rocket::custom).unwrap_or_else(rocket::ignite)`?

examples/todo/src/tests.rs
```diff
@@ -1,25 +1,37 @@
...
+            .extra("databases", databases)
+            .expect("valid config");
+
+        let (rocket, db) = super::rocket(Some(config));
+
         let $client = Client::new(rocket).expect("Rocket client");
         let $conn = db.expect("failed to get database connection for testing");
         $block
+
+        let _ = std::fs::remove_file(db_path);
```
SergioBenitez: If the test fails, won't this file stick around forever? Can we push this line up and have the removal happen on unwind?

jebrosen: "forever" in this case is until the temporary directory is cleared, but it's definitely not ideal. Now that you mention it, I'm afraid the file removal will unconditionally fail on windows because a connection is still open by this point. I'll have to think about this a bit more, maybe a `Drop` guard is needed to solve this.

jebrosen:
> What are we gaining from this, exactly? The tests now run in parallel, is that right? Aside from that, do we gain something else?

* Yes, the tests (should) run in parallel.
* It's now safe to `cargo run` the example, make some changes to the DB, then `cargo test` -- the tests don't clobber the database, and the tests still work.

I would have preferred a solution that uses a non-committed transaction, but it's *much* more complicated because the Rocket instance and the test code would need to share a single *connection* for that to work properly.

SergioBenitez:
My concern here is that 1) this is much more complicated than what we had previously, and 2) that this example won't generalize well which will confuse users.

jebrosen:
I'm going to try a different approach. Instead of messing with configuration, what if `run_test!` ran a `delete from tasks;` before executing the test body? This would still let us remove the bootstrap script and nothing would need to change in `main.rs`.

Another benefit of clearing the tasks before each test is that tests will be more clearly guaranteed to be isolated. This is currently true because `Task::all` sorts by ID, which isn't apparent from reading the tests themselves.

examples/todo/src/task.rs
```diff
@@ -50,4 +50,8 @@ impl Task {
     pub fn delete_with_id(id: i32, conn: &SqliteConnection) -> bool {
         diesel::delete(all_tasks.find(id)).execute(conn).is_ok()
     }
+
+    pub fn delete_all(conn: &SqliteConnection) -> bool {
```
SergioBenitez: Should we `#[cfg(test)]` this, to be safe?

SergioBenitez:
Merged in b9c3a5c64bf20b1b83bfb709d117970156b0fc00. Hoorah!

PR #864, opened 2018-12-13
Small revamp of pastebin guide

jtdowney:
I went through the guide and tried to update things that were out of date or missing. I wasn't able to find any big changes for 0.4. I did look at if the URL building in `upload` could use the new `uri!` but 1) it doesn't seem to help for the case of absolute URLs and 2) it would make the code unbuildable in that intermediate time before `retrieve` exists.


SergioBenitez:
Thanks so much for this! Merged in 9750ae67dd929c4ee09d198338a719e662eae4c6.

PR #874, opened 2018-12-26
Update README.md

andrewbanchich:
Update examples to use Rust 2018 syntax for importing macros.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,8 @@ expressibility, and speed. Here's an example of a complete Rocket application:
 ```rust
 #![feature(proc_macro_hygiene, decl_macro)]
 
-#[macro_use] extern crate rocket;
+extern crate rocket;
+use rocket::{get, routes};
 
 #[get("/<name>/<age>")]
 fn hello(name: String, age: u8) -> String {


jebrosen:
This has been on my mind for a while, and I have a lot of thoughts on this subject.

First, one thing should be cleared up, either in our documentation or by updates to the 2018 edition guide: both import styles should work the same way in both the 2015 and 2018 editions. `use rocket::get;` imports the `#[get]` attribute into the current module, and `#[macro_use] extern crate rocket;` imports the `#[get]` attribute crate-wide.

It is more idiomatic for the 2018 edition to import individual attributes (making the `extern crate` unnecessary), but `#[macro_use]` is still a convenient option for importing all of Rocket's macros at once everywhere if it's desired. This works great today for both 2018 edition crates and 2015 edition crates, including older projects that haven't migrated yet, so I'm not convinced we should change examples just yet. On the other hand, `cargo new` does default to the 2018 edition now.

But async/await is on the horizon, and it might tip the scales. If `await` is only available in Rust 2018, it might be worth encouraging the entire Rocket ecosystem to migrate to 2018 with the first release of Rocket that supports it.

The bottom line: I think this change is too early for the reasons above and not comprehensive enough because it only changes the `README`.

andrewbanchich:
Agreed! I wanted to see if this would be accepted before changing everything across the board.

> If await is only available in Rust 2018, it might be worth encouraging the entire Rocket ecosystem to migrate to 2018 with the first release of Rocket that supports it.

Wouldn't expecting devs to use Rust 2018 not be a big deal since Rocket is already dependent on `nightly`?

jebrosen:
> Wouldn't expecting devs to use Rust 2018 not be a big deal since Rocket is already dependent on `nightly`?

It's not a big deal, and new projects are likely to be on Rust 2018 anyway. But there are a lot of projects that are probably still on Rust 2015 and Rocket v0.3.

Thinking on it more, considering that `use rocket::<macro>;` a) is idiomatic for 2018 and b) works on  2015 makes me think maybe this isn't really a bad idea to change in more places. It is a pretty big change, though.

andrewbanchich:
True, it would change a lot of code / docs. Should we ask others about this?

Also, if it is something we want to update, I guess we'd just slowly transition to this whenever we can? Or should we try to update everything at once?

jebrosen:
I think all examples, rustdocs, and the guide should be updated in a single release or not at all. A lot of the code is similar or even copy and pasted between these and it will be confusing for readers to see mixed editions.

Rocket itself may be forced to transition to 2018 if it will make use of `async/await`, and some of the examples may want to use `async` as well for demonstration purposes. For that reason, I'm inclined to wait for `async`ification before doing anything with this: an examples-edition upgrade and an `async` upgrade at the same time could make for some painful merges or rebases. If `async` is a long way off however, it might be a good idea to forge ahead with docs and examples now.

@SergioBenitez, do you have any thoughts on `async`ification, or any preference on if and when to change edition idioms in the docs and examples?

SergioBenitez:
> I think all examples, rustdocs, and the guide should be updated in a single release or not at all. A lot of the code is similar or even copy and pasted between these and it will be confusing for readers to see mixed editions.

Agreed.

> @SergioBenitez, do you have any thoughts on `async`ification, or any preference on if and when to change edition idioms in the docs and examples?

I personally prefer using `#[macro_use]` to import Rocket's attributes, for convenience reasons, and will continue to do so even in my `2018` projects. I'm not sure whether we should advocate one way or another in the documentation. Clearly using `use` is the sanctioned approach in `2018`, but I worry that it will confuse users, but that's an unsubstantiated concern.

To answer your question: I don't think we should redo all of the examples for such a small change. I think we should wait until bigger changes are needed, like tossing `async` in places or making examples in the guide testable.

As such, I'm closing this PR. Let's revisit when we're ready to make bigger changes.

Issue #880, opened 2019-01-08
Matching Url-Escaped Routes


PSeitz:
Rocket version: 0.4
OS: Macos/Ubuntu 14

Some clients url escape before calling a url, but the rocket matcher does not cover these.
E.g. for the url `/$world`, some clients will create `/%24world`.
As far as I could tell `%24world` is also explicitly disallowed, so there is no way to match the route.

```rust
#![feature(proc_macro_hygiene, decl_macro)]
#[macro_use] extern crate rocket;

#[get("/$world")]
fn world() -> &'static str {
    "Hello, $world!"
}

fn main() {
    rocket::ignite().mount("/", routes![world]).launch();
}

```

#### Behaviour in express.js and Spring
I checked the behaviour in express.js and Spring. The machers there cover both routes `/$world` and `/%24world`.

```javascript
var express = require('express');
var app = express();

app.get('/\\$world', function (req, res) {
  res.send('Hello $World!');
});
app.listen(3000, function () {
  console.log('Example app listening on port 3000!');
});
```

jebrosen:
Rocket should percent decode either during all URI parsing or during route comparison, both of which are very tricky to do correctly. Doing it during parsing would likely also mean changing the `Origin` URI retrieved via request guard, but I suspect most routes that do care about the request URI would prefer the decoded form.

PSeitz:
We could also decode the url if there is a '%' char in it and then start with the matching. This should work if we internally store only the original route.

SergioBenitez:
Let's work on this in #998.

PR #882, opened 2019-01-10
match percent decoded urls #880

PSeitz:
Fixes https://github.com/SergioBenitez/Rocket/issues/880

### DIFF
--- a/core/lib/src/request/request.rs
+++ b/core/lib/src/request/request.rs
@@ -13,7 +13,7 @@ use request::{FromFormValue, FormItems, FormItem};
 use rocket::Rocket;
 use router::Route;
 use config::{Config, Limits};
-use http::{hyper, uri::{Origin, Segments}};
+use http::{hyper, uri::{Uri, Origin, Segments}};
 use http::{Method, Header, HeaderMap, Cookies};
 use http::{RawStr, ContentType, Accept, MediaType};
 use http::private::{Indexed, SmallVec, CookieJar};
@@ -789,10 +789,15 @@ impl<'r> Request<'r> {
         h_addr: SocketAddr,
     ) -> Result<Request<'r>, String> {
         // Get a copy of the URI for later use.
-        let uri = match h_uri {
+        let mut uri = match h_uri {
             hyper::RequestUri::AbsolutePath(s) => s,
             _ => return Err(format!("Bad URI: {}", h_uri)),
         };
+
+        // URI decode the uri, if applicable
+        if uri.contains('%'){
+            uri = Uri::percent_decode_lossy(uri.as_bytes()).to_string();
+        }
 
         // Ensure that the method is known. TODO: Allow made-up methods?
         let method = match Method::from_hyp(&h_method) {


PSeitz:
I wanted to add a test for this, but the percent decoding is only in the hyper conversion and does not apply here.


```rust

#[get("/$world")]
fn world() -> &'static str {
    "Hello, $world!"
}

    rocket::ignite().mount("/", routes![world])
}

mod tests {
    use super::*;
    use rocket::local::Client;
    use rocket::http::{Status};

    #[test]
    fn uri_get_original_route() {
        let client = Client::new(rocket()).unwrap();
        let mut response = client.get("/$world").dispatch();
        assert_eq!(response.body_string().unwrap(), format!("Hello, $world!"));
    }

    #[test]
    fn uri_get_url_encoded_route() {
        let client = Client::new(rocket()).unwrap();
        let mut response = client.get("/%24world").dispatch();
        assert_eq!(response.body_string().unwrap(), format!("Hello, $world!"));
    }
}
```

PSeitz:
Tests fail, because of unrelated rustup issue: https://github.com/rust-lang/rust/issues/57488

jebrosen:
This will need tests for sure if it's to be accepted. Even then, I'm hesitant to review any changes that might be rendered obsolete by the hyper upgrade so I'll wait to hear from @SergioBenitez on this.

SergioBenitez:
We definitely need tests. This, or anything like this, won't be accepted without them.

@PSeitz Do those tests pass now? I'd wager no, and that's a concern: we can't have the local client act differently than the "real" client, otherwise testing can't be trusted. In other words, we need to ensure that whatever fix we apply here applies universally.

jebrosen:
> @jebrosen I think whatever we do here we'll need to do when we upgrade hyper, so there's no harm in accepting this now, as far as I can see.

I meant to imply that hyper might even do this itself in newer versions, and that I'd rather we look carefully at the URL handling before implementing it.

That being said, it's a relatively small change in terms of code and any tests written now should apply in the future as well.

SergioBenitez:
> I meant to imply that hyper might even do this itself in newer versions, and that I'd rather we look carefully at the URL handling before implementing it.

I'd consider Hyper "wrong" if it did; it shouldn't be interpreting URIs in any way, given where it wants to sit in the stack. If it does, we should be able to work around it.

PSeitz:
@SergioBenitez I rechecked, but I couldn't find a common entry for the tests and hyper. `from_hyp` uses `Origin::parse_owned(string: String)`, the tests with local request use `Origin::parse(string: &'a str)`, meaning the tests would need to replicate the behaviour.

https://github.com/SergioBenitez/Rocket/blob/5ce43ed4e97261272bbe7d506c534750db71aa32/core/lib/src/local/request.rs#L402-L404

PSeitz:
Actually `Origin::parse(string: &'a str)` is a common entry point, but I can't apply the escaping there because of lifetime issues.

SergioBenitez (commented):
While taking a look at resolving the underlying issue, I took a look at this PR. Unfortunately this is incorrect. Decoding needs to be applied _after_ all delimiter-based splits have been applied or else risk introducing new splits. For example, `/a/b%2Fc/d` should result in being decoded in a path with 3 components `a`, `b/c`, `d`, but this PR would result in a path with 4 components `a`, `b`, `c`, and `d`. The same issue occurs with `?` and with `&` and `=` in query parts.

This is incredibly tricky to get right _and_ make perform well. I'm working now on a fix to this issue.

PR #885, opened 2019-01-12
Correctly quote version number in sample Cargo.toml.

austinhartzheim:
The version number needs to be quoted in `Cargo.toml`. This PR fixes an example in the docs that wasn't quoted.


jebrosen:
Thanks! Merged in a88bc952 along with a few stragglers that were still specified as `"*"`.

PR #887, opened 2019-01-14
Add support for Range requests

### LINKED ISSUE #806: [Request] - Support Range header for streaming media to Safari

Currently you can use Rocket to serve static assets such as Images, JS files, etc easily and it works, however if you try to serve video media (and potentially audio, haven't tested audio), it will only work in some browsers, because of browser support and leniency on the Range and RangeByte headers being respected and the correct information being returned.

Currently, it seems to send all bytes at the same time, even if the Range requested was only bytes 0-1 (which is the header request made by Safari to determine media type).

If you return all content back on that request, Chrome and other browsers are happy to play that content and ignore the mistake, however Safari refuses to play the content.

Using Rocket to serve static movie assets as part of your website is not currently possible, it currently requires either third-party (S3) or alternate server software for serving that asset, which isn't desirable if you want to lock the content behind auth walls etc in your application.

It can exist outside of Rocket, however it makes using Rocket as a static server unusable since a common type of asset is not able to be sent to a popular browser (Mobile safari + Desktop).

Drakulix:
This is an attempt to fix #806 (at least partially).

This transparently handles `Range` requests for `&[u8]`, `Vec<u8>`, `File` and `NamedFile`, when the total size is known, allowing partial downloads.

It also advertises this feature for the listed types by adding an `AcceptRanges` header.

The implementation follows the [MDN examples](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests) and the corresponding [RFC](https://tools.ietf.org/html/rfc7233), but only supports `Single part ranges`, which is compliant with the RFC, as a server may choose to ignore or reject any range request.

`Multipart ranges` would most likely be better implemented as a custom `Body` variant and would be more complicated. I have no use case for those, which is why I did not bother, but if that would be blocking and someone can provide some directions/suggestion on how to implement those (I am not very familiar with rocket's codebase), I would try to tackle that as well. But I would strongly prefer to do that in a separate PR.

I am looking for some initial feedback on the approach. I did implement it in the corresponding `Responder` implementations to have access to the `Range` header of the request and the response as well. It seemed fitting. But it would probably also be possible to just parse the header in the `Responder` and just store that in the `Response` or `Body` and deal with it later in the chain (when writing out the `Response` for example).

Docs and Tests are completely lacking, also formatting was not working well, `cargo fmt` changed the whole codebase. So I would appreciate some more comments on what I should provide on those subjects as well.

Thanks for your time :)

### DIFF
--- a/core/lib/src/response/responder.rs
+++ b/core/lib/src/response/responder.rs
@@ -272,6 +272,8 @@ impl<'r, B: io::Seek + io::Read + 'r> Responder<'r> for RangeResponder<B> {
 
                         let (start, end) = match ranges[0] {
                             ByteRangeSpec::FromTo(start, mut end) => {
+                                // make end exclusive
+                                end += 1;
                                 if end > size {
                                     end = size;
                                 }
@@ -304,7 +306,8 @@ impl<'r, B: io::Seek + io::Read + 'r> Responder<'r> for RangeResponder<B> {
                             .status(Status::PartialContent)
                             .header(AcceptRanges(vec![RangeUnit::Bytes]))
                             .header(ContentRange(ContentRangeSpec::Bytes {
-                                range: Some((start, end)),
+                                // make end inclusive again
+                                range: Some((start, end - 1)),
                                 instance_length: Some(size),
                             }))
                             .raw_body(Body::Sized(body, end - start))
--- a/core/lib/src/response/responder.rs
+++ b/core/lib/src/response/responder.rs
@@ -319,7 +319,11 @@ impl<'r> Responder<'r> for Vec<u8> {
 /// Returns a response with a sized body for the file. Always returns `Ok`.
 impl<'r> Responder<'r> for File {
     fn respond_to(self, req: &Request) -> response::Result<'r> {
-        RangeResponder(BufReader::new(self)).respond_to(req)
+        let (metadata, file) = (self.metadata(), BufReader::new(self));
+        match metadata {
+            Ok(_) => RangeResponder(file).respond_to(req),
+            Err(_) => Response::build().streamed_body(file).ok()
+        }        
     }
 }
 
--- a/core/lib/src/response/responder.rs
+++ b/core/lib/src/response/responder.rs
@@ -1,8 +1,10 @@
 use std::fs::File;
-use std::io::{Cursor, BufReader};
+use std::io::{self, Cursor, BufReader};
 use std::fmt;
+use std::str::FromStr;
 
-use http::{Status, ContentType, StatusClass};
+use http::{Status, ContentType, StatusClass, Method};
+use http::hyper::header::{AcceptRanges, Range, RangeUnit};
 use response::{self, Response, Body};
 use request::Request;
 
@@ -217,13 +219,100 @@ impl<'r> Responder<'r> for String {
     }
 }
 
+fn try_range_response<'r, B: io::Read + io::Seek + 'r>(req: Request<'r>, mut body: B) -> Result<response::Response<'r>, B> {
+    use std::cmp;
+    use http::hyper::header::{ContentRange, ByteRangeSpec, ContentRangeSpec};
+
+    //  A server MUST ignore a Range header field received with a request method other than GET.
+    if req.method() == Method::Get {
+        return Err(body);
+    }
+     
+    let range = req.headers().get_one("Range").and_then(Range::from_str).map_err(|_| body)?;
+         
+    match range {
+        Range::Bytes(ranges) => {
+            if ranges.len() == 1 {
+                let size = body.seek(io::SeekFrom::End(0))
+                    .expect("Attempted to retrieve size by seeking, but failed.");
+                
+                let (start, end) = match ranges[0] {
+                    ByteRangeSpec::FromTo(mut start, mut end) => {
+                        if end < start {
+                            return Ok(
+                                Response::build()
+                                    .status(Status::RangeNotSatisfiable)
+                                    .header(AcceptRanges(vec![RangeUnit::Bytes]))
+                                    .finalize()
+                            );
+                        }
+                        if start > size {
+                            start = size;
+                        }
+                        if end > size {
+                            end = size;
+                        }
+                        (start, end)
+                    },
+                    ByteRangeSpec::AllFrom(mut start) => {
+                        if start > size {
+                            start = size;
+                        }
+                        (start, size - start)
+                    },
+                    ByteRangeSpec::Last(len) => {
+                        // we could seek to SeekFrom::End(-len), but if we reach a value < 0, that is an error.
+                        // but the RFC reads:
+                        //      If the selected representation is shorter than the specified
+                        //      suffix-length, the entire representation is used.
+                        let start = cmp::max(size - len, 0);
+                        (start, size - start)
+                    }
+                };
+
+                body.seek(io::SeekFrom::Start(start))
+                    .expect("Attempted to seek to the start of the requested range, but failed.");
+
+                return Ok(
+                    Response::build()
+                        .status(Status::PartialContent)
+                        .header(AcceptRanges(vec![RangeUnit::Bytes]))
+                        .header(ContentRange(ContentRangeSpec::Bytes {
+                            range: Some((start, end)),
+                            instance_length: Some(end - start),
+                        }))
+                        .raw_body(Body::Sized(body, end - start))
+                        .finalize()
+                )
+            }
+            // A server MAY ignore the Range header field.
+        },
+        // An origin server MUST ignore a Range header field that contains a
+        // range unit it does not understand.
+        Range::Unregistered(_, _) => {},
+    };
+
+    Err(body)
+}
+
 /// Returns a response with Content-Type `application/octet-stream` and a
 /// fixed-size body containing the data in `self`. Always returns `Ok`.
 impl<'r> Responder<'r> for &'r [u8] {
-    fn respond_to(self, _: &Request) -> response::Result<'r> {
+    fn respond_to(self, req: &Request) -> response::Result<'r> {
+        let mut body = Cursor::new(self);
+        
+        match try_range_response(req, body) {
+            Ok(mut resp) => {
+                resp.set_header(ContentType::Binary);
+                return Ok(resp)
+            },
+            Err(old_body) => body = old_body,
+        };
+
         Response::build()
             .header(ContentType::Binary)
-            .sized_body(Cursor::new(self))
+            .header(AcceptRanges(vec![RangeUnit::Bytes]))
+            .sized_body(body)
             .ok()
     }
 }
@@ -231,20 +320,41 @@ impl<'r> Responder<'r> for &'r [u8] {
 /// Returns a response with Content-Type `application/octet-stream` and a
 /// fixed-size body containing the data in `self`. Always returns `Ok`.
 impl<'r> Responder<'r> for Vec<u8> {
-    fn respond_to(self, _: &Request) -> response::Result<'r> {
+    fn respond_to(self, req: &Request) -> response::Result<'r> {
+        let mut body = Cursor::new(self);
+
+        match try_range_response(req, body) {
+            Ok(mut resp) => {
+                resp.set_header(ContentType::Binary);
+                return Ok(resp)
+            },
+            Err(old_body) => body = old_body,
+        };
+
         Response::build()
             .header(ContentType::Binary)
-            .sized_body(Cursor::new(self))
+            .header(AcceptRanges(vec![RangeUnit::Bytes]))
+            .sized_body(body)
             .ok()
     }
 }
 
 /// Returns a response with a sized body for the file. Always returns `Ok`.
 impl<'r> Responder<'r> for File {
-    fn respond_to(self, _: &Request) -> response::Result<'r> {
-        let (metadata, file) = (self.metadata(), BufReader::new(self));
+    fn respond_to(self, req: &Request) -> response::Result<'r> {
+        let (metadata, mut file) = (self.metadata(), BufReader::new(self));
         match metadata {
-            Ok(md) => Response::build().raw_body(Body::Sized(file, md.len())).ok(),
+            Ok(md) => {
+                match try_range_response(req, body) {
+                    Ok(mut resp) => return Ok(resp),
+                    Err(old_body) => body = old_body,
+                };
+
+                Response::build()
+                .header(AcceptRanges(vec![RangeUnit::Bytes]))
+                .raw_body(Body::Sized(file, md.len()))
+                .ok()
+            },
             Err(_) => Response::build().streamed_body(file).ok()
         }
     }
--- a/core/lib/src/response/responder.rs
+++ b/core/lib/src/response/responder.rs
@@ -275,7 +275,7 @@ impl<'r, B: io::Seek + io::Read + 'r> Responder<'r> for RangeResponder<B> {
                             .header(AcceptRanges(vec![RangeUnit::Bytes]))
                             .header(ContentRange(ContentRangeSpec::Bytes {
                                 range: Some((start, end)),
-                                instance_length: Some(end - start),
+                                instance_length: Some(size),
                             }))
                             .raw_body(Body::Sized(body, end - start))
                             .ok()
--- a/core/lib/src/response/responder.rs
+++ b/core/lib/src/response/responder.rs
@@ -220,6 +220,40 @@ impl<'r> Responder<'r> for String {
     }
 }
 
+/// A special responder type, that may be used by custom [`Responder`][response::Responder]
+/// implementation to wrap any type implementing [`io::Seek`](std::io::Seek) and [`io::Read`](std::io::Read)
+/// to support [range requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests).
+///
+/// # **Note:**
+/// `&[u8]`, `Vec<u8>`, [`File`](std::fs::File) and [`NamedFile`](response::NamedFile)
+/// are implemented using `RangeResponder`. There is no need to wrap these types,
+/// they do support range requests out of the box.
+///
+/// # Example
+/// ```rust
+/// # #![feature(proc_macro_hygiene, decl_macro)]
+/// # #[macro_use] extern crate rocket;
+/// #
+/// # struct CustomIoType;
+/// # impl std::io::Read for CustomIoType {
+/// #     fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> { Ok(0) }
+/// # }
+/// # impl std::io::Seek for CustomIoType {
+/// #     fn seek(&mut self, pos: std::io::SeekFrom) -> std::io::Result<u64> { Ok(0) }
+/// # }
+/// #
+/// use rocket::request::Request;
+/// use rocket::response::{self, Response, Responder, RangeResponder};
+/// use rocket::http::ContentType;
+///
+/// impl<'r> Responder<'r> for CustomIoType {
+///     fn respond_to(self, req: &Request) -> response::Result<'r> {
+///         Response::build_from(RangeResponder(self).respond_to(req)?)
+///             .header(ContentType::Binary)
+///             .ok()
+///     }
+/// }
+/// ```
 pub struct RangeResponder<B: io::Seek + io::Read>(pub B);
 
 impl<'r, B: io::Seek + io::Read + 'r> Responder<'r> for RangeResponder<B> {
--- a/core/lib/src/response/responder.rs
+++ b/core/lib/src/response/responder.rs
@@ -271,19 +271,13 @@ impl<'r, B: io::Seek + io::Read + 'r> Responder<'r> for RangeResponder<B> {
                             .expect("Attempted to retrieve size by seeking, but failed.");
 
                         let (start, end) = match ranges[0] {
-                            ByteRangeSpec::FromTo(mut start, mut end) => {
-                                if start > size {
-                                    start = size;
-                                }
+                            ByteRangeSpec::FromTo(start, mut end) => {
                                 if end > size {
                                     end = size;
                                 }
                                 (start, end)
                             },
-                            ByteRangeSpec::AllFrom(mut start) => {
-                                if start > size {
-                                    start = size;
-                                }
+                            ByteRangeSpec::AllFrom(start) => {
                                 (start, size)
                             },
                             ByteRangeSpec::Last(len) => {
@@ -295,6 +289,13 @@ impl<'r, B: io::Seek + io::Read + 'r> Responder<'r> for RangeResponder<B> {
                                 (start, size)
                             }
                         };
+
+                        if start > size {
+                            return Response::build()
+                                .status(Status::RangeNotSatisfiable)
+                                .header(AcceptRanges(vec![RangeUnit::Bytes]))
+                                .ok()
+                        }
 
                         body.seek(io::SeekFrom::Start(start))
                             .expect("Attempted to seek to the start of the requested range, but failed.");
--- a/core/lib/src/response/responder.rs
+++ b/core/lib/src/response/responder.rs
@@ -228,7 +228,7 @@ impl<'r, B: io::Seek + io::Read + 'r> Responder<'r> for RangeResponder<B> {
 
         let mut body = self.0; 
         //  A server MUST ignore a Range header field received with a request method other than GET.
-        if req.method() != Method::Get {
+        if req.method() == Method::Get {
             let range = req.headers().get_one("Range").and_then(|x| Range::from_str(x).ok());
             match range {
                 Some(Range::Bytes(ranges)) => {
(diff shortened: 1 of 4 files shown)


core/lib/src/response/responder.rs
```diff
@@ -217,34 +219,156 @@ impl<'r> Responder<'r> for String {
...
+                    },
+                    ByteRangeSpec::AllFrom(mut start) => {
+                        let size = body.seek(io::SeekFrom::End(0))
+                            .expect("Attempted to retrieve size by seeking, but failed.");
+                        if start > size {
+                            start = size;
+                        }
+                        body.seek(io::SeekFrom::Start(start))
+                            .expect("Attempted to seek to the start of the requested range, but failed.");
+                        (start, size, size)
```
jebrosen: Should this be `(start, size - start, size)` ?

core/lib/src/response/responder.rs
```diff
@@ -217,34 +219,156 @@ impl<'r> Responder<'r> for String {
...
 }
 
+fn range_response<'r, B: io::Read + io::Seek + 'r>(range: Range, mut body: B) -> Result<response::Response<'r>, B> {
+    use std::cmp;
+    use http::hyper::header::{ContentRange, ByteRangeSpec, ContentRangeSpec};
+
+    match range {
+        Range::Bytes(ranges) => {
+            if ranges.len() == 1 {
+                let (start, end, total) = match ranges[0] {
```
jebrosen: Is `total` every anything other than `end - start`?

jebrosen: These match arms duplicate a decent amount of code (e.g. `let size = `, calling `seek()`). I think It will be easier to see the differences between the blocks if the common code was pulled of the match arms.

core/lib/src/response/responder.rs
```diff
@@ -217,34 +219,156 @@ impl<'r> Responder<'r> for String {
...
+                        if end < start {
+                            return Ok(
+                                Response::build()
+                                    .status(Status::RangeNotSatisfiable)
+                                    .header(AcceptRanges(vec![RangeUnit::Bytes]))
+                                    .finalize()
+                            );
+                        }
+                        let size = body.seek(io::SeekFrom::End(0))
+                            .expect("Attempted to retrieve size by seeking, but failed.");
```
jebrosen: I'm uncomfortable with this `expect`. I think it's fine for `Vec<u8>` and `[u8]` for example, but I'm not so sure about `File`.

core/lib/src/response/responder.rs
```diff
@@ -217,34 +219,156 @@ impl<'r> Responder<'r> for String {
...
+
 /// Returns a response with Content-Type `application/octet-stream` and a
 /// fixed-size body containing the data in `self`. Always returns `Ok`.
 impl<'r> Responder<'r> for &'r [u8] {
-    fn respond_to(self, _: &Request) -> response::Result<'r> {
+    fn respond_to(self, req: &Request) -> response::Result<'r> {
+        let mut body = Cursor::new(self);
+
+        //  A server MUST ignore a Range header field received with a request method other than GET.
+        if req.method() == Method::Get {
```
jebrosen: I think most of this code here could be broken out into a single function, maybe called `try_range_response`.

core/lib/src/response/responder.rs
Drakulix: How would you suggest to "fix" this?

To me this in an "Internal Server Error", because the state of the seek pointer is unknown at this point, which means we cannot fallback anymore, because this will also have an impact on `Read`.
Rocket would catch this (unlikely) panic anyway, but I guess I could create a 500-response instead and return that.

core/lib/src/response/responder.rs
Drakulix: fixed

core/lib/src/response/responder.rs
Drakulix: right, removed.

Drakulix: I have pulled out both seek calls.

core/lib/src/response/responder.rs
Drakulix: I have moved more of into `try_range_response`.

Drakulix:
@jebrosen Thanks for taking a quick look at it. I tried to address most of your concerns.

Another Responder similar to how `Stream` works today would also be possible. My reasoning was, that you probably want this behavior for any binary-data anyway. The only downside is, that it is not possible to make use of this functionality with any user-provided types, while theoretically any `Body` implementing `Seek` is applicable. But then again I cannot think of much more useful types implementing `Seek` then files and in-memory buffers.

Providing this functionality for all bodies on the other side might not be very helpful, but would introduce a good amount of additional complexity. (E.g. I cannot think of a good reason to request a certain byte-range of a json record. Most of the time this should be solved at the user-level.)

core/lib/src/response/responder.rs
jebrosen: Generally, I prefer to return some kind of `Error` directly rather than panicking. Although looking at it, `Response::set_sized_body` itself panics instead of returning `Result`.

I looked at a few of the platform implementations of `Seek` for `File`. It looks like *nix and Windows only panic when the offset is out of range for an `i64`, or for any platform-specific reasons (e.g. the `File` object is in reality a handle to a pipe or socket, not a file).

My original thinking was that in the case of a "seek to get size" error we could just seek back to the start, ignoring the range request, and/or return a range not satisfiable. But I'm not sure if that's really worth the complexity.

Drakulix: If one seek already has failed, it is very likely, that another to the beginning will fail as well.

The current approach is my favorite, but I if you would like to see something else, just tell me.

Drakulix:
This is starting to collect some dust, but I would still like to see this merged someday in any shape or form. Could I get another round of feedback please? Ping @jebrosen or @SergioBenitez

core/lib/src/response/responder.rs
```diff
@@ -217,34 +219,142 @@ impl<'r> Responder<'r> for String {
     }
 }
 
+fn try_range_response<'r, B: io::Read + io::Seek + 'r>(req: Request<'r>, mut body: B) -> Result<response::Response<'r>, B> {
```
jebrosen: This signature does feel really awkward, but that's mostly because of the `Seek` bound, right?

core/lib/src/response/responder.rs
```diff
@@ -217,34 +219,142 @@ impl<'r> Responder<'r> for String {
...
+}
+
 /// Returns a response with Content-Type `application/octet-stream` and a
 /// fixed-size body containing the data in `self`. Always returns `Ok`.
 impl<'r> Responder<'r> for &'r [u8] {
-    fn respond_to(self, _: &Request) -> response::Result<'r> {
+    fn respond_to(self, req: &Request) -> response::Result<'r> {
+        let mut body = Cursor::new(self);
+        
+        match try_range_response(req, body) {
```
jebrosen: Both of the `respond_to` methods are harder to read than without `Range` support, mostly because of that awkward function signature. I might play around with this and try to make things a bit nicer.

core/lib/src/response/responder.rs
```diff
@@ -217,34 +219,142 @@ impl<'r> Responder<'r> for String {
     }
 }
 
+fn try_range_response<'r, B: io::Read + io::Seek + 'r>(req: Request<'r>, mut body: B) -> Result<response::Response<'r>, B> {
+    use std::cmp;
+    use http::hyper::header::{ContentRange, ByteRangeSpec, ContentRangeSpec};
```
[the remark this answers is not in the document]
Drakulix: Is `typed-headers` an option for Rocket / are you planning to use this crate? In that case I might try to contribute an implementation for these headers in `typed-headers` instead of adding this directly to Rocket.

core/lib/src/response/responder.rs
Drakulix: Not really because of Seek (this implementation panics, if seeking fails), but to provide a way to recover of other errors. I could move those checks into `respond_to` instead. That would likely fix the awkwardness, but result in some code-duplication.

Drakulix:
> I would still love to provide a way for downstream crates to leverage this, but I'm not sure where exactly to put that. I think it would be possible to do struct RangeResponder<B: Read + Seek>(B); impl<'r, B: Read + Seek> Responder for RangeResponder, and that might also reduce some of the awkwardness I mention below.

While this likely would look a lot cleaner, it would duplicate a lot of code and I would like to make this transparent. There is no reason not to provide range-support for types implementing Seek, because this depends on client support anyway and is 100% backwards compatible.

core/lib/src/response/responder.rs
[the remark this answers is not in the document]
jebrosen: It looks like `RangeUnit` and `AcceptRanges` were previously in `typed-headers` and removed, but `Range` was never present. I think I'm okay with using hyper 0.10's typed headers for now, and crossing that bridge when we come to it.

core/lib/src/response/responder.rs
jebrosen: I should have explained that more clearly: I found it awkward because this isn't something like `Response` -> `Response`, but instead this function requires the caller to move the body in and out. That's why I was hoping a `Responder` bounded on `Seek` would make that go away.

jebrosen:
My idea was to replace the entire `try_range_request` function with the hypothetical `RangeResponder` -- it shouldn't duplicate any code, and it should make both `respond_to` methods shorter. If it worked, we could write this instead:

```rust
Response::build_from(RangeResponse(self))
    .header(ContentType::Binary)
    .ok()
```

That way `RangeResponder` would be responsible for both adding the `AcceptRanges` header and the fallback to the whole body (reducing the duplication in `respond_to`), and it would be easier to reuse and maybe even make it public. I'm happy to try this route myself, but it will be a while before I can seriously focus on it.

---

Aside from code, we will also need:

* Tests
* Documentation. I think this feature would be covered with a simple "this responder supports range requests" under https://api.rocket.rs/v0.4/rocket/response/trait.Responder.html#provided-implementations

Drakulix:
> My idea was to replace the entire `try_range_request` function with the hypothetical `RangeResponder` -- it shouldn't duplicate any code, and it should make both `respond_to` methods shorter. If it worked, we could write this instead:
> 
> ```rust
> Response::build_from(RangeResponse(self))
>     .header(ContentType::Binary)
>     .ok()
> ```
> 
> That way `RangeResponder` would be responsible for both adding the `AcceptRanges` header and the fallback to the whole body (reducing the duplication in `respond_to`), and it would be easier to reuse and maybe even make it public. I'm happy to try this route myself, but it will be a while before I can seriously focus on it.
> 

Alright, I have made some improvements:

- The first commit adds a new `range_response` function, that handles everything without returning the body, utilizing `Response::build_from` (I was not aware of this function, thanks. With this change alone it looks much clearer now.)
- The second commit adds the RangeResponder as a public struct as you were suggesting. I am not totally sure, that this is a good idea. People might get the impression that File/&[u8]/Vec<u8> do not support Range requests by default. But I think this is solvable with good docs, so everybody is aware that `RangeResponder` is a helper struct for new downstream implementations.

But I have also introduced a new issue in the process. The file handler is now much simpler, but not equivalent. It previously used `metadata` to construct a Sized body manually and otherwise construct a Streaming body. The new implementation does always use Seek to determine the size (because it just uses sized_body, if no Range is requested).

I do not know, why the old code, did not use Seek as well, I assume that this was some kind of optimization? I also assume, that it is an oversight in that case, that `sized_body` of the ResponseBuilder uses a File in the [documentation example](https://api.rocket.rs/v0.4/rocket/response/struct.ResponseBuilder.html#method.sized_body)? Because otherwise, the docs would encourage a different implementation than what Rocket does by default, which seems a little paradox?

I am not sure, there is a good way to restore the old implementation when implementing range responses like this. At least no good idea jumps to my mind. But we definitely want range responses for files by default in my opinion.

jebrosen:
Note that `RangeResponder` can be implemented outside of Rocket itself. It's my opinion that support for `Range` is nonetheless important enough to include for `File` and buffers (privately or publicly), and if not in rocket itself then in rocket_contrib - especially for StaticFiles.

core/lib/src/response/responder.rs
```diff
@@ -217,36 +219,108 @@ impl<'r> Responder<'r> for String {
...
+
+                        body.seek(io::SeekFrom::Start(start))
+                            .expect("Attempted to seek to the start of the requested range, but failed.");
+
+                        return Response::build()
+                            .status(Status::PartialContent)
+                            .header(AcceptRanges(vec![RangeUnit::Bytes]))
+                            .header(ContentRange(ContentRangeSpec::Bytes {
+                                range: Some((start, end)),
+                                instance_length: Some(end - start),
```
jebrosen: `Some(size)`

core/lib/src/response/responder.rs
```diff
@@ -217,36 +219,108 @@ impl<'r> Responder<'r> for String {
...
+                                if end > size {
+                                    end = size;
+                                }
+                                (start, end)
+                            },
+                            ByteRangeSpec::AllFrom(mut start) => {
+                                if start > size {
+                                    start = size;
+                                }
+                                (start, size - start)
```
jebrosen: `(start, size)`

core/lib/src/response/responder.rs
```diff
@@ -217,36 +219,108 @@ impl<'r> Responder<'r> for String {
...
+                            },
+                            ByteRangeSpec::Last(len) => {
+                                // we could seek to SeekFrom::End(-len), but if we reach a value < 0, that is an error.
+                                // but the RFC reads:
+                                //      If the selected representation is shorter than the specified
+                                //      suffix-length, the entire representation is used.
+                                let start = cmp::max(size - len, 0);
+                                (start, size - start)
```
jebrosen: `(start, size)`

core/lib/src/response/responder.rs
```diff
@@ -217,36 +219,108 @@ impl<'r> Responder<'r> for String {
...
+pub struct RangeResponder<B: io::Seek + io::Read>(pub B);
+
+impl<'r, B: io::Seek + io::Read + 'r> Responder<'r> for RangeResponder<B> {
+    fn respond_to(self, req: &Request) -> response::Result<'r> {
+        use std::cmp;
+        use http::hyper::header::{ContentRange, ByteRangeSpec, ContentRangeSpec};
+
+        let mut body = self.0; 
+        //  A server MUST ignore a Range header field received with a request method other than GET.
+        if req.method() != Method::Get {
```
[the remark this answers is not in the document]
Drakulix: fixed

core/lib/src/response/responder.rs
```diff
@@ -217,36 +219,108 @@ impl<'r> Responder<'r> for String {
...
+                                    start = size;
+                                }
+                                (start, size - start)
+                            },
+                            ByteRangeSpec::Last(len) => {
+                                // we could seek to SeekFrom::End(-len), but if we reach a value < 0, that is an error.
+                                // but the RFC reads:
+                                //      If the selected representation is shorter than the specified
+                                //      suffix-length, the entire representation is used.
+                                let start = cmp::max(size - len, 0);
```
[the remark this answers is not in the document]
Drakulix: fixed

core/lib/src/response/responder.rs
Drakulix: fixed

core/lib/src/response/responder.rs
Drakulix: fixed

core/lib/src/response/responder.rs
Drakulix: fixed

Drakulix:
@jebrosen I have:
- fixed all errors you pointed out
- discovered two more while writing tests (and fixed them)
- wrote tests
- wrote docs
- made `File` use `Stream` again, when `metadata` fails
- `RangeResponder` is now public

and travis now also seems to be happy.

Time for another review, I guess?

> Side note: hyper 0.12 removes typed headers, so in that porting process we will probably have to revive the header parsing code from the last time Sergio worked on this in the range branch.

Is this something I should take care of? Or is that just a note for whoever will work on the hyper upgrade?

jebrosen:
Awesome! I will try this out.

In addition to `&[u8]` `Vec<u8>` and `File`, any `Responder` that uses them as the underlying implementation automatically gets range request support. This includes at least `NamedFile` and `MsgPack`. I am not sure that automatically adding support to `&[u8]` and `Vec<u8>` are the right thing to do, especially as it's not possible to opt out. It might be expensive to regenerate the response data and/or likely to change between requests, which are both undesirable with Range request handling.

> Is this something I should take care of? Or is that just a note for whoever will work on the hyper upgrade?

Just something to be handled during the hyper upgrade.

core/lib/src/response/responder.rs
```diff
@@ -217,34 +220,144 @@ impl<'r> Responder<'r> for String {
...
+            let range = req.headers().get_one("Range").map(|x| Range::from_str(x));
+            match range {
+                Some(Ok(Range::Bytes(ranges))) => {
+                    if ranges.len() == 1 {
+                        let size = body.seek(io::SeekFrom::End(0))
+                            .expect("Attempted to retrieve size by seeking, but failed.");
+
+                        let (start, end) = match ranges[0] {
+                            ByteRangeSpec::FromTo(mut start, mut end) => {
+                                if start > size {
```
jebrosen: If we add `end += 1` before the `start > size` check, then `end` will be an exclusive position in all branches.

core/lib/src/response/responder.rs
```diff
@@ -217,34 +220,144 @@ impl<'r> Responder<'r> for String {
...
+                                    start = size;
```
jebrosen: If `start > size`, we should return 416 or 200 instead (see https://tools.ietf.org/html/rfc7233#section-4.4).

core/lib/src/response/responder.rs
```diff
@@ -217,34 +220,144 @@ impl<'r> Responder<'r> for String {
...
+                                }
+                                if end > size {
+                                    end = size;
+                                }
+                                (start, end)
+                            },
+                            ByteRangeSpec::AllFrom(mut start) => {
+                                if start > size {
+                                    start = size;
```
jebrosen: Same here with `start > size`.

core/lib/src/response/responder.rs
```diff
@@ -217,34 +220,144 @@ impl<'r> Responder<'r> for String {
...
+                        };
+
+                        body.seek(io::SeekFrom::Start(start))
+                            .expect("Attempted to seek to the start of the requested range, but failed.");
+
+                        return Response::build()
+                            .status(Status::PartialContent)
+                            .header(AcceptRanges(vec![RangeUnit::Bytes]))
+                            .header(ContentRange(ContentRangeSpec::Bytes {
+                                range: Some((start, end)),
```
jebrosen: At this point `end` is exclusive, so changing to `end-1` converts it to an inclusive position as the client wants.

Drakulix:
Should all be fixed. @jebrosen 

Thanks for all the testing and debugging. My use-case for this PR is long gone, I hope this gets merged soon. :smile:

Drakulix:
Forgot to fix up the tests. Hopefully travis is now happy again.

Drakulix:
@jebrosen Ping. Just in case you might have forgotten this PR. No rush.

Drakulix:
> I _think_ the only correct place for range request processing is as the last ("outermost") responder that runs, and this PR as written would break both `Compress<File>`, available in `rocket_contrib`, and `ETag<File>`, which isn't in rocket or `rocket_contrib` but implementations exist. I just wish I had realized this problem sooner :/

Okay, this sucks (but it is probably better to think about it now, rather than after the fact).
So we basically **cannot** do this transparently, but rather need to be able to do something like `Range<Compress<File>>` and add documentation to `Range` that it needs to be outmost?

Or is there a way to enforce the last constraint? Maybe something like a new trait `FinalResponder`, which can be returned by a `route` and is already implemented for every `Responder`, but cannot be wrapped by another responder, because those only accept the `Responder`-trait (not `FinalResponder`)??

This sounds really cumbersome.

jebrosen:
The problem is precisely with having `File` (or anything else) handle `Range` automatically - as it breaks every subsequent fairing or wrapping `Responder` that cares about the body.

The obvious solutions to my mind are a) making it strictly opt in or b) apply it automatically to `File`, in both cases explaining the ordering problem. Fairings and Responders that would modify the body should not operate when `Content-Range` header has been set. One significant downside to b) is that every implementer of `Responder` or `on_response` needs to be aware of this behavior.

Option c) is to do this in rocket itself automatically, but that will be a lot more work, possibly changing `Body` itself, to carry the `Seek` requirement all the way to response time. I haven't checked whether async will make this a terrible fit either (i.e. if `AsyncSeek` exists and/or how `AsyncRead` and `Seek` interact)

Option d) is something like `FinalResponder`, and I have a gut feeling that such an API wouldn't be fun to implement or use

Drakulix:
I think we are not doing anyone a favor with Option **b**.

I can either quickly change this PR to do Option **a** and then tool around with **d** (if rustc is able to infer a lot of this mess, it might be reasonable to use and just a little painful to implement)

or just go for **c**, but I have no idea about that async-stuff.

Do you have a preference for either of those? I think both have their cons and pros, but I would definitely try to tackle them. Time is not an issue, I rather do this right then to get this merged quickly.

jebrosen:
For now I think option **a** is the best choice, sadly. Unless a design for option **c** is relatively trivial (and I think it wouldn't be), I would like to avoid the headache of rewriting it again so soon for async.

The conundrum of order-sensitive `Responder`s seriously bothers me, so I'm going to spend some time thinking on it this weekend.

agersant:
@jebrosen Bumping this because I'm very interested and activity on this PR has died down.

jebrosen:
So I did write [a blog post with a sad conclusion](https://www.jebrosen.com/blog/1561262184/adventures-in-response-composition), but never really publicized or linked it.

I still think "Option a", the opt-in wrapping Responder along with clear documentation explaining potential pitfalls, is the right way to go. So far having range requests be done automatically for all responses (or even just `File`) feels like "too much" magic that can have risky interactions with other Responders or fairings.

agersant:
Oh, thank you for linking this - I had missed it in my searches.

As an end user, I would be very content with option A. It would match exactly what I've been doing except without having to maintain an extra [200 lines of code](https://github.com/agersant/polaris/blob/master/src/service/rocket/serve.rs) for the one endpoint on which I want to support range requests.

Drakulix:
Just a quick note, because I have stumbled on this (rather dead) PR again:
My previous iterations on this feature were motivated by work for my previous employer. While I would like to finish this, I do not have the free time to do it (at least right now). While I appreciate the very thoughtful review process, this simply took too much time.

I am happy to answer questions about the current implementation, just ping me. But other then that, I will not further participate. I encourage anyone reading through the pull request to continue, were we left of. All the work and discussion is public and collected at this place. I hope this feature will be finished some time in the future. Good luck to everyone involved. :)

richo:
Hi! I need to support a Range request in my application, so I'm basically trying to do the math on whether a cobbled implementation for one route in my application is the right move, or whether I could carry this PR across the line.

I've just read the discussion above (And thankyou @Drakulix for most of an implementation) and my understanding is:

* There's no real appetite for automatically implementing Range support for `File` and friends
* There are real problems using `Range` requests with responders that would either modify the body or which make assumptions about the body, it sounds as though for now folks are comfortable with the safety rails for these things being clear documentation that explains what a bad idea that is.

So to get this across the line I believe the actions are:

* Remove the auto implementations for `File`, `&[u8]` etc.
* Produce documentation that clearly articulates what this does and how it interacts with other wrapping responders.
* It looks like potentially write some more tests?

Does that sound accurate @jebrosen ?

jebrosen:
That sounds about right. Automatic compression via `Content-Encoding` and automatic `ETag` are two examples of relatively commonly used or requested middleware that are straightforward to apply to all responses, yet would do the wrong thing if the response was a partial range. The main options I see here are:

* Use automatic `Range` handling by default on `File` et al. Fairings or Responder types that inspect or change the existing response body have to be aware of `Content-Range` and skip processing those responses (losing out on the functionality they provide).
* Make automatic `Range` handling opt-in via APIs such as `RangeResponder`, `NamedFile::with_range_handling()`, etc., and describe the caveats in the documentation.

richo:
Ok that's a little different to what I was thinking of, I was planning to
add a Range wrapping responder for Seek + Read. It sounds like you're more
interested in special casing File with an extra constructor?

For once I am fairly unopinioated, I just want to make sure I work on
something you're interested in merging :)

jebrosen:
Ah, sorry - either way, I would like to keep a wrapping `RangeResponder` as an option. In that comment I was thinking more about possible defaults and/or convenient APIs, for the most common use cases.

richo:
Gotcha, I am with you now! I will try to get a PR up for a wrapping responder, once that makes it in maybe we can discuss how to make the ergnomics better?

AseCoder:
I know this has been dead for four years but I would really like this feature – or at least some guidance on how to implement Range requests – for my own Rocket project.

PR #890, opened 2019-01-18
Better explain return value of FromFormValue::default() for Option<T>.

dvic:


site/guide/4-requests.md
```diff
@@ -280,7 +280,8 @@ More specifically, types that return `Some` from their
 [`FromFormValue::default()`] implementations will validate even when a value for
 the given parameter is missing, using the value returned from the `default()`
 method as the parameter's value. On such type is `Option<T>`, which returns
```
jebrosen: Might as well fix this here, too: "On such type" -> "One such type"

site/guide/4-requests.md
```diff
@@ -280,7 +280,8 @@ More specifically, types that return `Some` from their
...
-`None` from its `default()` implementation. By using a type of `Option<T>` for a
+`Some(None)` from its `default()` implementation, indicating that `None` should
+be used in case the parameter is missing. By using a type of `Option<T>` for a
```
jebrosen: Reading this more closely, I feel like this whole paragraph is too detailed for the guide. This section would fit better if it described only the use of `Option<T>`; the more detailed documentation can go on `FromFormValue` and `FromFormValue::default()` with a hyperlink in the guide pointing to the docs.

dvic: I agree, it still sounded confusing, I took a shot at simplifying it, what do you think?

jebrosen: Now I think it's incorrect, unfortunately: `Option<T>` already returns `Some(None)` from `FromFormValue::default` (done by Rocket), it's not something that a user needs to do.

I see two ways to explain this:

* Use this new wording without "and return `Some(None)` in [`FromFormValue::default()`].". This is better for users who just want to make parameters optional; anyone implementing `FromFormValue` themselves can read further about `default` and implement it on their own types if they wish. I would also adjust some phrasing in the later paragraphs if we go this route.
* Explain that "some types that implement FromFormValue provide a default value, which is used if the parameter is missing from the query", and demonstrate that `Option` and `bool` do this. This is closer to the original text but without as much "type soup" in the paragraph.

As a first-time reader, which approach do you think would have made more sense to you?

Side note: "Other defaultable `FromFormValue` types include `Result<T, E>`" is wrong -- currently only `Option` and `bool` override `default()`.

jebrosen:
Merged in ecc17d1 with slightly different wording. Thanks!

PR #894, opened 2019-01-21
Support raw identifiers (e.g. r#type) in forms and in route functions.

### LINKED ISSUE #881: Enable the usage of raw identifier
There is a [known limitation](https://rocket.rs/v0.4/guide/requests/#field-renaming) of using reserved identifier, e.g. `type` or `struct` as a parameter name.

As of [Rust 1.30 one could use raw identifier](https://rust-lang-nursery.github.io/edition-guide/rust-2018/module-system/raw-identifiers.html) making it possible to use those reserved keywords as variable name, e.g. 

```rust
let r#type = "foo";
```

I would like it to be possible to use those, e.g. in this minimal example:

```rust

#[get("/<type>")]
pub fn offers_get(r#type: String) -> String {
    unimplemented!()
}
```

This current results in the following error:

```text
error: custom attribute panicked
 --> src/main.rs:4:1
  |
4 | #[get("/<type>")]
  | ^^^^^^^^^^^^^^^^^
  |
  = help: message: `"__rocket_param_r#type"` is not a valid identifier
```

[Serde has a detection for raw identifier and just strips them](https://github.com/serde-rs/serde/blob/fea4e8e5b69bf44c051066aa9621f80144275992/serde_derive/src/dummy.rs#L26-L32). This should also be possible in this case.

jebrosen:
Resolves #881.

Brief explanation:

* For forms, a field named `r#type` without a `#[form(name="something")]` will now match `&type=` in an input string, when it currently matches `&r#type=`. This test currently passes in `master`:
    ```rust
    #[derive(Debug, PartialEq, FromForm)]
   struct Hazard { r#type: String, }

    #[test]
    fn test_hazard() {
        let form: Result<Hazard, _> = strict("r#type=test");
        assert_eq!(form, Ok(Hazard { r#type: "test".into() }));
    }
    ```

* For routes, two things happen:
  * Prefixing is always done to an `unraw`ed ident. This is implemented in `prefix()` itself.
  * Comparison of route segment names (e.g. "<enum>") is done against the `unraw`ed idents instead of the original idents.

TODO:

* [x] Use `unraw` from syn (dtolnay/syn#625) once it's in a release, instead of implementing it here
* [x] rebase on top of latest `master` and update for 1.0 releases of proc_macro2 and syn
* [x] Unconditionally call `unraw()` in `prefix()` instead
* [x] Explore interactions when using a raw ident in one place and a normal ident in another. In particular, there is some tricky behavior with the `uri!` macro that might be more likely to show up when mixing 2015 code (think `async: bool`) with 2018 code (passing in `r#async: true` to `uri!`).
* [x] Comprehensive reform of handling raw identifiers (see https://github.com/SergioBenitez/Rocket/pull/894#pullrequestreview-259139845)

### DIFF
--- a/core/codegen/src/attribute/route.rs
+++ b/core/codegen/src/attribute/route.rs
@@ -7,7 +7,7 @@ use devise::ext::{SpanDiagnosticExt, TypeExt};
 use indexmap::IndexSet;
 
 use crate::proc_macro_ext::{Diagnostics, StringLit};
-use crate::syn_ext::IdentExt;
+use crate::syn_ext::{IdentExt, NameSource};
 use crate::proc_macro2::{TokenStream, Span};
 use crate::http_codegen::{Method, MediaType, RoutePath, DataSegment, Optional};
 use crate::attribute::segments::{Source, Kind, Segment};
@@ -48,7 +48,7 @@ struct Route {
     /// The parsed inputs to the user's function. The first ident is the ident
     /// as the user wrote it, while the second ident is the identifier that
     /// should be used during code generation, the `rocket_ident`.
-    inputs: Vec<(syn::Ident, syn::Ident, syn::Type)>,
+    inputs: Vec<(NameSource, syn::Ident, syn::Type)>,
 }
 
 fn parse_route(attr: RouteAttribute, function: syn::ItemFn) -> Result<Route> {
@@ -74,7 +74,7 @@ fn parse_route(attr: RouteAttribute, function: syn::ItemFn) -> Result<Route> {
         for segment in iter.filter(|s| s.is_dynamic()) {
             let span = segment.span;
             if let Some(previous) = set.replace(segment.clone()) {
-                diags.push(span.error(format!("duplicate parameter: `{}`", previous.name))
+                diags.push(span.error(format!("duplicate parameter: `{}`", previous.name.name()))
                     .span_note(previous.span, "previous parameter with the same name here"))
             }
         }
@@ -110,7 +110,7 @@ fn parse_route(attr: RouteAttribute, function: syn::ItemFn) -> Result<Route> {
         };
 
         let rocket_ident = ident.prepend(ROCKET_PARAM_PREFIX);
-        inputs.push((ident.clone(), rocket_ident, ty.with_stripped_lifetimes()));
+        inputs.push((ident.clone().into(), rocket_ident, ty.with_stripped_lifetimes()));
         fn_segments.insert(ident.into());
     }
 
@@ -118,7 +118,7 @@ fn parse_route(attr: RouteAttribute, function: syn::ItemFn) -> Result<Route> {
     let span = function.sig.paren_token.span;
     for missing in segments.difference(&fn_segments) {
         diags.push(missing.span.error("unused dynamic parameter")
-            .span_note(span, format!("expected argument named `{}` here", missing.name)))
+            .span_note(span, format!("expected argument named `{}` here", missing.name.name())))
     }
 
     diags.head_err_or(Route { attribute: attr, function, inputs, segments })
@@ -207,10 +207,10 @@ fn query_exprs(route: &Route) -> Option<TokenStream> {
     let query_segments = route.attribute.path.query.as_ref()?;
     let (mut decls, mut matchers, mut builders) = (vec![], vec![], vec![]);
     for segment in query_segments {
-        let name = &segment.name;
+        let name = segment.name.name();
         let (ident, ty, span) = if segment.kind != Kind::Static {
             let (ident, ty) = route.inputs.iter()
-                .find(|(ident, _, _)| ident == &segment.name)
+                .find(|(name, _, _)| name == &segment.name)
                 .map(|(_, rocket_ident, ty)| (rocket_ident, ty))
                 .unwrap();
 
@@ -327,8 +327,8 @@ fn generate_internal_uri_macro(route: &Route) -> TokenStream {
         .filter(|seg| seg.source == Source::Path || seg.source == Source::Query)
         .filter(|seg| seg.kind != Kind::Static)
         .map(|seg| &seg.name)
-        .map(|name| route.inputs.iter().find(|(ident, ..)| ident == name).unwrap())
-        .map(|(ident, _, ty)| quote!(#ident: #ty));
+        .map(|name| route.inputs.iter().find(|(name2, ..)| name2 == name).unwrap())
+        .map(|(name, _, ty)| { let id = name.ident().unwrap(); quote!(#id: #ty) });
 
     let mut hasher = DefaultHasher::new();
     route.function.sig.ident.hash(&mut hasher);
@@ -385,8 +385,8 @@ fn codegen_route(route: Route) -> Result<TokenStream> {
     let mut data_stmt = None;
     let mut req_guard_definitions = vec![];
     let mut parameter_definitions = vec![];
-    for (ident, rocket_ident, ty) in &route.inputs {
-        let fn_segment: Segment = ident.into();
+    for (name, rocket_ident, ty) in &route.inputs {
+        let fn_segment: Segment = name.ident().unwrap().into();
         match route.segments.get(&fn_segment) {
             Some(seg) if seg.source == Source::Path => {
                 parameter_definitions.push(param_expr(seg, rocket_ident, &ty));
--- a/core/codegen/src/attribute/route.rs
+++ b/core/codegen/src/attribute/route.rs
@@ -74,7 +74,7 @@ fn parse_route(attr: RouteAttribute, function: syn::ItemFn) -> Result<Route> {
         for segment in iter.filter(|s| s.is_dynamic()) {
             let span = segment.span;
             if let Some(previous) = set.replace(segment.clone()) {
-                diags.push(span.error(format!("duplicate parameter: `{}`", previous.name.name()))
+                diags.push(span.error(format!("duplicate parameter: `{}`", previous.name))
                     .span_note(previous.span, "previous parameter with the same name here"))
             }
         }
@@ -118,7 +118,7 @@ fn parse_route(attr: RouteAttribute, function: syn::ItemFn) -> Result<Route> {
     let span = function.sig.paren_token.span;
     for missing in segments.difference(&fn_segments) {
         diags.push(missing.span.error("unused dynamic parameter")
-            .span_note(span, format!("expected argument named `{}` here", missing.name.name())))
+            .span_note(span, format!("expected argument named `{}` here", missing.name)))
     }
 
     diags.head_err_or(Route { attribute: attr, function, inputs, segments })
@@ -207,7 +207,6 @@ fn query_exprs(route: &Route) -> Option<TokenStream> {
     let query_segments = route.attribute.path.query.as_ref()?;
     let (mut decls, mut matchers, mut builders) = (vec![], vec![], vec![]);
     for segment in query_segments {
-        let name = segment.name.name();
         let (ident, ty, span) = if segment.kind != Kind::Static {
             let (ident, ty) = route.inputs.iter()
                 .find(|(name, _, _)| name == &segment.name)
@@ -232,6 +231,7 @@ fn query_exprs(route: &Route) -> Option<TokenStream> {
             Kind::Static => quote!()
         };
 
+        let name = segment.name.name();
         let matcher = match segment.kind {
             Kind::Single => quote_spanned! { span =>
                 (_, #name, __v) => {
@@ -327,8 +327,9 @@ fn generate_internal_uri_macro(route: &Route) -> TokenStream {
         .filter(|seg| seg.source == Source::Path || seg.source == Source::Query)
         .filter(|seg| seg.kind != Kind::Static)
         .map(|seg| &seg.name)
-        .map(|name| route.inputs.iter().find(|(name2, ..)| name2 == name).unwrap())
-        .map(|(name, _, ty)| { let id = name.ident().unwrap(); quote!(#id: #ty) });
+        .map(|seg_name| route.inputs.iter().find(|(in_name, ..)| in_name == seg_name).unwrap())
+        .map(|(name, _, ty)| (name.ident(), ty))
+        .map(|(ident, ty)| quote!(#ident: #ty));
 
     let mut hasher = DefaultHasher::new();
     route.function.sig.ident.hash(&mut hasher);
@@ -386,7 +387,7 @@ fn codegen_route(route: Route) -> Result<TokenStream> {
     let mut req_guard_definitions = vec![];
     let mut parameter_definitions = vec![];
     for (name, rocket_ident, ty) in &route.inputs {
-        let fn_segment: Segment = name.ident().unwrap().into();
+        let fn_segment: Segment = name.ident().into();
         match route.segments.get(&fn_segment) {
             Some(seg) if seg.source == Source::Path => {
                 parameter_definitions.push(param_expr(seg, rocket_ident, &ty));
--- a/core/codegen/src/derive/from_form.rs
+++ b/core/codegen/src/derive/from_form.rs
@@ -1,6 +1,7 @@
 use devise::{*, ext::{TypeExt, Split3, SpanDiagnosticExt}};
 
 use crate::proc_macro2::{Span, TokenStream};
+use crate::syn_ext::NameSource;
 
 #[derive(FromMeta)]
 pub struct Form {
@@ -9,7 +10,7 @@ pub struct Form {
 
 pub struct FormField {
     pub span: Span,
-    pub name: String
+    pub name: NameSource,
 }
 
 fn is_valid_field_name(s: &str) -> bool {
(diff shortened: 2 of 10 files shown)


jebrosen:
> Perhaps we can propose such a change?

I'll work on such an implementation for this, and after iterating over it and fixing the inevitable bugs I can propose it.

core/codegen/src/ident.rs
```diff
@@ -0,0 +1,99 @@
...
+#[derive(Clone, Debug, Eq, Ord)]
+pub struct Ident(Ident2);
+
+impl Ident {
+    pub fn is_raw(&self) -> bool {
+        self.0.to_string().starts_with("r#")
+    }
+
+    pub fn name(&self) -> String {
+        self.0.unraw().to_string()
```
jebrosen: `unraw()` requires a newer version of `syn` than what `devise` depends on, so this would break anyone using minimal versions.

core/codegen/src/ident.rs
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    pub fn prepend(&self, string: &str) -> Ident {
+        // TODO: want to use r#{}{}, but that fails (`"r#__rocket_param_enum"` is not a valid identifier)
```
jebrosen: `proc_macro2::Ident` has a separate `new_raw` constructor for this, but it's super-extra-unstable.

core/codegen/src/derive/uri_display.rs
```diff
@@ -60,7 +60,7 @@ pub fn derive_uri_display_query(input: TokenStream) -> TokenStream {
         .try_map_field(|_, field| {
             let span = field.span().into();
             let accessor = field.accessor();
-            let tokens = if let Some(ref ident) = field.ident {
+            let tokens = if let Some(ref ident) = field.ident.as_ref().map(|i| crate::Ident::from(i.clone())) {
```
SergioBenitez: I would do the conversion inside:

```rust
let tokens = if let Some(ref ident) = field.ident {
    let ident = crate::Ident::from(ident);
    ...
}
```

core/codegen/src/ident.rs
```diff
@@ -0,0 +1,99 @@
+//! A variant of Ident with more ergonomic handling of raw identifiers.
+
+use std::cmp::Ordering;
+
+use devise::syn;
+use self::syn::ext::IdentExt as _;
+use crate::proc_macro2::{Ident as Ident2, Span};
+use quote::ToTokens;
+
+#[repr(transparent)]
```
SergioBenitez: Why have the `repr(transparent)`? Presumably we'd do this so we can cast between `&syn::Ident` and `&crate::Ident`, but we don't seem to do that.

core/codegen/src/ident.rs
```diff
@@ -0,0 +1,99 @@
...
+#[derive(Clone, Debug, Eq, Ord)]
+pub struct Ident(Ident2);
```
SergioBenitez: Would it make more sense to store a `String` instead of an `Ident2`? That's internally what the `Ident2` is, and it would allow us to implement `name` without the `unraw` and as `name(&self) -> &str`. The downside, of course, is then we can't have the `Deref -> Ident2`, so if we use lots of methods from `Ident2`, this is probably a bad idea.

core/codegen/src/lib.rs
```diff
@@ -97,7 +97,9 @@ mod attribute;
 mod bang;
 mod http_codegen;
 mod syn_ext;
+mod ident;
 
+use crate::ident::Ident;
```
SergioBenitez: Where do we use this in this file?

core/codegen/tests/route-raw.rs
```diff
@@ -0,0 +1,42 @@
+#![feature(proc_macro_hygiene, decl_macro)]
```
SergioBenitez: After a rebase, we no longer need `decl_macro`.

core/codegen/src/ident.rs
```diff
@@ -0,0 +1,99 @@
...
+    }
+}
+
+impl<T: ?Sized + AsRef<str>> PartialEq<T> for Ident {
+    fn eq(&self, other: &T) -> bool {
+        self.name() == other.as_ref()
+    }
+}
+
+impl PartialOrd for Ident {
```
SergioBenitez: Do we use this?

core/codegen/src/ident.rs
```diff
@@ -0,0 +1,99 @@
...
+    pub fn name(&self) -> String {
+        self.0.unraw().to_string()
+    }
+
+    pub fn prepend(&self, string: &str) -> Ident {
+        // TODO: want to use r#{}{}, but that fails (`"r#__rocket_param_enum"` is not a valid identifier)
+        Ident::from(syn::Ident::new(&format!("{}{}", string, self.unraw()), self.span()))
+    }
+
+    pub fn span(&self) -> Span {
```
SergioBenitez: Shouldn't this come for free through the `Deref` impl?

core/codegen/src/derive/from_form_value.rs
```diff
@@ -39,7 +39,7 @@ pub fn derive_from_form_value(input: TokenStream) -> TokenStream {
         .try_map_enum(null_enum_mapper)
         .try_map_variant(|_, variant| {
             let variant_str = Form::from_attrs("form", &variant.attrs)
-                .unwrap_or_else(|| Ok(Form { value: variant.ident.to_string() }))?
```
SergioBenitez: With the right `From` impl, this becomes `variant.ident.into()` instead.

core/codegen/src/derive/from_form.rs
```diff
@@ -84,7 +84,7 @@ pub fn derive_from_form(input: TokenStream) -> TokenStream {
         .try_map_fields(move |_, fields| {
             let (constructors, matchers, builders) = fields.iter().map(|field| {
                 let (ident, span) = (&field.ident, field.span().into());
-                let default_name = ident.as_ref().expect("named").to_string();
+                let default_name = crate::Ident::from(ident.as_ref().expect("named").clone()).to_string();
```
SergioBenitez: I don't think we need to convert this to `crate::Ident` since we just call `to_string`.

core/codegen/src/derive/from_form.rs
```diff
@@ -40,7 +40,7 @@ fn validate_struct(gen: &DeriveGenerator, data: Struct<'_>) -> Result<()> {
 
     let mut names = ::std::collections::HashMap::new();
     for field in data.fields().iter() {
-        let id = field.ident.as_ref().expect("named field");
+        let id = crate::Ident::from(field.ident.as_ref().expect("named field").clone());
```
SergioBenitez: Ideally we can write:

```suggestion
        let id: crate::Ident = field.ident.as_ref().expect("named field").into();
```

core/codegen/src/bang/mod.rs
```diff
@@ -3,15 +3,15 @@ use crate::proc_macro2::TokenStream as TokenStream2;
...
+use crate::syn_ext::syn_to_diag;
 use crate::{ROUTE_STRUCT_PREFIX, CATCH_STRUCT_PREFIX};
 
 mod uri;
 mod uri_parsing;
 
 crate fn prefix_last_segment(path: &mut Path, prefix: &str) {
     let mut last_seg = path.segments.last_mut().expect("syn::Path has segments");
-    last_seg.value_mut().ident = last_seg.value().ident.prepend(prefix);
+    last_seg.value_mut().ident = crate::Ident::from(last_seg.value().ident.clone()).prepend(prefix).into();
```
SergioBenitez: This is...rather long.

core/codegen/src/ident.rs
jebrosen: On a new enough compiler an `Ident2` contains only a `proc_macro::Ident` which is an opaque handle to the compiler. We probably could store a `String`, but then we also need to store a `Span` and this is where I didn't want to duplicate so much logic. There are some invariants that `Ident` (in `proc_macro` or `proc_macro2` or both, I would need to look again) upholds regarding what the name is.

core/codegen/src/lib.rs
jebrosen: Nowhere. It's here so that `crate::Ident` works instead of `crate::ident::Ident` everywhere else.

core/codegen/src/ident.rs
jebrosen: No; I kept it for feature parity with `Ident` from `proc_macro2`.

core/codegen/src/ident.rs
jebrosen: It should, but I ran into some kind of method resolution problem. I will take a closer look and document it.

core/codegen/src/derive/from_form.rs
jebrosen: This is a long way to get to `ident.unraw()`, which is all that is really needed here.

core/codegen/src/ident.rs
jebrosen: Without this `pub fn`, `.span()` resolves to a `proc_macro::Span` instead of a `proc_macro2::Span` and I can't figure out why.

jebrosen:
I've rebased this on `master` and adjusted for the `syn` 1.0 breaking changes, but I have not attempted to make use of any improvements that might be there.

SergioBenitez:
@jebrosen My apologies for dropping this. I'd love to merge this in after a rebase/updating to master.

jebrosen:
Rebased and all tests passing. However, it looks like there are a few opportunities for minor improvements that I have either not investigated or forgotten the answer to.

core/codegen/src/ident.rs
SergioBenitez: Is this the same as `#[derive(PartialOrd)]`?

SergioBenitez: And, should we be comparing `name()` instead, since that's what `PartialEq` is defined on?

core/codegen/src/ident.rs
```diff
@@ -0,0 +1,128 @@
+//! A variant of Ident with more ergonomic handling of raw identifiers.
+
+use std::cmp::Ordering;
+
+use devise::syn;
+use self::syn::ext::IdentExt as _;
+use crate::proc_macro2::{Ident as Ident2, Span};
+use quote::ToTokens;
+
+#[derive(Clone, Debug)]
```
SergioBenitez, line 10: We should have a doc-string explaining why this exists and what it intends to do.

core/codegen/src/ident.rs
SergioBenitez: It bothers me that effectively all of these functions require a `String` allocation or two due to the way this is implemented. Can we do something so that we don't allocate on each method call unless we _need_ to, i.e, some downstream use requires a `String`?

core/codegen/src/lib.rs
SergioBenitez: I see. Can we add `pub(crate)` to this to make it clear?

core/codegen/src/derive/from_form_value.rs
SergioBenitez: Did the `.into()` not work?

core/codegen/src/ident.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+}
+
+impl From<Ident2> for Ident {
+    fn from(ident: Ident2) -> Ident {
+        Ident(ident)
+    }
+}
+
+impl From<&Ident2> for Ident {
```
SergioBenitez, line 56: I find this `impl` kind of odd. Why not `clone()` at the `::from()` site?

core/codegen/src/ident.rs
jebrosen: I believe it's <https://docs.rs/devise/0.2.0/devise/trait.Spanned.html#tymethod.span> that is being selected instead of the `Deref` impl. I have added a commit that removes this function and updates the call sites instead.

core/codegen/src/ident.rs
jebrosen: I added the `impl` per <https://github.com/SergioBenitez/Rocket/pull/894#pullrequestreview-264465565> - I went ahead and used it because it was shorter than `.clone()`, but I don't have a strong opinion on this.

core/codegen/src/ident.rs
jebrosen: I've actually removed this - I expect sorting/comparing `Ident`s to be very unlikely and should be handled more thoughtfully.

core/codegen/src/ident.rs
SergioBenitez: We could use a [`Storage`](https://docs.rs/state/0.4.1/state/struct.Storage.html) and [`Storage::get_or_set()`](https://docs.rs/state/0.4.1/state/struct.Storage.html#method.get_or_set) to allocate the `String` only if needed.

jebrosen:
> Something doesn't feel right here.

I agree! I thought I said that already, but I must have closed the tab without finishing and hitting `Comment`.

> There are many many `crate::Ident::from()` that somehow feel unnecessary? Or perhaps fragile: if we forget to go from `syn::Ident` to `crate::Ident` when, say, using `.to_string()`, all of sudden everything is incorrect.

That's my issue with this as well. To recap:

* The `unraw()` operation is necessary when an identifier is going to be converted or compared to a string, where keeping `r#` would be incorrect or inconvenient. This primarily occurs with:
  * Route parameters, which are compared to the route string passed to the macro
  * Form field names, which are compared to the runtime-provided form string.
  * Any identifiers where a prefix is added. `prepend` must be aware of `r#`, to avoid creating the invalid identifier `rocket_generated_r#thing`
* We added `Ident` to encompass the comparison (`PartialEq`), conversion to string (`name` and `to_string`), and `prepend` into a single type. Each of those operations must first "unraw" the `Ident`, so it makes some sense for them to be together in this "raw-aware Ident" type.
  * I switched to `Ident` "everywhere" I could - the main exception is the types already provided by `syn` or `devise`, which we can't change here. For those types, `crate::Ident::from()` is needed before calling `to_string` or `prepend` or performing a rawness-affected equality check.
* It has also been over a year since I started work on this, and I have likely forgotten or misconstrued a few of the decisions that led to this point.

In some ways `Ident` now looks more painful than it is helpful. Instead of asking "do we need to call `unraw()` here", we ask "do we need to use `crate::Ident` here" - effectively, we have traded one inconvenience for a different one. If `syn` had `name()` and `prepend()` already, or if `devise` could somehow enforce using a modified version of `Ident` in more places, I think this would be a clearer win. But as far as I can tell from a search, dtolnay/syn#625 is the most that has been asked for and done there.

If we could *always* use raw identifiers that might be a viable solution, but it has the same problems wrt to not being provided by the upstream crates.

SergioBenitez:
Thanks for the recap. To re-recap, it sounds like there are two logical operations we've added `crate::Ident` for:

- To compare a user's identifier string (`"<foo>"`) to a raw ident (`#foo`).
- To correctly prefix a raw identifier.

If this is the case, I propose the following course of action:

- When we parse user identifier strings, store the resulting string in a new type, say, `DynamicParam`. Never return a `String` or `&str` for any of the low-level APIs so that this decision is viral and impossible to circumvent/forget/miss. 
- Implement `PartialEq<Ident>` for `DynamicParam` so that it properly handles raw identifiers.
- Have `devise` add an `Ident::prepend()` extension that properly handles raw identifiers.

I believe this would assuage my concerns.

jebrosen:
This PR has been entirely re-done, hopefully in the spirit of but not entirely to the letter of the previous comment:

* `prepend()` is already in `rocket_contrib` before this PR, so I have simply updated it. It can of course be moved to `devise` instead, which already has several `*Ext` traits.
* I named the type `NameSource`, which I am not completely happy with. It is intended to be used in any place that a string *or* identifier defines a "name" to match against later, whether it be another `String` or `Ident`.
  * In several places, a field `ident: Ident` becomes `name: NameSource`. Likewise, unnamed fields of type `NameSource` is pattern matched into a binding called `name`. I don't like this very much either.
  * `uri!` is one example where both items being compared are `Ident`s, but we want e.g. `r#blah` to match `blah`. This situation  appears to be unique to `uri!`; every other case has a `String` *somewhere* that is being compared to or produced from a `NameSource`.
  * There are a few `unwrap()`s used to "get back" the original `Ident`, which may or may not be present. It seems plausible to replace this with `span()`, which *all* `NameSources` carry unconditionally.

I only had one serious issue:

> Never return a String or &str for any of the low-level APIs so that this decision is viral and impossible to circumvent/forget/miss.

This seems difficult or impossible in practice: some of the "problematic" places are `Ident` without any go-between to `String`. So, I can't be sure that this approach exhaustively covers all places it might be necessary.

---

I know this is not exactly what you described, but I think this approach will be much easier to iterate on or fix than the ad-hoc `unraw` and `from` calls.

SergioBenitez:
I think we're finally ready to merge this. @jebrosen I squashed your commit with minimal changes to work on the latest master. I pushed a second commit with a few changes. If you agree with these, then we're ready to merge. Let me know!

core/codegen/src/attribute/route.rs
```diff
@@ -48,7 +48,7 @@ struct Route {
     /// The parsed inputs to the user's function. The first ident is the ident
```
jebrosen, line 48: "first ident" and "second ident" no longer matches the code.

jebrosen (commented):
I think the changes all look good to me, with one non-code nitpick:

SergioBenitez:
Awesome. Merged in 97f6bc5dc0c842609bf92a106a78844c8e346fdd!

Was really happy with the `NameSource` implementation, by the way. Absolutely resolved my prior reservations. Well done!

Issue #635, opened 2018-05-07
error[E0425]: cannot find value `rocket_param_id` in this scope


Boscop:
After updating my nightly (from 03-06) I'm getting this error, any idea how to fix it?
error[E0425]: cannot find value `rocket_param_id` in this scope

![image](https://user-images.githubusercontent.com/535593/39691658-8e69e74c-51de-11e8-847f-38bba92b27a6.png)

This error happens only for my `delete` handler which is in a macro that's used 11 times, so I get the same error 11 times:
```rust
			#[post("/delete/<id>")]
			fn delete(me: Customer, id: i64) -> Res<()> {
```
The other handlers outside of macros, and even the other handlers inside macros but with only guard args and no url args don't cause any errors. E.g. this one is also inside the same macro but causes no error:
```rust
			#[post("/list")]
			fn list(me: Customer) -> Res<Vec<View>> {
```

jebrosen:
Can you make a simple reproducible test case? Since at least one and probably two different kinds of macros are involved I wouldn't want to try to guess what's happening here. For issues with the codegen, building with `ROCKET_CODEGEN_DEBUG=1` can help reveal what's going on.

Boscop:
Thanks, with `ROCKET_CODEGEN_DEBUG=1` it looks like this. Any idea why it's not working? :)
```rust
--> C:\Users\me\.cargo\git\checkouts\rocket-8bf16d9ca7e90bdc\3894e4e\codegen\src\decorators\route.rs:286
Route params: RouteParams { annotated_fn: Function(Spanned { node: (delete#0, FnDecl { inputs: [Arg { ty: type(Customer), pat: pat(4294967295: me), id: NodeId(4294967295) }, Arg { ty: type(i64), pat: pat(4294967295: id), id: NodeId(4294967295) }], output: Ty(type(Res<()>)), variadic: false }), span: Span { lo: BytePos(37053), hi: BytePos(37311), ctxt: #0 } }), method: Spanned { node: Post, span: Span { lo: BytePos(37028), hi: BytePos(37032), ctxt: #0 } }, uri: Spanned { node: Uri { uri: "/delete/<id>", path: (0, 12), query: None, fragment: None, segment_count: 2 }, span: Span { lo: BytePos(37035), hi: BytePos(37049), ctxt: #0 } }, data_param: None, query_param: None, format: None, rank: None }

[...]

PR #896, opened 2019-01-22
Allow users to use `#[get]` inside of macros

oli-obk:


jebrosen:
I see this as a strict win (not compiling -> compiling with an imperfect diagnostic).

I think the test fits better in `core/codegen/tests/expansion.rs` that came about from the last nested-macro bug (#635).

SergioBenitez:
This is great. Thanks, @oli-obk, @jebrosen. Merged in c86f4312fb273c0380fb76a97bfb7fc227800542.

Issue #888, opened 2019-01-15
Use different version of database crate?


hamaluik:
Hello,

I'm used to using r2d2 with rocket, but the latest contrib (0.4.0) apparently has this functionality baked in all nicely, which is great! Unfortunately, I need to use a different version of the database crate than contrib is pulling in; is there any way to change what the contrib is pulling in while still using the contrib?

Specifically, I need `rusqlite 0.16.0` with the `sqlcipher` feature enabled, but if I include it I of course get errors:

```
➜ cargo build
error: multiple packages link to native library `sqlite3`, but a native library can be linked only once

package `libsqlite3-sys v0.9.1`
    ... which is depended on by `rusqlite v0.14.0`
    ... which is depended on by `rocket_contrib v0.4.0`
    ... which is depended on by `<my package>`
links to native library `sqlite3`

package `libsqlite3-sys v0.11.1`
    ... which is depended on by `rusqlite v0.16.0`
    ... which is depended on by `<my package>`
also links to native library `sqlite3`
```

Thanks!

hamaluik:
No worries, thanks for the response. I ended up doing just that and it works a treat!

PR #902, opened 2019-01-27
Upgrade rusqlite dependency to 0.16

NoraCodes:
This solves the specific issue raised by #888, though not the general problem of database versioning with native libraries.


jebrosen:
`r2d2_sqlite` needs to be bumped as well for this to work. The log output suggests a conflict with diesel but I think bumping `r2d2_sqlite` may resolve that.

However `rusqlite` 0.15 and 0.16 had breaking changes, as your changes to examples demonstrates. That makes this a semver major change, so it can't be released until at least 0.5.

SergioBenitez:
Merged in c6c0b3a6e13c27ae3f0cecff4ea1d77e9f1ad06d by @jebrosen!

PR #906, opened 2019-01-30
Change `rocket_contrib` to not depend on default features from `rocket`.

### LINKED ISSUE #905: rocket_contrib is missing private-cookies flag
I'm currently trying to prepare a bugfix release but I'm having trouble due to the ring dependency in the cookie crate. It's currently depending on an outdated ring version, but since I don't need cookies at all I would like to disable the need for ring.

It's already possible to build rocket without ring like this:
```toml
rocket = { version = "0.4", default-features = false }
```
This disables the private-cookies feature, which in turn disables the private-cookies feature in rocket_http, which disables the `secure` feature in `cookie`.
```toml
private-cookies = ["rocket_http/private-cookies"]
```

This works fine unless you also need rocket_contrib, which is the case for me. `rocket_contrib` depends on `rocket` with default features and no way to disable it, which means that the `ring` dependency in `cookie` is getting enabled again:
```toml
rocket = { version = "0.4.0", path = "../../core/lib/" }
```

Introducing a feature to disable ring in rocket_contrib would be the best solution for my problem, I'm currently trying to find an alternative solution since that specific version is [faulty on aarch64](https://github.com/kpcyrd/sn0int/issues/69).

Thanks!

jebrosen:
Fixes #905.

...this is technically a breaking change, I think? Any crate depending on `rocket` with `default_features = false` and also depending on `rocket_contrib` will lose access to private cookie functionality with this change.

### DIFF
--- a/contrib/lib/Cargo.toml
+++ b/contrib/lib/Cargo.toml
@@ -39,7 +39,7 @@ memcache_pool = ["databases", "memcache", "r2d2-memcache"]
 [dependencies]
 # Global dependencies.
 rocket_contrib_codegen = { version = "0.4.0", path = "../codegen", optional = true }
-rocket = { version = "0.4.0", path = "../../core/lib/" }
+rocket = { version = "0.4.0", path = "../../core/lib/", default-features = false }
 log = "0.4"
 
 # Serialization and templating dependencies.


kpcyrd:
I've accidentally duplicated this PR (doing too many things at the same time).

> Any crate depending on rocket with default_features = false and also depending on rocket_contrib will lose access to private cookie functionality with this change.

This is what I've posted (explanation follows):

> It seems the private-cookies feature isn't actually used by rocket_contrib at all, which means it can be safely disabled.
>
> This is not a breaking change because this feature is getting re-enabled in rocket as soon as any crate in the dependency tree depends on rocket with explicitly disabling default features, which is the case if private cookies are actually needed.
>
> Please consider publishing this as 0.4.1. Thanks!

Basically, the feature is not "disabled" but instead "not enabled". This means that any other crate in the dependency tree can enable this feature if needed by depending on rocket with default features.

Since rocket_contrib doesn't care it shouldn't try to enable the feature, which it currently does by not explicitly opting-out. It's a bit complicated to explain since it's a default feature, but it's getting clearer if you think about it as a feature that's explicitly enabled by rocket_contrib.

jebrosen:
Merged in 90296dd.

PR #912, opened 2019-02-02
Export derive macros alongside the traits in `rocket` that they derive.

### LINKED ISSUE #903: Consider reexporting derive macros at the same scope at their traits
Currently derive macros are exported only at the crate root. This is not an issue in the 2015 edition since macros are in a global scope accessible by all modules. In the 2018 edition though this can create confusion since both racer and intellij do not provide completion suggestions for importing. It would be useful if derive macros (such as `FromForm`) we exported at the same level as their traits so one could just `use rocket::request::FromForm` and get both the derive macro and trait in scope.

jebrosen:
Resolves #903.

Does not reexport `UriDisplayPath` or `UriDisplayQuery` because they are in `rocket_http`, and implementing those would mean that `rocket_http` would have depend on `rocket_codegen`.

TODO:

* [ ] How do these appear in the docs?

### DIFF
--- a/core/lib/src/request/form/from_form.rs
+++ b/core/lib/src/request/form/from_form.rs
@@ -1,5 +1,7 @@
 use request::FormItems;
 
+pub use rocket_codegen::FromForm;
+
 /// Trait to create an instance of some type from an HTTP form.
 /// [`Form`](::request::Form) requires its generic type to implement this trait.
 ///
--- a/core/lib/src/request/form/from_form.rs
+++ b/core/lib/src/request/form/from_form.rs
@@ -1,7 +1,5 @@
 use request::FormItems;
 
-pub use rocket_codegen::FromForm;
-
 /// Trait to create an instance of some type from an HTTP form.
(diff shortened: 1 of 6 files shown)


jebrosen:
This is how they currently appear in the docs: https://ptpb.pw/fZ0m.png ... I have no idea why it says reexport of `self::form::FromForm` for a reexport of `rocket_codegen::FromForm`.

jebrosen:
New screenshot: https://paste.rs/VEU.png. I believe I can't use `#[doc(hidden)]` on the `pub use` because that would also hide the trait. I'll try importing the derive macro in `request.rs` instead of `form.rs` - it should have the same end result but it would be possible to either hide it in the documentation or show up as a reexport of the more obvoius `rocket_codegen::FromForm`.

SergioBenitez:
Merged in https://github.com/SergioBenitez/Rocket/commit/6bf751fb226016cee7efda4056a1f9dfcbbea339. This was a nice change. Thanks, Jeb.

PR #919, opened 2019-02-07
Added PG Example to Databases Documentation (Rocket Contrib)

PrismaPhonic:
This PR adds a small example of connecting Postgres to the databases documentation, so that it's obvious how trivial it is to tie a non-sqlite database to Rocket.

### DIFF
--- a/contrib/lib/src/databases.rs
+++ b/contrib/lib/src/databases.rs
@@ -180,6 +180,13 @@
 //! ROCKET_DATABASES={my_db={url="db.sqlite"}}
 //! ```
 //!
+//! Multiple databases can be specified in the `ROCKET_DATABASES` environment variable
+//! as well by comma separating them:
+//!
+//! ```bash
+//! ROCKET_DATABASES={my_db={url="db.sqlite"},my_pg_db={url="postgres://root:root@localhost/pg_test"}}
+//! ```
+//!
 //! ## Guard Types
 //!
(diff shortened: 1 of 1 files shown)


contrib/lib/src/databases.rs
```diff
@@ -203,6 +210,21 @@
...
 //!
+//! Tieing your application to a different RDBMS is simple, and follows the same pattern.
+//! Here's an example using Postgres with Diesel as the driver:
+//!
+//! ```rust
+//! # extern crate rocket;
+//! # #[macro_use] extern crate rocket_contrib;
+//! # #[cfg(feature = "diesel_postgres_pool")]
+//! # mod test {
+//! use rocket_contrib::databases::diesel;
```
jebrosen: Can we use a non-Diesel driver here, just so it's not too similar?

PrismaPhonic: Sure, I'll go ahead and make the adjustment

PrismaPhonic: @jebrosen just updated with a new commit to use Rust-Postgres as the driver.

PrismaPhonic:
Just realized that prettier auto-formatted the code on save.  Will roll back a commit and turn off auto-save formatting and make adjustments again.  Sorry about that

jebrosen:
Merged in 622f4f3, thank you!

PR #920, opened 2019-02-09
Replace deprecated dependency

Deedasmi:
According to https://crates.io/crates/isatty isatty is no longer maintained.


SergioBenitez:
Looks good to me! Merged in 75d2ea96d926f5d34d9bd249aeea912e29738856.

Issue #749, opened 2018-08-23
Filtering requests


mmrath:
What I want is to inspect every request and if it is anything but POST to /api/auth/register or /api/auth/login then check for for auth header. If header is not present, then reply 401. I looked at Fairing - but Fairing cant respond to requests. I looked at request guard, it looks like for them I need to add a param to every route handler.

## Questions

Any questions _must_ include:

  1. The version of Rocket this question is based on, if any.
master

  2. What steps you've taken to answer the question yourself.
Looked at Fairing and request guard docs

  3. What documentation you believe should include an answer to this question.
not sure

jebrosen:
The pain point you are encountering is that request Fairings cannot alter the request handling flow by responding or by returning some kind of error type. This is a known issue that can and should be available in Rocket's API eventually but I don't see an open issue for it.

Adding a request guard to every route handler is annoying, but does make it more obvious which routes need authentication and which don't. If you do add a request guard to every route indicating its privilege level, you can require instances of those request guards in database or API code as a compile-time assurance that the necessary security properties are enforced, as well.

But if "every" route requires authentication, it can be annoying to do that. You *can* work around the fairing response situation by altering the request uri in the fairing, thereby changing which route it ends up going to. `rocket_cors` does this (https://github.com/lawliet89/rocket_cors/blob/master/src/fairing.rs#L61), for example. It is a hack and I'm hesitant to even bring it up, but it might be worth exploring its suitability for your problem.

mmrath:
From what I hear from you this issue should be a feature request than a question. I pretty much need everything except login screen to be authenticated. I think this is very common scenario.

SergioBenitez:
@mmrath I'm with you. We have a feature in mind that would resolve this problem, but alas, it is as of now unimplementable due to lacking features in Rust.

NoUsername:
@SergioBenitez would be interesting to know what features Rust would need so this could be implemented.
Keep up the good work 👍

mmrath:
@SergioBenitez Would you be able to give a bit more information on how the solution might look like?

SergioBenitez:
@NoUsername We'd need [`non_static_type_id`](https://github.com/rust-lang/rfcs/blob/master/text/1849-non-static-type-id.md), blocked in https://github.com/rust-lang/rust/issues/41875.

@mmrath The gist, from my design doc, is:

> ## Design
> 
> As the name implies, the guiding principle behind type-based catchers is that
[...]

PR #930, opened 2019-02-19
[WIP] Add a fairing for CORS support

derekdreery:
This commit adds a fairing to support CORS.

It works by adding CORS headers to all responses, and also catching 404 OPTIONS requests - returning a 200 response with the correct cors headers as per configuration.

It probably needs some work on the docs before it is merged, and also I didn't realise that my text editor runs `cargo fmt` by default, so either a separate commit needs to go in to just run `cargo fmt` on the existing codebase, or I need to revert all those changes :'(.

I'm opening it now to see how you feel about the approach. It works well for the project I'm using it in, but I'm interested to see if that is replicated in other projects.


jebrosen:
> This commit adds a fairing to support CORS.
> 
> It works by adding CORS headers to all responses, and also catching 404 OPTIONS requests - returning a 200 response with the correct cors headers as per configuration.

I haven't looked much at the implementation yet, partly because of the large diff. Have you seen https://github.com/lawliet89/rocket_cors or @SergioBenitez's `Responder` example in https://github.com/SergioBenitez/Rocket/issues/25#issuecomment-271065434? How does this compare to those? I assume from the documentation in this PR that it essentially only operates in the "Fairing" mode of `rocket_cors`.

Personally, I think it would be reasonable to provide a "works for the majority" CORS implementation in `contrib`, and let other crates such as `rocket_cors` provide more advanced functionality or unusual scenarios. I don't know well enough, however, where the line should be drawn between "works for the majority" vs "unusual scenarios".

> I didn't realise that my text editor runs `cargo fmt` by default, so either a separate commit needs to go in to just run `cargo fmt` on the existing codebase, or I need to revert all those changes :'(.

That is unfortunate. It shouldn't be too hard to "undo" entire trees that had format-only chages (e.g. `core/`, `contrib/codegen`), which should cover most of the noise. I'd be happy to give further advice or assistance with this effort.

SergioBenitez:
Closing due to inactivity.

derekdreery:
I've been working on another project, but now I'm back to rust web :).

> It may not, however, abort or respond directly to the request; these issues are better handled via request guards or via response callbacks. 

Does this mean that CORS should be implemented as a `RequestGuard`? The above statement seems to indicate *yes*, however you probably want the guard on all requests, so it seems safer to apply it as a blanket rather than on route callbacks where they could potentially be forgotten?

I don't know the right answer so interested to hear thoughts.

derekdreery:
I've thought about this some more, and I think `Fairing`s should be allowed to interrupt the usual flow and respond in the *request* stage of the middleware traversal. I know there are times that a guard is more appropriate, but CORS is something that in most cases works the same for the whole website, so implementing it at the fairing level seems to make sense, at least to me.

This would require having a way to say "stop and return a response now" in the request part of the fairing.

SergioBenitez:
@derekdreery I agree, and it's planned, but doing it the way other frameworks do it is not a path we'd like to follow for many reasons. See my commentary in https://github.com/SergioBenitez/Rocket/issues/749#issuecomment-432572197 for more.

PR #933, opened 2019-02-22
Missing "super"

crudolf:
There is a compilation error otherwise.

### DIFF
--- a/examples/json/src/tests.rs
+++ b/examples/json/src/tests.rs
@@ -1,4 +1,4 @@
-use rocket;
+use super::rocket;
 use rocket::local::Client;
 use rocket::http::{Status, ContentType};
 


jebrosen:
What is the compilation error? All tests should be running and are passing on the CI, so we need to track down and fix that too if that's the case.

crudolf:
error[E0423]: expected function, found module `rocket`
 --> src/tests.rs:7:30
  |
7 |     let client = Client::new(rocket()).unwrap();
  |                              ^^^^^^ not a function

crudolf:
I currently learn rust - all I can say is, that I assume, that the example code was either not fully maintained or a rust update requires this. I am using the latest nightly rust built.

jebrosen:
The test is passing on CI - https://travis-ci.org/SergioBenitez/Rocket/jobs/497261520#L1288

I think what has happened here is that your own code is using the 2018 edition, but all of Rocket's code and examples are still on the 2015 edition. The way use statements work was changed, and `use rocket;` in 2015 imports what is called `crate::rocket` (or `super::rocket`) in 2018.

Updating documentation to 2018 syntax has been discussed, but there are no plans to do so quite yet (see https://github.com/SergioBenitez/Rocket/pull/874#issuecomment-454621248).

Since this will be done later as part of a much larger change encompassing pretty much all of Rocket's code, I'm going to close this PR. Thank you for the reminder, though!

PR #941, opened 2019-03-06
Allow more unencoded characters in query strings

vladignatyev:


core/http/src/parse/uri/tables.rs
```diff
@@ -29,6 +29,40 @@ const PATH_CHARS: [u8; 256] = [
...
+
+const QUERY_CHARS: [u8; 256] = [
+    //  0      1      2      3      4      5      6      7      8      9
+        0,     0,     0,     0,     0,     0,     0,     0,     0,     0, //   x
+        0,     0,     0,     0,     0,     0,     0,     0,     0,     0, //  1x
+        0,     0,     0,     0,     0,     0,     0,     0,     0,     0, //  2x
+        0,     0,     0,  b'!',     0,     0,  b'$',  b'%',  b'&',     0, //  3x
+     b'(',  b')',  b'*',  b'+',  b',',  b'-',  b'.',  b'/',  b'0',  b'1', //  4x
+     b'2',  b'3',  b'4',  b'5',  b'6',  b'7',  b'8',  b'9',  b':',  b';', //  5x
+//      <             > (we allow this chars: while they are not valid in URLs but in URI)
```
jebrosen: The comment is confusing and maybe wrong - I don't know what "not valid in URLs but in URI" means.

But note that query might need the same distinction that path has between "valid always" vs "valid only in a route specification as a placeholder," which is why `<` and `>` are handled specially there.

core/http/src/parse/uri/tables.rs
```diff
@@ -29,6 +29,40 @@ const PATH_CHARS: [u8; 256] = [
...
+        1,  b'=',     1,  b'?',  b'@',  b'A',  b'B',  b'C',  b'D',  b'E', //  6x
+     b'F',  b'G',  b'H',  b'I',  b'J',  b'K',  b'L',  b'M',  b'N',  b'O', //  7x
+     b'P',  b'Q',  b'R',  b'S',  b'T',  b'U',  b'V',  b'W',  b'X',  b'Y', //  8x
+     b'Z',  b'[',  b'\\',  b']', b'^',  b'_',  b'`',  b'a',  b'b',  b'c', //  9x
+     b'd',  b'e',  b'f',  b'g',  b'h',  b'i',  b'j',  b'k',  b'l',  b'm', // 10x
+     b'n',  b'o',  b'p',  b'q',  b'r',  b's',  b't',  b'u',  b'v',  b'w', // 11x
+     b'x',  b'y',  b'z',  b'{',  b'|',  b'}',  b'~',     0,     0,     0, // 12x
```
jebrosen: I'm surprised (and somewhat disappointed) that all of these characters might appear unencoded in a query string.

core/http/src/parse/uri/tests.rs
```diff
@@ -89,6 +103,25 @@ fn bad_parses() {
...
+    assert_parses!("/some/path/?tags=[rocket,is,perfect]&clients={arenot}");
+    assert_parses!("/rocket/@user/");
+    assert_parses!("/rocket/@user/?tags=[rocket,%F0%9F%98%8B]");
+    assert_parses!("/rocket/?username=@sergio&tags=[rocket,%F0%9F%98%8B]");
+    assert_parses!("/rocket/?Key+With+Spaces=value+too");
+    assert_no_parse!("/rocket/?Key+With+\'");
+    assert_parses!("/rocket/?query=%3E5");
+}
+
+// #[test]
```
jebrosen: As above, this will probably have to split into "parse arbitrary URI" vs "parse a route specification".

core/http/src/parse/uri/tables.rs
vladignatyev: @jebrosen I mean that it is correct for declaration in usage code, but not in HTTP request

core/http/src/parse/uri/tables.rs
vladignatyev: @jebrosen same here)

Valodim:
I'm confused as to whether angle brackets `<` `>` are accepted now, or not?

As a specific use case, [GnuPG](https://gnupg.org) does not escape angle brackets as they might appear in user ids, and its documentation says to do. This leads to an [issue](https://gitlab.com/sequoia-pgp/hagrid/issues/94) with such queries in our rocket application.

jebrosen:
My quick test suggests `hyper` is rejecting `GET /a> HTTP/1.1` as an invalid HTTP header before Rocket even sees it. I'll add that to #998 to look at more closely.

marek-lach:
This is an important issue for multilangual applications indeed...

Toromino:
+1

Would be nice to have this in, I need it for some API endpoints.

core/http/src/parse/uri/tests.rs
```diff
@@ -89,6 +103,25 @@ fn bad_parses() {
...
+    assert_parses!("/path?param={value}", "/path/?param={value}", "/some/path/?param={forgot-to-replace-placeholder}");
+    assert_parses!("/path?param={value}&onemore={value}");
+    assert_parses!("/some/path/?tags=[]", "/some/path/?tags=[rocket,is,perfect]");
+    assert_parses!("/some/path/?tags=[rocket,is,perfect]&clients={arenot}");
+    assert_parses!("/rocket/@user/");
+    assert_parses!("/rocket/@user/?tags=[rocket,%F0%9F%98%8B]");
+    assert_parses!("/rocket/?username=@sergio&tags=[rocket,%F0%9F%98%8B]");
+    assert_parses!("/rocket/?Key+With+Spaces=value+too");
+    assert_no_parse!("/rocket/?Key+With+\'");
+    assert_parses!("/rocket/?query=%3E5");
```
igalic:
```rocket
assert_parses!("/rocket/~user/");
```

PR #949, opened 2019-03-14
clarify "reinterpreting" in guide/4-requests.md

seeekr:
Small change to clarify what is meant by the first sentence there.

### DIFF
--- a/site/guide/4-requests.md
+++ b/site/guide/4-requests.md
@@ -52,7 +52,7 @@ your application explicitly handles.
 
 ### Reinterpreting
 
-Because browsers can only send `GET` and `POST` requests, Rocket _reinterprets_
+Because HTML forms are only allowed to use `GET` and `POST` HTTP methods, Rocket _reinterprets_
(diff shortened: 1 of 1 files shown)


site/guide/4-requests.md
```diff
@@ -52,7 +52,7 @@ your application explicitly handles.
 
 ### Reinterpreting
 
-Because browsers can only send `GET` and `POST` requests, Rocket _reinterprets_
+Because HTML forms are only allowed to use `GET` and `POST` HTTP methods, Rocket _reinterprets_
```
jebrosen: I think the following makes the rationale more obvious than either of these:

"Because HTML forms can only be sent as `GET` or `POST` requests unless client-side scripts are used,"

seeekr: I agree, that works as well, no preference on my side. The confusion from the original wording would be gone either way.

jebrosen:
Merged in 06edd5f. Thanks for bringing this up!

PR #958, opened 2019-03-28
Allow base16-encoded secret keys

notriddle:
Why
---

It's basically a workaround for Heroku, which has built-in support for generating secret keys, but generates 256-bit hex-encoded strings, instead of 256-bit base64-encoded ones.

Right now, I have to work around it in my Procfile, which isn't bad, but it adds embarrasing friction to what would otherwise be a pretty clean 12-factor app wrapper.
<https://github.com/notriddle/more-interesting/blob/ac944cc97c26571f799d090bb8dae78c9a3b98cb/Procfile>

What alternatives are there?
----------------------------

We could, instead, lift the format requirements entirely. We'd run PKDIF2 or something on the secret_key, generating a 256-bit secret no matter what the input actually is. I don't think that's better, because it means the user could use predictable, memorable "passwords" instead of something truly random.

Leaving this alone is also an option, but it means punting to the application to handle converting this, and getting it wrong can result in subtle disasters (like generating a secret of `0` because of a typo).

### DIFF
--- a/core/lib/Cargo.toml
+++ b/core/lib/Cargo.toml
@@ -34,6 +34,7 @@ state = "0.4.1"
 time = "0.1"
 memchr = "2" # TODO: Use pear instead.
 base64 = "0.10"
+base16 = "0.2"
 pear = "0.1"
 atty = "0.2"
 
--- a/core/lib/src/config/config.rs
+++ b/core/lib/src/config/config.rs
@@ -10,7 +10,7 @@ use crate::config::{Table, Value, Array, Datetime};
 use crate::http::private::Key;
 
 use super::custom_values::*;
-use {num_cpus, base64};
+use {num_cpus, base16, base64};
 
 /// Structure for Rocket application configuration.
 ///
@@ -298,7 +298,7 @@ impl Config {
     ///   * **workers**: Integer (16-bit unsigned)
     ///   * **keep_alive**: Integer
     ///   * **log**: String
-    ///   * **secret_key**: String (256-bit base64)
+    ///   * **secret_key**: String (256-bit base64 or base16)
     ///   * **tls**: Table (`certs` (path as String), `key` (path as String))
     pub(crate) fn set_raw(&mut self, name: &str, val: &Value) -> Result<()> {
         let (id, ok) = (|val| val, |_| Ok(()));
@@ -423,11 +423,11 @@ impl Config {
     }
 
     /// Sets the `secret_key` in `self` to `key` which must be a 256-bit base64
-    /// encoded string.
+    /// or base16 encoded string.
     ///
     /// # Errors
     ///
-    /// If `key` is not a valid 256-bit base64 encoded string, returns a
+    /// If `key` is not a valid 256-bit encoded string, returns a
     /// `BadType` error.
     ///
     /// # Example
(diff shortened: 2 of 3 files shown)


jebrosen:
> It's basically a workaround for Heroku, which has built-in support for generating secret keys, but generates 256-bit hex-encoded strings, instead of 256-bit base64-encoded ones.

Is this documented? I have only been able to find documentation and blog posts that say things along the lines of "now set your SECRET_KEY to <example>". If you are referring to https://github.com/will/securekey, it has recently been switched (in the `master` branch) to hex-encoding instead of base16.

That being said, this looks okay to me minus the build failures.

notriddle:
> hex-encoding instead of base16

Hexadecimal *is base16*.

notriddle:
Looking at [this PR](https://github.com/will/securekey/pull/7), which is what I think you were talking about, what they did is migrate from generating a random number and encoding it as hex (which strips leading zeroes, and thus might produce a string that is shorter than 64 characters even though they're gathering 256 bits of entropy) to grabbing a 64-character hex number (which may start with leading zeroes).

notriddle:
Okay, I've updated the guide.

core/lib/src/config/config.rs
```diff
@@ -439,20 +439,31 @@ impl Config {
...
     /// assert!(config.set_secret_key(key).is_ok());
+    /// let key = "fe4c5b09a9ac372156e44ce133bc940685ef5e0394d6e9274aadacc21e4f2643";
+    /// assert!(config.set_secret_key(key).is_ok());
     /// assert!(config.set_secret_key("hello? anyone there?").is_err());
     /// ```
     pub fn set_secret_key<K: Into<String>>(&mut self, key: K) -> Result<()> {
         let key = key.into();
         let error = self.bad_type("secret_key", "string",
-                                  "a 256-bit base64 encoded string");
+                                  "a 256-bit base64/base16 encoded string");
```
jebrosen: Nit: `base64/base16` -> `base64 or base16`

site/guide/9-configuration.md
```diff
@@ -92,9 +92,9 @@ limits = { forms = 32768 }
 The `workers` and `secret_key` default parameters are computed by Rocket
 automatically; the values above are not valid TOML syntax. When manually
 specifying the number of workers, the value should be an integer: `workers =
-10`. When manually specifying the secret key, the value should a 256-bit base64
-encoded string. Such a string can be generated using a tool such as openssl:
-`openssl rand -base64 32`.
+10`. When manually specifying the secret key, the value should a randum 256-bit
```
jebrosen: `randum` -> `random`

core/lib/Cargo.toml
```diff
@@ -33,6 +33,7 @@ state = "0.4.1"
 time = "0.1"
 memchr = "2" # TODO: Use pear instead.
 base64 = "0.10"
+base16 = "0.1"
```
jebrosen: Let's update this to base16 0.2

notriddle:
That should resolve all of those questions.

notriddle:
The test suite is busted by the uitest thing. [Here's an identical failure in a pull request other than mine.](https://dev.azure.com/SergioBenitez/Rocket/_build/results?buildId=246&view=logs&jobId=fe4e408a-5576-5fa1-f2ab-61ac24819df6&taskId=7f0864f1-1a96-5cff-86e3-45ee63058371&lineStart=1343&lineEnd=1361&colStart=1&colEnd=1)

notriddle:
https://github.com/SergioBenitez/Rocket/pull/1172 is this same PR, rebased onto the async branch.

jebrosen:
Can you rebase this on `master` to trigger a CI run?

notriddle:
@jebrosen K

notriddle:
The Linux and Windows builds are passing, but something's up with macOS. It doesn't look like anything I did, though.

jebrosen:
Merged in 121210c. Thanks so much for your patience on this and for the reminders to look at it.

MacOS CI has broken in a new way - it looks like something went wrong with a path for linking `mysqlclient`.

PR #968, opened 2019-04-10
Implementing FromIterator on JsonValue

adamAndMath:
Exposing the `FromIterator` implementation from `serde_json::Value` on JsonValue makes it more rare to need to unpack it.

This makes code cleaner without overhead.

### DIFF
--- a/contrib/lib/src/json.rs
+++ b/contrib/lib/src/json.rs
@@ -280,7 +280,7 @@ impl From<serde_json::Value> for JsonValue {
 
 impl<T> FromIterator<T> for JsonValue where serde_json::Value: FromIterator<T> {
     fn from_iter<I: IntoIterator<Item=T>>(iter: I) -> Self {
-        JsonValue(iter.into_iter().collect())
+        JsonValue(serde_json::Value::from_iter(iter))
     }
 }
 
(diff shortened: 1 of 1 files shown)


contrib/lib/src/json.rs
```diff
@@ -277,6 +278,12 @@ impl From<serde_json::Value> for JsonValue {
     }
 }
 
+impl<T> FromIterator<T> for JsonValue where serde_json::Value: FromIterator<T> {
+    fn from_iter<I: IntoIterator<Item=T>>(iter: I) -> Self {
+        JsonValue(iter.into_iter().collect())
```
jebrosen: This `collect()` looks unnecessary and potentially expensive for large items -- `JsonValue(serde_json::Value::from_iter(iter))` should suffice.

adamAndMath:
Needing explicit `JsonValue(...)` breaks the linearity of iterator manipulation.
```rust
JsonValue(
    list.iter()
        .filter(...)
        .map(...)
        ...
        .collect()
)
```
This can be circumvented with an `into()`, but I find my self needing to wrap my value further, and type inference for `into().into()` is not possible.

jebrosen:
Implemented in fd05f99.

PR #976, opened 2019-04-22
fix README example

Kjuvi:
The route points to "/\<name>/\<age>" but the main function mount "/hello"


jebrosen:
This change would result in `/hello/hello/<name>/<age>` - `/hello` passed into `mount()` is combined with the route path.

Kjuvi:
woops, seems I haven't watch this close enough, my bad!

Issue #975, opened 2019-04-17
ring dependency out of date


brndnmtthws:
I'm trying to get gRPC w/ rustls working, and recently bumped into a conflict with Rocket and tokio-rustls. Specifically, the ring dependency for Rocket (and all the deps it pulls in) is 0.13, whereas everything else is on 0.14.

```
error: failed to select a version for `ring`.
    ... required by package `cookie v0.11.0 (https://github.com/alexcrichton/cookie-rs.git#c59d94e1)`
    ... which is depended on by `rocket_http v0.4.0 (https://github.com/SergioBenitez/Rocket#7ab1c427)`
    ... which is depended on by `rocket v0.4.0 (https://github.com/SergioBenitez/Rocket#7ab1c427)`
    ... which is depended on by `api-frontend v0.1.0 (api-frontend)`
versions that meet the requirements `^0.14.0` are: 0.14.6, 0.14.5, 0.14.4, 0.14.3, 0.14.2, 0.14.1, 0.14.0

the package `ring` links to the native library `ring-asm`, but it conflicts with a previous package which links to `ring-asm` as well:
package `ring v0.13.5`
    ... which is depended on by `rustls v0.14.0`
    ... which is depended on by `hyper-sync-rustls v0.3.0-rc.4`
    ... which is depended on by `rocket_http v0.4.0 (https://github.com/SergioBenitez/Rocket#7ab1c427)`
    ... which is depended on by `rocket v0.4.0 (https://github.com/SergioBenitez/Rocket#7ab1c427)`
    ... which is depended on by `api-frontend v0.1.0 (api-frontend)`
```

It seems like @SergioBenitez and @alexcrichton own these repos. Can we get the versions bumped in git?

Maybe you can consider adding a service like dependabot to your repos as well, which will keep the dependencies up to date.

jebrosen:
Updating ring's major version is a breaking change (in each project affected), so it cannot happen until at least cookie 0.12 and Rocket 0.5. It would be annoying to do that version bump solely because of ring's version, but it might be necessary if all of 0.13 does get yanked.

In the meantime, is there an older version of tokio-rustls that you can depend on that uses ring 0.13? If not, I think you would have to fork and/or patch some dependencies as long as the versions are not compatible.

---

I typed out the above without realizing your comment about git. Yes, it should be "safe" to do this in the master branches of these repositories.

brndnmtthws:
[...]

PR #986, opened 2019-05-02
Bump cookie crate to 0.12.

brndnmtthws:
This fixes the dependency issues with cookie (as mentioned in #975).

### DIFF
--- a/core/http/Cargo.toml
+++ b/core/http/Cargo.toml
@@ -26,7 +26,7 @@ time = "0.1"
 indexmap = "1.0"
 rustls = { version = "0.14", optional = true }
 state = "0.4"
-cookie = { version = "0.11", features = ["percent-encode"] }
+cookie = { version = "0.12", features = ["percent-encode"] }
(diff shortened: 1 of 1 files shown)


brndnmtthws:
I forgot about that conflict with `hyper-sync-rustls`. On my build I just disable TLS because it's broken.

jebrosen:
Yeah, `hyper-sync-rustls` would need to be bumped first. I'm pretty sure that and `cookie` are the only things that use `ring`.

SergioBenitez:
Just published `hyper-sync-rustls v0.3.0-rc.5`.

brndnmtthws:
Updated with latest `hyper-sync-rustls`.

Pierozi:
I'm using `jsonwebtoken` which are in conflict. I cannot even downgrade to `5.0.1` because the feature related to encryption ES256 are not present.

jebrosen:
Merged in ba6aa2f along with other *ring* dependencies.

PR #973, opened 2019-04-13
Add error handling for too large JSON payloads

Nilix007:
Hi,

This PR adds error handling for too large JSON payloads. It also adds a new `JsonError` `PayloadTooLarge` so that this case can be handled properly. Fixes #972.

Old behaviour for too large JSON payloads:
```
POST / application/json:
    => Matched: POST / (x)
    => Warning: Data left unread. Force closing network stream.
    => Error: Couldn't parse JSON body: Error("EOF while parsing a string", line: 1, column: 1048576)
    => Outcome: Failure
    => Warning: Responding with 400 Bad Request catcher.
    => Response succeeded.
```

New behaviour:
```
POST / application/json:
    => Matched: POST / (x)
    => Error: JSON payload too large.
    => Warning: Data left unread. Force closing network stream.
    => Outcome: Failure
    => Warning: Responding with 413 Payload Too Large catcher.
    => Response succeeded.
```

PR #990, opened 2019-05-06
[WIP] Add "read with limit" functionality to `DataStream`

Nilix007:
Hey there,

as suggested in https://github.com/SergioBenitez/Rocket/pull/973#issuecomment-482831877, Rocket should provide an easy way to read from the HTTP body with an upper limit on the body size. (As this is needed for `Form`, `JSON`, etc.)

Let's start off with this proof of concept which adds `DataStream::read_to_string_with_limit` and `DataStream::read_to_end_with_limit` together with a new error type `LimitReadError`.

@jebrosen: What do you think?

### DIFF
--- a/core/lib/src/data/data_stream.rs
+++ b/core/lib/src/data/data_stream.rs
@@ -16,6 +16,43 @@ pub type InnerStream = Chain<Cursor<Vec<u8>>, BodyReader>;
 /// be used as an opaque [`Read`] structure.
 pub struct DataStream(crate InnerStream);
 
+impl DataStream {
+    pub fn read_to_string_with_limit(
+        &mut self,
+        buf: &mut String,
+        limit: usize,
+    ) -> Result<usize, LimitReadError> {
+        self.do_with_limit(limit, |r| r.read_to_string(buf))
+    }
+
+    pub fn read_to_end_with_limit(
+        &mut self,
+        buf: &mut Vec<u8>,
+        limit: usize,
+    ) -> Result<usize, LimitReadError> {
+        self.do_with_limit(limit, |r| r.read_to_end(buf))
+    }
+
+    fn do_with_limit<F: FnOnce(&mut io::Take<&mut Self>) -> io::Result<T>, T>(
+        &mut self,
+        limit: usize,
+        f: F,
+    ) -> Result<T, LimitReadError> {
+        let mut r = self.by_ref().take(limit as u64 + 1);
+        let s = f(&mut r).map_err(LimitReadError::Io)?;
+        if r.limit() > 0 {
+            Ok(s)
+        } else {
+            Err(LimitReadError::LimitReached)
+        }
+    }
+}
+
+pub enum LimitReadError {
+    LimitReached,
+    Io(io::Error),
+}
+
 // TODO: Have a `BufRead` impl for `DataStream`. At the moment, this isn't
 // possible since Hyper's `HttpReader` doesn't implement `BufRead`.
 impl Read for DataStream {


core/lib/src/data/data_stream.rs
```diff
@@ -16,6 +16,43 @@ pub type InnerStream = Chain<Cursor<Vec<u8>>, BodyReader>;
...
+
+    pub fn read_to_end_with_limit(
+        &mut self,
+        buf: &mut Vec<u8>,
+        limit: usize,
+    ) -> Result<usize, LimitReadError> {
+        self.do_with_limit(limit, |r| r.read_to_end(buf))
+    }
+
+    fn do_with_limit<F: FnOnce(&mut io::Take<&mut Self>) -> io::Result<T>, T>(
```
hellow554: why `Take` and not `&mut dyn Read`?

jebrosen: `F: FnOnce(impl Read)` is not currently allowed and a desugaring like `for<R: Read> FnOnce(R)` doesn't exist either as far as I know. Exposing `Take` isn't something we should do if this API were to be made public, so I think `&mut dyn Read` is the only way to erase the type here.

core/lib/src/data/data_stream.rs
```diff
@@ -16,6 +16,43 @@ pub type InnerStream = Chain<Cursor<Vec<u8>>, BodyReader>;
...
+        &mut self,
+        limit: usize,
+        f: F,
+    ) -> Result<T, LimitReadError> {
+        let mut r = self.by_ref().take(limit as u64 + 1);
```
jebrosen: Using `take` with 1 more than the real limit is clever. Maybe a bit too clever -- I want to double check the edge cases.

Nilix007:
Let me add some motivation for my API proposal:
- All the code I've seen so far that deals with body data in Rocket uses either `Read::read_to_end` or `Read::read_to_string`.  Also, all of these have some kind of size limit (for obvious reasons).
- Probably most of the code wants to handle the specific error case of a too large payload, eg. for returning a 429 status code. `io::Error` does not seem a good error type for this case because the API consumers would need to check the error kind and try to downcast the inner error. A separate enum based error type is IMHO more convenient to use.

All in all, I'd start off with just the two functions and keep the `do_with_limit` private.  If different use cases evolve, we can add a more comprehensive API.

Nilix007:
@jebrosen: If you're OK with this approach, I would continue with adding documentation, some tests for the edge cases, and update all the body data consumers in Rocket and contrib.

(I'm also unsure whether `take(limit + 1).read_to_end(&mut buf)` might overallocate by a factor of 2 when the input data exceeds the limit)

jebrosen:
I think it's a good idea to add this in some form, maybe with some name/API changes. Users have often wanted  `Vec<u8>` or `String` in release builds; this will provide that flexibility without the same degree of risk and also make it easier to write certain `FromData`/`FromDataSimple` implementations with limit support.

In the documentation, I would definitely want to 1) encourage any potential users of these functions to use a user-configurable [Limit](https://api.rocket.rs/v0.4/rocket/config/struct.Limits.html#method.limit) and 2) strongly caution against using something like `usize::max_value()` as the limit because it's a trivial denial of service vector.

jebrosen:
I want to revisit this, and I have some quick thoughts on the API:

* I think the methods should be on `Data` and take `self` by move
* I also think these methods should allocate a new `String` or `Vec` themselves

For example, `rocket_contrib::msgpack`'s `FromData` implementation:

```rust
   let size_limit = r.limits().get("msgpack").unwrap_or(LIMIT);
   let mut buf = Vec::new();
   let mut reader = d.open().take(size_limit);
   match reader.read_to_end(&mut buf).await {
       Ok(_) => Borrowed(Success(buf)),
       Err(e) => Borrowed(Failure((Status::BadRequest, Error::InvalidDataRead(e)))),
   }
```

becomes

```rust
    let size_limit = r.limits().get("msgpack").unwrap_or(LIMIT);
    match d.read_to_end_with_limit(size_limit).await {
        Ok(buf) => Borrowed(Success(buf)),
        Err(e) => Borrowed(Failure((Status::BadRequest, Error::InvalidDataRead(e))))
    }
```

This way the methods would do what a lot of users do already, and anyone who needs to do something more complicated can still call `open` and `read_to_end` manually.

@SergioBenitez any thoughts on the above?

SergioBenitez:
I think we should go even further. Here's my proposed API:

```rust
impl Data {
    fn open(limit: usize) -> DataStream;
    fn peek(&self) -> &[u8];
    fn peek_complete(&self) -> bool;
}

impl DataStream {
    fn stream_to<W: Write>(self, writer: &mut W) -> io::Result<u64>;
    fn stream_to_file<P: AsRef<Path>>(self, path: P) -> io::Result<u64>;
    fn stream_to_string(self) -> io::Result<String>;
    fn stream_to_vec(self) -> io::Result<Vec<u8>>;
}

impl Read for DataStream { /* ... */ }
```

I can't think of a valid reason to _ever_ read without a limit. So, let's enforce it.

We might even consider abusing `unsafe` by adding an `open_unsafe()` method that sets no limit on the returned `DataStream`.

jebrosen:
I like those methods as well. I suspect `open(usize::MAX)` (please never do that) is close enough to `open_unsafe` it's not worth a separate method.

The name `stream_to_string` feels a bit wrong to me compared to `read_to_string` -- but I do like that it gives four methods whose names all start with `stream_to_`.

SergioBenitez:
Given that `DataStream` implements `Read`, it'll also have the `read_to` methods. I wanted to avoid the naming conflict.

SergioBenitez:
I'd like a comprehensive approach to this. I've opened https://github.com/SergioBenitez/Rocket/issues/1325 to track such an effort. Let's move the discussion there until we have a concrete plan.

PR #997, opened 2019-05-08
Set the 'Server' header to a default value only if it isn't already set

lu4nm3:


SergioBenitez:
Thanks for this! Can you add a regression test for this in `core/lib/tests`?

lu4nm3:
This should do it! :)

core/lib/tests/conditionally-set-server-header-997.rs
```diff
@@ -0,0 +1,43 @@
+#![feature(proc_macro_hygiene, decl_macro)]
+
+#[macro_use]
+extern crate rocket;
+extern crate rocket_http;
+
+use rocket::response::Redirect;
```
jebrosen: Travis warns that this line is unused.

SergioBenitez:
Merged in https://github.com/SergioBenitez/Rocket/commit/0a978fdb5dae84d0f98c4352b144a2a50b1e3b56 with a bit of test clean-up in https://github.com/SergioBenitez/Rocket/commit/9f75e0f443b3ccf678fe915668e0e7f5e1f89273. Thanks so much!

lu4nm3:
Thank you for the quick turn around! Pardon my ignorance as I'm fairly new to Rust and am still learning about the ecosystem and the community but, what would be the fastest way for me to uptake these changes into my application? Not sure if it can only be done once the next official release is made or if there is perhaps a nightly version that gets published somewhere.

SergioBenitez:
@lu4nm3 The easiest thing to do is to depend on Rocket from the git repository instead of an actual version (in Cargo.toml):

```diff
- rocket = "0.4"
+ rocket = { git = "https://github.com:SergioBenitez/Rocket.git" }
```

See the [cargo docs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories) for more.

Issue #90, opened 2017-01-01
Native WebSocket support


net:
I think Rocket would benefit heavily from native WebSocket support. Even better would be something similar to [Phoenix's channels](http://www.phoenixframework.org/docs/channels).

It's great to see such a comprehensive web framework for Rust coming together!

jsen-:
@net did you manage to combine `rocket` with eg. [websocket](https://crates.io/crates/websocket) on the same `host:port`?
I'm trying to `upgrade` connection to `ws`, but no luck so far :(

SergioBenitez:
This is indeed planned.

wagenet:
@SergioBenitez got any suggestions for how one could manually add web socket support?

SergioBenitez:
@wagenet Do you mean to Rocket or to your web application? The former is subtle is it requires a lot of design work, but the latter should be straightforward. You should be able to use the [`websocket`](https://github.com/cyderize/rust-websocket) or [`ws-rs`](https://github.com/housleyjk/ws-rs) crates as they are intended. You won't be able to use the same port for both servers, of course. And, you'll need to spawn one of the two servers in a separate thread so that one server doesn't block the other one from starting. Should you choose to do this, your `main` function would look something like:

```rust
fn main() {
    thread::spawn(|| {
        listen("127.0.0.1:3012", |out| {
            move |msg| {
                out.send(msg)
           }
        })
    })

    rocket.ignite()..more()..launch()
}
```

wagenet:
@SergioBenitez thanks :) I ended up discovering this sort of approach myself.

jhpratt:
@SergioBenitez is this waiting on a future version of Rocket that migrates to Actix or some other async framework?

sparky8251:
Is there no way to add web socket support on the same port as HTTP?

If not, this is quite problematic for me.

jhpratt:
@sparky8251 Not currently, at least easily.

sparky8251:
Well, that sucks. Liked the readability of Rocket but without the ability to have websockets on the same port Rocket is useless to me.

Actix just doesnt have the same readability after all... If you happen to know a way to do web sockets on the same port, even if its not clean, I'd love to hear it. If I end up porting to Actix I won't be returning to Rocket but if I can make the web socket stuff work, hopefully I can stick around until native support lands.

jebrosen:
[...]

PR #1008, opened 2019-05-19
Upgrade to hyper 0.12 (WIP)

schrieveslaach:
In order to support websockets, it is necessary to upgrade to the newest hyper version (see [this comment](https://github.com/SergioBenitez/Rocket/issues/90#issuecomment-488267042) in #90). This PR is a simple proof of concept to upgrade to hyper 0.12 which was a bit difficult and I hope, that we can use this PR as a WIP PR to upgrade to hyper 0.12. 

I hope that you can support me to finish the upgrade.

### DIFF
--- a/core/http/Cargo.toml
+++ b/core/http/Cargo.toml
@@ -22,7 +22,9 @@ private-cookies = ["cookie/secure"]
 [dependencies]
 smallvec = "0.6"
 percent-encoding = "1"
-hyper = { version = "0.10.13", default-features = false }
+hyper = { version = "0.12.31", default-features = false, features = ["tokio"] }
+http = "0.1.17"
+mime = "0.3.13"
 time = "0.1"
 indexmap = "1.0"
 rustls = { version = "0.15", optional = true }
--- a/core/http/src/content_type.rs
+++ b/core/http/src/content_type.rs
@@ -6,7 +6,7 @@ use std::fmt;
 use crate::header::Header;
 use crate::media_type::{MediaType, Source};
 use crate::ext::IntoCollection;
-use crate::hyper::mime::Mime;
+use mime::Mime;
 
 /// Representation of HTTP Content-Types.
 ///
@@ -281,11 +281,11 @@ impl From<Mime> for ContentType {
     #[inline]
     fn from(mime: Mime) -> ContentType {
         // soooo inefficient.
-        let params = mime.2.into_iter()
+        let params = mime.params().into_iter()
             .map(|(attr, value)| (attr.to_string(), value.to_string()))
             .collect::<Vec<_>>();
 
-        ContentType::with_params(mime.0.to_string(), mime.1.to_string(), params)
+        ContentType::with_params(mime.type_().to_string(), mime.subtype().to_string(), params)
     }
 }
 
--- a/core/lib/src/rocket.rs
+++ b/core/lib/src/rocket.rs
@@ -1,14 +1,22 @@
 use std::collections::HashMap;
-use std::str::from_utf8;
+use std::convert::From;
+use std::str::{from_utf8, FromStr};
 use std::cmp::min;
 use std::io::{self, Write};
 use std::time::Duration;
 use std::mem;
+use std::net::{IpAddr, SocketAddr, ToSocketAddrs};
+use std::sync::Arc;
+
+use futures::{Future, Stream};
+use futures::future::{self, FutureResult};
 
 use yansi::Paint;
 use state::Container;
+use tokio::net::TcpListener;
+use tokio::prelude::{Future as _, Stream as _};
 
-#[cfg(feature = "tls")] use crate::http::tls::TlsServer;
+#[cfg(feature = "tls")] use crate::http::tls::TlsAcceptor;
 
 use crate::{logger, handler};
 use crate::ext::ReadExt;
@@ -37,23 +45,59 @@ pub struct Rocket {
     fairings: Fairings,
 }
 
+struct RocketArcs {
+    config: Arc<Config>,
+    router: Arc<Router>,
+    default_catchers: Arc<HashMap<u16, Catcher>>,
+    catchers: Arc<HashMap<u16, Catcher>>,
+    state: Arc<Container>,
+    fairings: Arc<Fairings>,
+}
+
+impl<Ctx> hyper::MakeService<Ctx> for RocketArcs {
+    type ReqBody = hyper::Body;
+    type ResBody = hyper::Body;
+    type Error = hyper::Error;
+    type Service = RocketHyperService;
+    type Future = FutureResult<Self::Service, Self::MakeError>;
+    type MakeError = Self::Error;
+
+    fn make_service(&mut self, _: Ctx) -> Self::Future {
+        future::ok(RocketHyperService::new(self))
+    }
+}
+
+#[derive(Clone)]
+pub struct RocketHyperService {
+    config: Arc<Config>,
+    router: Arc<Router>,
+    default_catchers: Arc<HashMap<u16, Catcher>>,
+    catchers: Arc<HashMap<u16, Catcher>>,
+    state: Arc<Container>,
+    fairings: Arc<Fairings>,
+}
+
 #[doc(hidden)]
-impl hyper::Handler for Rocket {
+impl hyper::Service for RocketHyperService {
+    type ReqBody = hyper::Body;
+    type ResBody = hyper::Body;
+    type Error = hyper::Error;
+    //type Future = FutureResult<hyper::Response<Self::ResBody>, Self::Error>;
+    type Future = Box<future::Future<Item = hyper::Response<Self::ResBody>, Error = Self::Error> + Send>;
+
     // This function tries to hide all of the Hyper-ness from Rocket. It
     // essentially converts Hyper types into Rocket types, then calls the
     // `dispatch` function, which knows nothing about Hyper. Because responding
     // depends on the `HyperResponse` type, this function does the actual
     // response processing.
-    fn handle<'h, 'k>(
-        &self,
-        hyp_req: hyper::Request<'h, 'k>,
-        res: hyper::FreshResponse<'h>,
-    ) {
-        // Get all of the information from Hyper.
-        let (h_addr, h_method, h_headers, h_uri, _, h_body) = hyp_req.deconstruct();
+    fn call<'h>(
+        &mut self,
+        hyp_req: hyper::Request<Self::ReqBody>,
+    ) -> Self::Future {
+        let (parts, body) = hyp_req.into_parts();
 
         // Convert the Hyper request into a Rocket request.
-        let req_res = Request::from_hyp(self, h_method, h_headers, h_uri, h_addr);
+        let req_res = Request::from_hyp(&self.config, &self.state, &parts);
         let mut req = match req_res {
             Ok(req) => req,
             Err(e) => {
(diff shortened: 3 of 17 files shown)


jebrosen:
Seeing this open I realize I should have shared my progress sooner. I have an async fork of Rocket that supports async routes, `FromData`, and (IIRC) `Response` and `Body` but still has some serious blockers before a thorough review can be done. It looks like this PR leaves out a few key things, but I might not be able to do a real comparison for a while still.

schrieveslaach:
@jebrosen, I didn't know that you are working on an update as well. BTW, it sounds very nice what you are doing.

Do you think we should continue this PR and should use your insights to finish this update? This PR has some issue in it and Rocket could benefit from your and others point of view. Based on the upgrade Rocket could start integrate async routes, etc.

SergioBenitez:
@jebrosen @schrieveslaach Can I task you with working together to create a comprehensive PR? We should do this in the open to avoid duplicating efforts in the future. Perhaps we can start by having you share your work, @jebrosen, and perhaps you can take the lead to incorporate @schrieveslaach contributions into your ongoing effort.

jebrosen:
I have rebased and updated the comments on my branch: https://github.com/SergioBenitez/Rocket/compare/master...jebrosen:async?expand=1. I have been working on it on and off for several months now, but churn around `futures_api` and `await_macro` made it a bit challenging to power through.

Here's what I was thinking, roughly in order:

1. Update at least core and codegen, but probably all crates in Rocket to the 2018 edition. I personally think it would be a good idea to also add `#![warn(rust_2018_idioms)]`. This is needed for the `async_await` feature which is a significant benefit for Rocket and crates that use it. `cargo fix --edition` works mostly fine for this task.
2. Combine the efforts of this PR with what I have worked on. I skimmed this PR and I *think* my work includes (in some form) most of these changes and more, with the exception of how `MakeService` is handled. I should be able to read though these changes more carefully this weekend at the latest and see if I missed anything else.
3. There are a few things in `core` that are unfinished and need to be worked on:
    * `Cookies` now uses a weird `Mutex` hack that I'm pretty sure is wrong. I just did it to get past the errors and work on fixing other more interesting things
    * Some low-level things including keep-alive, closing the data stream on drop, and retrieving the client address of an incoming connection
    * Everything to do with `Client`, `LocalRequest`, etc. used everywhere in tests.
4. `Response`/`Body` needs some design work. As I have written it, all responses are buffered into a `Vec` before being sent to the client. This is *extremely wrong* (but at least it does it asynchronously! 😄)
    * We additionally want to provide an `AsyncWrite`-like interface in particular. Hyper has [`Body::channel`](https://docs.rs/hyper/0.12.21/hyper/struct.Body.html#method.channel), which we probably don't want to use directly long-term but might be usable to hash out design or serve as some inspiration for something better.
5. Work on making the API nicer. I am tired of repeatedly typing out things like `Pin<Box<Future<Output = T> + Send + 'a>>`, for example.
6. Fix `rocket_contrib` and examples.

... and more after that. I have more notes and ideas of things that need to be done, but I have to leave it at this for right now.

schrieveslaach:
@SergioBenitez, sure. :smiley: 

@jebrosen, thanks for your comments. 
In my PR I tried to make as less breaking changes as possible but I noticed very quickly that the [buffering of the request body was necessary for this](https://github.com/schrieveslaach/Rocket/commit/18ef5afabe0c2700d05e21a3d3ebe6a5c89744bc#diff-a9ea99dbf2f3bc38a7d7e4c36a99a288R106). :see_no_evil: 

How do you propose to continue? Should I wait for your comments and than I should start to adopt this PR?

jebrosen:
If you have the time and interest, I would appreciate if you could take a look at my work and share your opinions on anything we both did but in different ways. I'll do the same, but that might be a few days away still and this would help me know what to watch out for.

I will want to be careful about merging our code together, maybe in an "async" branch in this repository. You deserve credit for starting this, and I also want to keep the other work I've done.

schrieveslaach:
Okay, I will have a look at your work. 

> You deserve credit for starting this, and I also want to keep the other work I've done.

Thanks for your appreciation. I agree with you that we should incorporate our development branches.

core/http/src/content_type.rs
```diff
@@ -281,11 +281,11 @@ impl From<Mime> for ContentType {
     #[inline]
     fn from(mime: Mime) -> ContentType {
         // soooo inefficient.
-        let params = mime.2.into_iter()
```
jebrosen: I had dropped `mime` support entirely.

core/http/Cargo.toml
```diff
@@ -21,7 +21,9 @@ private-cookies = ["cookie/secure"]
 [dependencies]
 smallvec = "0.6"
 percent-encoding = "1"
-hyper = { version = "0.10.13", default-features = false }
+hyper = { version = "0.12.28", default-features = false, features = ["runtime"] }
```
jebrosen: I chose to use `tokio` manually instead of via the `runtime` feature (for easier TLS handling, I think).

core/lib/src/rocket.rs
```diff
@@ -22,38 +25,68 @@ use outcome::Outcome;
...
-    fairings: Fairings,
+    crate config: Arc<Config>,
+    router: Arc<Router>,
+    default_catchers: Arc<HashMap<u16, Catcher>>,
+    catchers: Arc<HashMap<u16, Catcher>>,
+    crate state: Arc<Container>,
+    fairings: Arc<Fairings>,
+}
+
+impl<Ctx> hyper::MakeService<Ctx> for Rocket {
```
jebrosen: This is one of the key structural differences between our approaches - I used `hyper::service_fn` as in https://github.com/jebrosen/Rocket/commit/27a59ef90dd6e1b978b77be8c0b675b8d547fb08#diff-a9ea99dbf2f3bc38a7d7e4c36a99a288R786. Do you have an opinion on which is better? I prefer `hyper::service_fn` because it's far shorter, but is there a benefit to explicitly specifying the associated types by implementing `MakeService` manually?

crusadergo:
really looking forward when this MR will be ready :)

core/http/Cargo.toml
schrieveslaach: Okay, I can change this as well.

core/lib/src/rocket.rs
schrieveslaach: I had some struggle with `hyper::service_fn` and the idea of using had seemed to me very handy and more robust against further changes in hyper. But that's only a gut feeling. Nevertheless, I think we have to use some kind of cloning of `Arc<SomeRocketStruct>` here.

core/http/src/content_type.rs
schrieveslaach: From your comment above, I think you want to keep `mime` support, correct?

jebrosen: Yes, if there's no downside, at least. I think it would just add non-async-related churn to an async development branch for now.

core/http/Cargo.toml
jebrosen: Feel free to change it in your branch, or not - I foresee a big project of a merge one of the coming weekends either way!

nicoburns: @jebrosen One thing to keep in mind is that people will definitely want to run stuff other than Rocket on the Tokio event loop (e.g. async database calls using tokio-postgres). So either:

1. Rocket should (optionally) accept a tokio reactor to run on
2. Rocket should be able to expose it's own reactor for use by other code
3. The tokio runtime should be used (I *think* this allows for this).

Probably #1 with an option for #3 would be ideal?

jhpratt:
I probably wouldn't be of much help in porting the code to 0.12, but I can certainly take a look over code before merging if wanted.

SergioBenitez:
Rocket is now on Rust 2018! Looking forward to seeing progress here.

@jebrosen What are the next steps?

core/http/Cargo.toml
jebrosen: IIRC hyper 0.12 actually requires tokio runtime specifically (hyper 0.13 might not.) I definitely like the idea of allowing a user-configured runtime, though.

jebrosen:
@schrieveslaach, do you have any changes you made since this PR was opened?

The next step on my list is to rebase this PR on top of master, then my changes on top of that. That should give us master -> partial hyper 0.12 upgrade -> full async/await in core and codegen. Since it's a massive and incomplete breaking change that breaks tests left and right, I had the idea to start an `async` branch that can be pushed to / have PRs opened on.

Ideally we would have a way to divvy up tasks so there's not too much duplication of work and interested parties could pitch in -- @SergioBenitez do you have any opinions here as far as a tracking issue(s) or PR submission/review?

schrieveslaach:
@jebrosen, @SergioBenitez, I'll have time on sunday and I'll work on it again. I keep you posted.

schrieveslaach:
@jebrosen, @SergioBenitez, I rebased my fork with the current master and I improved my code a bit. For example, I restored some code that I had to disable. 

I also switched to tokio runtime. I used the code @jebrosen's fork. 

This PR still uses `MakeService` with some helper structs (eg. `RocketArcs` and `RocketHyperService`). Do you think we should keep this way or should I switch to `hyper::service_fn`?

jebrosen:
I will start working on merging these together now. I will likely make a few changes - e.g. using a single `Arc<Rocket>` instead of individual `Arc`s, so that it will be easier to combine them. But this looks great so far.

schrieveslaach:
@jebrosen, I'm looking forward to get this done. 

Just out of curiosity: do you create an async branch and do you accept this PR by merging it into the async branch?

apiraino:
> I also switched to tokio runtime. I used the code @jebrosen's fork

Just asking out of curiosity: why did you both end up choosing tokio instead of trying something else (e.g. Runtime)? Perhaps was the async story already started at that time and you just restarted from there?

Me too, I'm trying to figure out the "right" library to use and I find tokio pretty hard to use (disclaimer: not an experienced Rustacean).

Anyways, thank you so much for pushing this forward, I'm really looking forward to it :+1:

lnicola:
`runtime` is a very thin wrapper over `tokio` and `romio`, the latter being more of an experiment in API design for the new version of futures. Now `tokio` is moving to the standard library futures, and it's designed to offer much better performance and functionality than the alternatives (of which there aren't many, as I mentioned). So there's no reason not to use it.

In any case, the choice will be more or less invisible to Rocket users.

apiraino:
awesome @lnicola thanks for explaining! As long as the async story is ergonomic to use for "the rest of us", I'm super happy with whatever choice ;-)

jebrosen:
@schrieveslaach My `async` branch now includes your commit, but I still want to go back and re-resolve a few merges with "yours" instead of "mine".

---

I originally went with `tokio` because `hyper` uses it by default. Somehow I ended up thinking that `hyper` even depended specifically on `tokio`, which I don't think is the case. Assuming it works, we should allow the use of any runtime. And yes, for the vast majority of use cases the specific underlying runtime should not need to be worried about.

jebrosen:
Update: I got busy with a few other things, bit I plan to come back to this very soon. Thank you everyone for your patience.

My `async` branch is pretty close to being a fully async/await API, but some features are still unimplemented or seriously unvetted. I also want to take a look at the `master` branch of hyper, which is now using futures 0.3 and might let us drop all uses of `Compat`.

jebrosen:
I've merged this and my code as the `async` branch and opened #1065. Let's continue work starting there.

PR #1010, opened 2019-05-21
Add rustfmt to do nothing

derekdreery:
Basically this PR tells cargo fmt to do nothing if run. Would have saved me trouble when I accidentally ran it and so messed up some work I was doing.

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,2 @@
+unstable_features = true
+ignore = ["src", "contrib", "core", "examples"]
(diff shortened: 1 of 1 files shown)


SergioBenitez:
I don't believe we should commit files that take into affect only under some developer environments. This should be configured on a per-IDE basis. If your editor is running `rustfmt` automatically, perhaps it can be configured not to do unless a `rustfmt` file is present?

Thank you for the PR, however; I'm sure others will find this useful as documentation.

PR #1012, opened 2019-05-23
Implement Content Security Policy header

fcsonline:
The HTTP Content-Security-Policy response header allows web site administrators to control resources the user agent is allowed to load for a given page. With a few exceptions, policies mostly involve specifying server origins and script endpoints. This helps guard against cross-site scripting attacks (XSS).

More information:

- https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP
- https://w3c.github.io/webappsec-csp/2/

### DIFF
--- a/contrib/lib/src/helmet/policy.rs
+++ b/contrib/lib/src/helmet/policy.rs
@@ -96,6 +96,7 @@ impl_policy!(Frame, "X-Frame-Options");
 impl_policy!(Hsts, "Strict-Transport-Security");
 impl_policy!(ExpectCt, "Expect-CT");
 impl_policy!(Referrer, "Referrer-Policy");
+impl_policy!(ContentSecurityPolicy, "Content-Security-Policy");
 
 /// The [Referrer-Policy] header: controls the value set by the browser for the
 /// [Referer] header.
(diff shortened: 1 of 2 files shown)


jebrosen:
A similar PR was closed fairly recently: https://github.com/SergioBenitez/Rocket/pull/869#issuecomment-454625244

fcsonline:
Ouch! I didn't see the closed PR.

I think is interesting to have CSP included by default in helmet. It is one of the most important headers regarding the number of attacks it could block.

Although I can understand @SergioBenitez arguments about complexity of configuration, I don't understand the false sense of security argument. All security headers in helmet require a bit of understanding about which security implications have each one. What is more, I think CSP header is a bit easier to understand than other implemented headers.

We could work on the documentation to explain better each one of the headers giving an example of the vector attacks it tries to mitigate.

SergioBenitez:
I would be in favor of a type-safe representation of CSP that does not allow malformed versions of the header to be constructed, like the new `Permission` header. The implementation should not allocate unless necessary and should _strictly_ follow the RFC https://www.w3.org/TR/CSP3/#csp-header. Unfortunately this implementation doesn't quite adhere to this standard of implementation. As it is now significantly out of date (fault of mine!), I am closing this out.

PR #1014, opened 2019-05-25
Allow non_snake_case in query parameter codegen.

### LINKED ISSUE #1003: `non_snake_case` warning on generated route parameter
- Rocket Version: 0.4.1
- Question: `non_snake_case` warning for the rocket generated route parameter, e.g. `__rocket_param__id`.

## Actual Code:

``` rust
// The normal `check` route requiring `AdminUser` to be presented
#[get("/admin/data/check?<id>")]
pub fn check(_user: AdminUser, id: String) -> Result<Template, Redirect> { ... }

// The `check` route where `AdminUser` doesn't exist, simply redirect to "admin/login"
#[get("/admin/data/check?<_id>", rank=2)]
pub fn check_fail(_id: String) -> Redirect { Redirect::to("/admin/login") }
```

And the warning generated by rust:

```
warning: variable `__rocket_param__id` should have a snake case name
  --> src/routes/pages/admin/data/check.rs:34:19
   |
34 | pub fn check_fail(_id: String) -> Redirect {
   |                   ^^^ help: convert the identifier to snake case: `__rocket_param_id`
   |
   = note: #[warn(non_snake_case)] on by default
```

In this example, I'm trying to implement `check_fail` backing up the normal `check` route when there's no `AdminUser` presented. (I think this is almost identical to one of the example usage of `route` so I guess I don't need to elaborate a ton on this.)

Of course I don't need the parameter `id` in my `check_fail` function because I'm just redirecting it to `/admin/login` (as there's no `AdminUser` presented). But I still need the `?<id>` part in the route because otherwise this wouldn't match the original admin route. If I use `id` here then rust compiler will be yelling at me saying that "you are not using this `id` in the function, so you should change it to `_id`". Ok now I change the parameter name to `_id`, and rust is complaining that the rocket generated parameter name `__rocket_param__id` not matching the snake case (because there are two consecutive `_` before `id`.)

Usually this kind of unused variable problem will be fixed just by prepending `_`. But right now even if I add `_` it wouldn't get through the warning. Rocket also does not allow me to use an ignored `_` as the variable name. So what do we usually do in this case?

**Potential Solution**: Add a check on Rocket's side checking whether a parameter starts with an underscore (`_`). If it is then you don't append a trailing `_` at the end of `__rocket_param`. So if my variable is called `_id`, the generated variable name will be `__rocket_param_id` instead of `__rocket_param__id`. Therefore the rust compiler will not complain about the snake case violation.

jebrosen:
Fixes #1003.

Tested by adding the following to `examples/hello_person/src/main.rs`:

```rust
#![deny(non_snake_case)]

#[get("/hello2?<_name>")]
pub fn hello2(_name: String) -> String {
    _name
}
```

Although, it might be applying `allow()`s in more places than are strictly necessary.

This commit can be safely backported to 0.4 as well.

### DIFF
--- a/core/codegen/tests/route.rs
+++ b/core/codegen/tests/route.rs
@@ -1,5 +1,7 @@
 #![feature(proc_macro_hygiene, decl_macro)]
 
+#![deny(non_snake_case)]
+
 #[macro_use] extern crate rocket;
 
 use std::path::PathBuf;
@@ -63,6 +65,10 @@ fn post2(
     let uri = uri!(post2: a, name.url_decode_lossy(), path, sky, query.into_inner());
 
     format!("({}) ({})", string, uri.to_string())
+}
+
+#[post("/<_unused_param>?<_unused_query>", data="<_unused_data>")]
+fn test_unused_params(_unused_param: String, _unused_query: String, _unused_data: Data) {
 }
 
 #[test]
--- a/core/codegen/tests/route.rs
+++ b/core/codegen/tests/route.rs
@@ -1,5 +1,8 @@
 #![feature(proc_macro_hygiene, decl_macro)]
 
+// Rocket sometimes generates mangled identifiers that activate the
+// non_snake_case lint. We deny the lint in this test to ensure that
+// code generation uses #[allow(non_snake_case)] in the appropriate places.
 #![deny(non_snake_case)]
 
 #[macro_use] extern crate rocket;
--- a/core/codegen/src/attribute/route.rs
+++ b/core/codegen/src/attribute/route.rs
@@ -220,9 +220,11 @@ fn query_exprs(route: &Route) -> Option<TokenStream2> {
 
         let decl = match segment.kind {
             Kind::Single => quote_spanned! { span =>
+                #[allow(non_snake_case)]
                 let mut #ident: Option<#ty> = None;
             },
             Kind::Multi => quote_spanned! { span =>
+                #[allow(non_snake_case)]
                 let mut #trail = #SmallVec::<[#request::FormItem; 8]>::new();
             },
             Kind::Static => quote!()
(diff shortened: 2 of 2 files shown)


jebrosen:
Are you in favor of adding `#![deny(non_snake_case)]` and a few other similar lints to tests in codegen to prevent any regressions in the future?

SergioBenitez:
Do we need `unreachable_patterns`, `unreachable_code` in all of those locations?

> Are you in favor of adding `#![deny(non_snake_case)]` and a few other similar lints to tests in codegen to prevent any regressions in the future?

Absolutely. I'd be happy to have a test in codegen that has a `deny` set up. Since it shouldn't be the case that `rustc` breaks this for us, it would serve well as a regression test.

jebrosen:
> Do we need `unreachable_patterns`, `unreachable_code` in all of those locations?

Probably not. I copied them around from other places in the code (that might not need them either)

> > Are you in favor of adding `#![deny(non_snake_case)]` and a few other similar lints to tests in codegen to prevent any regressions in the future?
> 
> Absolutely. I'd be happy to have a test in codegen that has a `deny` set up. Since it shouldn't be the case that `rustc` breaks this for us, it would serve well as a regression test.

I'll try deny()ing unreachable as well, so we can get a witness for that.

SergioBenitez:
> I'll try deny()ing unreachable as well, so we can get a witness for that.

I'd use the `route.rs` test as a starting point since its intention is to use all of codegen at once.

jebrosen:
Narrowing down the unreachable code is a bit trickier so I will work on that later. My guess is that it was primarily added in the case of implementations of `FromRequest` or `FromData` with `!` as the Error type.

SergioBenitez:
> My guess is that it was primarily added in the case of implementations of `FromRequest` or `FromData` with `!` as the Error type.

That's exactly right.

core/codegen/tests/route.rs
```diff
@@ -1,5 +1,7 @@
 #![feature(proc_macro_hygiene, decl_macro)]
 
+#![deny(non_snake_case)]
```
SergioBenitez: Can we add a comment explaining why we've added this?

SergioBenitez:
Merged in https://github.com/SergioBenitez/Rocket/commit/2d924f0f51bdf8cc9f248a743691a4203f124008.

PR #1024, opened 2019-06-04
Implement UriDisplay for contrib Uuid

blakesmith:
Hello there!

I use the contrib Uuid wrapper, and bind uuid types to my route handlers. When trying to use Uuid types in conjunction with the `uri!` macro, with code like this:

```rust
Redirect::to(uri!(show: instance_membership_id = instance_membership_id))
```

Where `instance_membership_id` is a contrib `Uuid` wrapper type, and the route handler function signature looks something like:

```rust
use rocket_contrib::uuid::Uuid as RocketUuid;

#[get("/instances/<instance_membership_id>/topics/<topic_name>")]
pub fn show(
    instance_membership_id: RocketUuid,
    topic_name: String,
    db: WebappDb,
    user: User,
) -> Result<Template, ExecuteError> {
   ...
}
```

I get an error like this:

```
error[E0277]: the trait bound `rocket_contrib::uuid::Uuid: rocket::http::uri::FromUriParam<rocket::http::uri::Path, _>` is not satisfied                                       
   --> webapp/src/instances.rs:319:58
    |
319 |         Redirect::to(uri!(show: instance_membership_id = instance_membership_id)),                                                                                       
    |                                                          ^^^^^^^^^^^^^^^^^^^^^^ the trait `rocket::http::uri::FromUriParam<rocket::http::uri::Path, _>` is not implemented for `rocket_contrib::uuid::Uuid`
```

This PR adds trait implementations for `UriDisplay`, and their corresponding `FromUriParam` identity implementation. Everything type checks and works after local testing. A few questions / notes:

1. I attempted to use the `impl_with_display!` macro, since UUIDs should be able to use the exact same `Display` implementation for their Uri representation, but had trouble with the macro visibility.
2. I wasn't sure about the best way to unit test this, with the uri `Formatter`: I had trouble understanding how Uri `Formatters` are constructed, given that they seem to be built using macro code-gen. Any advice on how to unit test this would be helpful.

Please let me know if there's a better way to implement this, especially if it makes sense to hook into the `impl_with_display!` macro.

Thanks, and thanks for Rocket!

### DIFF
--- a/contrib/lib/src/uuid.rs
+++ b/contrib/lib/src/uuid.rs
@@ -17,11 +17,12 @@
 pub extern crate uuid as uuid_crate;
 
 use std::fmt;
-use std::str::FromStr;
 use std::ops::Deref;
+use std::str::FromStr;
 
-use rocket::request::{FromParam, FromFormValue};
-use rocket::http::RawStr;
+use rocket::http::uri::{Formatter, Path, Query, UriDisplay};
+use rocket::http::{impl_from_uri_param_identity, RawStr};
+use rocket::request::{FromFormValue, FromParam};
 
 pub use self::uuid_crate::parser::ParseError;
 
@@ -100,6 +101,23 @@ impl fmt::Display for Uuid {
     }
 }
 
+impl UriDisplay<Path> for Uuid {
+    #[inline(always)]
+    fn fmt(&self, f: &mut Formatter<Path>) -> fmt::Result {
+        f.write_value(&format!("{}", self.0))
+    }
+}
+
+impl UriDisplay<Query> for Uuid {
+    #[inline(always)]
+    fn fmt(&self, f: &mut Formatter<Query>) -> fmt::Result {
+        f.write_value(&format!("{}", self.0))
+    }
+}
+
+impl_from_uri_param_identity!([Path] Uuid);
+impl_from_uri_param_identity!([Query] Uuid);
+
 impl<'a> FromParam<'a> for Uuid {
     type Error = ParseError;
 
@@ -149,9 +167,9 @@ impl PartialEq<uuid_crate::Uuid> for Uuid {
 #[cfg(test)]
 mod test {
     use super::uuid_crate;
-    use super::Uuid;
     use super::FromParam;
     use super::FromStr;
+    use super::Uuid;
 
     #[test]
     fn test_from_str() {


jebrosen:
The big question that comes to mind is which string format to use. `FromParam` for `Uuid` delegates to `FromStr` for `Uuid`, which will parse any format. The default for converting to a string appears to be lowercase hex with no hyphens. Perhaps we could implement `FromUriParam` for the various [format adapters](https://docs.rs/uuid/0.7.4/uuid/adapter/index.html) so the choice is explicit? Although I think that would no longer be able to use the convenience of `impl_from_uri_param_identity`, so let's discuss before doing all that work manually.

> 1. I attempted to use the `impl_with_display!` macro, since UUIDs should be able to use the exact same `Display` implementation for their Uri representation, but had trouble with the macro visibility.

That macro is not exposed (including to `rocket_contrib`). It could be, but it's also so trivial that it's probably not worth it.

> 2. I wasn't sure about the best way to unit test this, with the uri `Formatter`: I had trouble understanding how Uri `Formatters` are constructed, given that they seem to be built using macro code-gen. Any advice on how to unit test this would be helpful.

I believe all of the unit tests for `FromUriParam` are done by calling `uri!`.

Since the `Path` and `Query` versions are the same, I think both the `impl UriDisplay` and `impl_from_uri_param_identity` can be done with `P: UriPart` instead of writing out both implementations.

blakesmith:
Thanks for the response @jebrosen!

I like the idea of using the various adapters. Where are you thinking about explicitly setting the adapter? I guess `uri!` could take another parameter, but that might be a bit clumsy? Not sure how else we'd do it though.

jebrosen:
My thought was to have all of these:

```rust
impl<P: UriPart> FromUriParam<P, Hyphenated> for Uuid {}
impl<'a, P: UriPart> FromUriParam<P, HyphenatedRef<'a>> for Uuid {}
impl<P: UriPart> FromUriParam<P, Simple> for Uuid {}
impl<'a, P: UriPart> FromUriParam<P, SimpleRef<'a>> for Uuid {}
impl<P: UriPart> FromUriParam<P, Urn> for Uuid {}
impl<'a, P: UriPart> FromUriParam<P, UrnRef<'a>> for Uuid {}
```

... and then I guess multiply each of those by three (`T`, `&T`, `&mut T`) for parity with other `FromUriParam` implementations. This would allow something like:

```rust
#[get("/a/<uuid>?<other_uuid>")]
fn route_a(uuid: Uuid, other_uuid: Uuid) { }

#[get("/b/<uuid>")]
fn route_b(uuid: Uuid) {
    // suppose uuid is 936DA01F9ABD4d9d80C702AF85C822A8

    // this would generate /a/936DA01F-9ABD-4d9d-80C7-02AF85C822A8/936DA01F9ABD4d9d80C702AF85C822A8
    uri!(route_a: uuid = uuid.to_hyphenated_ref(), other_uuid = uuid.to_simple_ref());
}
```

The main reason to support all of these would be to allow an application to produce URIs in a certain format for compatibility. On the other hand, it looks like *a lot* of impls for relatively small gains.

blakesmith:
@jebrosen Thanks, this is helpful. Let me play around with it a bit, and see how it looks. This might be another good use of a macro for all the different implementations, since they're all pretty boiler-plate. Since we can't impl traits outside of our own module in Rust, this might be the best option. The current workaround I'm using is to just manually interpolate the uri with `format!` inside of the route handler, which isn't ideal.

jebrosen:
> This might be another good use of a macro for all the different implementations, since they're all pretty boiler-plate.

Good point, they should all be able to use `Display`. I feel slightly less bad pointing it out now!

jebrosen:
Ping - what is the status of this?

After re-reading the past conversation, I still think it would be great to have implementations for the various adapters, for user control over the formatting used in the generated URI:

```rust
impl<P: UriPart> FromUriParam<P, Hyphenated> for Uuid {}
impl<'a, P: UriPart> FromUriParam<P, HyphenatedRef<'a>> for Uuid {}
impl<P: UriPart> FromUriParam<P, Simple> for Uuid {}
impl<'a, P: UriPart> FromUriParam<P, SimpleRef<'a>> for Uuid {}
impl<P: UriPart> FromUriParam<P, Urn> for Uuid {}
impl<'a, P: UriPart> FromUriParam<P, UrnRef<'a>> for Uuid {}
```

It's my opinion that we should *definitely* have a `impl<P: UriPart> FromUriParam<P, Uuid> for Uuid {}`. I don't have strong feelings on which format to use, but I am slightly inclined toward `Hyphenated`.

SergioBenitez:
Closing due to inactivity, but otherwise open to it.

Piping:
Is it possible to include `uuid` type in where `FromFrom` `UriDisplay` defined? The PR code works for me to add a typed uri with `uuid` input. 

However, I have to vendor the rocket and rocket_http crates so that I can avoid rust `foreign trait implemented for foreign type` error.

For example,  the `diesel` library derive macro are implemented for `uuid` type from crates.io, not the wrapped type in Rocket_Contrib.

PR #1026, opened 2019-06-07
[WIP/experiment] automatic route mounting

macpp:
This pr allows user to ommit mounting each route manually. Instead all available routes can be automatically detected and mounted during launch. Additionaly, user is able to provide base path for all routes in module, or disable automatic mounting for all routes in module by using `mod_auto_mount!()` macro (see example). 

The main goal here is convinience of use - having to scroll to the `routes![]` macro after adding new route quickly becomes cumbersome.

`mod_auto_mount!()` probably could be an attribute macro that is applied to module, but for now i can't get it working 100% of the time when project is split into multiple files.

Next steps:

- [ ] better documentation
- [ ] better name for  `mod_auto_mount!()`  :smile: 
- [ ] limit ammount of calls to Rocket::mount

[inspired by](https://docs.servicestack.net/routing#auto-route-generation-strategies)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ members = [
   "core/http/",
   "contrib/lib",
   "contrib/codegen",
+  "examples/auto_mounting",
   "examples/cookies",
   "examples/errors",
   "examples/form_validation",
--- a/core/codegen/Cargo.toml
+++ b/core/codegen/Cargo.toml
@@ -14,6 +14,10 @@ build = "build.rs"
 [lib]
 proc-macro = true
 
+[features]
+default=[]
+auto-mount = []
+
 [dependencies]
 indexmap = "1.0"
 quote = "0.6.1"
--- a/core/codegen/src/attribute/route.rs
+++ b/core/codegen/src/attribute/route.rs
@@ -337,6 +337,23 @@ fn generate_internal_uri_macro(route: &Route) -> TokenStream2 {
     }
 }
 
+fn generate_auto_mount(generated_struct_name: &syn::Ident) -> TokenStream2 {
+
+    #[cfg(feature="auto-mount")]
+    quote! {
+        #[allow(unused_imports)]
+        use ::rocket::auto_mount::__default_auto_mount_info::*;
+
+        rocket::inventory::submit!{
+            #![crate = rocket]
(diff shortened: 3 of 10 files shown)


SergioBenitez:
I have no issue with automatically collecting routes for the user. Historically, this has been avoided because of 1) the complexity (though not difficulty) of a portable implementation, and 2) security/correctness implications.

I'll ignore 1) for now since it's largely subjective, but I want to expand on 2). The main concern in 2) is that routes defined in dependencies of a Rocket application will be auto-mounted in the top-level Rocket application. For example, if my crate R with route `r` depends on crate A with route `a`, auto-mounting routes in `R` should not mount `a`. Unless I'm mistaken, this implementation, which depends on the implementation in the [`inventory`](https://docs.rs/inventory/) crate, will have this flaw. This is both a security issue (routes can be injected into unsuspecting applications) and a correctness issue (external, and especially crate-private routes, should never be mounted unknowingly).

Any solution we accept must have the property that only routes defined in the "local" crate are "auto-mounted". Such a solution must be secure from malicious dependencies. Unfortunately, I do not know of a such a solution.

A note on the API: I believe that the API we'd want to expose is that of a mutable "auto-collected route set". In other words, I believe we should allow the user to call a function/method/macro to get a mutable `Set` of all of the routes in the application. After any modifications, that set can then be mounted. This would enable excluding routes at will as well as mounting routes multiple times and with differing options without a "magic" macro.

In summary, the security/correctness issue _must_ be addressed before any such solution is accepted. I'd also like to see the API modified as detailed above.

macpp:
Thanks for detailed response. It seems that i was overenthusiastic about this, and forgot about potential security issues :smile: I will try to rethink this -i have at least one idea that might work, but it requires some experimentation first.

SergioBenitez:
@macpp Wonderful! I'm closing this PR for now, simply to clear up the queue. Would love to see you open a new PR if you find a secure implementation!

macpp:
After some tinkering, i think that secure automatic mounting will require some additional configuration from user. For example, my poor (but working!) prototype requires that user: 

- puts `routes_inventory!();` macro invocation at the crate root (this defines  functions that allow user to get sets of routes)
- puts `println!("cargo:rustc-cfg=rocket_codegen_auto_mounting");` in build script, to avoid breaking upstream crates that use rocket but don't use auto mounting

Is a  solution that requires this much (or less) configuration acceptable?

PR #1029, opened 2019-06-13
Add a deployment guide

anurag:
This PR adds a deployment guide in response to https://github.com/SergioBenitez/Rocket/issues/171.


ivegotasthma:
I don't think the guide should serve as an advertisement platform for deployment services.

The guide should be neutral, and explain a basic deployment with optional TLS and specificities for Linux / Windows / MacOS. All of that is missing, except the explicit parts for Render, which IMO are a niche use-case.

**Render** should have the guide on deploying Rocket applications, not the other way around.

jebrosen:
I am opposed to including in the guide any content dependent on a specific deployment service, except in the presence of a prenegotiated agreement. I think *linking* to tutorials for several platforms or services would be okay, as long as it does not look like an endorsement of any particular platform over any other.

I *would* be okay with including generic examples of creating docker or other containers, uploading binaries to an external server and configuring startup with an initscript or systemd service, and so on. I would also be fine with including a section along the lines of "platforms that respect `rust-toolchain` files, such as Render, this or that heroku buildpack, or QuuxBar".

anurag:
Makes sense. We already have a deployment guide for Rust at https://render.com/docs/deploy-rocket-rust, and we're happy to contribute a guide with standard examples that don't rely on Render.

PR #1030, opened 2019-06-15
add a  keyword because it must be explicit now

hbina:
Implicit dynamic return is going to be deprecated.


jebrosen:
This (and I think a few more instances of `dyn`) is included in #1013, by the way.

PR #1034, opened 2019-06-20
Add impl FromParam,FromFormValue for NonZero* types

dima74:


jebrosen:
My gut reaction is "why?", but I suppose "why not?" is equally valid considering we implement `FromParam` for pretty much everything that implements `FromStr`.

Docs would need to be updated as well (https://api.rocket.rs/v0.4/rocket/request/trait.FromParam.html#provided-implementations).

dima74:
@jebrosen yes, docs were updated, also added FromFormValue implementations.

SergioBenitez:
In the documentation, can we split the one point line with all of the types into two? That is, have one line for non-`NonZero` types and another specifically calling them out.

dima74:
@SergioBenitez yes, like this:
image

or maybe make three groups? Like this:
image

SergioBenitez:
I like the three groups! Let's go with that one.

dima74:
@SergioBenitez ok, updated

SergioBenitez:
Merged in c100a9212735928ed5ada2a38754a21247c86677. Thank you!

PR #1051, opened 2019-07-14
Update handlebars to 2.0

sunng87:


SergioBenitez:
@jebrosen I think that's a great idea. I'll take that on.

SergioBenitez:
Merged in 93c979eb4314c88480539c26677a1c87eea58107. Thank you!

PR #1052, opened 2019-07-18
Fix and update benchmarks.

jebrosen:
These have been broken since 61f107f when `#![feature(test)]` was inadvertently removed. That same commit changed collision rules for `format = ` on `get` routes.


jebrosen:
Merged in e7e84a6d.

PR #1060, opened 2019-07-28
Remove `crate_visiblity_modifier` feature gate, others that are unused

jhpratt:
Simple change from `crate` to `pub(crate)` where necessary. I've also removed other feature gates that were completely unused.

Related: #19


jhpratt:
This will likely be done after merging `async` into `master`, possibly at the same time as #1114.

SergioBenitez:
Can you rebase this on master? Also, we should consider which of these can be made `pub` (not `pub(crate)`) while remaining private due to higher-level privacy constraints, for instance, a `pub struct` in a private module that is never reexported.

SergioBenitez:
I'm going to close this since it's now rather outdated. I would _love_ a PR for this that takes into account  https://github.com/SergioBenitez/Rocket/pull/1060#issuecomment-530165362, however. Please let me know if you intend to work on it as I'll likely get to it myself sooner than later.

jhpratt:
I can certainly get around to it if you'd like. My intent was to wait to see if there will be any action on rust-lang/rust#53120, which could possibly have a PR land before a 0.5 release.

Perhaps @Centril (or someone else) could kickstart that? Otherwise I can certainly go through and see what the minimum change would be.

jebrosen:
Side note: I believe I had expressed a preference toward doing this only after `async` and `master` had merged, but looking at the changes again I think the merge conflicts will not be too bad after all.

Centril:
@jhpratt realistically I probably don't have the time to get to `crate fn` sadly. :(

jhpratt:
Alright, no problem! I'll see what I can do with regard to removing this feature gate later today.

PR #1061, opened 2019-07-28
Remove `label_break_value` feature

jhpratt:
This amounts to creating a `loop` that only runs once through.

If this feature ever stabilizes (which seems unlikely at this point), it
will be trivial to revert this to the previous state.

### DIFF
--- a/core/lib/src/lib.rs
+++ b/core/lib/src/lib.rs
@@ -2,7 +2,6 @@
 #![feature(try_trait)]
 #![feature(proc_macro_hygiene)]
 #![feature(crate_visibility_modifier)]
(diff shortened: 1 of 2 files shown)


jebrosen:
My `async` branch also removes the use of this feature, but loses the benefit of only specifying `Transform::Borrowed` in one place.

jhpratt:
Yeah, I just preferred the `loop` to preserve Sergio's desire of only specifying it once. It's pretty trivial to inline it, though.

jhpratt:
Closing in favor of #1060, which also removes unused feature gates.

PR #1073, opened 2019-08-02
Use async/await for msgpack

Redrield:
Similar to what was already done for JSON


jebrosen:
Can you also update the tests in the msgpack example? I should have mentioned that in the main tracking issue.

In `examples/msgpack/`, `cargo test` should have all tests passing.

Redrield:
I had to switch from `'static` to `'r` defined in the impl for MsgPack's `Responder` to get everything to compile properly, but the tests pass now.

jebrosen:
Merged as e33078f. Beautiful!

Issue #1071, opened 2019-08-02
(async) Use the futures 0.3 - compatible versions of hyper and tokio


jebrosen:
hyper 0.13 and tokio 0.2 are slated to use futures 0.3. We should use those, ideally by the first release of async-capable Rocket. This would allow us to remove our own dependencies on the futures 0.1 ecosystem and the futures 0.1-0.3 compatibility layer.

The following issues make that inconvenient today:
  * These are currently the master branches of those projects, planned for hyper 0.13 and tokio 0.2. The port to futures 0.3 is still fresh, and we don't want to point to hyper@master until it settles a bit.
  * ~~We need `hyper::body::Sender` to implement the new `Sink` trait, or find a different approach for that section of code.~~ Using a different approach.
  * ~~ hyper's requirements on executors - see also #1068 ~~ hyper requires a tokio runtime as its executor, so this is resolved
  * Keep in mind that *this update should only affect internal APIs*. If the futures 0.1 ecosystem has already "leaked" into rocket's public APIs, we have done something wrong!

jebrosen:
I've made a proof of concept / WIP: https://github.com/SergioBenitez/Rocket/compare/async...jebrosen:async-futures03-hyper?expand=1

This makes use of https://github.com/Nemo157/futures-tokio-compat because tokio's `AsyncRead` is different from the one in `futures-preview`. Given this and other incompleteness (see https://github.com/tokio-rs/tokio/issues/1201), I would still like to wait a bit longer to resolve this.

jhpratt:
Tokio's 0.2 alpha has been released, which includes support for `std::future::Future`s.

jebrosen:
Unfortunately hyper is completely broken for a few days due to some changes in tokio. It will maybe be fixed by hyperium/hyper#1890.

jebrosen:
I have updated the async branch to use `tokio 0.2.0-alpha.2` and an arbitrary compatible version of hyper.

As the commit message explains, tokio-rs/tokio#1356 causes some tests to fail. My own other mistake is *also* causing some of the same tests to fail, and I would like to wait for the upstream tokio issue to be resolved before trying too hard to fix those.

PR #1076, opened 2019-08-03
[WIP] convert serve module to async/await

Redrield:
Converting serve.rs and associated tests to use async/await

### DIFF
--- a/contrib/lib/src/serve.rs
+++ b/contrib/lib/src/serve.rs
@@ -14,12 +14,12 @@
 //! features = ["serve"]
 //! ```
 
-use std::path::{PathBuf, Path};
+use std::path::{Path, PathBuf};
 
-use rocket::{Request, Data, Route};
-use rocket::http::{Method, uri::Segments};
-use rocket::handler::{Handler, Outcome};
+use rocket::handler::{Handler, HandlerFuture, Outcome};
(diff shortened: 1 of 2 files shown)


Redrield:
I've marked this as WIP for now as I'm having issues getting the tests to pass. Trying to block on the file IO is giving errors somewhere down the line on my machine, but I can't pinpoint where. [Example of what it's doing](https://0x0.st/zOrv.png)

jebrosen:
Oh bother, I have thought and forgot about this little detail before. The problem here is that I had used tokio's `File` to implement the `File` responder, which doesn't work outside of tokio. But `.body_bytes_wait()` uses the `futures-rs` executor. Fixing this should be a relatively small code change but is a pretty significant design issue to consider.

Redrield:
Will it be something considered in discussion on the tracking issue? I can put this on hold until something is reached and then either add the change to this PR or open a new one with the change to fix the tests for this.

jebrosen:
Let's put this one on hold for the moment - and anything else to do with `File`, sadly. It's harder to fix than I realized because `tokio` 0.1's "block_on" equivalent requires the future to be `'static`. #1071 should make this easier to fix.

jebrosen:
I merged this without rustfmt-induced changes in 5b1125b. Thanks!

PR #1079, opened 2019-08-07
Graceful shutdown

### LINKED ISSUE #180: Clean shutdown?
Is there any way to perform a "clean" shutdown of Rocket, such that any requests currently being handled can finish (and such that managed state can be cleanly dropped)? I'm assuming right now that pressing ^C just kills the program without any graceful shutdown, since I would expect Rocket to print something to stdout otherwise.

Besides not wanting to have a bad experience for users, I'm primarily concerned with the fact that the program I'm writing is using SQLite and therefore killing it abruptly in the middle of any database mutations is probably a bad idea.

jhpratt:
This PR implements a mandatory graceful shutdown on a Rocket instance, which can be called at runtime.

Calling `.get_shutdown_handle()` before launch will return a `ShutdownHandle`. As shutting things down before starting isn't too useful most of the time, there is also `rocket::shutdown::ShutdownHandle`, which implements `FromRequest`, so it can be used as a request guard. In either situation, calling `.shutdown()` will signal the server to shut down.

Per a preexisting comment in the code, I've also converted `rocket.launch()` to return `Result<(), LaunchError>`.

Resolves #180.


core/lib/src/rocket.rs
```diff
@@ -805,12 +825,22 @@ impl Rocket {
...
 
-        unreachable!("the call to `block_on` should block on success")
+        Ok(())
+    }
+
+    /// Gracefully shut down the instance of Rocket. Returns a `Result`,
+    /// indicating whether the shutdown request was successfully processed.
+    #[inline(always)]
+    pub fn shutdown(&mut self) -> Result<(), ()> {
+        // TODO How do we want to handle a possible error here?
```
jhpratt: Will need a resolution at some point before 0.5 on what we want to do with a possible error.

jhpratt:
Just a note, it won't be too difficult to add signal handling (<kbd>ctrl</kbd> + <kbd>C</kbd>), as the `signal-hook` crate provides for simple abstractions. There's also `tokio-signal`, which is similar. That can be done in this PR if wanted.

core/lib/src/rocket.rs
jebrosen: IIRC Senders and Receivers only error if the other side is gone. The sender(s) theoretically can't be dropped before the receiver, so I think `recv()` should never fail. And the sender(s) should ignore send errors, because that can only mean the server already is/has shut down.

jhpratt: If that's the case, should it even return a `Result`, or should I just `.expect(...)` the value?

jhpratt:
Is there anything current using managed stated within Rocket itself? If not, wouldn't it be best to preserve that expectation?

Having a `ShutdownHandle` is definitely feasible, I'd actually just need to revert one of the commits where I inclined the changes. Off the top of my head, an mpsc sender is cloneable, so it _should be_ relatively easy to implement `FromRequest`. We could then expose `FromRequest` (the exact path TBD), and use it like any other guard from the clients perspective.

With regard to handling the sender/received errors, should the channel be left open after receiving a shutdown signal? I see no reason to, as it would provide a clear indication that the method is being called more than once.

jhpratt:
Should work as a request guard now.

NB: I haven't yet checked how Hyper interacts with Rocket here — ideally Rocket will return 503 service unavailable (if it even responds).

core/lib/src/rocket.rs
jhpratt: In the latest commits, I have replaced it with `.expect("Error sending shutdown signal")`. If someone runs across this, we'll definitely know it.

jhpratt:
Going to hold off on signal handling for the time being. I'd like to see tokio land support that doesn't require a pinned version of `futures-preview`, allowing Rocket to use its desired version. `signal-hook` doesn't seem too easy to use on a per-instance level due to its requirement of the `'static` lifetime.

jhpratt:
@jebrosen I believe this is in sufficient condition to merge.

I just tested it locally with a `/shutdown` route, which included a 15 second delay to allow requesting other routes in the meantime. The server successfully stayed alive while waiting for the 15 seconds to expire (before responding), and acted like it was shut down for any new requests.

Here's the full code if you'd like to test it yourself.

```rust
#![feature(proc_macro_hygiene, async_await)]

use futures::compat::Future01CompatExt; // futures-preview
use rocket::{get, routes, shutdown::ShutdownHandle};
use std::time::{Duration, Instant};
use tokio::timer::Delay;

async fn sleep(secs: u64) {
    Delay::new(Instant::now() + Duration::from_secs(secs))
        .compat()
        .await
        .unwrap()
}

#[get("/immediate")]
fn immediate() {}

#[get("/shutdown")]
async fn shutdown(handle: &ShutdownHandle<'_>) {
    handle.shutdown();
    sleep(15).await;
}

fn main() {
    let _ = rocket::ignite()
        .mount("/", routes![shutdown, immediate])
        .launch();
}
```

core/lib/src/rocket.rs
```diff
@@ -720,7 +723,8 @@ impl Rocket {
 
     /// Starts the application server and begins listening for and dispatching
     /// requests to mounted routes and catchers. Unless there is an error, this
-    /// function does not return and blocks until program termination.
+    /// function returns `()` and blocks until program termination or graceful
```
jebrosen: This should be reworked. Something like "this function does not return unless a shutdown is requested via a `ShutdownHandle`." I will look into whether hyper can return an error at any time other than launch, because that would also be a change in expectations here.

core/lib/src/rocket.rs
```diff
@@ -805,12 +809,36 @@ impl Rocket {
...
+    /// # Example
+    ///
+    /// ```rust
+    /// # #![feature(proc_macro_hygiene, async_await)]
+    /// #
+    /// let rocket = rocket::ignite();
+    /// let handle = rocket.get_shutdown_handle();
+    ///
+    /// // Likely called from another thread.
+    /// handle.shutdown();
```
jebrosen: Let's make the example appear (with some invisible `if false {` magic) to start a thread that sleeps 10 seconds and then calls `handle.shutdown()`, actually calling shutdown immediately.

core/lib/src/shutdown.rs
```diff
@@ -0,0 +1,92 @@
...
+/// fn main() {
+///     # if false {
+///     rocket::ignite()
+///         .mount("/", routes![shutdown])
+///         .launch();
+///     # }
+/// }
+/// ```
+#[derive(Debug, Clone, Copy)]
+pub struct ShutdownHandle<'a> {
```
jebrosen: I can't guess how likely this will be, but I will not be surprised if someone asks for a `ShutdownHandle` that is `'static`.

core/lib/src/shutdown.rs
```diff
@@ -0,0 +1,92 @@
...
+    /// Notify Rocket to shut down gracefully.
+    ///
+    /// This method should only need to be called once.
+    #[inline]
+    pub fn shutdown(&self) {
+        // Clone to avoid requiring `self` to be mutable.
+        self.sender
+            .clone()
+            .try_send(())
+            .expect("Error sending shutdown signal")
```
jebrosen: I think we should purposefully ignore this error - with a good comment explaining why. An error means that either the send queue is full (too many outstanding shutdown requests) or the receiver was dropped (already shut down).

core/lib/src/shutdown.rs
```diff
@@ -0,0 +1,92 @@
...
+/// ```
+#[derive(Debug, Clone, Copy)]
+pub struct ShutdownHandle<'a> {
+    sender: &'a mpsc::Sender<()>,
+}
+
+impl ShutdownHandle<'_> {
+    /// Notify Rocket to shut down gracefully.
+    ///
+    /// This method should only need to be called once.
```
jebrosen: "should" isn't very strong. There is no reason calling once would fail but twice would work, and calling more than once is not an error.

core/lib/src/shutdown.rs
```diff
@@ -0,0 +1,92 @@
...
+}
+
+impl Future for Receiver {
+    type Output = ();
+
+    #[inline]
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        self.receiver
+            .poll_next_unpin(cx)
+            .map(|opt| opt.unwrap_or(()))
```
jebrosen: This is a bit weird. When could the future need to be polled after all senders are already gone?

core/lib/src/shutdown.rs
jhpratt: "This method only needs to be called once"? Keep it straight to the point, while not even hinting at what could/does happen when called a second time.

core/lib/src/shutdown.rs
jhpratt: IIRC I had to do this because it's a channel, so there are error types associated. In this situation, I don't believe it's possible to poll after all senders are `drop`ped.

core/lib/src/shutdown.rs
jhpratt: How should this be reconciled with the migration to managed state?

jhpratt: Also, the reason I did this was to avoid unnecessary clones. It wouldn't be much cost, but all I'd have to do is revert a single commit.

core/lib/src/rocket.rs
jhpratt: Pushed a change. Leaving open with regard to hyper shutting down at other times.

core/lib/src/shutdown.rs
```diff
@@ -0,0 +1,78 @@
+use futures::{channel::mpsc, stream::StreamExt};
+use std::{
+    future::Future,
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+/// Wrapper around an mpsc channel.
+#[derive(Debug)]
+crate struct Shutdown {
```
jebrosen: I think we can avoid this struct and impls: we can inline the `mpsc::channel(1)` construction into `rocket.rs` (storing the Option<Receiver> directly in `Rocket`) and then use `.with_graceful_shutdown(async move { receiver.next().await; })`. That should save quite a bit of code.

core/lib/src/request/from_request.rs
```diff
@@ -424,6 +425,14 @@ impl FromRequest<'_, '_> for SocketAddr {
     }
 }
 
+impl FromRequest<'_, '_> for ShutdownHandle {
```
jebrosen: Let's move this `impl` to live next to `ShutdownHandle` itself.

core/lib/src/rocket.rs
```diff
@@ -794,24 +806,52 @@ impl Rocket {
...
             .expect("Cannot build runtime!");
 
-        // TODO.async: Use with_graceful_shutdown, and let launch() return a Result<(), Error>
         match self.spawn_on(&runtime) {
-            Ok(fut) => match runtime.block_on(fut) {
-                Ok(_) => unreachable!("the call to `block_on` should block on success"),
-                Err(err) => err,
-            }
-            Err(err) => Box::new(err),
+            Ok(fut) => runtime.block_on(fut).map(|_| ()),
```
jebrosen: This `.map(|_| ())` shouldn't be needed -- what is the type of this `_`?

jhpratt:
With regard to the error type, I think the "issue" is that hyper's errors aren't exposed anywhere, so there's not currently any way to represent that. My suggestion (before and now) is to create an `enum` that we can then wrap up hyper's errors along with any launch errors.

core/lib/src/shutdown.rs
jhpratt: Should just be a slight compilation time improvement, as I imagine rustc is smart enough to inline it. Can do, though.

core/lib/src/rocket.rs
jhpratt: Looks like it wasn't needed. Removed.

jebrosen:
Let's use this error type for `launch()` for the time being:

```rust
enum Error {
    Launch(LaunchError),
    Run(Box<dyn std::error::Error>),
}
```

jhpratt:
Would it be better to use `hyper::Error` over `Box<dyn std::error::Error>`, given that we know that's the type it will be? After taking a quick look at things, `hyper::Error` is actually already public via `LaunchErrorKind`, so it wouldn't be newly exposed.

`.spawn_on()` could also return `hyper::Error` over the boxed, since you had wanted to return a `Result<impl Future<Result<(), ??>, LaunchError>`, per previous discussion. The advantage of that signature is not having to `.await` to determine if launch failed.

jebrosen:
Interesting, I really thought hyper was "erased" from the error type. In that case, I don't mind `hyper::Error` instead of the boxed error after all.

core/lib/src/rocket.rs
```diff
@@ -764,20 +780,64 @@ impl Rocket {
...
+                        futures::future::ready(())
+                    });
+
+                    // Race the futures against each other. In doing so, we
+                    // ensure a future is dropped from the runtime upon
+                    // completion of the other. This prevents a potential memory
+                    // leak where the server shuts down by means other than an
+                    // intercepted signal, leaving the signal handler being
+                    // polled indefinitely.
+                    futures::select!(
```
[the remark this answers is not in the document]
jhpratt: Probably not. I was searching for something to that effect, and apparently didn't see it.

core/lib/src/rocket.rs
```diff
@@ -764,20 +780,64 @@ impl Rocket {
                 }))
             }
         });
+        let (ctrl_c_sender, ctrl_c_receiver) = oneshot::channel();
```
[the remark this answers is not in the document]
jhpratt: Opting for `cancel_ctrl_c_listener_{sender|receiver}`. It's pretty long, but certainly descriptive.

core/lib/src/shutdown.rs
```diff
@@ -0,0 +1,51 @@
...
+impl ShutdownHandle {
+    /// Notify Rocket to shut down gracefully.
+    ///
+    /// This method only needs to be called once.
+    #[inline]
+    pub fn shutdown(&self) {
+        // Intentionally ignore any error, as the only scenarios this can happen
+        // is sending too many shutdown requests or we're already shut down.
+        //
+        // Clone to avoid requiring `self` to be mutable.
```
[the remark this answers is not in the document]
jhpratt: Yup, that's a remnant of when I had everything with references. At the time, I preferred routes not having to declare `mut`. That's no longer an issue.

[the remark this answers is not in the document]
jhpratt: Changed to take ownership, eliminating the clone. Leaving open for discussion.

core/lib/src/error.rs
```diff
@@ -106,11 +115,14 @@ impl LaunchError {
...
     /// ```rust
+    /// use rocket::error::Error;
     /// # if false {
-    /// let error = rocket::ignite().launch();
-    ///
-    /// // This line is only reached if launch failed.
-    /// let error_kind = error.kind();
+    /// if let Err(error) = rocket::ignite().launch() {
+    ///     match error {
+    ///         Error::Launch(err_kind) => println!("Found a launch error: {}", err_kind),
```
jebrosen: This isn't quite right for a doc test for `kind()`. I think we should use `::Launch(err) => println!("Found a launch error: {}", err.kind())`

core/lib/src/rocket.rs
```diff
@@ -445,14 +449,22 @@ impl Rocket {
...
             default_catchers: catcher::defaults::get(),
             catchers: catcher::defaults::get(),
             state: Container::new(),
             fairings: Fairings::new(),
-        }
+            shutdown_handle: ShutdownHandle(shutdown_sender),
+            shutdown_receiver: Some(shutdown_receiver),
+        };
+
+        rocket.state.set(rocket.shutdown_handle.clone());
```
jebrosen: Can we put a `// TODO` here regarding the type that will be put into managed state? I'm not sure we want to expose `State<ShutdownHandle>` as a way to get one.

core/lib/src/rocket.rs
```diff
@@ -765,19 +781,58 @@ impl Rocket {
...
 
-        let (future, handle) = server.remote_handle();
         runtime.spawn(future);
-        Ok(handle.err_into())
+
+        #[cfg(feature = "ctrl_c_shutdown")]
+        match tokio::net::signal::ctrl_c() {
+            Ok(mut ctrl_c) => {
+                runtime.spawn(async move {
+                    // Race the futures against each other. In doing so, we
```
jebrosen: This is a bit long-winded explanation. Can we simplify this to "Stop listening for Ctrl-C if the server shuts down externally to avoid waiting forever?"

core/lib/src/rocket.rs
```diff
@@ -793,25 +848,54 @@ impl Rocket {
...
-                Ok(_) => unreachable!("the call to `block_on` should block on success"),
-                Err(err) => err,
-            }
-            Err(err) => Box::new(err),
+            Ok(fut) => runtime.block_on(fut).map_err(Error::Run),
+            Err(err) => Err(Error::Launch(err)),
         }
     }
 
+    /// Returns a [`ShutdownHandle`], which can be used to gracefully terminate
```
jebrosen: Something like this reads easier, IMO:

> Returns a [`ShutdownHandle`] which can be used to gracefully terminate
> this instance of Rocket. In routes, you can use the [`ShutdownHandle`]
> request guard.

core/lib/src/shutdown.rs
```diff
@@ -0,0 +1,47 @@
+use crate::request::{FromRequest, Outcome, Request};
```
jebrosen: This module is quite short (only a single `struct` and two `impls`). If you had to inline this somewhere, where do you think you would choose?

core/lib/src/rocket.rs
jhpratt: Resolving in favor of a newtype.

core/lib/src/shutdown.rs
jhpratt: Just summing up what I said in IRC/Matrix earlier for anyone else following this: After briefly looking at the other directories and files that exist in the `core/lib`, I honestly don't see anywhere where it would fit sensibly. Though it is comparatively small, the module is significantly different than anything else that currently exists in `core/lib`.

jebrosen:
Merged as 49709a6e, with some error handling followup in a3719c4. Thanks for all the hard work on this!

PR #1082, opened 2019-08-08
Use `impl Trait` where possible

jhpratt:
In some situatations, ther are nested `impl`s. Other places may be
possible, but would require explicitly specifying the `Self` type.


SergioBenitez:
Closing this out since it's rather out-of-date. What's more, I don't see a clear advantage to this change, so I don't believe we should proceed with this.

PR #1084, opened 2019-08-10
Convert two `Pin<Box<_>>` functions to `async`

jhpratt:


core/lib/src/fairing/fairings.rs
```diff
@@ -68,12 +65,10 @@ impl Fairings {
     }
 
     #[inline(always)]
-    pub fn handle_response<'a, 'r>(&'a self, request: &'a Request<'r>, response: &'a mut Response<'r>) -> Pin<Box<dyn Future<Output=()> + Send + 'a>> {
-        Box::pin(async move {
-            for &i in &self.response {
-                self.all_fairings[i].on_response(request, response).await;
-            }
-        })
+    pub async fn handle_response<'a, 'r>(&'a self, request: &'a Request<'r>, response: &'a mut Response<'r>) {
```
jebrosen: I wonder if we can omit all of these lifetimes and let rustc elide them however is appropriate - I think I added them just to be able to put them in the output type.

jhpratt: Quick check shows we still need `'r` to tie the Request and Response lifetimes together. Others can be omitted.

jebrosen: > to tie the Request and Response lifetimes together

Ah, right. IIRC I also introduced *that* to satisfy some other annoying trait bound requirements and it wasn't there in the synchronous version. I spelled this out for `Respond::respond_to` in the main tracking issue but forgot about the equivalent for fairings.

jebrosen:
Thanks! Merged as f2aaab0.

PR #1092, opened 2019-08-13
Correct `Handler` implementation for `StaticFiles`

jhpratt:
This is the last part out of core & contrib that didn't compile (aside from examples), so it's 100% compiling now!


jhpratt:
Just saw #1076. Closing as a duplicate. Probably should rebase :smile:

PR #1097, opened 2019-08-18
Allow launching on a custom Tokio runtime

### LINKED ISSUE #1068: (async) Support launching on a custom runtime
It might be desired for a user to set up a custom runtime to launch rocket on, for example if one binary is starting multiple servers or is running a rocket server plus other tasks.

We could create a `launch_on(runtime: Runtime)` or similar function. Note that this must be a `tokio` runtime because hyper uses timers and likely other functionality from tokio that will only work inside tokio's event loop.

### LINKED ISSUE #1069: (async) Access to the runtime/executor Rocket is running on
Routes might want to spawn futures on a runtime. We could provide access to a runtime via a handle that implemented both `FromRequest` and `Spawn`. Spawning a future on this handle would spawn it on the same runtime Rocket itself is running on.

For example, it could be used to track a pool of background tasks:
```rust
struct BackgroundTasks(Arc<Mutex<Vec<Task>>>);

#[post("/thing")]
fn thing(exec: ExecutorHandle, tasks: State<BackgroundTasks>) {
    tasks.add(exec.spawn(async {
            // whatever
    }))
}
```

On the other hand, maybe we should not expose such a handle and require users to create their own thread pool.

jhpratt:
Currently only on a Tokio runtime for reasons described in #1068. In the future, it would naturally be ideal to allow launching on _any_ runtime; such a change would necessitate larger code changes to provide for a common API.

Resolves #1068.

### DIFF
--- a/core/lib/src/rocket.rs
+++ b/core/lib/src/rocket.rs
@@ -680,7 +680,8 @@ impl Rocket {
         Ok(self)
     }
 
-    /// Identical to `launch()`, but using a custom Tokio runtime.
+    /// Identical to `launch()`, but using a custom Tokio runtime. The runtime
+    /// has no other restrictions, and can have other tasks running on it.
     ///
     /// # Example
     ///
@@ -690,16 +691,16 @@ impl Rocket {
     /// let runtime = tokio::runtime::Runtime::new();
     ///
     /// # if false {
-    /// rocket::ignite().launch_on(runtime);
+    /// let fut = rocket::ignite().launch_on(&runtime);
+    /// fut.then(|_| {
+    ///     // do things after the server shuts down
+    /// });
     /// # }
     /// ```
-    pub fn launch_on(mut self, mut runtime: tokio::runtime::Runtime) -> LaunchError {
+    pub async fn launch_on(mut self, runtime: &tokio::runtime::Runtime) -> Result<(), LaunchError> {
         #[cfg(feature = "tls")] use crate::http::tls;
 
-        self = match self.prelaunch_check() {
-            Ok(rocket) => rocket,
-            Err(launch_error) => return launch_error
-        };
+        self = self.prelaunch_check()?;
 
         self.fairings.pretty_print_counts();
 
@@ -707,7 +708,7 @@ impl Rocket {
         let addrs = match full_addr.to_socket_addrs() {
             Ok(a) => a.collect::<Vec<_>>(),
             // TODO.async: Reconsider this error type
-            Err(e) => return From::from(io::Error::new(io::ErrorKind::Other, e)),
+            Err(e) => return Err(From::from(io::Error::new(io::ErrorKind::Other, e))),
         };
 
         // TODO.async: support for TLS, unix sockets.
@@ -715,7 +716,7 @@ impl Rocket {
 
         let mut incoming = match hyper::AddrIncoming::bind(&addrs[0]) {
             Ok(incoming) => incoming,
-            Err(e) => return LaunchError::new(LaunchErrorKind::Bind(e)),
+            Err(e) => return Err(LaunchError::new(LaunchErrorKind::Bind(e))),
         };
 
         // Determine the address and port we actually binded to.
@@ -760,10 +761,9 @@ impl Rocket {
             .executor(runtime.executor())
             .serve(service);
 
-        // TODO.async: Use with_graceful_shutdown, and let launch() return a Result<(), Error>
-        runtime.block_on(server).expect("TODO.async handle error");
-
-        unreachable!("the call to `block_on` should block on success")
+        let (future, handle) = server.remote_handle();
+        runtime.spawn(future);
+        Ok(handle.await?)
     }
 
     /// Starts the application server and begins listening for and dispatching
@@ -792,7 +792,11 @@ impl Rocket {
             .build()
             .expect("Cannot build runtime!");
 
-        self.launch_on(runtime)
+        // TODO.async: Use with_graceful_shutdown, and let launch() return a Result<(), Error>
+        match runtime.block_on(self.launch_on(&runtime)) {
+            Ok(_) => unreachable!(),
+            Err(err) => err,
+        }
     }
 
     /// Returns an iterator over all of the routes mounted on this instance of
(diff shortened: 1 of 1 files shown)


jebrosen:
`launch_on` should take `&Runtime` or possibly `TaskExecutor`; moving `Runtime` by value probably precludes most of the useful reasons to do this.

`block_on` shouldn't be called inside `launch_on`, for a similar reason. I suspect `launch_on` will need to return some kind of future or spawn handle.

Ideally we could have a method that took only `self` and returned a future that could be spawned by the user on any tokio runtime, but we currently need the executor "ahead of time" in order to pass it to hyper.

jhpratt:
Ah, I originally assumed this was just to be able to customize the runtime, not the ability to launch on a runtime that may already exist and does other things. That's certainly a different situation.

I've opted to accept a `&Runtime` over an `&Executor`, as that will be helpful when #1069 is implemented.

core/lib/src/rocket.rs
```diff
@@ -703,54 +703,43 @@ impl Rocket {
...
     /// # if false {
-    /// rocket::ignite().launch();
+    /// let fut = rocket::ignite().launch_on(&runtime);
+    /// fut.then(|_| {
+    ///     // do things after the server shuts down
+    /// });
     /// # }
     /// ```
-    pub fn launch(mut self) -> LaunchError {
+    pub async fn launch_on(mut self, runtime: &tokio::runtime::Runtime) -> Result<(), LaunchError> {
```
jebrosen: `async` functions do nothing until the first `poll()`, so this would *require* the caller to do something with the returned future. I think this signature would be more honest if it was a non-`async` fn named `spawn_on` returning `Result<impl Future<Output = ()>, LaunchError>` (by returning an `async` block at the very end of the function).

Another reason for this is error handling: AFAICT `handle.await` can return non-launch `Error`s, but this would report them as `LaunchError`.

core/lib/src/rocket.rs
```diff
@@ -792,10 +781,42 @@ impl Rocket {
...
+        // TODO.async What meaning should config.workers have now?
+        // Initialize the tokio runtime
+        let runtime = tokio::runtime::Builder::new()
+            .core_threads(self.config.workers as usize)
+            .build()
+            .expect("Cannot build runtime!");
+
+        // TODO.async: Use with_graceful_shutdown, and let launch() return a Result<(), Error>
+        match runtime.block_on(self.launch_on(&runtime)) {
+            Ok(_) => unreachable!(),
```
jebrosen: This should explain why: `unreachable!("a listening server should never complete")`

jhpratt: Whoops, thought I copied the text over from before the commits. There used to be a message there, I'll use that.

core/lib/src/rocket.rs
jhpratt: As you mentioned, `handle.await` can return other errors. Should the `Future` output a `Result` as well? The only other option would be ignoring it or panicking.

jebrosen: I suppose that should return a `Result` as well, yes. That would be necessary for a use case like "log a message and restart the server after a failure"

jhpratt: Done in recent commit.

core/lib/src/rocket.rs
```diff
@@ -703,54 +703,49 @@ impl Rocket {
...
     ///
     /// ```rust
+    /// // This gives us the default behavior. Alternatively, we could use a
+    /// // `tokio::runtime::Builder` to configure with greater detail.
+    /// let runtime = tokio::runtime::Runtime::new();
+    ///
     /// # if false {
-    /// rocket::ignite().launch();
+    /// let fut = rocket::ignite().spawn_on(&runtime).expect("error launching server");
+    /// fut.then(|result| {
```
jebrosen: This assert will never be reached (because futures are lazy): something like `runtime.block_on(fut)` would be needed in order to accomplish that. It happens to be the case that servers cannot (yet) terminate successfully either, but we can fix that aspect of this example in #1079.

core/lib/src/rocket.rs
```diff
@@ -703,54 +703,49 @@ impl Rocket {
...
+    ///     // do things after the server shuts down
+    ///     assert!(result.is_ok());
+    /// });
     /// # }
     /// ```
-    pub fn launch(mut self) -> LaunchError {
+    pub fn spawn_on(
+        mut self,
+        runtime: &tokio::runtime::Runtime,
+    ) -> Result<impl Future<Output = Result<(), rocket_http::hyper::Error>>, LaunchError> {
```
jebrosen: We shouldn't expose `rocket_http::hyper::Error`. Let's have it return `Box<dyn std::error::Error>` for now with a TODO for us to design or rename a concrete error type. I think returning `handle.err_into()` from `TryFutureExt` should work nicely to adapt the error once the type signature is changed.

core/lib/src/rocket.rs
```diff
@@ -792,10 +787,45 @@ impl Rocket {
...
+        Ok(handle)
+    }
+
+    /// Starts the application server and begins listening for and dispatching
+    /// requests to mounted routes and catchers. Unless there is an error, this
+    /// function does not return and blocks until program termination.
+    ///
+    /// # Error
+    ///
+    /// If there is a problem starting the application, a [`LaunchError`] is
```
jebrosen: Note that we now return a `LaunchError` if there's an error *after* launch as well. Happy to address this documentation with #1079 instead, which will change most of these signatures around again.

jhpratt: Hmm...should we be returning a `LaunchError` if it's not necessarily related to launch? I don't believe Hyper's errors were directly exposed before — would it make sense to create an `enum` of the various error types and return that?

core/lib/src/rocket.rs
jhpratt: Futures being lazy keeps coming back to bite me. I need to stop relying on my JavaScript background 😁 … It would certainly make sense to tack a `.shutdown` between the two.

core/lib/src/rocket.rs
jhpratt: Returning a boxed error here requires the same on `.launch()`. Moving ahead with that change as a temporary solution.

jebrosen:
Merged as 70ca157.

PR #1099, opened 2019-08-20
Remove `async_await` feature gate

jhpratt:
`async_await` was recently stabilized on nightly in rust-lang/rust#63209.


jhpratt:
Note that tests will certainly fail, as this requires a nightly which will be released tonight.

jebrosen:
Thanks! Merged in 02a98fe.

PR #1104, opened 2019-08-23
Enable/disable the colorful log

m19c:
This Pull-Request introduces a new way to enable/disable the colorful log output.
Instead of using the environment variable `ROCKET_CLI_COLORS` you can also use the Config itself to do so.

Example:

```
config.enable_colorful_logging();

// or

config.disable_colorful_logging();
```

### DIFF
--- a/core/lib/src/config/config.rs
+++ b/core/lib/src/config/config.rs
@@ -11,6 +11,10 @@ use crate::http::private::Key;
 
 use super::custom_values::*;
 use {num_cpus, base64};
+use yansi::Paint;
+use std::env;
+
+crate const COLORS_ENV: &str = "ROCKET_CLI_COLORS";
 
 /// Structure for Rocket application configuration.
 ///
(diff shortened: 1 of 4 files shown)


jhpratt:
Following the existing pattern in that module, an API of `.set_colorful_logging(bool)` seems more in line than `.{enable|disable}_colorful_logging()`.

m19c:
You are right. There is no enable/disable so far... I'll change it to `set_colorful_logging`.

should i add tests?

m19c:
done.

if I have to add tests:
where would be a good place for it (as far as I can see there are no tests for `Config`)?

m19c:
Ping

jebrosen:
This looks relatively complex for something that feels like it should be a simple change.

What is the motivation for having this rather than doing (for example) `std::env::set_var("ROCKET_CLI_COLORS", "off");`?

PR #1106, opened 2019-08-29
Prefer async-std over tokio

jhpratt:
This currently only affects tests and simple APIs for files. It does let fetching file metadata be asynchronous, so it's not a completely trivial change.

This also mostly eliminates the need for the tokio compatibility layer (though it would've been temporary anyways), with the only usage now being in `rocket.rs` for wrapping the executor.

I have also pinned tokio to 0.2.0-alpha.2 to avoid breaking hyper. This can be removed in the future.

### DIFF
--- a/core/lib/Cargo.toml
+++ b/core/lib/Cargo.toml
@@ -28,7 +28,7 @@ rocket_codegen = { version = "0.5.0-dev", path = "../codegen" }
 rocket_http = { version = "0.5.0-dev", path = "../http" }
 futures-preview = "0.3.0-alpha.18"
 futures-tokio-compat = { git = "https://github.com/Nemo157/futures-tokio-compat", rev = "8a93702" }
-tokio = "0.2.0-alpha.2"
+tokio = "=0.2.0-alpha.2"
 yansi = "0.5"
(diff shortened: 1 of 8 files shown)


jebrosen:
Pro: `File` now works on not-tokio runtimes.
Con: Now there are two runtimes in use.

I'm in favor of this (at least for now) also for the sake of not needing `tokio-futures-compat`.

I already have some ideas about longer-term plans for runtime compatibility and usage in general, I just need to write them down.

jhpratt:
Is it a different runtime, though? `async-std` just provides various types with `Future` implementations, which can then be run on any runtime. At least that was my understanding.

jebrosen:
Merged as 2606c64. Thanks!

zmlzm:
img

async-std will stable 1.0 at 2019-09-26 ? It is a good choose!

jhpratt:
@zmlzm Why would it need to be stable by then? async/await won't land until November, and Rocket 0.5 won't be released until after that point anyway. We have plenty of time to let things settle.

zmlzm:
@jhpratt AaHa, I find that at there  announcing-blog : https://async.rs/blog/announcing-async-std/

PR #1114, opened 2019-09-02
Linting

jhpratt:
Most of these lints are either built into rustc (and are allowed by default) or are available via clippy. A couple I've done by hand by using ripgrep to find instances (an example of this is `use foo` declarations, which isn't available via tooling).

Commit messages, which describe what was changed:

- Remove unnecessary braces on import
- Use shorter path where already imported
  In these situations, the relevant path is already imported into the current scope. As such, we can remove parts of the path without cost.

- Remove `use foo`, which is a noop
  This is likely left over from the conversion to 2018, where `extern crate` would have been converted to this.

- Use struct field shorthand where possible

- Remove identity function
  `.params()` already returns an iterator

- Restrict visibility where not externally visible
  Use `crate` over `pub`. This way, rustc can let us know if something is actually unused, rather than assuming it may be used publicly.

- Remove `self` from path where possible
  Doing this is allowed in newer versions of the compiler.


jebrosen:
> rustc reported all branches other than the first as unreachable, with the reason being that it was matched on an ident.

The enum variants in question should still be in scope there, so that *is* concerning. Which lint says this?

Which lints were all of these, in general? We may want to `#[warn()]` on some of them more permanently, at least the ones don't require clippy.

> Remove self from path where possible
> Doing this is allowed in newer versions of the compiler.

Is it a newer version than we already require? I *think* it's the same version #964 bumped to (`uniform_paths`?), but I want to be sure.

---

I think I will want to hold off on some of the more disruptive changes until `async` is merged into `master`, to avoid making that task much harder. But a lot of these look fine.

jhpratt:
The non-clippy lints come from [this list](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html). I enabled them one at a time to avoid information overload; there's a few false positives.

## rustc lints

- `anonymous_parameters`
  No warnings.

- `bare_trait_objects` & `elided_lifetimes_in_paths`
  I believe this falls under `rust_2018_idoms`, which is already `warn`. As such, no warnings.

- `missing_copy_implementations`
  3 warnings. I don't believe any of these strictly need `copy`, as they're only really used in situations where a specific structure is needed (and not shared).

- `missing_debug_implementations`
  14 warnings. No harm in adding these in where possible, but I didn't do so.

- `missing_docs`
  7 warnings. All are new in the async migration.

- `single_use_lifetimes`
  Many warnings. After I tried a few situations, I believe all are false positives.

- `trivial_casts`
  6 warnings, all false positives.

- `trivial_numeric_casts`
  No warnings.

- `unreachable_pub`
  Tons of warnings. This lint is one I definitely wish was enabled by default. This is where I replace `pub` with `crate`.

- `unsafe_code`
  One warning, the usage of `unsafe` is documented.

- `unused_extern_crates`
  No warnings, as `extern crate` is rarely used.

- `unused_import_braces`
  Two warnings. Strictly stylistic, but trivial to fix.

- `unused_qualifications`
  10 warnings. All of these are changed except the one inside a macro.

- `unused_results`
  Plenty of warnings, I didn't bother to touch any of these.

- `variant_size_differences`
  No warnings. Could be good to enable to ensure memory usage doesn't grow a lot unexpectedly.

## clippy lints

Just including the ones that have warnings.

- [`redundant_field_names`](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_field_names)
  Stylistic. Trivial to fix.

- [`module_inception`](https://rust-lang.github.io/rust-clippy/master/index.html#module_inception)
  A few warnings, but I don't think it's a big deal.

- [`large_enum_variant`](https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant)
  Apparently a different threshold than rustc.

- [`identity_conversion`](https://rust-lang.github.io/rust-clippy/master/index.html#identity_conversion)
  Zero reason to not straight up `deny` this one, it's literally useless. One warning, which is fixed.

- [`should_implement_trait`](https://rust-lang.github.io/rust-clippy/master/index.html#should_implement_trait)
  Currently warning on an `into_iter()` method. While it wouldn't hurt to implement the trait, it would require a bit of work. There's also a `from_str()` method, though the return type is not the same.

- [`if_same_then_else`](https://rust-lang.github.io/rust-clippy/master/index.html#if_same_then_else)
  I don't think this is an issue, as the compiler almost certainly optimizes this away. It allows for clearer commenting.

- [`redundant_pattern_matching`](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_pattern_matching)
  Useless lint here, as it's exclusively because of pear's proc macros.

- [`cast_lossless`](https://rust-lang.github.io/rust-clippy/master/index.html#cast_lossless)
  Wouldn't hurt to change this, though I didn't in this PR.

---

> Is it a newer version than we already require? I think it's the same version #964 bumped to (uniform_paths?), but I want to be sure.

Nope. I don't remember the minimum nightly, I know we require at least an August compiler because of async/await stabilization, and I have used this before then on my own projects. #964 sounds right.

> I think I will want to hold off on some of the more disruptive changes until async is merged into master, to avoid making that task much harder. But a lot of these look fine.

Fair. I'm sure it'll already be quite interesting to merge the branch in.

> The enum variants in question should still be in scope there, so that is concerning. Which lint says this?

Looking into this again, I believe I was mistaken. I recall experimentally removing the `use` statement in the beginning of the file, which naturally would have taken them out of scope. I must have changed this before re-adding the `use`. I've just pushed a commit reverting that change.

jhpratt:
Per a brief discussion on IRC/Matrix, this will be held off on until after merging `async` into `master` to make things a bit easier. At that point, I'll likely force push similar changes, if Jeb doesn't mind it being a single PR (something he mentioned). Otherwise, I'll split it apart and create multiple PRs.

SergioBenitez:
Closing this out to clear the queue. Let's resume the discussion once `async` is merged into `master`.

jhpratt:
Sounds good @SergioBenitez. Let me know when async gets merged and I'll create another PR.

PR #1119, opened 2019-09-06
Implement fmt::display for Json<T>

InstinctBas:
First submission so I'm sure there is room for improvement here, definitely not sure if the error condition should be handled the way it is but I'm pretty new to Rust still so...

This simply implements fit::display for Json<T> allowing you to use to_string to convert the data structure back to a string.

In my use case I'm receiving Json data from the front end, have Rocket parse it so I know I'm receiving valid data, then do additional validation checks on the data, and finally send the JSON to a stored function in PostgreSQL. It is here where I need the data as a string again so being able to just call to_string is a nice convenience here.

### DIFF
--- a/contrib/lib/src/json.rs
+++ b/contrib/lib/src/json.rs
@@ -14,6 +14,7 @@
 //! features = ["json"]
 //! ```
 
+use std::fmt;
 use std::ops::{Deref, DerefMut};
 use std::io::{self, Read};
 use std::iter::FromIterator;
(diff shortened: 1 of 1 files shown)


jebrosen:
I looked into the error handling issue you mention, and it's my opinion that we should not implement this in `rocket_contrib`. This change would indeed allow `input.to_string()` instead of `serde_json::to_string(&*input)` which is more convenient, but `.to_string()` and `format!` assume that the `Display` implementation never errors. Since `serde_json::to_string` can return an error if anything inside of the data fails to serialize, I think that providing this will hide a possible panic condition for a very minor ergonomics benefit.

If you prefer the convenience of `to_string()` and know your data can't fail to serialize, you should be able to write an extension trait like so:

```rust
trait JsonToStringExt {
    fn to_string(&self) -> String;
}

impl<T: Serialize> JsonToStringExt for Json<T> {
    fn to_string(&self) -> String {
        serde_json::to_string(&**self).expect("failed to serialize")
    }
}
```

PR #1121, opened 2019-09-06
Update known_media_types.rs with .mp3 support

tomgie:


core/http/src/known_media_types.rs
```diff
@@ -40,6 +40,7 @@ macro_rules! known_media_types {
         MOV (is_mov): "quicktime video", "video", "quicktime",
         MP4 (is_mp4): "MPEG4 Video", "video", "mp4",
         ZIP (is_zip): "ZIP archive", "application", "zip",
+        MP3 (is_mp3): "MP3 Audio", "audio", "mp3",
```
jebrosen: This MIME type should be `audio/mpeg`, not `audio/mp3`.

SergioBenitez: Why was this marked as resolved?

core/http/src/known_media_types.rs
```diff
@@ -40,6 +40,7 @@ macro_rules! known_media_types {
         MOV (is_mov): "quicktime video", "video", "quicktime",
         MP4 (is_mp4): "MPEG4 Video", "video", "mp4",
         ZIP (is_zip): "ZIP archive", "application", "zip",
+        MP3 (is_mp3): "MP3 Audio", "audio", "mpeg",
```
jebrosen: Although better, this won't write be right after all because it's not a one to one mapping: `audio/mpeg` might not be an MP3 file. But accommodating that and similar cases probably requires restructuring the entire way these impls are generated...

PR #1123, opened 2019-09-06
Add `impl Resonder` for either::Either

jhpratt:


core/lib/src/response/responder.rs
```diff
@@ -321,6 +327,18 @@ impl<'r, R: Responder<'r> + Send + 'r, E: Responder<'r> + fmt::Debug + Send + 'r
     }
 }
 
+/// Responds with the wrapped `Responder`.
+impl<'r, L: Responder<'r> + Send + 'r, R: Responder<'r> + Send + 'r> Responder<'r> for either::Either<L, R> {
+    fn respond_to(self, req: &'r Request<'_>) -> response::ResultFuture<'r> {
+        Box::pin(async move {
```
jebrosen: We shouldn't need to make another box here, since each branch already creates a boxed future

jhpratt: Isn't that also the case in other implementations in the same module?

jhpratt: Quick check says yes. I'll do that as a separate PR.

core/lib/Cargo.toml
```diff
@@ -40,6 +40,7 @@ base64 = "0.10"
 pear = "0.1"
 atty = "0.2"
 async-std = "0.99.4"
+either = "1.5.2"
```
SergioBenitez: Is there a reason to depend on `1.5.2` vs. just `1.0`?

core/lib/src/response/responder.rs
```diff
@@ -98,6 +98,12 @@ use crate::request::Request;
 ///     to the client. If the `Result` is `Err`, the wrapped `Err` responder is
 ///     used to respond to the client.
 ///
+///   * **::either::Either&lt;L, R>** _where_ **L: Responder** and **R:
```
SergioBenitez: Why not just call it `Either`? That is, why use `either::Either`?

SergioBenitez:
I think we should use `Either` from the `either` crate via a re-export in the `response` module. This enables us to inline the documentation into the Rocket crate while also enabling compatibility with the rest of the ecosystem.

The plan has been to expose a kind of `Either` type to remove the dependency on the `specialization` feature by replacing `Result` with `Either` in appropriate cases. I believe the `impl Responder for Result` is the only place we use specialization at the moment. As such, after merging this, we should investigate a migration strategy away from `specialization` and the current use of `Result`. This is a silent breaking change, so the strategy should include some way to protect current dependencies of the `Result` specialization functionality from the silent breaking change. One idea, for example, is to enable specialization via a (default) feature flag and emit a deprecation warning at runtime if the specialization functionality is used.

jebrosen:
>  is the only place we use specialization at the moment

There are a few others - a few that look like performance optimizations, and IIRC a specialization for `T: Responder + Hash` for `Created` responses.

> One idea, for example, is to enable specialization via a (default) feature flag and emit a deprecation warning at runtime if the specialization functionality is used.

Adding a feature flag arguably counts as a breaking change (for anyone using `default-features=false`), and my understanding is we don't want `0.5` to use it at all so we will surely be able to compile on stable.

Is it possible to `#[deprecated]` an `impl` at compile time?

SergioBenitez:
> > is the only place we use specialization at the moment
> 
> There are a few others - a few that look like performance optimizations, and IIRC a specialization for `T: Responder + Hash` for `Created` responses.

Ah right. Bummer.

The `Created` one looks straightforward to work around via two different constructors for `Created`, one which saves the hash a priori. I think we can apply the same "just use different constructors" trick to the other two cases as well. I'd be happy to tackle this.

> > One idea, for example, is to enable specialization via a (default) feature flag and emit a deprecation warning at runtime if the specialization functionality is used.
> 
> Adding a feature flag arguably counts as a breaking change (for anyone using `default-features=false`), and my understanding is we don't want `0.5` to use it at all so we will surely be able to compile on stable.

That's true; we could still go the flag route but disable it by default.

> Is it possible to `#[deprecated]` an `impl` at compile time?

This would have the same issue of needing a nightly compiler.

core/lib/Cargo.toml
jhpratt: It's just what was added when I used `cargo add`. There shouldn't be any particular reason to require the higher version.

core/lib/src/response/responder.rs
jhpratt: In this case because it's external unlike the others. Given that you suggested re-exporting it, that reasoning no longer applies.

jebrosen:
> > Is it possible to #[deprecated] an impl at compile time?
> 
> This would have the same issue of needing a nightly compiler.

I should have been more specific. I was thinking of backporting this functionality in a 0.4.x and simultaneously deprecating the "bad" `impl`, then removing it entirely in 0.5.

jhpratt:
I've addressed Sergio's concerns in the most recent commit.

I think backporting a `#[deprecated]` to avoid unnecessary specialization makes sense, especially given that he upgrade should be as simple as an `.into()` — `Either` implements `From<Result<R, L>>`.

SergioBenitez:
> > > Is it possible to #[deprecated] an impl at compile time?
> > 
> > 
> > This would have the same issue of needing a nightly compiler.
> 
> I should have been more specific. I was thinking of backporting this functionality in a 0.4.x and simultaneously deprecating the "bad" `impl`, then removing it entirely in 0.5.

Ah, got it. I like it!


> I think backporting a `#[deprecated]` to avoid unnecessary specialization makes sense, especially given that he upgrade should be as simple as an `.into()` — `Either` implements `From<Result<R, L>>`.

That's right. We should mention this in the deprecation notice as well.

jhpratt:
Would you prefer the deprecation be done in this PR? Then this could be cherry-picked for the backport.

SergioBenitez:
A separate commit. I'm planning to address the deprecation in tandem with the removal of the other uses of specialization shortly.

jebrosen:
> I think backporting a `#[deprecated]` to avoid unnecessary specialization makes sense, especially given that the upgrade should be as simple as an `.into()` — `Either` implements `From<Result<R, L>>`.

I hate being "that guy", but I'm guessing a lot of people do use `?` and it won't be quite so straightforward to switch.

jhpratt:
Hmm...that is true. I guess that's the advantage of a deprecation over immediate removal; give people some time to rewrite their code.

SergioBenitez:
> > I think backporting a `#[deprecated]` to avoid unnecessary specialization makes sense, especially given that the upgrade should be as simple as an `.into()` — `Either` implements `From<Result<R, L>>`.
> 
> I hate being "that guy", but I'm guessing a lot of people do use `?` and it won't be quite so straightforward to switch.
> 
> EDIT: Which is not to say it shouldn't be deprecated, just that we might want to provide some additional migration guidance.

Yeah, this is the most unfortunate consequence. What's more, `Try` does not seem to be on the path to near-term specialization - it hasn't received any attention whatsoever recently - so we couldn't make this easier if we wanted to.

One thing I had considered was to keep the `Result<T, E> where E: Debug + Responder` implementation instead, which would enable the use of `?`, and which would be a compile-time breaking change, which seems like the better approach. Of course, this means that all implementations where `E` _isn't_ a `Responder` now fail. If only we had some usage metrics.

maxdeviant:
> > I think backporting a `#[deprecated]` to avoid unnecessary specialization makes sense, especially given that the upgrade should be as simple as an `.into()` — `Either` implements `From<Result<R, L>>`.
> 
> I hate being "that guy", but I'm guessing a lot of people do use `?` and it won't be quite so straightforward to switch.
> 
> EDIT: Which is not to say it shouldn't be deprecated, just that we might want to provide some additional migration guidance.

@jebrosen Sorry if I'm lacking context here, but are you saying that with this change people would no longer be able to use `?` in their routes?

jebrosen:
@maxdeviant Since 2537a11 this PR and conversation are somewhat outdated. That commit implements `Responder Result<T: Responder, E: Responder>` and adds a `Debug` type to opt into the previous "500 error for non-Responder" behavior, so `?` should work as-is in most cases.

jebrosen:
Do we still need this after 2537a11? I think this implementation is now identical to the one we ended up with for `Result`.

jhpratt:
IMO it wouldn't hurt to at least have this behind a feature flag. Using `Err` for something that isn't an error just seems odd to me.

jebrosen:
Assuming I have the syntax right, `#[cfg(feature = "either")]` would work in conjunction with `optional = true` for `either`.

I'm personally neutral on merging this. @SergioBenitez, yay or nay?

zlnpgtdtfezf:
Any update on this?

SergioBenitez:
If `either` is already in our dependency tree, let's add this. If it's not, let's not. I don't think we should add a feature flag for this, in either case.

SergioBenitez:
In any case, let's get a new PR with just the change we're looking for.

jebrosen:
`either` is in the dependency tree of `rocket_contrib` via `r2d2_redis`, when that database is enabled - but not anywhere else, currently.

SergioBenitez:
@jebrosen Thanks for checking. In the case, especially because `either` has `into_result()`, I don't think we should add `either` as a dependency for this reason alone.

PR #1125, opened 2019-09-06
Use `BoxFuture` type for readability

jhpratt:


jebrosen:
Merged as 8b419c1. These look *much* better than before.

PR #1143, opened 2019-09-20
Remove `crate_visibility_modifier` feature

jhpratt:
In some situations, `pub(crate)` wasn't strictly necessary, as it was
within a context where it would not be exposed regardless of scope. In
these situations, `pub` is used per request.

Pinging @SergioBenitez and @jebrosen for review. This should be looked through carefully to ensure nothing is inadvertently exposed or hidden. Checking the documentation should be sufficient for this.

### DIFF
--- a/core/codegen/src/lib.rs
+++ b/core/codegen/src/lib.rs
@@ -1,5 +1,4 @@
 #![feature(proc_macro_diagnostic, proc_macro_span)]
-#![feature(crate_visibility_modifier)]
 #![recursion_limit="128"]
 
 #![doc(html_root_url = "https://api.rocket.rs/v0.5")]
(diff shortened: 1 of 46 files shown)


core/codegen/src/lib.rs
```diff
@@ -114,14 +113,14 @@ mod syn_ext;
 
 use crate::http::Method;
 use proc_macro::TokenStream;
-crate use devise::proc_macro2;
+pub(crate) use devise::proc_macro2;
```
jebrosen: Since it's in a crate root, this should have the same effect as no visibility modifier at all as access is always granted to children. If that is the case, let's remove the `pub(crate)` here to stay consistent with other `use`s.

jhpratt: Good catch. There's this in a couple places with `mod` instead of `use` as well.

jhpratt:
For reproducing my check, you'll want to generate docs for the current master and this PR. Move the `target/doc` directories to some other (temporary) folder; I'll call these `master` and `pr`.

In both of these directories, run `rm *.*` and `rm -r src`. This will remove any artefacts that are unrelated to the diff and the file for source code — we know that one will change, we're only interested in the doc changes.

Next, run `diff -r main pr > diff.txt`. Open `diff.txt` up in a text editor. Since we're looking for the addition or deletion of functions, modules, constants, etc. the only lines relevant here are those with a difference in length. To quickly check this, I removed all irrelevant lines (`---`, `0c0`, etc.), leaving just the file names and diff.

As of now, there are no lines with differing lengths, indicating that nothing was exposed or hidden accidentally.

core/codegen/src/lib.rs
```diff
@@ -114,14 +113,14 @@ mod syn_ext;
...
-crate use devise::proc_macro2;
+use devise::proc_macro2;
 
-crate static ROUTE_STRUCT_PREFIX: &str = "static_rocket_route_info_for_";
-crate static CATCH_STRUCT_PREFIX: &str = "static_rocket_catch_info_for_";
-crate static CATCH_FN_PREFIX: &str = "rocket_catch_fn_";
-crate static ROUTE_FN_PREFIX: &str = "rocket_route_fn_";
-crate static URI_MACRO_PREFIX: &str = "rocket_uri_macro_";
-crate static ROCKET_PARAM_PREFIX: &str = "__rocket_param_";
+pub(crate) static ROUTE_STRUCT_PREFIX: &str = "static_rocket_route_info_for_";
```
jebrosen: I think it's the same situation (don't need any visibility modifier) for these `static`s.

jhpratt: Fixed.

jebrosen:
Test failures look unrelated -- UI test regression possibly from rust-lang/rust#64498.

SergioBenitez:
@jebrosen That looks like a serious regression. I'm not sure if that's the PR to blame, but we should figure out what is.

jebrosen:
> you'll want to generate docs for the current master and this PR

You would actually need to do this with all features enabled in both debug and release mode. But that difference only applies to one or two types, and I also manually checked that everything is okay.

I'll merge once my local test run finishes.

jebrosen:
Awesome! Merged as 4e6a7dd.

PR #1149, opened 2019-09-30
Remove stringified number in macro call

jhpratt:
As the code is otherwise identical, we can use `stringify!` to generate this in the documentation.


jebrosen:
This approach doesn't actually work (and I think is part of the motivation for the horrifying-yet-charming `docify!` macro used elsewhere in the crate).

jhpratt:
Hmm...you're right. I naïvely assumed that `stringify!` was eagerly evaluated.

jhpratt:
Upsteam: rust-lang/rust#52607

PR #1153, opened 2019-10-03
Add support for RON format

liketechnik:
This PR adds support for [RON](https://github.com/ron-rs/ron) as an alternative to JSON. I plan to add comments to the code asap, but wanted to have the code already here so if there's something that needs to be changed I don't have to write comments twice (especially because they wouldn't differ from the comments for JSON, except for the format itself). 

The whole code is very strictly oriented at the already existing JSON code, I tested its functionality in a hobby project (which is not yet released, if ever), but didn't write any tests, as I was not able to find any JSON specific tests either.

### DIFF
--- a/contrib/lib/Cargo.toml
+++ b/contrib/lib/Cargo.toml
@@ -20,6 +20,7 @@ databases = ["r2d2", "tokio/blocking", "tokio/rt-threaded", "rocket_contrib_code
 default = ["json", "serve"]
 json = ["serde", "serde_json", "tokio/io-util"]
 msgpack = ["serde", "rmp-serde", "tokio/io-util"]
+ron = ["serde", "ron_crate"]
 tera_templates = ["tera", "templates"]
 handlebars_templates = ["handlebars", "templates"]
 helmet = ["time"]
@@ -51,6 +52,7 @@ log = "0.4"
 serde = { version = "1.0", optional = true }
 serde_json = { version = "1.0.26", optional = true }
 rmp-serde = { version = "0.14.0", optional = true }
+ron_crate = { version = "0.5.1", optional = true, package = "ron" }
 
 # Templating dependencies.
 handlebars = { version = "2.0", optional = true }
--- a/contrib/lib/Cargo.toml
+++ b/contrib/lib/Cargo.toml
@@ -52,7 +52,7 @@ log = "0.4"
 serde = { version = "1.0", optional = true }
 serde_json = { version = "1.0.26", optional = true }
 rmp-serde = { version = "0.14.0", optional = true }
-ron_crate = { version = "0.5.1", optional = true, package = "ron" }
+ron_crate = { version = "0.6.0", optional = true, package = "ron" }
 
 # Templating dependencies.
 handlebars = { version = "2.0", optional = true }
--- a/contrib/lib/src/ron.rs
+++ b/contrib/lib/src/ron.rs
@@ -95,5 +95,3 @@ impl<T> DerefMut for Ron<T> {
         &mut self.0
     }
 }
-
-
--- a/contrib/lib/src/ron.rs
+++ b/contrib/lib/src/ron.rs
@@ -69,10 +69,10 @@ impl<'a, T: Deserialize<'a>> FromData<'a> for Ron<T> {
 
 impl<'r, T: Serialize> Responder<'r> for Ron<T> {
     fn respond_to(self, req: &'r Request<'_>) -> response::ResultFuture<'r> {
-        match ron::ser::to_string(&self.0) {
+        match ron::ser::to_string_pretty(&self.0, ron::ser::PrettyConfig::default()) {
             Ok(string) => Box::pin(async move { Ok(content::Ron(string).respond_to(req).await.unwrap()) }),
             Err(e) => Box::pin (async move {
-                error_!("JSON failed to serialize: {:?}", e);
+                error_!("RON failed to serialize: {:?}", e);
                 Err(Status::InternalServerError)
             })
         }
--- a/contrib/lib/src/ron.rs
+++ b/contrib/lib/src/ron.rs
@@ -69,7 +69,7 @@ impl<'a, T: Deserialize<'a>> FromData<'a> for Ron<T> {
 
 impl<'r, T: Serialize> Responder<'r> for Ron<T> {
     fn respond_to(self, req: &'r Request<'_>) -> response::ResultFuture<'r> {
-        match ron::ser::to_string_pretty(&self.0, ron::ser::PrettyConfig::default()) {
+        match serialize(&self.0) {
             Ok(string) => Box::pin(async move { Ok(content::Ron(string).respond_to(req).await.unwrap()) }),
             Err(e) => Box::pin (async move {
                 error_!("RON failed to serialize: {:?}", e);
@@ -79,6 +79,14 @@ impl<'r, T: Serialize> Responder<'r> for Ron<T> {
     }
 }
 
+fn serialize<T>(value: &T) -> ron::ser::Result<String>
+where
+    T: Serialize,
+{
+    let mut s = ron::ser::Serializer::new(Some(ron::ser::PrettyConfig::default()), true);
+    value.serialize(&mut s)?;
+    Ok(s.into_output_string())
+}
 
 impl<T> Deref for Ron<T> {
     type Target = T;
--- a/contrib/lib/src/ron.rs
+++ b/contrib/lib/src/ron.rs
@@ -154,20 +154,37 @@ impl<'a, T: Deserialize<'a>> FromData<'a> for Ron<T> {
     }
 }
 
-
 /// Serializes the wrapped value into RON. Returns a response with Content-Type
 /// Text and a fixed-size body with the serialized value. If serialization
 /// fails, an `Err` of `Status::InternalServerError` is returned.
 impl<'r, T: Serialize> Responder<'r> for Ron<T> {
     fn respond_to<'a, 'x>(self, req: &'r Request<'a>) -> BoxFuture<'x, response::Result<'r>>
-        where 'a: 'x, 'r: 'x, Self: 'x
+    where
+        'a: 'x,
+        'r: 'x,
+        Self: 'x,
     {
-        match ron_crate::ser::to_string_pretty(&self.0, ron_crate::ser::PrettyConfig::default()) {
-            Ok(string) => Box::pin(async move { Ok(content::Plain(string).respond_to(req).await.unwrap()) }),
-            Err(e) => Box::pin (async move {
+        let mut output = Vec::new();
+        match ron_crate::ser::Serializer::new(
+            &mut output,
+            Some(ron_crate::ser::PrettyConfig::default()),
+            true,
+        ) {
+            Ok(mut serializer) => {
+                match self.0.serialize(&mut serializer) {
+                    Ok(_) => Box::pin(async move {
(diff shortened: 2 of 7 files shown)


contrib/lib/Cargo.toml
```diff
@@ -19,6 +19,7 @@ databases = ["r2d2", "rocket_contrib_codegen/database_attribute"]
 # User-facing features.
 default = ["json", "serve"]
 json = ["serde", "serde_json"]
+serde_ron = ["serde", "ron"]
```
liketechnik: I'm not 100% sure about the naming here, but just 'ron' won't work because the crate already uses that name.

core/lib/src/response/content.rs
```diff
@@ -90,5 +90,6 @@ ctrs! {
     Html: HTML, "HTML", "text/html",
     Plain: Plain, "plain text", "text/plain",
     Css: CSS, "CSS", "text/css",
-    JavaScript: JavaScript, "JavaScript", "application/javascript"
+    JavaScript: JavaScript, "JavaScript", "application/javascript",
+    Ron: RON, "RON", "text/x-ron"
```
liketechnik: I couldn't find any hint on what mimetype to use for Ron, so I went with this. Only problem I noticed with this: At least firefox doesn't show the data in browser, you have to download the response as text to view it. I imagine this to be annoying in the long run when debuggin/developing.

contrib/lib/src/ron.rs
```diff
@@ -0,0 +1,179 @@
...
+
+impl<T> Ron<T> {
+    #[inline(always)]
+    pub fn into_inner(self) -> T {
+        self.0
+    }
+}
+
+
+//// No idea what the default limit for ron is. Using Json's default of 1MB here.
```
jebrosen: This default limit is set by and for `rocket_contib`; JSON itself and presumably ron don't have an inherent limit that I know of.

contrib/lib/src/ron.rs
```diff
@@ -0,0 +1,179 @@
...
+            Ok(string) => Box::pin(async move { Ok(content::Ron(string).respond_to(req).await.unwrap()) }),
+            Err(e) => Box::pin (async move {
+                error_!("RON failed to serialize: {:?}", e);
+                Err(Status::InternalServerError)
+            })
+        }
+    }
+}
+
+fn to_string<T>(value: &T) -> ron::ser::Result<String>
```
jebrosen: `to_string` confused me earlier (I thought it referred to the function in `ron` until I got the). Can we give this a clearer name or inline it?

contrib/lib/src/ron.rs
```diff
@@ -0,0 +1,179 @@
...
+where
+    T: Serialize,
+{
+    let mut s = ron::ser::Serializer::new(Some(ron::ser::PrettyConfig::default()), true);
```
jebrosen: What is the reasoning behind using a pretty formatter here?

contrib/lib/src/ron.rs
```diff
@@ -0,0 +1,179 @@
...
+
+impl<T> DerefMut for Ron<T> {
+    #[inline(always)]
+    fn deref_mut(&mut self) -> &mut T {
+        &mut self.0
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct RonValue(pub ron::value::Value);
```
jebrosen: We may not need `RonValue`; IIRC `JsonValue` exists primarily for some reasons related to macro reexporting.

core/lib/src/response/content.rs
jebrosen: I believe we have stuck with only registered MIME types in the past, in which case we shouldn't add this and use something else like `text/plain` instead.

contrib/lib/Cargo.toml
jebrosen: I'm tempted to do a feature named `ron` and use `package=` to rename the crate locally. I want to check what other crates do before doing anything, though.

contrib/lib/src/ron.rs
liketechnik: The responses are easier to read for humans.

contrib/lib/src/ron.rs
liketechnik: So I completely remove this part?

core/lib/src/response/content.rs
liketechnik: That sounds totally reasonable, only doubt I have is that it won't be easy to determine the format the responses are sent from the outside.

contrib/lib/src/ron.rs
jebrosen: This and all trait implementations. But let me make sure it's actually unnecessary and I didn't miss something.

core/lib/src/response/content.rs
jebrosen: Applications can always use `Content` and set their own arbitrary content types for resources if they need to. Since browsers don't have any built-in support for RON, setting a media type has limited value in comparison to other file formats.

contrib/lib/Cargo.toml
jebrosen: Here's what I think: the feature cannot be named the same as the dependency, we should be consistent with the other features, and we can't rename those. Therefore I propose we use the feature `ron = ["serde", "ron_crate"]` along with the dependency `ron_crate = { version = "0.5.1", optional = true, package = "ron" }`. Code would refer to `ron_crate::whatever`.

contrib/lib/src/ron.rs
jebrosen: Let's change this comment to `/// Default limit for RON is 1MB.` for consistency with the other formats

contrib/lib/src/ron.rs
```diff
@@ -0,0 +1,179 @@
...
+    }
+}
+
+
+//// No idea what the default limit for ron is. Using Json's default of 1MB here.
+const LIMIT: u64 = 1 << 20;
+
+#[derive(Debug)]
+pub enum RonError<'a> {
+    Io(io::Error),
```
jebrosen: These should have doc comments (can probably copy almost verbatim from the ones in JSON).

contrib/lib/src/ron.rs
jebrosen: `JsonValue` was added in a019f0d. I don't believe the same concern applies here; afaict `RonValue` is essentially identical to `Ron<Value>`.

core/http/src/known_media_types.rs
```diff
@@ -40,6 +40,7 @@ macro_rules! known_media_types {
         MOV (is_mov): "quicktime video", "video", "quicktime",
         MP4 (is_mp4): "MPEG4 Video", "video", "mp4",
         ZIP (is_zip): "ZIP archive", "application", "zip",
+        RON (is_ron): "Rusty Object Notation", "text", "plain",
```
jebrosen: This isn't the right way to go about this; nothing should change in `known_media_types.rs` at all since there's no media type for RON. Checks against RON should probably either require/expect `text/plain` or preferably not check the MIME type at all, and code that outputs RON should either say it is `text/plain` or not make any claim.

jebrosen:
@liketechnik any progress on this? I plan to merge this (perhaps with cleanups) by this weekend if I don't hear something sooner.

liketechnik:
I've almost finished to apply your suggestions and will hopefully push until tomorrow.

jebrosen:
This is looking good so far. For tests, let's copy `examples/json/` to `examples/ron` and adjust it accordingly.

liketechnik:
I've fixed the implementation in contrib (imported the wrong crate, the feature flag for exporting the module had the wrong name), bumped the ron version to `0.6.0` (latest stable release) and adjusted the json tests. Let me know if there's anything missing.

contrib/lib/src/ron.rs
```diff
@@ -0,0 +1,189 @@
...
+
+
+/// Serializes the wrapped value into RON. Returns a response with Content-Type
+/// Text and a fixed-size body with the serialized value. If serialization
+/// fails, an `Err` of `Status::InternalServerError` is returned.
+impl<'r, T: Serialize> Responder<'r> for Ron<T> {
+    fn respond_to<'a, 'x>(self, req: &'r Request<'a>) -> BoxFuture<'x, response::Result<'r>>
+        where 'a: 'x, 'r: 'x, Self: 'x
+    {
+        match ron_crate::ser::to_string_pretty(&self.0, ron_crate::ser::PrettyConfig::default()) {
```
jebrosen, line 165: Somewhere along the way we lost struct names. As far as I can tell this does not matter for deserialization, but we should probably either keep them for readability, or use entirely non-pretty output if speed/size is preferred.

examples/ron/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+name = "ron"
+version = "0.0.0"
+workspace = "../../"
+edition = "2018"
+publish = false
+
+[dependencies]
+rocket = { path = "../../core/lib" }
+serde = "1.0"
+ron = "0.5.1"
```
jebrosen, line 11: We can remove this line, since it's now the wrong version of `ron` and the example doesn't actually need it.

examples/ron/src/main.rs
```diff
@@ -0,0 +1,83 @@
...
+// We're going to store all of the messages here. No need for a DB.
+type MessageMap = Mutex<HashMap<ID, String>>;
+
+#[derive(Serialize, Deserialize)]
+struct Message {
+    id: Option<ID>,
+    contents: String
+}
+
+#[derive(Serialize, Deserialize)]
```
jebrosen, line 26: This can be only `derive(Serialize)`, since it's never used as input to the server.

examples/ron/src/main.rs
```diff
@@ -0,0 +1,83 @@
...
+struct ErrorMessage {
```
jebrosen, line 27: Is this added because it's harder to construct a `ron::Value` without shorthand like `json!`?

Let's rename this to `StatusMessage`, since it also encodes `ok` values. Additionally, we can change this to `status: &'static str`  and all of the `.to_owned()` calls will go away.

liketechnik: That's correct, ron doesn't have a `ron!` equivalent yet. I still have my problems with the way rust handles strings, `.to_owned()` usually just works, but with `&'static str` it's a lot better.

examples/ron/src/main.rs
liketechnik: Oops, didn't realize that.

contrib/lib/src/ron.rs
liketechnik: It now serializes correctly with struct names. And yes, the struct names are completely optional for ron.

liketechnik:
I guess I'm rather the one who should thank you for the patience, a lot of things could have been handled better by me in the first place. But you're welcome nevertheless.

liketechnik:
I can't replicate those build failures locally. All builds run through, the `scripts/tests.sh` script runs through, I updated my rust version, no idea what's going wrong there.

jebrosen:
`Responder` was [just changed](https://github.com/SergioBenitez/Rocket/pull/1334#issuecomment-647470370) several hours ago - see in particular https://github.com/SergioBenitez/Rocket/commit/4f792551d300e45738ddb2cfcd044b0f00379a40#diff-b3a72cd4afeba324b4b81b6bee07fc46R171-R179 for `json`. CI attempts to rebase or merge before running tests, which I assume you haven't done locally which is why it seems to work for you.

> the `scripts/tests.sh` script runs through

Incidentally, we should also add a line to `test.sh` to test `ron` independently of the other features.

PR #1163, opened 2019-10-26
Add initial local_cache_async implementation

Follpvosten:
Hello!

This is just an initial working implementation as of now; two things need to be decided still:
1. How would I go about adding doc tests to this? Should there be an async `Request::example` as well? Since the Future returned by `local_cache_async` also has to be awaited, we have to show that in the usage example.
2. Should we keep the sync variant of local_cache or go for async only?
  My take here is that I would keep both, because while passing an async block which executes synchronously is easy, you have to then call `executor::block_on` to get a value from it, adding complexity to sync FromRequest impls.

### DIFF
--- a/core/lib/src/request/request.rs
+++ b/core/lib/src/request/request.rs
@@ -1,5 +1,6 @@
 use std::sync::{Arc, RwLock, Mutex};
 use std::net::{IpAddr, SocketAddr};
+use std::future::Future;
 use std::fmt;
 use std::str;
 
@@ -562,6 +563,23 @@ impl<'r> Request<'r> {
                 self.state.cache.set(f());
                 self.state.cache.get()
             })
+    }
+
+    /// Retrieves the cached value for type `T` from the request-local cached
+    /// state of `self`. If no such value has previously been cached for this
+    /// request, `fut` is `await`ed to produce the value which is subsequently
+    /// returned.
+    pub async fn local_cache_async<T, F>(&self, fut: F) -> &T
+        where F: Future<Output = T>,
+              T: Send + Sync + 'static
+    {
+        match self.state.cache.try_get() {
+            Some(s) => s,
+            None => {
+                self.state.cache.set(fut.await);
+                self.state.cache.get()
+            }
+        }
     }
 
     /// Retrieves and parses into `T` the 0-indexed `n`th segment from the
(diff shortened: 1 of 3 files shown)


jebrosen:
> How would I go about adding doc tests to this? Should there be an async `Request::example` as well?

Possibly. Let's try `Request::example(method, uri, |req| rocket::async_test(async { /* use req */ })` first as a proof of concept, though.

> while passing an async block which executes synchronously is easy, you have to then call `executor::block_on` to get a value from it,

`FromRequest` implementations should still not perform long-blocking operations, since `from_request` is executed inside an async context. In fact, I believe `block_on` inside `from_request` will panic. That's another thing to solve with documentation, or perhaps by unconditionally annotating `FromRequestAsync::from_request`'s call to `FromRqeuest::from_request` as blocking.

Follpvosten:
After a bit of debugging (is there a way to run a specific doctest?), I got my test to pass. I've also specified a lifetime for `&self` and the returned reference since that looks much better in the docs than having two different elided lifetimes.

As for the `FromRequest` issue - should I take any action there?

Also, I just realized that keeping the sync local_cache also means it can be used from `FromRequestAsync` impls still - in case you have a sync operation you know will complete fast (like the request-local state example from the guide), it avoids an unnecessary async block and await operator.

jebrosen:
CI failures look like my fault; I didn't fix them correctly before. I'll see how quickly I can fix that.

> is there a way to run a specific doctest?

Yes! I think `cargo test --doc local_request_async` would do it.

> As for the `FromRequest` issue - should I take any action there?

Let's keep it out of this PR, but I'll expand on the concern a bit on the main tracking issue.

> it avoids an unnecessary async block and await operator.

Sounds good; let's keep both versions.

Follpvosten:
So, looking at this, changing `Guard1` to `FromRequestAsync` wouldn't really work, and `Guard2` doesn't really do anything, so there's nothing it could to async...in short, changing just one of them isn't possible as far as I can see.

So either we change both to `FromRequestAsync` and await `Guard1::from_request` in `Guard2`, or we copy the whole example and make the new variant async.

Follpvosten:
This would be my first idea for showing how to use the async local cache (including a passing test, of course).

jebrosen:
Merged in baef291. Thanks for requesting and implementing this!

In the future it might be worth revamping the whole example as simply incrementing atomics is far from a typical use case, but keeping similar tests to the synchronous version is great for now.

PR #1164, opened 2019-10-27
Update uuid, unicode-xid and base64

paolobarbolini:
Updates uuid, unicode-xid and base64 to reduce duplicate crate versions in projects.


jebrosen:
Can you set the base branch for this to `master` instead of `async`?

jebrosen:
@SergioBenitez the UI tests are a pain since (I think) rust-lang/rust#64151 - see e.g. https://dev.azure.com/SergioBenitez/Rocket/_build/results?buildId=246&view=logs&jobId=fe4e408a-5576-5fa1-f2ab-61ac24819df6&taskId=7f0864f1-1a96-5cff-86e3-45ee63058371&lineStart=1343&lineEnd=1361&colStart=1&colEnd=1

This has two problems. First, there is trailing whitespace after one of the `|`, so `test.sh` will complain after blessing the UI test. Second, that path isn't normalized/normalizable (at least not with `$DIR`). I had "fixed" this in the `async` branch by commenting out the complian-about-trailing-whitespace but the non-normalized path slipped my attention and it doesn't work.

paolobarbolini:
Unfortunately `uuid` made a breaking change by changing `uuid::parser::ParseError` to `uuid::Error` so it wasn't possible to support both of them.

contrib/lib/src/uuid.rs
```diff
@@ -23,7 +23,7 @@ use std::ops::Deref;
 use rocket::request::{FromParam, FromFormValue};
 use rocket::http::RawStr;
 
-pub use self::uuid_crate::parser::ParseError;
+pub use self::uuid_crate::Error as ParseError;
```
jebrosen: So, I think we could actually backwards compatibly support both `0.7` and `0.8` after all:

`type ParseError = <self::uuid_crate::Uuid as FromStr>::Err;`

(although `cargo` would still always choose `0.8` during a `cargo update` unless `--precise` is used)

jebrosen: (cc @SergioBenitez)

jebrosen:
Can you rebase this on `master` to trigger a CI run?

paolobarbolini:
It looks like the macOS build is failing because it's not finding mysqlclient

jebrosen:
Thank you! Merged in cd5f96a.

PR #1170, opened 2019-11-23
Updated catch_type_error Test Expectations

0xcaff:
Tests are failing on the async branch. This is the first step to fixing the failures.

It seems the errors from the rust compiler changed in a recent version, but the expectations for errors weren't updated. This PR updates the expectations.

This leaves two tests failing. It seems that sometimes environment specific file paths get used in the comparison of compiler errors causing it to always fail for a couple of tests:
* https://github.com/SergioBenitez/Rocket/blob/188b8d352863af5bd0e00f21c11fc5090e2abd01/core/codegen/tests/ui-fail/responder-types.stderr#L7

* https://github.com/SergioBenitez/Rocket/blob/188b8d352863af5bd0e00f21c11fc5090e2abd01/core/codegen/tests/ui-fail/typed-uri-bad-type.stderr#L114

* https://github.com/SergioBenitez/Rocket/blob/188b8d352863af5bd0e00f21c11fc5090e2abd01/core/codegen/tests/ui-fail/typed-uri-bad-type.stderr#L128

This is likely caused by an upstream bug in compiletest (the crate). They may be already fixed upstream (in the rust project) and could be fixed fixed when this lands (in the crate): https://github.com/laumann/compiletest-rs/pull/201

### DIFF
--- a/core/codegen/tests/ui-fail/catch_type_errors.rs
+++ b/core/codegen/tests/ui-fail/catch_type_errors.rs
@@ -15,8 +15,9 @@ fn f2(_request: &Request) -> bool {
 }
 
 #[catch(404)]
+//~^ ERROR mismatched types
 fn f3(_request: bool) -> usize {
-    //~^ ERROR mismatched types
+    //~^ ERROR usize: rocket::response::Responder
     10
 }
 
--- a/core/codegen/tests/ui-fail/catch_type_errors.stderr
+++ b/core/codegen/tests/ui-fail/catch_type_errors.stderr
@@ -18,10 +18,7 @@ error[E0308]: mismatched types
   --> $DIR/catch_type_errors.rs:17:1
    |
 17 | #[catch(404)]
-   | ^^^^^^^^^^^^^ expected bool, found reference
-   |
-   = note: expected type `bool`
-              found type `&'_b rocket::Request<'_>`
+   | ^^^^^^^^^^^^^ expected `bool`, found `&rocket::Request<'_>`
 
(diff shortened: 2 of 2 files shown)


jebrosen:
I'm hesitant to make these the "desired" errors; it's only due to an accident that I had trouble fixing (see https://github.com/SergioBenitez/Rocket/commit/9c4167a7efd9907ffe922eca57f97957654a6ca7#diff-e00e6dba1cc1a7479917f658ea6c9257R85 )

For the path issues, I started to fix this in `compiletest` in https://github.com/laumann/compiletest-rs/issues/198#issuecomment-552118193 but didn't like where it was going. `trybuild` has a better solution in the form of `$WORKSPACE` normalization, but @SergioBenitez has tried it and run into some issues.

For the time being I'm inclined to disable the UI tests on the `async` branch until things have settled.

0xcaff:
Oops, missed that context. +1 to disabling tests until things stabilize a bit more with compiletest. These tests are flaky. It seems like this fix doesn't work. The tests are still failing because line numbers for errors changed between the last nightly versions (Nov 22nd and Nov 21st).

jebrosen:
I ended up going ahead and updating these errors and others in addition to disabling UI tests, in the interest of having working CI for any future PRs to the `async` branch. I still hope to fix the error message location, but there are some strange interactions between `async fn` and function pointer syntax that are making it difficult.

0xcaff:
Woo! Awesome! Looking forward to playing with the async stuff next.

PR #1176, opened 2019-11-29
Add support for rusqlite bundled via sqlite_bundled

FylmTM:
[Rusqlite](https://github.com/jgallagher/rusqlite/releases) provides `bundled` feature to use bundles SQLite source, instead of requiring user to install SQLite development library.
It is very convenient (especially for Windows users) to use it.

Unfortunately Cargo does not support enabling features for transitive dependencies, therefore bundled option should be re-exported from `rocket_contrib`.


jebrosen:
> Unfortunately Cargo does not support enabling features for transitive dependencies

I'm not sure exactly what you tried, but it's true that some of the syntaxes don't allow for this.

However, you can still enable it as a dependency of your own crate as described in [the guide](https://rocket.rs/v0.4/guide/state/#usage). Features are unified across all uses of the same version of the same dependency, so this should do exactly what you want:

```toml
rusqlite = { version = "0.16.0", features = ["bundled"] }
```

---

Since there's another way to accomplish this and we don't have individual features for features of any  other database engines, I am going to close this. Thank you, though!

Issue #464, opened 2017-10-20
Warn on launch if {unused managed state} or {attempted use of unmanaged state}


lucasem:
Right now, rocket compiles and runs this:

```rust
#![feature(plugin)]
#![plugin(rocket_codegen)]

extern crate rocket;
use std::sync::atomic::{Ordering, AtomicUsize};

struct Counter {
    c: AtomicUsize,
}
impl Counter {
    fn new() -> Result<Self, ()> {
        Ok(Counter { c: AtomicUsize::new(0) })
    }
    fn update(&self) {
        self.c.fetch_add(1, Ordering::SeqCst);
        println!("count: {}", self.c.load(Ordering::SeqCst))
    }
}

#[get("/")]
fn hello(c: rocket::State<Counter>) -> String {
    c.update();
    "hello".to_string()
}

fn main() {
    rocket::ignite()
        .mount("/", routes![hello])
        .manage(Counter::new()) // whoops, should've unwrapped!
        .launch();
}
```

There are two problems here:
- The managed state `Result<Counter, ()>` is unused, there should be a warning about that.
- Use of unmanaged state `Counter` is attempted, there should be a warning about that.*

*500 is returned from a request to `/` because of unmanaged state. But one wouldn't know the state is unmanaged unless the particular route is being tested (other requests like `/foo` function as expected).

I'm not sure we could easily get this to warn/error during compilation, but we certainly should be able to print a warning _upon launch_ for both of these cases. This would improve usability and reduce bug potential.

SergioBenitez:
Rocket actually _did_ issue a warning at compile-time about unmanaged state in `0.2` and earlier versions of `0.3`. This was accomplished using lints, another kind of syntax extension. Unfortunately, the lint API is incredibly unstable, and lots of nightly-related breakage ensued as a result.

The idea was to bring back these lints in due time. In fact, the [guide](https://rocket.rs/guide/state/#unmanaged-state) still alludes to the lints for just this reason. Unfortunately, it seems unlikely given that the lint API will, with high probability, never stabilize. But one possibility is to introduce a new crate `rocket_lints`, that includes this very unstable code. See #389 for that idea.

[...]

PR #1182, opened 2019-12-09
warn about unused and unmanaged state

msehnout:
Hi,

this PR would fix the issue #464 . I focused on not changing the public API, but it feels a little bit like a spaghetti code, since the types are forwarded from the `StaticRouteInfo`, to `Route`, and finally to the `Rocket` struct. 

Other issues with current solution:
 * requires just another nightly only feature
 * mapping from `TypeId` to human readable type name
 * reading the comments in `std` docs, I'm not sure it is encouraged to use `TypeId`
 * allows only 16 managed states per handler (is it really an issue?)

The mapping can be created during the codegen phase and stored alongside
the `TypeId`. Type names for unused managed states can be obtained using
`std::any::type_name::<T>()`.

I also included an expanded `main.rs` from the `state` example so that you can easily check the generated code. I'll, of course, remove it from the commit. Do you have any suggestions how to improve the approach of communicating the types from endpoint handlers to the `Rocket` struct?


jebrosen:
Thank you for the contribution!

I think this will be a low priority in the short term because of the unstable feature, but I will definitely take a look at the overall approach and provide some feedback.

core/codegen/src/attribute/route.rs
```diff
@@ -414,6 +429,25 @@ fn codegen_route(route: Route) -> Result<TokenStream> {
...
     let format = Optional(route.attribute.format);
+    let mut states: Vec<Option<&syn::Type>> = vec![None; 16];
+
+    // Find all managed states
+    let mut iter = 0;
+    for (_arg_ident, _, ty) in &route.inputs {
+        match ty {
+            syn::Type::Path(type_path) => {
+                if let Some(path_segment) = type_path.path.segments.first() {
+                    if path_segment.ident == "State" {
```
jebrosen: This is the most fragile part of the implementation, since it assumes the `State` type is actually named `State`. There might be some type-level tricks we could employ, but it would probably be pretty convoluted.

Assuming we only match by exact name (which I am not super happy about), this should be checking that the *last* segment (before generic arguments) is `State`.

examples/fairings/src/main.rs
```diff
@@ -1,4 +1,5 @@
 #![feature(proc_macro_hygiene)]
+#![feature(const_type_id)]
```
jebrosen: I can't seem to find a tracking issue for this feature (the unstable points to the catch-all https://github.com/rust-lang/rust/issues/57563).

core/lib/src/codegen.rs
```diff
@@ -19,6 +19,8 @@ pub struct StaticRouteInfo {
     pub handler: StaticHandler,
     /// The route's rank, if any.
     pub rank: Option<isize>,
+    /// The states requested by this route
+    pub states: [Option<std::any::TypeId>; 16],
```
jebrosen: I think `&'static [Option<std::any::TypeId>]` should work here, simplifying some other code as well.

core/codegen/src/attribute/route.rs
msehnout: Yeah, I'm worried about this as well, but I couldn't think of any other way to compare the types.

msehnout:
Thanks for the suggestions, I'll look into it in couple of days! :-)

core/codegen/src/attribute/route.rs
jebrosen: Here's a totally untested concept, but I think the `impl` rules allow this:

```rust
// Usual "do not use this outside of Rocket" warning comment
#[doc(hidden)]
trait GetStateTypeId {
    const fn get_typeid() -> Option<TypeId>;
}

impl<T> GetStateTypeId for T {
    const fn get_typeid() -> Option<TypeId> { None }
}

impl<T> GetStateTypeId for State<T> {
    const fn get_typeid() -> Option<TypeId> { Some(type_id) }
}
```

Then emit code like `state_types: &[#( <#ty as #GetStateTypeId>::get_typeid() )]`. This is a bit inefficient to work with because of `None`, but those checks should only happen once at launch.

msehnout: Unfortunately that would require [specialization (RFC 1210)](https://github.com/rust-lang/rfcs/blob/master/text/1210-impl-specialization.md), which was marked as not needed any more in Rocket (https://github.com/SergioBenitez/Rocket/issues/19):
` Update (Sep 9, 2019): specialization is no longer used.`
and I'd love to keep it that way because I don't know if there is a timeline for this RFC but it seems pretty convoluted.

I found this article which describes a code that achieves specialization by using autoref in a specific way: https://github.com/dtolnay/case-studies/blob/master/autoref-specialization/README.md
I wrote this code that implements it for `State`, but I haven't yet had time to think about integration with this PR:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4b2bceca8dbe0e744aee5af4080b3628

jebrosen: That's disappointing; I misremembered and thought that what I posted didn't need specialization. But anyway it still only solves `State` itself and not state used internally in `FromRequest` impls; maybe a better solution exists that solves both.

SergioBenitez:
Closing due to inactivity, but I'd be super excited about this!

Issue #1065, opened 2019-08-02
Tracking Issue for Async I/O migration


jebrosen:
## Status

Rocket's `master` branch is currently **fully asynchronous and supports async/await**!

## Pending

  - [ ] #1067 - Remove or replace typed headers (PR: #1535)
  - [ ] #1329 - Documentation fixes, additions, improvements.
  - [ ] #1070 - Expose APIs for customizing the underlying I/O, instead of being restricted to TCP with optional TLS.
  - [ ] Detect blocking I/O in async: it seems the best we can do for now is to document the issues.
      A "blocking-in-async" lint in `clippy` or `rustc` would be much nicer, if feasible, but would occur upstream...someday.
  - [x] Miscellaneous small TODOs (tagged as `TODO.async`) throughout code.
  - [x] Response bodies are passed back to hyper using a channel.

## Done

  - [x] Async I/O everything!
  - [x] #1325 - Data reading and connection limits.
  - [x] Keep-alive and idle client timeouts
  - [x] Handling of blocking database connections.
  - [x] Gracefully handle panics with 500. (PR #1491)
  - [x] #1066 - Add a Write-like interface for responses, and consideration of `AsyncRead` vs `Stream<Item=Result<Chunk, Error>>`
  - [x] #180, #1374 - fixing graceful shutdown when using "infinite" responders
  - [x] #33 - Server-Sent Events
  - [x] #1117 - Asynchronous database pools, or at the very least forward compatibility with them.

## Design

The `async` APIs are based on rust's async/await feature and `std::future::Future`, running on the `tokio` (0.2) runtime.

* High-level APIs (routes and catchers):
  * Routes can now be defined as `#[get|post|route(...)] async fn(params...) -> R` where `R` is any `Responder` type. Routes that do not `await` anything can remain unchanged. `#[catch]`ers can also be defined as an `async fn`.
  * Methods on `Data` now return `Future`s that must be `.await`ed. As a consequence, any route that uses `Data` directly must be `async`. Routes that use data guards such as `Json` are not affected by this change.
  * Routes that make their own network requests or other I/O should be `async`, as it will allow other requests to continue while waiting for the operation.
* Mid and low-level APIs (`FromTransformedData` and `FromData`, `Fairing`, `Responder`)
  * Several traits are now defined using `async-trait`, re-exported as `#[rocket::async_trait]`.
[...]

PR #1184, opened 2019-12-10
Update to Tokio 0.2 Hyper 0.13

paolobarbolini:
WIP on updating to Tokio 0.2, Hyper 0.13

Currently the `ctrl_c_shutdown` feature makes my compiler crash, and I have to figure out how the new Hyper runtime api works (currently the docs.rs build of hyper fails so I'll have to dig in to the code).
Commenting out the line with `.executor(tokio::executor::DefaultExecutor::current())` in `core/lib/src/rocket.rs` makes the code build but it won't run.

### DIFF
--- a/contrib/codegen/src/database.rs
+++ b/contrib/codegen/src/database.rs
@@ -73,7 +73,7 @@ pub fn database_attr(attr: TokenStream, input: TokenStream) -> Result<TokenStrea
     let databases = quote_spanned!(span => ::rocket_contrib::databases);
     let Poolable = quote_spanned!(span => #databases::Poolable);
     let r2d2 = quote_spanned!(span => #databases::r2d2);
-    let tokio_executor = quote_spanned!(span => #databases::tokio_executor);
+    let tokio = quote_spanned!(span => #databases::tokio);
     let request = quote!(::rocket::request);
 
     let generated_types = quote_spanned! { span =>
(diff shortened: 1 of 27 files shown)


jebrosen:
I already nearly finished this yesterday, along with some other important changes (https://github.com/SergioBenitez/Rocket/issues/1065#issuecomment-562895589). My guess is a lot of it's similar, but I'll make sure I didn't miss anything here.

jhpratt:
I can take a look at the shutdown once other things are resolved.

jebrosen:
Once I get back to my computer, - I will compare this to my code (turns out I did push my WIP at https://github.com/jebrosen/Rocket/commit/58d8dcb1c20e65ede23ec7ad9162d7af61547f51) and merge and push some combination of the two.

We should change the shutdown approach a bit (i.e. not solely a feature flag), but let's keep that separate.

paolobarbolini:
@jhpratt After spending some time with it I finally figured it out. I couldn't understand why [futures_util::select!](https://docs.rs/futures-util/0.3.1/futures_util/macro.select.html) wouldn't work so I used [futures_util::future::select](https://docs.rs/futures-util/0.3.1/futures_util/future/fn.select.html) instead, which by the way was already being used previously.

All tests are now passing.

jebrosen:
This commit (a merger of this version and my version), the promised tweaks to trait lifetimes, and a rebase on `master` are now in the `async` branch.

PR #1186, opened 2019-12-15
Fix missing decl_macro in README.md

navarroaxel:


jebrosen:
The requirement for `decl_macro` was removed in #964, but it's not in a release. A better fix for this might be to have the main page show the `v0.4` branch by default instead of `master`, but I'm not sure how well that works.

PR #1190, opened 2019-12-20
Add YAML (application/x-yaml) as a known media type.

dkellner:


jebrosen:
What is the precedent for `application/x-yaml`? YAML isn't in the media types registry at all, and I would not want to introduce a media type without either that or widespread recognition.

dkellner:
You're right. In retrospect I opened this PR prematurely without suitable reasoning for an unobvious change, sorry!

Some arguments for `application/x-yaml`:
- Ruby on Rails uses it ([source](https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/http/mime_types.rb))
- SwaggerUI uses `application/yaml`, but I think the prefix `x-` is suitable as the official IANA registry does not contain YAML (yet).

I will research further after the holidays and at least update the commit message.

dkellner:
I've been looking at this from two different angles: First, what would be the right media type to use, and second, how others are handling this.

As you've said already, there is no "right" media type to use, as it is not in the official media types registry. Furthermore, RFC 4288 (https://tools.ietf.org/html/rfc4288#section-3.4) states that the usage of "x-" prefix is discouraged.

Both Nginx and Apache don't define a media type for YAML: https://github.com/nginx/nginx/blob/master/conf/mime.types, https://svn.apache.org/viewvc/httpd/httpd/branches/2.4.x/docs/conf/mime.types?view=markup. Others (including Ruby on Rails and SwaggerUI mentioned above) don't agree on a single one, but use one of `application/x-yaml`, `application/yaml`, `text/yaml` and `text/x-yaml`.

I recommend to not merge this, so I'm closing this PR.

jebrosen:
That's unfortunate; hopefully one will be converged on in the future. Thank you nonetheless!

PR #1191, opened 2019-12-23
Add Responder impl support for Either.

zlnpgtdtfezf:
Add optional support for Either. 

Some operation during their normal function want to return multiple different return types. E.g.: HTTP PUT can have several different result from 200, 204 or sometimes: 403, 404. 

Currently methods can return multiple types using Result. If the returned value is Result::Err(_) it gets logged. Also because of this the error type must implement Debug. Because of all these using Result is not optimal when you have multiple return types all of which are normal behaving results. 

Currently users can opt to implement their own Responder types but that is a little bit more boilerplate code. Using Either this can be avoided. 

Pls merge.

### DIFF
--- a/core/lib/Cargo.toml
+++ b/core/lib/Cargo.toml
@@ -19,7 +19,7 @@ edition = "2018"
 all-features = true
 
 [features]
-default = ["private-cookies"]
+default = ["private-cookies", "either"]
 tls = ["rocket_http/tls"]
 private-cookies = ["rocket_http/private-cookies"]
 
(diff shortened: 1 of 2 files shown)


jebrosen:
Closing as a duplicate of #1123.

> If the returned value is Result::Err(_) it gets logged.

This is only the case when the error type does *not* implement `Responder`. When it does implement `Responder`, `Result` already works the same as this implementation for `Either` and uses whichever `Responder` it is. In `master`, the support for and requirement for the error type to implement `Debug` has been moved to a new `Responder` type of `rocket::response::Debug`, making `Result` entirely equivalent to `Either`.

PR #1192, opened 2019-12-23
Add some more response wrappers for common HTTP status codes.

zlnpgtdtfezf:
For the most common response codes add response wrappers.

### DIFF
--- a/core/lib/src/response/status.rs
+++ b/core/lib/src/response/status.rs
@@ -213,47 +213,26 @@ impl<'r, R: Responder<'r>> Responder<'r> for Accepted<R> {
 
 /// Sets the status of the response to 204 (No Content).
 ///
-/// If a responder is supplied, the remainder of the response is delegated to
-/// it. The body will be empty even if the responder sets a non-empty body.
+/// The response body will be empty.
 ///
+/// # Example
+///
+/// A 204 No Content response:
+///
+/// ```rust
+/// use rocket::response::status;
+///
+/// # #[allow(unused_variables)]
+/// let response = status::NoContent;
+/// ```
 #[derive(Debug, Clone, PartialEq)]
-pub struct NoContent<R>(pub Option<R>);
+pub struct NoContent;
 
-/// Sets the status code of the response to 204 No Content. If the responder is
-/// `Some`, it is used to finalize the response.
-/// The body will be empty even if the responder sets a non-empty body.
-impl<'r, R: Responder<'r>> Responder<'r> for NoContent<R> {
-    fn respond_to(self, req: &Request<'_>) -> Result<Response<'r>, Status> {
+/// Sets the status code of the response to 204 No Content.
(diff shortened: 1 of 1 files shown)


core/lib/src/response/status.rs
```diff
@@ -211,6 +211,52 @@ impl<'r, R: Responder<'r>> Responder<'r> for Accepted<R> {
     }
 }
 
+/// Sets the status of the response to 204 (No Content).
+///
+/// If a responder is supplied, the remainder of the response is delegated to
+/// it. The body will be empty even if the responder sets a non-empty body.
+///
+#[derive(Debug, Clone, PartialEq)]
+pub struct NoContent<R>(pub Option<R>);
```
jebrosen: Let's simplify this to `struct NoContent;` to avoid allowing a body that won't even be used.

core/lib/src/response/status.rs
```diff
@@ -275,6 +363,70 @@ impl<'r, R: Responder<'r>> Responder<'r> for NotFound<R> {
...
+    }
+}
+
+/// Sets the status of the response to 451 (Unavailable For Legal Reasons).
+///
+/// If a responder is supplied, the remainder of the response is delegated to
+/// it. If there is no responder, the body of the response will be empty.
+///
+#[derive(Debug, Clone, PartialEq)]
+pub struct UnavailableForLegalReasons<R>(pub Option<R>);
```
jebrosen: `UnavailableForLegalReasons(R)` is such a specific error code I don't expect this one to be used much; my guess is that this status code is more likely to appear in in custom-implemented or derived `Responder`s. Of course `rocket::response::status::Custom(Status::UnavailableForLegalReasons, error)` is always an option too.

core/lib/src/response/status.rs
zlnpgtdtfezf: I think you are right. My logic was that Responders might be used to set some http response headers, however I don't think this would be so common. Since these wrappers are there for as just simple tools, it's probably not a big deal to say that anyone who wants something like setting an additional header for `No Content` might write their own responders or wrappers to do so anyway.

zlnpgtdtfezf: (The same is true for ResetContent's responder)

zlnpgtdtfezf:
I will update the PR with examples, remove the optional responder of `NoContent` and remove `ResetContent`, `UnavailableForLegalReasons` and `Gone` in a few days.

core/lib/src/response/status.rs
jebrosen: They could also still wrap in the opposite direction if they need that, assuming the Responder that adds the headers really only adds headers: `MyHeaders<NoContent>`.

zlnpgtdtfezf:
@jebrosen I have completed the requested modifications.

jebrosen:
Thanks! Merged this as 859224a.

PR #1196, opened 2020-01-07
Migrate over to AsyncSeek

notriddle:
This is a fairly complicated change, because the rest of the framework
wants to know the size of the sized body right away, but we need to
fetch it asynchronously. I'm not sure if this is an elegant-enough
solution.

This commit changes the `Response::set_sized_body` in the obvious way,
but to preserve the builder pattern, `ResponseBuilder` is changed so
that the act of finalizing it is asynchronous, rather than attempting to
make the `sized_body()` function asynchronous.

### DIFF
--- a/core/lib/src/response/response.rs
+++ b/core/lib/src/response/response.rs
@@ -105,16 +105,9 @@ impl<T> fmt::Debug for Body<T> {
     }
 }
 
-// error[E0225]: only auto traits can be used as additional traits in a trait object
-//   --> core/lib/src/response/response.rs:173:52
-//    |
-//186 |     pending_sized_body: Option<Box<dyn AsyncRead + AsyncSeek + Send + 'r>>,
-//    |                                        ---------   ^^^^^^^^^
-//    |                                        |           |
-//    |                                        |           additional non-auto trait
-//    |                                        |           trait alias used in trait object type (additional use)
-//    |                                        first non-auto trait
-//    |                                        trait alias used in trait object type (first use)
+/// Internal workaround for `Box<dyn AsyncRead + AsyncSeek>` not being allowed.
+/// 
+/// https://github.com/rust-lang/rfcs/issues/2035
 trait AsyncReadAsyncSeek: AsyncRead + AsyncSeek + Unpin + Send {}
 impl<T: AsyncRead + AsyncSeek + Unpin + Send> AsyncReadAsyncSeek for T {}
 
--- a/core/lib/src/response/response.rs
+++ b/core/lib/src/response/response.rs
@@ -611,22 +611,17 @@ impl<'r> Future for ResponseBuilder<'r> {
     type Output = Response<'r>;
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         let this = self.get_mut();
-        let mut fut = mem::replace(&mut this.fut, None);
-        let poll = if let Some(ref mut fut) = fut {
-            fut.as_mut().poll(cx)
-        } else {
+        if this.fut.is_none() {
             let mut response = mem::replace(&mut this.response, Response::new());
             let pending_sized_body = mem::replace(&mut this.pending_sized_body, None);
-            let mut fut = Box::pin(async {
+            this.fut = Some(Box::pin(async {
                 if let Some(sb) = pending_sized_body {
                     response.set_sized_body(sb).await;
                 }
                 response
-            });
-            fut.as_mut().poll(cx)
-        };
-        mem::replace(&mut this.fut, fut);
-        poll
+            }));
+        }
+        this.fut.as_mut().expect("this.fut.is_none() checked and assigned Some").as_mut().poll(cx)
     }
 }
 
--- a/core/lib/src/response/response.rs
+++ b/core/lib/src/response/response.rs
@@ -2,8 +2,9 @@ use std::{io, fmt, str};
 use std::borrow::Cow;
 use std::future::Future;
 use std::pin::Pin;
+use std::task::{Context, Poll};
 
-use tokio::io::{AsyncRead, AsyncReadExt};
+use tokio::io::{AsyncRead, AsyncReadExt, AsyncSeek, AsyncSeekExt};
 use futures_util::future::FutureExt;
 
 use crate::response::{Responder, ResultFuture};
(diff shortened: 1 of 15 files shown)


jebrosen:
I've been working on https://github.com/tokio-rs/tokio/pull/1986 before tackling this, since the `seek()` adapter is currently broken - this causes all tests to time out.

> rather than attempting to make the `sized_body()` function asynchronous.

Why don't you like that approach? I think it would be a less invasive change to do that, but I haven't looked in a while. Making `ResponseBuilder` a `Future` does look a bit awkward.

notriddle:
> Why don't you like that approach?

Both approaches are awkward. The other approach winds up with `ResponseBuilder` getting one, and only one, `async` function. You have to memorize which one it is, it stands out like a sore thumb since it's kind of contrary to the way builders normally work (builders normally don't *DO* anything until you consume them), and if we add more than one, the natural code is something like this:

```rust
Response::build()
    .sized_body(file).await
    .event_connection(conn).await // <- pointless serialization
    .finalize()
```

It's a minor increase in the complexity of Rocket's code, but it means the default behaviour of the library consumer's code is better.

```rust
Response::build()
    .sized_body(file)
    .event_connection(conn)
    .await // <- event_connection and sized_body are set up CONCURRENTLY!!!111
```

jebrosen:
Sounds good. I'm going to try and get my `seek()` fix merged tonight so we can start working through these changes - I have a few vague ideas for simplifying this that I want to try.

jebrosen:
Ok, tokio 0.2.9 is out. If you update to that in `core/lib/Cargo.toml` tests should all be green - or show relevant failures instead of timing out.

I'll do a full review of the code changes in a day or so - the overall structure looks good, but I think we can probably reduce the amount of churn.

core/lib/src/response/response.rs
```diff
@@ -522,46 +573,57 @@ impl<'r> ResponseBuilder<'r> {
...
-    pub fn ok<T>(&mut self) -> Result<Response<'r>, T> {
-        Ok(self.finalize())
+    pub async fn ok<E>(&mut self) -> Result<Response<'r>, E> {
+        Ok(self.await)
+    }
+}
+
+impl<'r> Future for ResponseBuilder<'r> {
+    type Output = Response<'r>;
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
```
jebrosen: I think we should keep `finalize()` but convert it to an `async fn`, primarily for future extensibility and readability.

core/lib/src/response/response.rs
```diff
@@ -1018,7 +1080,8 @@ impl<'r> Response<'r> {
 
     /// Sets the body of `self` to be the fixed-sized `body`. The size of the
     /// body is obtained by `seek`ing to the end and then `seek`ing back to the
-    /// start.
+    /// start. Since this is an asynchronous operation, it returns a promise
```
jebrosen: promise -> `Future`. We can probably leave out the part about "holds onto this Response" as well, since that's the case for all `self`-taking `async fn`.

core/lib/src/response/response.rs
```diff
@@ -354,15 +383,21 @@ impl<'r> ResponseBuilder<'r> {
...
     ///     .sized_body(File::open("body.txt").await?)
-    ///     .finalize();
+    ///     .await;
     /// # Ok(())
     /// # }
     /// ```
     #[inline(always)]
-    pub fn sized_body<B>(&mut self, body: B) -> &mut ResponseBuilder<'r>
-        where B: AsyncRead + io::Seek + Send + Unpin + 'r
+    pub fn sized_body<B: 'r>(&mut self, mut body: B) -> &mut ResponseBuilder<'r>
```
jebrosen: These `'r` and `mut` are extraneous, right? (maybe the `async` should be `async move` if they aren't).

core/lib/src/response/response.rs
```diff
@@ -354,15 +383,21 @@ impl<'r> ResponseBuilder<'r> {
...
+        where B: AsyncRead + AsyncSeek + Send + Unpin + 'r
     {
-        self.response.set_sized_body(body);
+        self.sized_body = Some(Box::pin(async {
```
jebrosen: This duplicates the logic in `Response`. Let's try changing this to `self.pending_sized_body` as a `Box<dyn AsyncRead + AsyncSeek + Send + Unpin + 'r>`, and call `response.set_sized_body` inside `async fn finalize`. If necessary we can split out another method or two to reduce these extraneous `Box`es.

core/lib/src/response/response.rs
```diff
@@ -522,46 +573,57 @@ impl<'r> ResponseBuilder<'r> {
...
+    ///     .ok().await;
     ///
     /// assert!(response.is_ok());
+    /// 
+    /// # })
     /// ```
     #[inline(always)]
-    pub fn ok<T>(&mut self) -> Result<Response<'r>, T> {
-        Ok(self.finalize())
+    pub async fn ok<E>(&mut self) -> Result<Response<'r>, E> {
```
jebrosen: I'm tempted but not quite convinced to make `ok()` return a `Pin<Box<dyn Future<Output=Result<Response<'r>, E>>>>` instead. That would allow the common case of `Box::pin(async move { /* build */.ok().await })` to become just `/* build */.ok()` again.

core/lib/src/response/response.rs
notriddle: The idea behind making ResponseBuilder a future itself came from <https://blog.yoshuawuyts.com/async-finalizers/>, based on the idea that `.finalize().await` is kind of redundant. If you really want to keep it a separate function, I'll do it, but I'm not sure how it aids in extensibility (you can already see that other modes like `ok()` still work).

jebrosen: I would be happy to tackle this part last. My main hesitation is that adding functionality to the manual implementation will be harder than adding it to an `async fn`, and implementing `Future` directly on `ResponseBuilder` is a strong commitment in comparison.

notriddle: My new implementation of it is actually based on an async block anyway, so it shouldn't be a problem.

jebrosen:
For posterity:
* changing `ok()` in the way I wanted is difficult because the builder uses `&mut self` everywhere instead of moving `self`.
* Avoiding the double boxing is hard: either `pending_sized_body` is `Box<AsyncReadAsyncSeek>`, or a generic type `PB: AsyncRead + AsyncSeek`. The first option offers no easy way to get just a `Box<AsyncRead>` back out (as required internally). The second option is a really weird API wart.
  * My original idea to make `sized_body` an `async fn` avoids this issue, but is inconsistent as discussed earlier.

jebrosen:
Thanks for working on this! Merged with some style/format changes as 9cb5af5.

Issue #1189, opened 2019-12-20
request to adding database support for mongodb 0.9.0


phantomxe:
Mongodb officially released the rust driver.

TheBestJohn:
Deadpool will be able to do this for you I believe. Take a look at this post: https://www.reddit.com/r/rust/comments/e9n6mx/creating_and_using_a_postgres_database_pool_is/

Deadpool basically sets up an async pool for any connection to data. This example uses Postgres but the setup looks almost identical to MongoDB.

jebrosen:
This can be updated once `r2d2-mongodb` is updated, but it won't land until 0.5.

mongodb is synchronous, so r2d2 is probably a better choice than deadpool if you need to use it now. This can be done with a manually managed pool, as described in https://rocket.rs/v0.3/guide/state/#databases.

Figments:
The official driver includes its own database pooling support. Is R2D2 still required to make official support possible?

jebrosen:
I did not see that the new client has pool support built-in; thanks for mentioning that. The current database support (`#[database]`) is built on r2d2 and I don't think it would be very easy to change - definitely not in 0.4.

Since one of the main benefits of `#[database]` is the connection pooling, it probably makes more sense (at least for now) to put a mongodb `Client` in managed state yourself, optionally with a request guard (i.e. [the 0.3 database example](https://rocket.rs/v0.3/guide/state/#databases)).

Figments:
It might, then, be beneficial to include MongoDB-specific documentation in the interim in case moving away from R2D2 in the far future is possible. That way anyone that wants to use Rocket with the officially-supported driver doesn't have to dig all that far to find one (as documentation for manually-managed database state is currently buried in an older version of the documentation). 

I can try to write something after I figure out how to implement it, if that's alright.

jebrosen:
I do think we should inline and/or link to the 0.3 database documentation, for the case of an unsupported database engine or different version (see also #1179).

MongoDB is a bit different now by having its own connection pool: it can be used just as any other managed state via `State<'_, Client>`. Creating a custom connection guard is optional and beneficial mostly in the case of connecting to multiple mongodb servers.

jebrosen:
[...]

PR #1201, opened 2020-01-14
Upgraded dependencies

### LINKED ISSUE #1195: Update to tera 1.0
Current Tera version (0.11) in contrib is nearly 1 year old and there are missing features, which are present in Tera documentation. 
Because of this, it's difficult to know what is available and not.

As an example, I'm getting this error:    ` => Error: Filter 'float' not found`
The filter `float` is present in standard Tera doc, but not in the (oldish) version Rocket uses.

Is there a plan to bump Tera version to current one?

Razican:
Doing the commit directly to the `async` branch, since it seems the most up to date one :)

I upgraded most of the dependencies. This also fixes #1195. This does not upgrade mongodb, for the reasons explained in #1189. I didn't upgrade `time`, because it requires a fundamental change on how `cookie` works, and I didn't upgrade `rusqlite` and its `r2d2` crate because `diesel` does not support it yet.

I didn't upgrade `percent-encoding` because `Rocket` depends on a trait that no longer exists in the new version, and I didn't have the time to fully understand our use and do a proper upgrade.

`compiletest_rs` drops the `stable` feature (it still exists but does nothing), so I removed it.

I also ran `rustfmt` on the files I modified.

### DIFF
--- a/core/codegen/src/lib.rs
+++ b/core/codegen/src/lib.rs
@@ -1,8 +1,10 @@
 #![feature(proc_macro_diagnostic, proc_macro_span)]
-#![recursion_limit = "128"]
+#![recursion_limit="128"]
+
 #![doc(html_root_url = "https://api.rocket.rs/v0.5")]
 #![doc(html_favicon_url = "https://rocket.rs/v0.5/images/favicon.ico")]
 #![doc(html_logo_url = "https://rocket.rs/v0.5/images/logo-boxed.png")]
+
 #![warn(rust_2018_idioms)]
 
 //! # Rocket - Code Generation
@@ -56,8 +58,8 @@
 //! ROCKET_CODEGEN_DEBUG=1 cargo build
 //! ```
 
-#[macro_use]
-extern crate quote;
+#[macro_use] extern crate quote;
+extern crate proc_macro;
 
 use rocket_http as http;
 
@@ -105,15 +107,15 @@ macro_rules! define_vars_and_mods {
 
 #[macro_use]
 mod proc_macro_ext;
+mod derive;
 mod attribute;
 mod bang;
-mod derive;
 mod http_codegen;
 mod syn_ext;
 
 use crate::http::Method;
-use devise::proc_macro2;
 use proc_macro::TokenStream;
+use devise::proc_macro2;
 
 static ROUTE_STRUCT_PREFIX: &str = "static_rocket_route_info_for_";
 static CATCH_STRUCT_PREFIX: &str = "static_rocket_catch_info_for_";
@@ -123,7 +125,7 @@ static URI_MACRO_PREFIX: &str = "rocket_uri_macro_";
 static ROCKET_PARAM_PREFIX: &str = "__rocket_param_";
 
 macro_rules! emit {
-    ($tokens:expr) => {{
+    ($tokens:expr) => ({
         let tokens = $tokens;
         if std::env::var_os("ROCKET_CODEGEN_DEBUG").is_some() {
             proc_macro::Span::call_site()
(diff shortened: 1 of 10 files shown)


jebrosen:
> I also ran rustfmt on the files I modified.

Can this be undone? It often makes reviews and syncing back with `master` significantly harder.

Most of these should be straightforward, but a few specific ones like `tera` and `handlebars` might need additional work in non-Rust files. Hopefully the tests will catch those if that's the case.

core/codegen/src/lib.rs
```diff
@@ -58,8 +56,8 @@
 //! ROCKET_CODEGEN_DEBUG=1 cargo build
 //! ```
 
-#[macro_use] extern crate quote;
-extern crate proc_macro;
```
jebrosen: This particular line also bumps the minimum required cargo version.

Razican:
> > I also ran rustfmt on the files I modified.
> 
> Can this be undone? It often makes reviews and syncing back with `master` significantly harder.

I will try. I didn't do many changes to the code, so it shouldn't be too difficult.

Maybe, to avoid the automated formatting of tools like VScode it might be worth to add a `#[rustfmt::skip]` at the crate level.

Still, it would make sense to pass the whole project through `cargo fmt` (with a [custom][custom_fmt] `rustfmt.toml`, if needed), so that the code style is homogeneous in the project.

> Most of these should be straightforward, but a few specific ones like `tera` and `handlebars` might need additional work in non-Rust files. Hopefully the tests will catch those if that's the case.

`handlebars` was not updated. The `tera` template required a small change in one comment, and a bit of code when rendering, since it will now by default use a custom `Context` for context, instead of a `Serialize` struct. It did provide a simple method to render from a serializable structure, so I think it will be fine. All tests passed in my local machine, let's see in the CI. It seems that Windows tests were failing, but that the errors seemed unrelated to the PR.

[custom_fmt]: https://rust-lang.github.io/rustfmt/

core/codegen/src/lib.rs
Razican: This was suggested by `clippy`, I can revert this part, but Rocket requires Rust nightly, and I would say that most of the time Rust is installed with the same cargo version.

jebrosen: I'm okay bumping the minimum version, but it has to be accompanied by a change to the corresponding `build.rs` file(s). I can do that when I merge, too.

Razican: I have reverted it for now. I think that this PR should be about the dependencies. Maybe it would make sense to upgrade the syntax to 2018 edition, but that would include removing all `extern crate` references, and importing macros where they are used.

Razican:
I reverted the formatting changes.

core/codegen/src/lib.rs
jebrosen: Everything is already upgraded to the 2018 edition. `proc_macro` was a special case that still needed `extern crate` until a change in `cargo` around mid-December.

jebrosen:
Looks good overall. I'll do the more thorough checking this week.

Razican:
Good, thanks! Something that is holding back many duplicate crates in the dependency tree is the `pear` dependency. I opened [a PR][pear-pr] in the repo to upgrade its dependencies, but ideally it should be published in crates.io.

[pear-pr]: https://github.com/SergioBenitez/Pear/pull/21

jebrosen:
Progress update: I've been working on updating compiletest only and fixing the disabled/outdated UI tests at the same time; after that I'll go through the rest.

Somewhere in there I will see if I can figure out why Windows CI is busted.

jebrosen:
Thank you! Merged as f4bb8bb (in `master`). It will show up in the `async` branch in the next merge or rebase.

PR #1204, opened 2020-01-15
Extended example of parameters.

### LINKED ISSUE #1203: How to support hyphen in urls?
(I'd ask this on riot/matrix but couldn't find the room for some reason)

If we already have a query parameter unkind-rust-name=foo, is there a way to encode this as a route? 

```rust
#[get(/bar?unkind-rust-name=<kind_rust_name>)]
fn bar(kind_rust_name: String) {}
```

Many thanks in advance!

gilescope:
Shows how to handle parameter renaming for the next person that is looking for it (Closes #1203 )

### DIFF
--- a/examples/query_params/src/main.rs
+++ b/examples/query_params/src/main.rs
@@ -8,6 +8,8 @@ use rocket::request::{Form, LenientForm};
 
 #[derive(FromForm)]
 struct Person {
+    /// Use the `form` attribute to expect an invalid Rust identifier in the HTTP form.
+    #[form(field = "first-name")]
     name: String,
     age: Option<u8>
 }
--- a/examples/query_params/src/tests.rs
+++ b/examples/query_params/src/tests.rs
@@ -11,12 +11,12 @@ macro_rules! run_test {
 
 #[test]
 fn age_and_name_params() {
-    run_test!("?age=10&name=john", |mut response: Response<'_>| {
+    run_test!("?age=10&first-name=john", |mut response: Response<'_>| {
         assert_eq!(response.body_string(),
             Some("Hello, 10 year old named john!".into()));
     });
 
-    run_test!("?age=20&name=john", |mut response: Response<'_>| {
+    run_test!("?age=20&first-name=john", |mut response: Response<'_>| {
         assert_eq!(response.body_string(),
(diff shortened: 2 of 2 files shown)


jebrosen:
Tests are failing because the expected name was renamed but the tests weren't changed to match.

gilescope:
Ah sorry didn't spot the tests. Will fixup.

gilescope:
Great backtrace doesn't compile on an azure windows box. Not sure what I can do about that?

jebrosen:
The Windows CI failure is caused by `mysql 17` -> `mysql_common 0.19.2` -> `lexical 4` -> `cfg-if = 0.1.9`. This breaks `backtrace 0.3.42`, which should say it depends on `cfg-if ^0.1.10` but did not. Since the version correction appears to be the only thing changed in `backtrace 0.3.43`, I will ask if yanking `0.3.42` would be possible.

README.md
```diff
@@ -158,3 +158,4 @@ Rocket is licensed under either of the following, at your option:
  * MIT License ([LICENSE-MIT](LICENSE-MIT) or http://opensource.org/licenses/MIT)
 
 The Rocket website source is licensed under [separate terms](site#license).
+
```
SergioBenitez: This shouldn't be here.

examples/query_params/src/main.rs
```diff
@@ -8,6 +8,7 @@ use rocket::request::{Form, LenientForm};
 
 #[derive(FromForm)]
 struct Person {
+    #[form( field = "first-name" )]  // renaming parameter to valid rust identifier
```
SergioBenitez:
```suggestion
    /// Use the `form` attribute to expect an invalid Rust identifier in the HTTP form.
    #[form(field = "first-name")]
```

gilescope:
@SergioBenitez all good?

gilescope:
@SergioBenitez I can't see what changes are outstanding? What's left to merge this?

gilescope:
Great, @SergioBenitez can you merge in please? I don't have permissions to do that bit?

SergioBenitez:
Merged in ff4b63c0be85c602feb4d8426a5c777db7ad6d87. Thank you!

Issue #1210, opened 2020-01-21
Docs: Consider running `cargo fmt` on `examples/*`


benjaminch:
First of, thanks a lot for this very nice project :)

## Documentation?
I have been playing quite a bit with rocket and have enjoyed all the very detailed examples listed [here](https://github.com/SergioBenitez/Rocket/tree/master/examples).
It appears though that they are not all following rustfmt convention.
I though it will be nice to apply formatting on those.

Thanks a lot !

SergioBenitez:
Please see #1211. Thank you!

PR #1211, opened 2020-01-21
Apply fmt on examples/*

benjaminch:
CL addressing issue: #1210

### DIFF
--- a/examples/config/tests/common/mod.rs
+++ b/examples/config/tests/common/mod.rs
@@ -1,8 +1,8 @@
-use rocket::{self, State};
-use rocket::fairing::AdHoc;
 use rocket::config::{self, Config, Environment, LoggingLevel};
+use rocket::fairing::AdHoc;
 use rocket::http::Status;
 use rocket::local::Client;
+use rocket::{self, State};
 
 struct LocalConfig(Config);
 
@@ -10,7 +10,7 @@ struct LocalConfig(Config);
 fn check_config(config: State<'_, LocalConfig>) -> Option<()> {
     let environment = match std::env::var("ROCKET_ENV") {
         Ok(name) => name,
-        Err(_) => return None
+        Err(_) => return None,
     };
 
     let config = &config.0;
--- a/examples/config/tests/development.rs
+++ b/examples/config/tests/development.rs
@@ -1,6 +1,7 @@
 #![feature(proc_macro_hygiene)]
 
-#[macro_use] extern crate rocket;
+#[macro_use]
+extern crate rocket;
 
 mod common;
 
--- a/examples/content_types/src/main.rs
+++ b/examples/content_types/src/main.rs
@@ -1,14 +1,20 @@
 #![feature(proc_macro_hygiene)]
 
-#[macro_use] extern crate rocket;
-#[macro_use] extern crate serde_derive;
+#[macro_use]
+extern crate rocket;
+#[macro_use]
+extern crate serde_derive;
 
-#[cfg(test)] mod tests;
+#[cfg(test)]
+mod tests;
 
 use std::io::{self, Read};
 
-use rocket::{Request, data::Data};
-use rocket::response::{Debug, content::{Json, Html}};
+use rocket::response::{
+    content::{Html, Json},
+    Debug,
+};
+use rocket::{data::Data, Request};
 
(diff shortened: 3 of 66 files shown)


jebrosen:
Last time we tried this (https://github.com/SergioBenitez/Rocket/pull/931#issuecomment-491468449) we were still not too happy with the results.

However it may be okay to do it for only examples, which should run into those problems less often. I'll take a look at what kind of changes are involved here.

benjaminch:
Thanks @jebrosen ! Let see how it goes then. Happy if I can help anyways.

examples/config/tests/development.rs
```diff
@@ -1,6 +1,7 @@
 #![feature(proc_macro_hygiene)]
 
-#[macro_use] extern crate rocket;
+#[macro_use]
+extern crate rocket;
```
jebrosen: Is there any configuration that would keep these together on one line? Same with some `#[cfg]` lines.

examples/content_types/src/main.rs
```diff
@@ -27,7 +33,10 @@ struct Person {
 #[get("/<name>/<age>", format = "json")]
 fn get_hello(name: String, age: u8) -> Json<String> {
     // NOTE: In a real application, we'd use `rocket_contrib::json::Json`.
-    let person = Person { name: name, age: age, };
+    let person = Person {
+        name: name,
+        age: age,
+    };
```
jebrosen: These lines are far too short for `rustfmt` to be trying to split them!

examples/config/tests/common/mod.rs
```diff
@@ -1,16 +1,16 @@
...
+use rocket::{self, State};
 
 struct LocalConfig(Config);
 
 #[get("/check_config")]
 fn check_config(config: State<'_, LocalConfig>) -> Option<()> {
     let environment = match std::env::var("ROCKET_ENV") {
         Ok(name) => name,
-        Err(_) => return None
+        Err(_) => return None,
```
jebrosen: :+1:

examples/cookies/src/main.rs
```diff
@@ -1,15 +1,16 @@
...
-#[macro_use] extern crate rocket;
+#[macro_use]
+extern crate rocket;
 
 #[cfg(test)]
 mod tests;
 
 use std::collections::HashMap;
 
+use rocket::http::{Cookie, Cookies};
```
jebrosen: `use` sorting - :+1:

examples/config/tests/development.rs
benjaminch: I guess so, let me dig into this one

benjaminch: Looks like `inline_attribute_width` should do the trick
(Reference)[https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#inline_attribute_width]

examples/content_types/src/main.rs
benjaminch: Well, I agree but it also ease updates reducing merge conflicts down the road. Having those inlined will generate merge conflict in case more than one person is adding / removing fields.

jebrosen: What I am more worried about with these lines is that a future release of rustfmt will want to format this differently (i.e. merge the lines back together because it is short)

benjaminch: Well, in that case, we will be able to change the settings I guess. Also, why do you think it's gonna change soon ?

jebrosen: In this specific case rustfmt has the option `struct_lit_single_line`, which should fit the literal on one line. Since the default for that setting is `true`, this change looks like a bug in rustfmt. I am against enforcing style changes that are due to bugs, especially when the knobs that control those formatting options don't have any stability guarantees.

More generally, rustfmt makes *many* changes that appear to be due to bugs or surprising heuristics for "small" that actually make code (IMO) less readable, and even configuring a bunch of unstable options doesn't lead to correct results.

benjaminch: Fair enough :)
Thanks for explanations !

SergioBenitez:
While I'm certainly in favor of having a consistent style, I am not in favor of using `cargo fmt` on the basis that it is buggy, inconsistent, and simultaneously too opinionated (about some things) and not opinionated enough (about others). Thank you for the PR, however.

Issue #1213, opened 2020-01-24
Cookies from error handlers are ignored


SergioBenitez:
Consider the following application running under `0.4.2`:

```rust
#![feature(proc_macro_hygiene, decl_macro)]

#[macro_use] extern crate rocket;

use rocket::request::Request;
use rocket::http::{Cookie, Cookies};

#[catch(404)]
fn not_found(request: &Request) -> &'static str {
    request.cookies().add(Cookie::new("not_found", "hi"));
    "404 - Not Found"
}

#[get("/")]
fn index(mut cookies: Cookies) -> &'static str {
    cookies.add(Cookie::new("index", "hi"));
    "Hello, world!"
}

fn main() {
    rocket::ignite()
        .register(catchers![not_found])
        .mount("/", routes![index])
        .launch();
}
```

Cookies set by the `index` route will be reflected as normal. Cookies set by `net_found`, however, are not reflected. Rocket should reflect cookies set by error handlers.

PR #1214, opened 2020-01-24
Backport 1213

Ebrithil95:
Backports #1213 to v0.4


core/http/Cargo.toml
```diff
@@ -22,16 +22,16 @@ private-cookies = ["cookie/secure"]
 smallvec = "0.6"
 percent-encoding = "1"
 hyper = { version = "0.10.13", default-features = false }
-time = "0.1"
+time = "0.2.4"
 indexmap = "1.0"
-rustls = { version = "0.14", optional = true }
+rustls = { version = "0.16", optional = true }
```
jebrosen: Changing these versions is a breaking change due to the `ring` dependency. Backporting this to 0.4 would instead require backporting the `reset_delta` to the cookie 0.11 series - @SergioBenitez, is that likely or unlikely to be accepted?

SergioBenitez: Happy to accept a backport to cookie 0.11 and subsequently to 0.4.

Ebrithil95: I created a PR https://github.com/SergioBenitez/cookie-rs/pull/135

Ebrithil95: I have updated cookies to 0.11.2 now

Ebrithil95:
@jebrosen @SergioBenitez Anything else you want me to change?

SergioBenitez:
@Ebrithil95 To backport fixes, you should cherry-pick the relevant commits from the upstream branch, fixing merges along the way as needed to work in the older branch, and finally adding a commit if necessary for non-merge related fixes. We want to preserve as much of the relevant upstream commit as possible, not create brand new commits that are nearly identical.

I've done exactly this and merged it in into the v0.4 branch as 0c56ecc181b66861d2db4fd348e0da17afc60b3e. Thank you so much for initiating this!

Issue #95, opened 2017-01-02
Alllow customizing HTTP headers for `NamedFile`, `StaticFiles`


eliovir:
This has been converted into a tracking issue for "Alllow customizing HTTP headers for NamedFile, StaticFiles". Original text after the fold.

This tracking issue is a merger of the following:

  *  Static files should be sent with Last-Modified header #989  
  *  add support for Last-Modified and If-Modified-Since headers #1376 
  *  RFC for Attachments #791 
  *  Default content-type for StaticFiles #1397 
  *  Cache-Control and other headers for static served files #1522 

In short, `NamedFile`, and by implication `StaticFiles`, should support options to compute and set more response headers. In summary, these are:

  * [ ] [`Last-Modified`]
  * [ ] [`If-Modified-Since`]
  * [ ] [`ETag`]
  * [ ] [`If-Match`]
  * [ ] [`Content-Disposition`]
  * [ ] [`Cache-Control`]
  * [ ] [`Content-Range`], [`If-Range`], [`Range`]
  * [ ] Fallback `Content-Type`

[`Last-Modified`]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Last-Modified
[`If-Modified-Since`]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since
[`ETag`]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag
[`If-Match`]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Match
[`Content-Disposition`]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition
[`Cache-Control`]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control
[`If-Range`]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Range
[`Content-Range`]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range
[`Range`]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range

------------

Hi,

happy new year with successfull Rust projects!

[...]

Issue #1067, opened 2019-08-02
(async) Remove or replace of typed headers


jebrosen:
Typed headers were removed between hyper 0.10 and hyper 0.12. `typed-headers` does not appear to have all the headers hyper 0.10 did, but it looks like `hyperx` might.

Options:
* Re-export headers from one or the other library - this is what we did with hyper 0.10
* Remove typed header support - this is currently implemented in the `async` branch; only the header names are available in hyper and we re-export those.
* Implement `From<X> for Header` for every header in one or both of `hyperx` and `typed-headers`, perhaps behind a feature flag

davidbarsky:
Typed headers exist at [hyperium/headers](https://github.com/hyperium/headers), and they can be used with the latest releases of `hyperium/http` & `hyperium/hyper`—you might not need to write your own bridging.

jebrosen:
Thanks for pointing that out, I missed and/or forgot about it when I checked for existing implementations.

davidbarsky:
It’s not well advertised, and it requires importing an extension trait so that the `http::header::HeaderMap` can use the typed headers.

The derive headers in the repository only supports headers defined within `hyperium/header`. I’d like to make them more extensible/usable outside the crate, but I haven’t had the time, unfortunately.

jebrosen:
I'm inclined towards trying out the headers from the `headers` crate first, re-exporting and/or `From` impls - possibly under a feature flag. In `0.4`, the headers are both re-exported and have implementations for `From<X> for rocket::http::Header<'static>` and I think we should start with the same approach for now to minimize disruption.

jebrosen:
This should be an easy issue to help with, and first it's a bit of a research question - how complete is `headers`, and what kind of integration could be done with Rocket?

I previously opened hyperium/headers#48 because the types I looked at (`Location`, `ETag`) didn't seem to have public constructors. Since that time:

* `ETag` has gained a public constructor via `FromStr`. However it is still of limited utility, because the comparison functions i.e. `EntityTag::weak_eq` are `pub(crate)`.
* I realized that I picked two unlucky examples to check - `ContentRange` and `CacheControl` are two headers that *do* have public constructors and can be useful.
[...]

PR #1219, opened 2020-02-01
Supporting Last-Modified Header for StaticFiles

schrieveslaach:
This commit adds basic support of Last-Modified and If-Modified-Since headers in order to enable HTTP caching while serving static files. #95

### DIFF
--- a/contrib/lib/src/serve.rs
+++ b/contrib/lib/src/serve.rs
@@ -133,6 +133,12 @@ impl Options {
     /// that would be served is a directory.
     pub const NormalizeDirs: Options = Options(0b0100);
 
+    /// `Options` enabling caching based on the `Last-Modified` header. When
+    /// this is enabled, the [`StaticFiles`] handler will at the `Last-Modified`
+    /// header baesd on the modification datetime of the files in the served
+    /// directory.
+    pub const LastModifiedHeader: Options = Options(0b1000);
+
     /// Returns `true` if `self` is a superset of `other`. In other words,
     /// returns `true` if all of the options in `other` are also in `self`.
     ///
@@ -381,11 +387,19 @@ impl Handler for StaticFiles {
                     return Outcome::forward(data);
                 }
 
-                let index = NamedFile::open(p.join("index.html")).await.ok();
+                let index = named_file(p.join("index.html"), &self.options).await;
                 Outcome::from_or_forward(req, data, index)
             },
-            Some(p) => Outcome::from_or_forward(req, data, NamedFile::open(p).await.ok()),
+            Some(p) => Outcome::from_or_forward(req, data, named_file(p, &self.options).await),
             None => Outcome::forward(data),
         }
+    }
+}
+
+async fn named_file<P: AsRef<Path>>(path: P, options: &Options) -> Option<NamedFile> {
+    if options.contains(Options::LastModifiedHeader) {
+        NamedFile::with_last_modified_date(path).await.ok()
+    } else {
+        NamedFile::open(path).await.ok()
     }
 }
--- a/core/lib/Cargo.toml
+++ b/core/lib/Cargo.toml
@@ -45,6 +45,7 @@ indexmap = { version = "1.0", features = ["serde-1"] }
 tempfile = "3"
 async-trait = "0.1.43"
 multer = { version = "2", features = ["tokio-io"] }
+headers = "0.3"
 
 [dependencies.async-stream]
 git = "https://github.com/SergioBenitez/async-stream.git"
--- a/core/lib/src/response/named_file.rs
+++ b/core/lib/src/response/named_file.rs
@@ -1,17 +1,24 @@
 use std::io;
-use std::path::{Path, PathBuf};
 use std::ops::{Deref, DerefMut};
+use std::path::{Path, PathBuf};
+use std::time::SystemTime;
 
+use headers::{Header as HeaderTrait, HeaderValue, IfModifiedSince};
 use tokio::fs::File;
 
+use crate::http::{ContentType, Header, Status};
 use crate::request::Request;
 use crate::response::{self, Responder};
-use crate::http::ContentType;
+use crate::Response;
 
 /// A file with an associated name; responds with the Content-Type based on the
 /// file extension.
 #[derive(Debug)]
-pub struct NamedFile(PathBuf, File);
+pub struct NamedFile {
+    path: PathBuf,
+    file: File,
+    modified: Option<SystemTime>,
+}
 
 impl NamedFile {
     /// Attempts to open a file in read-only mode.
@@ -39,7 +46,33 @@ impl NamedFile {
         // all of those `seek`s to determine the file size. But, what happens if
         // the file gets changed between now and then?
         let file = File::open(path.as_ref()).await?;
-        Ok(NamedFile(path.as_ref().to_path_buf(), file))
+        Ok(NamedFile {
+            path: path.as_ref().to_path_buf(),
+            file,
+            modified: None,
+        })
+    }
+
+    /// Attempts to open a file in the same manner as `NamedFile::open` and
+    /// reads the modification timestamp of the file that will be used to
+    /// respond with the `Last-Modified` header. This enables HTTP caching by
+    /// comparing the modification timestamp with the `If-Modified-Since`
+    /// header when requesting the file.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use rocket::response::NamedFile;
+    ///
+    /// # #[allow(unused_variables)]
+    /// # rocket::async_test(async {
+    /// let file = NamedFile::with_last_modified_date("foo.txt").await;
+    /// # });
+    /// ```
+    pub async fn with_last_modified_date<P: AsRef<Path>>(path: P) -> io::Result<NamedFile> {
+        let mut named_file = NamedFile::open(path).await?;
+        named_file.modified = named_file.metadata().await?.modified().ok();
+        Ok(named_file)
     }
 
     /// Retrieve the underlying `File`.
@@ -57,7 +90,7 @@ impl NamedFile {
     /// ```
     #[inline(always)]
     pub fn file(&self) -> &File {
-        &self.1
+        &self.file
     }
 
     /// Retrieve a mutable borrow to the underlying `File`.
@@ -75,7 +108,7 @@ impl NamedFile {
     /// ```
     #[inline(always)]
     pub fn file_mut(&mut self) -> &mut File {
-        &mut self.1
+        &mut self.file
     }
 
     /// Take the underlying `File`.
@@ -93,7 +126,7 @@ impl NamedFile {
     /// ```
     #[inline(always)]
     pub fn take_file(self) -> File {
-        self.1
+        self.file
     }
 
     /// Retrieve the path of this file.
@@ -111,7 +144,7 @@ impl NamedFile {
     /// ```
     #[inline(always)]
     pub fn path(&self) -> &Path {
-        self.0.as_path()
+        self.path.as_path()
     }
 }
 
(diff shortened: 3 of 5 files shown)


core/lib/src/response/named_file.rs
```diff
@@ -80,13 +105,30 @@ impl NamedFile {
...
+        let mut response = self.file.respond_to(req)?;
+        if let Some(ext) = self.path.extension() {
             if let Some(ct) = ContentType::from_extension(&ext.to_string_lossy()) {
                 response.set_header(ct);
             }
         }
 
+        if let Some(last_modified) = last_modified {
+            response.set_header(Header::new("Last-Modified", last_modified));
+            response.set_header(Header::new("Cache-Control", "public, max-age=31536000"));
```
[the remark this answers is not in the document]
schrieveslaach: I'm not 100% sure if this can be omitted but it seems that it is not necessary. According to RFC7232 the server must send it in 304 when it has been send in the 200 response. 

> The server generating a 304 response MUST generate any of the
> following header fields that would have been sent in a 200 (OK)
> response to the same request: Cache-Control, Content-Location, Date,
> ETag, Expires, and Vary.

I tested it with Firefox and it worked.

schrieveslaach:
@jebrosen, I addressed your comments and update the PR.

core/lib/src/response/named_file.rs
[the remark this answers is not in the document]
jebrosen: I looked at the definitions again, and I think that setting `Cache-Control` at all is too opinionated of a decision to do directly in `NamedFile`. This particular value states the content is not expected to change for a year, after which point the browser should re-request it.

contrib/lib/src/serve.rs
```diff
@@ -14,12 +14,12 @@
 //! features = ["serve"]
 //! ```
 
-use std::path::{PathBuf, Path};
+use std::path::{Path, PathBuf};
```
jebrosen: `rustfmt` strikes again... would you mind reverting the formatting-only changes so it's easier to review what did change?

core/lib/Cargo.toml
```diff
@@ -37,6 +37,7 @@ base64 = "0.11"
 base16 = "0.2"
 pear = "0.1"
 atty = "0.2"
+httpdate = "0.3"
```
jebrosen: `hyper` should already have all the date parsing we need, e.g. in <https://docs.rs/headers/0.3.1/headers/struct.IfModifiedSince.html>

core/lib/src/response/named_file.rs
```diff
@@ -80,13 +96,38 @@ impl NamedFile {
...
+            if let Some(last_modified) = &self.modified {
+                let if_modified_since = match httpdate::parse_http_date(if_modified_since) {
+                    Ok(if_modified_since) => if_modified_since,
+                    Err(_err) => return Response::build().status(Status::BadRequest).ok(),
+                };
+
+                if last_modified
+                    .duration_since(if_modified_since)
+                    .unwrap_or(Duration::from_secs(60))
+                    <= Duration::from_secs(1)
```
jebrosen: Can this be a simpler comparison, like `if last_modified <= if_modified_since`? My guess is it can be, as long as both dates are `HttpDate`s.

core/lib/Cargo.toml
schrieveslaach: Do you know how to parse a `&str` into `IfModifiedSince`? I could not find a way a of creating an instance of `IfModifiedSince` (other headers implement `FromStr`). Additionally, the master branch depends on an old version of hyper so these features might not be available at all.

jebrosen: In hyper 0.10 I believe we would want to go through [`Header::parse_header`](https://docs.rs/hyper/0.10.16/hyper/header/trait.Header.html#tymethod.parse_header), and for hyper 0.13 there is [`Header::decode`](https://docs.rs/headers/0.3.1/headers/trait.Header.html#tymethod.decode).

contrib/lib/src/serve.rs
schrieveslaach: Do you have any special configuration that I'm missing? If I run `cargo fmt` in the root, `git status` shows me more than 200 modified files.

core/lib/Cargo.toml
schrieveslaach: Parsing works with `Header::parse_header`. However, the current version of hyper use time 0.1.x (with very limited API) and if I'm using time in `named_file.rs` I'm using time 0.2.x which makes it hard to write nice code when comparing the dates. Do you have any advice?

core/lib/src/response/named_file.rs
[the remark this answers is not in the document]
schrieveslaach: It has been removed.

contrib/lib/src/serve.rs
jebrosen: You shouldn't be running it at all, is my point. We can't configure `cargo fmt` in a way that is compatible with the current style of the codebase and also consistent, as much as we want it to work.

core/lib/Cargo.toml
jebrosen: I didn't realize it had a dependency on `time 0.1` for that... that means the difference in our code for hyper 0.10 and 0.13 is more than I expected, which is disappointing. I had been thinking of going ahead and applying this to the `master` branch and then updating it when we merge `async`, but that might be too complex of an approach if we are running into issues like these.

schrieveslaach: I think in this case I should wait until you finished #1065

SergioBenitez:
Besides needed a rebase, what is the status of this?

schrieveslaach:
@jebrosen, the PR was stuck due to incompatible APIs (see comments above). I updated the PR and pushed a new version that use the headers crate (suggested by @jebrosen).

core/lib/src/response/named_file.rs
```diff
@@ -38,25 +45,35 @@ impl NamedFile {
...
+        Ok(NamedFile {
+            path: path.as_ref().to_path_buf(),
+            file,
+            modified: None,
+        })
+    }
+
+    pub async fn with_last_modified_date<P: AsRef<Path>>(path: P) -> io::Result<NamedFile> {
+        let mut named_file = NamedFile::open(path).await?;
+        named_file.modified = Some(named_file.path().metadata()?.modified()?);
```
jebrosen, line 57: This can be improved two ways:

```rust
named_file.modified = named_file.metadata().await?.modified().ok()
```

* get the metadata of the actually opened file, asynchronously
* fall back to `None` instead of an error on platforms where `modified()` isn't supported

core/lib/src/response/named_file.rs
```diff
@@ -87,27 +104,55 @@ impl NamedFile {
 /// implied by its extension, use a [`File`] directly.
 impl<'r> Responder<'r, 'static> for NamedFile {
     fn respond_to(self, req: &'r Request<'_>) -> response::Result<'static> {
-        let mut response = self.1.respond_to(req)?;
-        if let Some(ext) = self.0.extension() {
+        if let Some(last_modified) = &self.modified {
+            if let Some(if_modified_since) = req.headers().get_one("If-Modified-Since") {
+                let if_modified_since = match parse_if_modified_since(if_modified_since) {
+                    Ok(if_modified_since) => if_modified_since,
+                    Err(_err) => return Response::build().status(Status::BadRequest).ok(),
```
jebrosen, line 111: I think we should ignore the header in this case, instead of sending a 400:

> A recipient MUST ignore the If-Modified-Since header field if the
>   received field-value is not a valid HTTP-date, or if the request
>    method is neither GET nor HEAD.
(<https://tools.ietf.org/html/rfc7232#section-3.3>)

jebrosen (changes requested):
Overall LGTM; I requested a few minor changes. The encoding/decoding is a little bit wonky, but I think that would best be resolved by implementing #1067 with the `headers` crate.

core/lib/src/response/named_file.rs
schrieveslaach: Thanks for pointing out the section in the RFC 7232. I wasn't aware of it.

core/lib/src/response/named_file.rs
schrieveslaach: Done

contrib/lib/src/serve.rs
```diff
@@ -104,6 +104,9 @@ impl Options {
     /// directory, rather than its parent. This is _not_ enabled by default.
     pub const NormalizeDirs: Options = Options(0b0100);
 
+    /// `Options` enabling caching based on the `Last-Modified` header.
```
jebrosen, line 107: This should mention the `304` response with `If-Modified-Since`, maybe in another sentence or two.

core/lib/src/response/named_file.rs
```diff
@@ -38,25 +45,35 @@ impl NamedFile {
...
         let file = File::open(path.as_ref()).await?;
-        Ok(NamedFile(path.as_ref().to_path_buf(), file))
+        Ok(NamedFile {
+            path: path.as_ref().to_path_buf(),
+            file,
+            modified: None,
+        })
+    }
+
+    pub async fn with_last_modified_date<P: AsRef<Path>>(path: P) -> io::Result<NamedFile> {
```
jebrosen, line 55: A doc comment on this method should explain that it works like `NamedFile::open` and that it sets and checks `Last-Modified` and `If-Modified-Since`, and include an example.

jebrosen (approved):
This looks great to me, with a small expansion to the docs.

I think this PR already satisfies my concerns from <https://github.com/SergioBenitez/Rocket/pull/1376#pullrequestreview-500744280> by being opt-in via a new constructor. So my vote is for this PR, which also includes tests already.

core/lib/src/response/named_file.rs
schrieveslaach: Done

contrib/lib/src/serve.rs
schrieveslaach: Done

schrieveslaach:
@SergioBenitez, did you have the chance to review this pull request?

schrieveslaach:
@SergioBenitez, @jebrosen do you have any idea why the build [“Linux Debug (nightly)”](https://github.com/SergioBenitez/Rocket/pull/1219/checks?check_run_id=2599567514) after I rebased this PR? The error message seems unrelated to the changes in this PR.

SergioBenitez:
Likely a change in how diagnostics are rendered. I'll take a close look now.

SergioBenitez:
The unit tests should now pass. Is this ready for a review?

schrieveslaach:
@SergioBenitez, thanks. It works again.

contrib/lib/src/serve.rs
```diff
@@ -133,6 +133,12 @@ impl Options {
     /// that would be served is a directory.
     pub const NormalizeDirs: Options = Options(0b0100);
 
+    /// `Options` enabling caching based on the `Last-Modified` header. When
+    /// this is enabled, the [`StaticFiles`] handler will at the `Last-Modified`
```
LionsAd, line 137:
```suggestion
    /// this is enabled, the [`StaticFiles`] handler will add the `Last-Modified`
```

contrib/lib/src/serve.rs
```diff
@@ -133,6 +133,12 @@ impl Options {
...
+    /// header baesd on the modification datetime of the files in the served
```
LionsAd, line 138:
```suggestion
    /// header based on the modification datetime of the files in the served
```

SergioBenitez:
I don't think the API or implementation here are ones that I'm in favor of. Specifically:

  * This API does not leave room for growth into headers which are not simply flags, that is, which require a value like `Cache-Control`.
  * This API reuses `Options` from `StaticFiles` for `NamedFile` when only one bitfield actually applies.
  * This implementation introduces a dependency on `header` for two types.
  * The constructor approach (`with_last_modified_date()`) results in lengthy invocations.
  * The `Responder` impl indexes into a vector without checking it contains a value and then `unwrap()`s. This can trivially lead to panics.
  * There are a many `vec![]`, `to_string()` allocations. This functionality should require none.

Before revisting an implementation for this and #95 in general, I would like to agree on an API design and implementation strategy that has none of these drawbacks. Until then, I am closing this pull request.

PR #1230, opened 2020-02-13
Use Results instead of booleans in the todo example's error handling

cedric-h:
Using Booleans to convey success is often very verbose and not particularly idiomatic. Using Results to express the possible failure is much more explicit and lends the error handling impetus to the callee.

One part of this I'm not particularly happy with is the new need for `.unwrap()` on many places where `Task::all` was used previously; perhaps we also want to bubble errors in these areas? `.unwrap()` in the tests is simply unavoidable, as far as I can see.

### DIFF
--- a/examples/todo/src/main.rs
+++ b/examples/todo/src/main.rs
@@ -32,11 +32,11 @@ struct Context<'a, 'b>{ msg: Option<(&'a str, &'b str)>, tasks: Vec<Task> }
 
 impl<'a, 'b> Context<'a, 'b> {
     pub fn err(conn: &DbConn, msg: &'a str) -> Context<'static, 'a> {
-        Context{msg: Some(("error", msg)), tasks: Task::all(conn)}
+        Context{msg: Some(("error", msg)), tasks: Task::all(conn).unwrap()}
     }
 
     pub fn raw(conn: &DbConn, msg: Option<(&'a str, &'b str)>) -> Context<'a, 'b> {
-        Context{msg: msg, tasks: Task::all(conn)}
+        Context{msg: msg, tasks: Task::all(conn).unwrap()}
     }
 }
 
@@ -45,7 +45,7 @@ fn new(todo_form: Form<Todo>, conn: DbConn) -> Flash<Redirect> {
     let todo = todo_form.into_inner();
     if todo.description.is_empty() {
         Flash::error(Redirect::to("/"), "Description cannot be empty.")
-    } else if Task::insert(todo, &conn) {
+    } else if Task::insert(todo, &conn).is_ok() {
         Flash::success(Redirect::to("/"), "Todo successfully added.")
     } else {
         Flash::error(Redirect::to("/"), "Whoops! The server failed.")
@@ -54,20 +54,22 @@ fn new(todo_form: Form<Todo>, conn: DbConn) -> Flash<Redirect> {
 
 #[put("/<id>")]
 fn toggle(id: i32, conn: DbConn) -> Result<Redirect, Template> {
-    if Task::toggle_with_id(id, &conn) {
-        Ok(Redirect::to("/"))
-    } else {
-        Err(Template::render("index", &Context::err(&conn, "Couldn't toggle task.")))
-    }
+    Task::toggle_with_id(id, &conn).map_err(|e| Template::render(
+        "index",
+        &Context::err(&conn, &format!("Couldn't toggle task: {}", e))
+    ))?;
+
+    Ok(Redirect::to("/"))
 }
 
 #[delete("/<id>")]
 fn delete(id: i32, conn: DbConn) -> Result<Flash<Redirect>, Template> {
-    if Task::delete_with_id(id, &conn) {
-        Ok(Flash::success(Redirect::to("/"), "Todo was deleted."))
-    } else {
-        Err(Template::render("index", &Context::err(&conn, "Couldn't delete task.")))
-    }
+    Task::delete_with_id(id, &conn).map_err(|e| Template::render(
+        "index",
+        &Context::err(&conn, &format!("Couldn't delete task: {}", e))
+    ))?;
+
+    Ok(Flash::success(Redirect::to("/"), "Todo was deleted."))
 }
 
 #[get("/")]
--- a/examples/todo/src/main.rs
+++ b/examples/todo/src/main.rs
@@ -28,14 +28,14 @@ embed_migrations!();
 pub struct DbConn(SqliteConnection);
 
 #[derive(Debug, Serialize)]
-struct Context<'a, 'b>{ msg: Option<(&'a str, &'b str)>, tasks: Vec<Task> }
+struct Context<'a>{ msg: Option<(&'a str, &'a str)>, tasks: Vec<Task> }
 
-impl<'a, 'b> Context<'a, 'b> {
-    pub fn err(conn: &DbConn, msg: &'b str) -> Context<'static, 'b> {
+impl<'a> Context<'a> {
+    pub fn err(conn: &DbConn, msg: &'a str) -> Context<'a> {
         Self::raw(conn, Some(("error", msg)))
     }
 
-    pub fn raw<'x, 'y>(conn: &DbConn, msg: Option<(&'x str, &'y str)>) -> Context<'x, 'y> {
+    pub fn raw(conn: &DbConn, msg: Option<(&'a str, &'a str)>) -> Context<'a> {
         match Task::all(conn) {
             Ok(tasks) => Context{msg: msg, tasks},
             Err(_) => Context{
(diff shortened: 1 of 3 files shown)


jebrosen:
I agree with switching to `Result`.

> One part of this I'm not particularly happy with is the new need for `.unwrap()` on many places where Task::all was used previously

Yes. We should ideally not `unwrap` in handlers - it looks like it's only done in `Context`. Although, `Task::all` itself was unwrapping before. Perhaps `Context` could return no tasks and/or an error message if `Task::all` fails?

> `.unwrap()` in the tests is simply unavoidable, as far as I can see.

`unwrap` in tests is fine (and even idiomatic!)

cedric-h:
> Perhaps `Context` could return no tasks and/or an error message if `Task::all` fails?
Implemented!
```rs
            Err(_) => Context{
                msg: Some(("error", "Couldn't access the task database.")),
                tasks: vec![]
            }
```
I considered actually sending the user the Diesel error in the message, but I decided that might reveal some private information, and it would mess up the lifetime juggling we do, since that would require allocating a new string and we're operating purely on references here. I did want to log the error, but there wasn't a logging solution already set up so I opted to just ignore it.

```rs
    pub fn raw<'x, 'y>(conn: &DbConn, msg: Option<(&'x str, &'y str)>) -> Context<'x, 'y> {
```
I had to make `Task::raw` dynamic over the lifetimes passed to it so that I could just call `Task::raw` from `Task::err`, which let me get around copying the error handling into both functions.

cedric-h:
Would it be advisable to `cargo fmt` the examples?

examples/todo/src/main.rs
```diff
@@ -31,12 +31,18 @@ pub struct DbConn(SqliteConnection);
 struct Context<'a, 'b>{ msg: Option<(&'a str, &'b str)>, tasks: Vec<Task> }
```
jebrosen: Since we're in this file already, I'm pretty sure this can just be `Context<'a>`. That we only have one lifetime to name, simplifying a lot of the function signatures.

examples/todo/src/main.rs
```diff
@@ -54,20 +60,22 @@ fn new(todo_form: Form<Todo>, conn: DbConn) -> Flash<Redirect> {
 
 #[put("/<id>")]
 fn toggle(id: i32, conn: DbConn) -> Result<Redirect, Template> {
-    if Task::toggle_with_id(id, &conn) {
-        Ok(Redirect::to("/"))
-    } else {
-        Err(Template::render("index", &Context::err(&conn, "Couldn't toggle task.")))
-    }
+    Task::toggle_with_id(id, &conn).map_err(|e| Template::render(
```
jebrosen: I find this (and `delete`) easier to read as a match than with `?`, since it's the only expression:

```rust
match Task::toggle_with_id(id, &conn) {
    Ok(()) => Ok(Redirect::to("/")),
    Err(e) => Err(Template::render("index", &Context::err(&conn, format!("Couldn't toggle task: {}", e)))),
}
```

examples/todo/src/main.rs
```diff
@@ -31,12 +31,18 @@ pub struct DbConn(SqliteConnection);
 struct Context<'a, 'b>{ msg: Option<(&'a str, &'b str)>, tasks: Vec<Task> }
 
 impl<'a, 'b> Context<'a, 'b> {
-    pub fn err(conn: &DbConn, msg: &'a str) -> Context<'static, 'a> {
-        Context{msg: Some(("error", msg)), tasks: Task::all(conn)}
+    pub fn err(conn: &DbConn, msg: &'b str) -> Context<'static, 'b> {
+        Self::raw(conn, Some(("error", msg)))
```
jebrosen: I do think we should preserve the original error message; `tasks: Task::all(conn).unwrap_or_default()` is probably the clearest way.

examples/todo/src/main.rs
cedric-h: I always think `Ok(_) => Ok(` is a little weird, but I understand where you're coming from, this is definitely easier to read for most people

cedric-h:
It looks like the test only failed because it timed out for some reason?

examples/todo/src/task.rs
```diff
@@ -27,32 +28,39 @@ pub struct Todo {
...
-        updated_task.set(task_completed.eq(new_status)).execute(conn).is_ok()
+        updated_task.set(task_completed.eq(new_status)).execute(conn)?;
+
+        Ok(())
     }
 
-    pub fn delete_with_id(id: i32, conn: &SqliteConnection) -> bool {
-        diesel::delete(all_tasks.find(id)).execute(conn).is_ok()
+    pub fn delete_with_id(id: i32, conn: &SqliteConnection) -> Result<()> {
+        match diesel::delete(all_tasks.find(id)).execute(conn) {
```
SergioBenitez: This is just:

```suggestion
        diesel::delete(all_tasks.find(id)).execute(conn).map(|_| ())
```

What's the type of the `Ok`, and, should we just throw it away? Is there any utility to returning it?

examples/todo/src/task.rs
```diff
@@ -27,32 +28,39 @@ pub struct Todo {
...
+            Ok(_) => Ok(()),
+            Err(e) => Err(e)
+        }
     }
 
     #[cfg(test)]
-    pub fn delete_all(conn: &SqliteConnection) -> bool {
-        diesel::delete(all_tasks).execute(conn).is_ok()
+    pub fn delete_all(conn: &SqliteConnection) -> Result<()> {
+        match diesel::delete(all_tasks).execute(conn) {
```
SergioBenitez: Same here.

examples/todo/src/task.rs
cedric-h: https://docs.rs/diesel/1.4.3/diesel/query_dsl/trait.RunQueryDsl.html#method.execute

Evidently this returns the number of rows affected.
Some sort of behind the scenes database modification metrics could be saved potentially, but I think for an example it's best to leave that kind of functionality out and simply leave a comment explaining that the number of rows changed are returned for those who want them?

cedric-h: Map was being used here, but Jebrosen indicated in some other places that he found `matches` more readable in general, so I changed here as well for consistency. In retrospect, I probably should've evaluated this on a case by case basis instead of trying to achieve all around consistency.

examples/todo/src/task.rs
```diff
@@ -1,4 +1,5 @@
-use diesel::{self, prelude::*};
+use diesel::{self, result::Error as DieselError, prelude::*};
+type Result<T> = std::result::Result<T, DieselError>;
```
SergioBenitez, line 2: Isn't this `diesel::result::QueryResult`? We should import and return that from these functions instead of duplicating it here.

examples/todo/src/task.rs
```diff
@@ -27,32 +28,38 @@ pub struct Todo {
...
+        all_tasks.order(tasks::id.desc()).load::<Task>(conn)
     }
 
-    pub fn insert(todo: Todo, conn: &SqliteConnection) -> bool {
+    pub fn insert(todo: Todo, conn: &SqliteConnection) -> Result<()> {
         let t = Task { id: None, description: todo.description, completed: false };
-        diesel::insert_into(tasks::table).values(&t).execute(conn).is_ok()
+        diesel::insert_into(tasks::table).values(&t).execute(conn)?;
+
+        Ok(())
```
SergioBenitez, line 39: I don't see the harm in returning the result from the call to `insert_into` directly.

examples/todo/src/task.rs
```diff
@@ -27,32 +28,38 @@ pub struct Todo {
...
+    pub fn toggle_with_id(id: i32, conn: &SqliteConnection) -> Result<()> {
+        let task = all_tasks.find(id).get_result::<Task>(conn)?;
 
-        let new_status = !task.unwrap().completed;
+        let new_status = !task.completed;
         let updated_task = diesel::update(all_tasks.find(id));
-        updated_task.set(task_completed.eq(new_status)).execute(conn).is_ok()
+        updated_task.set(task_completed.eq(new_status)).execute(conn)?;
+
+        Ok(())
```
SergioBenitez, line 49: Same here. Why toss away potentially useful information?

examples/todo/src/task.rs
```diff
@@ -27,32 +28,38 @@ pub struct Todo {
...
     }
 
-    pub fn delete_with_id(id: i32, conn: &SqliteConnection) -> bool {
-        diesel::delete(all_tasks.find(id)).execute(conn).is_ok()
+    pub fn delete_with_id(id: i32, conn: &SqliteConnection) -> Result<()> {
+        // `.execute` returns a usize representing the number of
+        // database rows affected by this modification.
+        // In a larger app these metrics may be useful,
+        // but for a simple example we'll ignore them.
+        diesel::delete(all_tasks.find(id)).execute(conn).map(|_| ())
```
SergioBenitez, line 57: And here. The caller can ignore the result if they'd like. I don't see the utility or reason to doing so here.

examples/todo/src/task.rs
```diff
@@ -27,32 +28,38 @@ pub struct Todo {
...
     }
 
     #[cfg(test)]
-    pub fn delete_all(conn: &SqliteConnection) -> bool {
-        diesel::delete(all_tasks).execute(conn).is_ok()
+    pub fn delete_all(conn: &SqliteConnection) -> Result<()> {
+        // see comment in `.delete_with_id`.
+        diesel::delete(all_tasks).execute(conn).map(|_| ())
```
SergioBenitez, line 63: And here.

examples/todo/src/main.rs
```diff
@@ -28,15 +28,21 @@ embed_migrations!();
...
 #[derive(Debug, Serialize)]
-struct Context<'a, 'b>{ msg: Option<(&'a str, &'b str)>, tasks: Vec<Task> }
+struct Context<'a>{ msg: Option<(&'a str, &'a str)>, tasks: Vec<Task> }
 
-impl<'a, 'b> Context<'a, 'b> {
-    pub fn err(conn: &DbConn, msg: &'a str) -> Context<'static, 'a> {
-        Context{msg: Some(("error", msg)), tasks: Task::all(conn)}
+impl<'a> Context<'a> {
+    pub fn err(conn: &DbConn, msg: &'a str) -> Context<'a> {
+         Context{msg: Some(("error", msg)), tasks: Task::all(conn).unwrap_or_default()}
```
SergioBenitez, line 35:
```suggestion
        Context{msg: Some(("error", msg)), tasks: Task::all(conn).unwrap_or_default()}
```

examples/todo/src/main.rs
```diff
@@ -45,28 +51,26 @@ fn new(todo_form: Form<Todo>, conn: DbConn) -> Flash<Redirect> {
     let todo = todo_form.into_inner();
     if todo.description.is_empty() {
         Flash::error(Redirect::to("/"), "Description cannot be empty.")
-    } else if Task::insert(todo, &conn) {
-        Flash::success(Redirect::to("/"), "Todo successfully added.")
+    } else if let Err(e) = Task::insert(todo, &conn) {
+        Flash::error(Redirect::to("/"), &format!("Database error: {}", e))
```
SergioBenitez, line 55:
```suggestion
        Flash::error(Redirect::to("/"), format!("Database error: {}", e))
```

SergioBenitez, line 55: What do these database error messages actually look like? In a real application, you likely wouldn't want to divulge database information to the public, and so I worry that this is encouraging bad practice, depending on what information is actually contained in the error. Instead, we probably just want to log the error for private access.

SergioBenitez:
Merged in da34b66eb442cd151ba3efdb3bae6c7015bf2e4e with follow-up in a4f5be93ef52340724674aaf3270137336e58f66. Thank you!

Issue #852, opened 2018-12-09
Revamp Configuration


svenstaro:
As per discussion on IRC, I think the config file handling should be revamped. The current three fixed environment system doesn't scale and if you need more you have to work around that limit. That's not nice. Also, the current extras aren't typed and are cumbersome to use.

Below are some general goals that a new system should fulfill.

## Goals

- [ ] Get rid of the current rigid enum `Environment` structure.
- [ ] Allow the config to be provided via environment variables, a single config file, separate config files, or plain code.
- [ ] Allow for default environments still but you should easily be able to define your own set of environments.
- [ ] Allow for strongly typed user configs (instead of the current `extras`).

## Considerations

- Maybe use [config-rs](https://github.com/mehcode/config-rs)?
- The current three environments fulfill the purposes of providing a bunch of defaults as well as safety checks. If we get rid of those rigid environments and if the environment are dynamic, it won't be possible to do this for a specific pre-defined environment anymore. Therefore, in order to keep a safety switch on by default, I propose we make everything safe by default. For local development (and perhaps tests and such) where we don't care about safe defaults, we can have a `ROCKET_DEBUG` env var. Since the configs are isomorphic, this will translate to `config.debug` in code.
- Generally speaking, I don't think we should be talking about _Rocket_ configuration but about _app_ configuration and Rocket should then just be able to take what it needs from the app configuration.

  Currently the user is supposed to stuff all of their custom config into Rocket like it's a weird thing to have extra config. I think this is the wrong way around. Rather than that, there should some general app config and Rocket should just take from that whatever it needs (like port, address, etc). For instance, if we provided a TOML config file, we might have this in our new, Rocket-unspecific `app.toml`:

  ```
  [general]
  app_name = "hello"
  twitter_api_key = "something"

  [rocket]
  address = "0.0.0.0"
  port = 8080
  ```

  This illustration obviously glances over the problem with environments but it's how I imagine this to work. Likewise, if we're talking about environment variables, the above example would look like this:

  ```
  APP_NAME="hello"
  TWITTER_API_KEY="something"
  ROCKET_ADDRESS="0.0.0.0"
  ROCKET_PORT=8080
  ```
[...]

PR #1231, opened 2020-02-13
Add ability to override path to Rocket.toml configuration file.

chrismoos:
Right now a single Rocket.toml is allowed with a limited amount of environments. This change allows an environment variable, ROCKET_CONFIG_FILE to be set, allowing a user to override environment details with additional configuration files.

This will alleviate some of the concerns in #852.


SergioBenitez:
This wasn't quite the correct implementation, so I went ahead and added it in f7b1089b4916667fc0e0c13f21b672e2e89c054b. Thank you for opening this, however!

PR #1237, opened 2020-02-21
Add a way to cooperatively shut down the server

### LINKED ISSUE #180: Clean shutdown?

notriddle:
Closes #180

### DIFF
--- a/core/lib/src/rocket.rs
+++ b/core/lib/src/rocket.rs
@@ -7,7 +7,7 @@ use std::sync::Arc;
 
 use futures_util::future::{Future, FutureExt, BoxFuture};
 use futures_util::stream::StreamExt;
-use tokio::sync::{mpsc, oneshot};
+use tokio::sync::{broadcast, oneshot};
 
 use yansi::Paint;
 use state::Container;
@@ -40,7 +40,7 @@ pub struct Rocket {
     pub(crate) state: Container,
     fairings: Fairings,
     shutdown_handle: ShutdownHandle,
-    shutdown_receiver: Option<mpsc::Receiver<()>>,
+    shutdown_receiver: Option<broadcast::Receiver<()>>,
 }
 
 // This function tries to hide all of the Hyper-ness from Rocket. It
@@ -433,7 +433,7 @@ impl Rocket {
                           Paint::default(LoggedValue(value)).bold());
         }
 
-        let (shutdown_sender, shutdown_receiver) = mpsc::channel(1);
+        let (shutdown_sender, shutdown_receiver) = broadcast::channel(1);
 
         let rocket = Rocket {
             config,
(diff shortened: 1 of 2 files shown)


jebrosen:
This specifically implements this approach from https://github.com/SergioBenitez/Rocket/issues/180#issuecomment-588483359:

> provide a mechanism to cooperatively stop responses, etc.

The implementation looks good, but it would be good to have more test coverage. I am also not yet sure it's the approach we want to go with; I would like to see more discussion before we commit to an API.

notriddle:
I'm almost certain that I'd want cooperative connection teardown, for a couple of reasons:

* Everything else in async-land is based on cooperative resource management, so it seems arbitrary to use anything else here.

* The whole point of having "clean" shutdown is to avoid closing a connection until after you're done sending your response. If clean shutdown doesn't provide that, then it doesn't have much reason to exist. This means not stopping a request before it's done. But eternal requests are *never* done (unless, as implemented in this PR, it decides it's done because Rocket asks).

* You are probably running underneath a service manager like systemd, kubernetes, heroku, runit... Those systems all work by sending you service a signal to cooperatively shut down, which would trigger the code written in this PR, and then if you don't shut yourself down after ten seconds or so they would hard-kill you. No point in duplicating their functionality.

jebrosen:
> But eternal requests are never done (unless, as implemented in this PR, it decides it's done because Rocket asks).

This is pretty much what my uncertainty hinges on: this mechanism would require that all potentially infinite `Responder` implementations "know" about shutdown. In general, I am uncomfortable with APIs that must be "opted into" in order to be correct.

How to integrate this can be difficult to determine; for example a modification of [my rocket-rooms example](https://git.jebrosen.com/jeb/rocket-rooms/src/commit/6b6f833ec8c8bdacea2ca7258809181765bfd147/src/sse/v2.rs#L76) would require a call to `self.shutdown_notification_handle.poll()` somewhere, but it's not entirely obvious where that should be and is use-case dependent. It would be easier to add it to [the user-side code](https://git.jebrosen.com/jeb/rocket-rooms/src/commit/6b6f833ec8c8bdacea2ca7258809181765bfd147/src/main.rs#L37) by `select!`ing between events and the shutdown notification, but I think it should be the responsibility of the underlying `Responder`.

notriddle:
Then what's the point of having a clean-shutdown API other than guaranteeing that connections aren't closed until they're done?

jebrosen:
> Then what's the point of having a clean-shutdown API other than guaranteeing that connections aren't closed until they're done?

Clean shutdown isn't only about the connections; it is also about the rest of the application state and for some users that aspect can be more important. For example an application using the `async` branch can already be assured that `Drop` handlers will run on shutdown.

PR #1242, opened 2020-02-26
Async: on_attach fairings

jebrosen:
This PR makes `on_attach` fairings asynchronous. It transitively requires several other functions to become asynchronous, including the "entry point" (usually `launch()` or `Client::new()`). This specific implementation is a balance between needing to do asynchronous work in `on_attach`, such as database setup, versus the verbosity of making too many builder-type functions require `.await`s.

Some other similarly sweeping changes that I have not made yet, and can be added to this or made as a followup PR:

* A synchronous version of `Client` (with an internal `Runtime`, that would `block_on` each operation).
  * This can simplify *many* tests by making `.await`s internal - especially in e.g. https://github.com/SergioBenitez/Rocket/compare/async...jebrosen:async-on-attach?expand=1#diff-048d6496580899aa2aec5fc6a244faf0
* Changing the return type of `launch()` so that the most common usage does not require `let _ =` to suppress an unused `Result` warning.

Some other things we may want, off the top of my head:

* An explanation of why `#[rocket::main]` and `#[rocket::async_test]` exist (short version: an attribute is worth it, and reexports of `tokio::main` and `tokio::test` [don't work](hyperium/hyper#1878))


core/lib/src/rocket.rs
```diff
@@ -43,13 +55,21 @@ pub struct Rocket {
     shutdown_receiver: Option<mpsc::Receiver<()>>,
 }
 
+/// A readonly view into the current state of an unlaunched [`Rocket`].
+///
+/// The `Rocket` type only provides a single method for inspection,
+/// [`Rocket::inspect`], which returns this type. An `Inspector` instance can be
+/// used to access configuration, managed state, and mounted routes.
+#[derive(Copy, Clone)]
+pub struct Inspector<'r>(pub(crate) &'r RocketInner);
```
SergioBenitez: What do we get out of wrapping `RocketInner` in the `Inspector` struct? Do we lose any safety by passing around `&Manifest` to end-users?

core/lib/src/rocket.rs
```diff
@@ -33,6 +33,18 @@ use crate::http::uri::Origin;
...
+}
+
+enum BuildOperation {
+    Mount(Origin<'static>, Vec<Route>),
+    Register(Vec<Catcher>),
+    Manage(Box<dyn FnOnce(RocketInner) -> RocketInner + Send + Sync + 'static>),
+    Attach(Box<dyn Fairing>),
+}
+
+pub(crate) struct RocketInner {
```
SergioBenitez: The name `RocketInner`, for whatever reason, doesn't sit well with me. Could we name it something more indicative of the fact that it contains all of the information about the application? `Manifest`, perhaps, in keeping with the theme? Then we can have `rocket.manifest().await`.

core/lib/src/rocket.rs
```diff
@@ -33,6 +33,18 @@ use crate::http::uri::Origin;
 /// The main `Rocket` type: used to mount routes and catchers and launch the
 /// application.
 pub struct Rocket {
+    pub(crate) inner: Option<RocketInner>,
+    pending: Vec<BuildOperation>,
+}
+
+enum BuildOperation {
```
SergioBenitez: This is very nice.

core/lib/src/rocket.rs
```diff
@@ -610,22 +750,26 @@ impl Rocket {
...
     /// ```
     #[inline]
-    pub fn manage<T: Send + Sync + 'static>(self, state: T) -> Self {
-        if !self.state.set::<T>(state) {
-            error!("State for this type is already being managed!");
-            panic!("Aborting due to duplicately managed state.");
-        }
-
+    pub fn manage<T: Send + Sync + 'static>(mut self, state: T) -> Self {
+        self.pending.push(BuildOperation::Manage(Box::new(|rocket| {
```
SergioBenitez: Clever! Do we have tests that ensure this `panic` actually happens?

core/lib/src/rocket.rs
```diff
@@ -610,22 +750,26 @@ impl Rocket {
...
+            if !rocket.state.set::<T>(state) {
+                error!("State for this type is already being managed!");
+                panic!("Aborting due to duplicately managed state.");
+            }
+            rocket
```
SergioBenitez: Style: we should have an empty line above here.

core/lib/src/rocket.rs
```diff
@@ -610,22 +750,26 @@ impl Rocket {
...
+        })));
         self
```
SergioBenitez: Style: we should have an empty line above here.

core/lib/src/rocket.rs
```diff
@@ -641,135 +785,76 @@ impl Rocket {
...
-            }
-        });
-
-        #[derive(Clone)]
-        struct TokioExecutor;
-
-        impl<Fut> hyper::Executor<Fut> for TokioExecutor where Fut: Future + Send + 'static, Fut::Output: Send {
-            fn execute(&self, fut: Fut) {
-                tokio::spawn(fut);
+    pub(crate) fn finish(&mut self) -> BoxFuture<'_, ()> {
```
SergioBenitez: Can this not be an `async fn`? Also, the name `finish`, to me, implies some finality, but we call this function multiple times. Maybe `forward` or `advance` or `fire` or something?

core/lib/src/rocket.rs
```diff
@@ -641,135 +785,76 @@ impl Rocket {
...
+        Box::pin(async move {
+            while !self.pending.is_empty() {
```
SergioBenitez: Can this be `while let Some(op) = self.pending.pop()`?

core/lib/src/rocket.rs
```diff
@@ -641,135 +785,76 @@ impl Rocket {
...
     /// ```
-    pub async fn serve(self) -> Result<(), crate::error::Error> {
+    pub async fn launch(mut self) -> Result<(), crate::error::Error> {
         use std::net::ToSocketAddrs;
 
         use crate::error::Error::Launch;
 
-        let full_addr = format!("{}:{}", self.config.address, self.config.port);
+        self.finish().await;
+
```
SergioBenitez: Lots of vertical space here...

core/lib/src/rocket.rs
```diff
@@ -842,41 +928,40 @@ impl Rocket {
...
-    /// requests to mounted routes and catchers. This function does not return
-    /// unless a shutdown is requested via a [`ShutdownHandle`] or there is an
-    /// error.
-    ///
-    /// This is a convenience function that creates a suitable default runtime
-    /// and launches the server on that runtime. If you already have a runtime,
-    /// use the [`Rocket::serve`] method instead.
-    ///
-    /// # Error
+    pub(crate) fn take_inner(&mut self) -> RocketInner {
```
SergioBenitez: Can we make this safer? Ideally we'd like to make it easy to prove that `self.inner` is "always" `Some` after some logically atomic operation. I _think_ we'd like something to the effect of:

```rust
fn with_inner(&mut self, f: impl FnOnce(RocketInner) -> RocketInner) {
    let inner = self.inner.take().expect("...");
    self.inner = Some(f(inner));
}
```

core/lib/src/rocket.rs
```diff
@@ -641,135 +785,76 @@ impl Rocket {
...
     /// async fn main() {
     /// # if false {
-    ///     let result = rocket::ignite().serve().await;
+    ///     let result = rocket::ignite().launch().await;
     ///     assert!(result.is_ok());
     /// # }
     /// }
     /// ```
-    pub async fn serve(self) -> Result<(), crate::error::Error> {
+    pub async fn launch(mut self) -> Result<(), crate::error::Error> {
```
SergioBenitez: I know most the contents of this function were introduced in some other commits, but it has become very difficult to follow. The combination of macros and `cfg`s, in particular, really help to obfuscate the functionality. We really must improve this before we release 0.5.

scripts/test.sh
```diff
@@ -60,7 +60,7 @@ ensure_tab_free
 #ensure_trailing_whitespace_free
 
 echo ":: Updating dependencies..."
-cargo update
+#cargo update
```
SergioBenitez: Why was this commented out?

SergioBenitez:
This is really excellent work, @jebrosen! Please see my inline comments. The two biggest suggestions are:

  1. Renaming `RocketInner` to `Manifest`, exposing `Manifest` directly instead of wrapping it in an `Inspector`.
  2. Changing the signature of `on_attach` to allow returning unboxed futures.

core/lib/src/fairing/ad_hoc.rs
```diff
@@ -66,10 +66,10 @@ impl AdHoc {
     /// use rocket::fairing::AdHoc;
     ///
     /// // The no-op attach fairing.
-    /// let fairing = AdHoc::on_attach("No-Op", |rocket| Ok(rocket));
+    /// let fairing = AdHoc::on_attach("No-Op", |rocket| Box::pin(async move { Ok(rocket) }));
     /// ```
     pub fn on_attach<F: Send + 'static>(name: &'static str, f: F) -> AdHoc
-        where F: FnOnce(Rocket) -> Result<Rocket, Rocket>
+        where F: FnOnce(Rocket) -> BoxFuture<'static, Result<Rocket, Rocket>>
```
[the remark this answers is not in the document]
jebrosen: Yeah, it should.

It might be nice to do that for `on_request` and `on_response` too, but the lack of support for `for<'a> Fn(&'a Request) -> Fut, Fut: 'a` may make it harder or impossible for those particular functions.

contrib/codegen/tests/ui-fail/database-types.rs
```diff
@@ -7,6 +7,8 @@ struct Unknown;
 //~^ ERROR Unknown: rocket_contrib::databases::Poolable
 //~^^ ERROR is private
 //~^^^ ERROR no method named `get`
+//~^^^^ ERROR no function or associated item
+//~^^^^^ ERROR Unknown: rocket_contrib::databases::Poolable
```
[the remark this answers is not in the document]
jebrosen: They might be indirectly caused by these changes. I have noticed that in general rustc has been more and more likely to continue with further error checking over the last few months, so it is hard to say exactly.

core/lib/src/rocket.rs
jebrosen: Good point - privacy should be enough to make this safe. I didn't want to expose the name `RocketInner` or use the name `Inspector` as the underlying type name, but the name `Manifest` solves both of those problems.

core/lib/src/rocket.rs
jebrosen: I like the type name `Manifest`. I'm not quite sold on the method name `manifest()` because of potential confusion over "to *manifest* a `Rocket`" vs "to retrieve a `Rocket`'s *manifest*`.

core/lib/src/rocket.rs
jebrosen: I don't see such a test, but it's easy to add.

core/lib/src/rocket.rs
jebrosen: I *believe* this is not an `async fn` in order to break recursion with `_attach` - if that's indeed the case I will make a note of that.

Agreed on naming and I will think on that some more.

core/lib/src/rocket.rs
jebrosen: I believe some callers can and some callers can't handle that signature. I will see about making those more clear.

scripts/test.sh
jebrosen: Whoops, just got committed accidentally - needed it to run tests offline.

core/lib/src/fairing/ad_hoc.rs
```diff
@@ -153,17 +153,20 @@ impl Fairing for AdHoc {
         Info { name: self.name, kind }
     }
 
-    fn on_attach(&self, rocket: Rocket) -> Result<Rocket, Rocket> {
+    async fn on_attach(&self, rocket: Rocket) -> Result<Rocket, Rocket> {
         if let AdHocKind::Attach(ref mutex) = self.kind {
-            let mut opt = mutex.lock().expect("AdHoc::Attach lock");
-            let f = opt.take().expect("internal error: `on_attach` one-call invariant broken");
-            f(rocket)
+            let f = mutex
```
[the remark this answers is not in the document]
jebrosen: Would you prefer it to look more like the previous version, or to go further and remove this `let` binding?

core/lib/src/rocket.rs
jebrosen: While working on this I noticed some docs discrepancies: `mount` can panic immediately or on `finish`, `manage` won't panic until `finish`, and an `attach`ed `on_attach` fairing is no longer executed immediately. I'll make sure to address those.

core/lib/src/rocket.rs
jebrosen: I added a comment why it's not - to preserve order.

jebrosen:
I've added commits resolving several comments, but I am not quite finished yet / have a few follow-up questions.

core/lib/src/rocket.rs
```diff
@@ -641,135 +781,79 @@ impl Rocket {
...
-
-        impl<Fut> hyper::Executor<Fut> for TokioExecutor where Fut: Future + Send + 'static, Fut::Output: Send {
-            fn execute(&self, fut: Fut) {
-                tokio::spawn(fut);
+    // NB: Not an `async fn` because _attach() calls finish(), recursing.
+    pub(crate) fn finish(&mut self) -> BoxFuture<'_, ()> {
+        Box::pin(async move {
+            while !self.pending.is_empty() {
+                let op = self.pending.remove(0);
+                let manifest = self.manifest.take().expect("TODO error message");
```
jebrosen: Write a better error message (and in the other few places this appears). The error condition would be along the lines of ".inspect() was called but not awaited"

core/lib/src/local/client.rs
```diff
@@ -139,11 +144,13 @@ impl Client {
...
+    /// # };
     /// ```
     #[inline(always)]
-    pub fn untracked(rocket: Rocket) -> Result<Client, LaunchError> {
-        Client::_new(rocket, false)
+    pub async fn untracked(rocket: Rocket) -> Result<Client, LaunchError> {
+        Client::_new(rocket, false).await
     }
 
     /// Returns the instance of `Rocket` this client is creating requests for.
```
jebrosen: This documentation and the the example don't make as much sense now

contrib/codegen/tests/ui-fail/database-types.rs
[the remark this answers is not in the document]
jebrosen: As far as I can tell, the `no function or associated item` is due to attempting to call `.map(#conn_type::pool)` - incidentally, that should probably be `<#conn_type as #Poolable>::pool`.

This playground does some of the same things that `#[database]` does when the types don't implement `Poolable`:

https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=6c3e950eafb9094cf567495fb5d0cddd

Notice that invoking a nonexistent method on `b` in an async block is enough to raise an error and also suppress the errors indicating that `A` and `B` are not iterators, but calling it directly in a function doesn't do the same (seemingly because `S` itself is determined to be malformed before even typechecking anything in `impl S`). It's likely that the number of errors will continue to go up and down over time with minor tweaks to resolution order of types and traits as it has in the past.

core/lib/src/fairing/ad_hoc.rs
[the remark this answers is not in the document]
jebrosen: So, this is actually necessary so that `opt` (`MutexGuard`) is not in scope across an `await` point. The cleaner-looking `drop(opt)` before the `await` doesn't work due to rust-lang/rust#57478.

This is the other way that works:

```rust
            let f = {
                let mut opt = mutex.lock().expect("AdHoc::Attach lock");
                opt.take().expect("internal error: `on_attach` one-call invariant broken")
            };
            f(rocket).await
```

jebrosen:
This is ready for another review.

core/lib/src/rocket.rs
SergioBenitez: What's the status of the discrepancies that you found?

core/lib/src/rocket.rs
```diff
@@ -641,135 +781,79 @@ impl Rocket {
...
-
-        impl<Fut> hyper::Executor<Fut> for TokioExecutor where Fut: Future + Send + 'static, Fut::Output: Send {
-            fn execute(&self, fut: Fut) {
-                tokio::spawn(fut);
+    // NB: Not an `async fn` because _attach() calls finish(), recursing.
+    pub(crate) fn finish(&mut self) -> BoxFuture<'_, ()> {
+        Box::pin(async move {
+            while !self.pending.is_empty() {
+                let op = self.pending.remove(0);
+                let manifest = self.manifest.take().expect("internal error: manifest was taken and not replaced. Was `inspect()` called but not polled to completion?");
```
SergioBenitez, line 807: This needs to be broken into multiple lines. Use `\` to break up a string without adding space:

```rust
"this is a very \
long string"
```

core/lib/src/rocket.rs
```diff
@@ -337,6 +353,149 @@ impl Rocket {
...
+        self.fairings.pretty_print_counts();
+
+        // Determine the address and port we actually binded to.
+        self.config.port = listener.local_addr().map(|a| a.port()).unwrap_or(0);
+
+        let proto = if self.config.tls.is_some() {
+            "https://"
+        } else {
+            "http://"
+        };
```
SergioBenitez, line 442:
```suggestion
        let proto = self.config.tls.map_or("http://", |_| "https://");
```

core/lib/src/rocket.rs
SergioBenitez: I'm not sure I follow. Why would making the function an `async fn` lead to recursion where it otherwise wouldn't? Also, any suggestions on a name change? Perhaps `process` or `process_manifest`?

core/lib/src/rocket.rs
```diff
@@ -641,135 +781,79 @@ impl Rocket {
...
-            }
-        });
-
-        #[derive(Clone)]
-        struct TokioExecutor;
-
-        impl<Fut> hyper::Executor<Fut> for TokioExecutor where Fut: Future + Send + 'static, Fut::Output: Send {
-            fn execute(&self, fut: Fut) {
-                tokio::spawn(fut);
+    // NB: Not an `async fn` because _attach() calls finish(), recursing.
```
SergioBenitez, line 802: Can we have a more detailed comment on the expected effect (pre/post conditions) of this function as well as a description of where/when/why it's called?

core/lib/src/rocket.rs
```diff
@@ -641,135 +781,79 @@ impl Rocket {
...
-            .map_err(|e| crate::error::Error::Run(Box::new(e)))
+    pub(crate) async fn finish_and_take_manifest(mut self) -> Manifest {
+        self.finish().await;
+        self.manifest.take().expect("internal error: finish() should have replaced self.manifest")
     }
 
-    /// Returns a `Future` that drives the server and completes when the server
-    /// is shut down or errors. If the `ctrl_c_shutdown` feature is enabled,
-    /// the server will shut down gracefully once `Ctrl-C` is pressed.
+    /// Returns a `Future` that drives the server, listening for and dispathcing
```
SergioBenitez, line 823:
```suggestion
    /// Returns a `Future` that drives the server, listening for and dispatching
```

core/lib/src/rocket.rs
```diff
@@ -641,135 +781,79 @@ impl Rocket {
...
+    /// requests to mounted routes and catchers. The `Future` completes when the
+    /// server is shut down (via a [`ShutdownHandle`] or encounters a fatal
```
SergioBenitez, line 825:
```suggestion
    /// server is shut down, via a [`ShutdownHandle`], or encounters a fatal
```

core/lib/src/rocket.rs
```diff
@@ -641,135 +781,79 @@ impl Rocket {
...
+    /// error.  If the `ctrl_c_shutdown` feature is enabled, the server will
+    /// also shut down once `Ctrl-C` is pressed.
+    ///
+    /// # Error
+    /// If there is a problem starting the application, an [`Error`] is
```
SergioBenitez, line 830:
```suggestion
    /// # Error
    ///
    /// If there is a problem starting the application, an [`Error`] is
```

core/lib/src/rocket.rs
```diff
@@ -641,135 +781,79 @@ impl Rocket {
...
 
         use crate::error::Error::Launch;
 
-        let full_addr = format!("{}:{}", self.config.address, self.config.port);
+        let mut manifest = self.finish_and_take_manifest().await;
+        manifest.prelaunch_check().map_err(crate::error::Error::Launch)?;
+
+        let config = manifest.config();
+
+        let full_addr = format!("{}:{}", config.address, config.port);
```
SergioBenitez, line 856:
```suggestion
        let config = manifest.config();
        let full_addr = format!("{}:{}", config.address, config.port);
```

core/lib/src/rocket.rs
```diff
@@ -842,41 +927,32 @@ impl Rocket {
...
-    /// unless a shutdown is requested via a [`ShutdownHandle`] or there is an
-    /// error.
-    ///
-    /// This is a convenience function that creates a suitable default runtime
-    /// and launches the server on that runtime. If you already have a runtime,
-    /// use the [`Rocket::serve`] method instead.
-    ///
-    /// # Error
+    pub(crate) fn _manifest(&self) -> &Manifest {
+        self.manifest.as_ref().expect("internal error: manifest was taken and not replaced. Was `inspect()` called but not polled to completion?")
```
SergioBenitez, line 931: Same here with splitting the line up.

core/lib/src/rocket.rs
```diff
@@ -842,41 +927,32 @@ impl Rocket {
...
+
+    /// Access the current state of this `Rocket` instance.
     ///
-    /// If there is a problem starting the application, a [`LaunchError`] is
-    /// returned. Note that a value of type `LaunchError` panics if dropped
-    /// without first being inspected. See the [`LaunchError`] documentation for
-    /// more information.
+    /// The `Mnaifest` type provides methods such as [`Manifest::routes`]
+    /// and [`Manifest::state`]. This method is called to get an `Manifest`
+    /// instance.
```
SergioBenitez, line 938:
```suggestion
    /// The `Manifest` type provides methods such as [`Manifest::routes()`]
    /// and [`Manifest::state()`]. This method is called to get a `Manifest`
    /// instance.
```

core/lib/src/rocket.rs
```diff
@@ -952,11 +1032,13 @@ impl Rocket {
...
+    /// assert_eq!(rocket.inspect().await.state::<MyState>(), Some(&MyState("hello!")));
     ///
-    /// let client = rocket::local::Client::new(rocket).expect("valid rocket");
-    /// assert_eq!(client.rocket().state::<MyState>(), Some(&MyState("hello!")));
+    /// let client = rocket::local::Client::new(rocket).await.expect("valid rocket");
+    /// assert_eq!(client.manifest().state::<MyState>(), Some(&MyState("hello!")));
+    /// # });
     /// ```
     #[inline(always)]
     pub fn state<T: Send + Sync + 'static>(&self) -> Option<&T> {
```
SergioBenitez, line 1044: Perhaps we should also have a `async state()` method on `Rocket`?

site/guide/3-overview.md
```diff
@@ -214,9 +215,9 @@ You can find async-ready libraries on [crates.io](https://crates.io) with the
 
 ! note
 
-  Rocket 0.5 uses the tokio (0.2) runtime. `Rocket::launch()` will automatically
-  start a runtime for you, or you can create a runtime yourself and use
-  `Rocket::spawn()`.
+  Rocket 0.5 uses the tokio (0.2) runtime. The runtime is started for you
+  if you use `#[rocket::main]`, but you can still `launch()` a rocket
+  instance on a custom-built `Runtime`.
```
SergioBenitez, line 220: This requires some pointers on how to do so. Is that material available?

core/lib/src/fairing/ad_hoc.rs
```diff
@@ -1,8 +1,8 @@
 use std::sync::Mutex;
 
-use futures_util::future::BoxFuture;
+use futures_util::future::{Future, BoxFuture};
```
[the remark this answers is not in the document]
jebrosen: I made this change and removed the direct dependency on `futures-util`, but I opted for `use futures::` instead of `use crate::futures::` based on length.

site/guide/3-overview.md
SergioBenitez: We should perhaps open up an issue, if one doesn't already exist, to document all of the things that need to be documented before we release `async`.

core/lib/src/rocket.rs
SergioBenitez: That'd be great. Let's do it.

core/lib/src/request/state.rs
```diff
@@ -99,9 +101,11 @@ use crate::http::Status;
 ///     state.0.to_string()
 /// }
 ///
-/// let rocket = rocket::ignite().manage(MyManagedState(127));
-/// let state = State::from(&rocket).expect("managing `MyManagedState`");
+/// # rocket::async_test(async {
+/// let mut rocket = rocket::ignite().manage(MyManagedState(127));
+/// let state = State::from(rocket.inspect().await).expect("managing `MyManagedState`");
```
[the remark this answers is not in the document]
SergioBenitez: That's fair. I think the idea here was to use that the two were equivalent. The addition of `.inspect()` and `.await` makes this feel overly verbose, however.

core/lib/src/fairing/ad_hoc.rs
```diff
@@ -66,12 +66,14 @@ impl AdHoc {
     /// use rocket::fairing::AdHoc;
     ///
     /// // The no-op attach fairing.
-    /// let fairing = AdHoc::on_attach("No-Op", |rocket| Ok(rocket));
+    /// let fairing = AdHoc::on_attach("No-Op", |rocket| async move { Ok(rocket) });
```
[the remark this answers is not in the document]
SergioBenitez: I think it serves as good documentation that nothing is being moved into the closure. I wouldn't be surprised if Rust introduced a lint for it at some point, in fact, `unncessary_async_move`.

core/lib/src/fairing/ad_hoc.rs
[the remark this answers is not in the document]
SergioBenitez: Yeah, that's perfect.

SergioBenitez:
> > * I recall us discussing allowing the minimal Rocket application to look like (thing without a `.await`)
> 
> I think this risks too much confusion. These are what I'm aiming for here:
> 
> * Have a single function to document and explain - i.e. no split between `async fn serve()` and `fn launch()`.
> * Have it be `async`, to work better next to other async code (like setting up another server) and for better composability like `select!(rocket.launch(), timeout)` or `join!(rocket.launch(), websocket_server.run())`
> * Be consistent with other functions that return futures, and with usage inside an `async fn` that is not `main` or decorated with an attribute. I think it would be confusing to need to `.await` or not `.await` `launch()` for different reasons than for other futures.

Your last point actually convinces me that we _should_ allow a non-`.launch().await` version _if and only if_ the main function is declared as non-`async`. This is consistent with routes which, when not declared `async`, must not return a future. If your `main` isn't doing anything `async`, there's no reason why you should need to `async fn` it.

The challenge, of course, is that `launch()` needs to have one type signature: I don't think we should employ any magic here. Perhaps the solution is to have `main` return a type and have Rocket do the appropriate thing with it. Perhaps we allow returning a `Rocket` and a `impl Future<Output=Rocket>` (or something similar)?

core/lib/src/catcher.rs
```diff
@@ -153,7 +153,7 @@ macro_rules! default_catchers {
         let mut map = HashMap::new();
 
         $(
-            fn $fn_name<'r>(req: &'r Request<'_>) -> futures_util::future::BoxFuture<'r, response::Result<'r>> {
+            fn $fn_name<'r>(req: &'r Request<'_>) -> futures::future::BoxFuture<'r, response::Result<'r>> {
```
SergioBenitez, line 156: Should this be `$crate::futures::`, to be safe?

core/lib/src/rocket.rs
SergioBenitez: `actualize_manifest`, `execute_manifest`, `consume_manifest`?

core/lib/src/rocket.rs
jebrosen: Done in c94e53d.

core/lib/src/catcher.rs
jebrosen: Since this macro isn't usable in any other module(s), the only conflict would be with another valid path `futures::future::BoxFuture` reachable from `mod defaults` below. IMO `$crate` will look verbose/out of place here with no benefit.

core/lib/src/rocket.rs
jebrosen: Of those three, `actualize_manifest` sounds the closest to what it actually does - I'll go with that one.

jebrosen:
> The challenge, of course, is that launch() needs to have one type signature: I don't think we should employ any magic here. Perhaps the solution is to ask main to return a value and have Rocket do the appropriate thing with it. Perhaps we allow returning a Rocket and a impl Future<Output=Rocket> (or something similar)?

Is this set of examples in line with that proposal?

```rust
// Pulled out of main() for illustrative purposes.
fn rocket() -> Rocket {
    rocket::ignite()
        .attach(AdHoc::on_attach(|rocket| async {
            info!("Look! async!");
            Ok(rocket)
        }))
        .mount("/", rocket::routes![a, b, c])
}

// No setup helpers. launch() is an async fn, so it returns a Future that can be passed to block_on.
// Also notice the semicolon - this example assumes we fix launch() to return
// a type that is easily convertible to Result, is *not* must_use, and automatically logs
// an error if not inspected (like LaunchError in 0.4).
fn main() {
    create_a_runtime().block_on(rocket().launch());
}

// Using tokio::main
#[tokio::main]
async fn main() {
    async_setup().await;
    rocket().launch().await;
}
// This (roughly) expands to:
fn main() {
    create_runtime().block_on(async {
        async_setup().await;
        rocket().launch().await;
    })
}

// Using rocket::main with an async fn
#[rocket::main]
async fn main() -> Rocket {
    async_setup().await;
    rocket()
}

// Using rocket::main with a non-async fn.
#[rocket::main]
fn main() -> Rocket {
    rocket()
}

// These could expand to (roughly):
fn main() {
    create_runtime().block_on(async {
        [async] fn __inner_main() -> Rocket { /* original contents of main() */ }
        let rocket: ::rocket::Rocket = __inner_main()[.await];
        rocket.launch().await;
    })
}
```

The last example is the closest to what you described, and the desugaring is very similar to what is already done for routes to be either `async fn` or non-`async fn`. This API has these notable properties:
* The `#[rocket::main]` style is *extremely* easy for getting started
* There is no way to access the launch result programmatically - it relies entirely on the "print on drop" behavior
* `#[tokio::main]` or custom-constructed `Runtime`s are also easy to use if necessary, by adding `.launch().await`.
  * The future returned by `launch()` can also easily be used in a `join!` or `select!`
* It does **not** satisfy "we *should* allow a non-`.launch().await` version *if and only if* the main function is declared as non-`async`.
  * I did try, and was not super happy with, the version where changing `fn main` to `async fn main` *additionally* required adding `.launch().await`. IMO, that option would complicate both documentation and codegen.

jebrosen:
I think I have addressed or replied to all comments at this point and this is ready for another review. I am finding GitHub's PR interface more difficult to navigate than I remember it, so I might have missed something.

SergioBenitez:
Doing a final review now. I _really_ like your proposal for a `-> Rocket` returning `main`.

> There is no way to access the launch result programmatically - it relies entirely on the "print on drop" behavior

If the function _doesn't_ return a value, I think we should have `#[rocket::main]` mean exactly `#[tokio::main]`. That way, a user need not import `tokio`, and they can still access the launch result programmatically, albeit by needing to add `.launch().await`, which they'd need to do anyway.

contrib/lib/src/databases.rs
```diff
@@ -548,16 +553,17 @@ pub enum ConfigError {
 /// #     .extra("databases", databases)
 /// #     .expect("custom config okay");
 /// #
-/// # rocket::custom(config).attach(AdHoc::on_attach("Testing", |rocket| {
+/// # rocket::custom(config).attach(AdHoc::on_attach("Testing", |mut rocket| async {
 /// # {
-/// let config = database_config("my_db", rocket.config()).unwrap();
+/// let manifest = rocket.inspect().await;
```
SergioBenitez, line 558: Can all of these use `rocket.config().await` now?

core/lib/benches/simple-routing.rs
```diff
@@ -109,7 +109,7 @@ mod benches {
 
         // Hold all of the requests we're going to make during the benchmark.
         let mut requests = vec![];
-        for route in client.rocket().routes() {
+        for route in client.manifest().routes() {
```
SergioBenitez, line 112: I apparently can't comment on lines that haven't been modified, but note that this no longer compiles due to a missing `.await` on `Client::new()`. Can we even do async benches?

examples/form_kitchen_sink/src/tests.rs
```diff
@@ -14,31 +14,29 @@ impl fmt::Display for FormOption {
...
+    let mut res = client.post("/")
+        .header(ContentType::Form)
+        .body(form_str)
+        .dispatch().await;
+
+    assert_eq!(res.body_string().await, Some(expected));
 }
 
-fn assert_valid_form(client: &Client, input: &FormInput<'_>) {
+async fn assert_valid_form(client: &Client, input: &FormInput<'_>) {
```
SergioBenitez, line 26: It would be nice to make this and the other assert macros so that we don't have to call `.await` every time, instead calling it in the macro itself.

contrib/codegen/src/database.rs
```diff
@@ -118,8 +118,8 @@ pub fn database_attr(attr: TokenStream, input: TokenStream) -> Result<TokenStrea
             /// Retrieves a connection of type `Self` from the `rocket`
             /// instance. Returns `Some` as long as `Self::fairing()` has been
             /// attached and there is at least one connection in the pool.
-            pub fn get_one(rocket: &::rocket::Rocket) -> Option<Self> {
-                rocket.state::<#pool_type>()
+            pub fn get_one(manifest: &::rocket::Manifest) -> Option<Self> {
+                manifest.state::<#pool_type>()
```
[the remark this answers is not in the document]
jebrosen: Actually, taking an `&mut Rocket` should be (mostly) fine:

```rust
let mut rocket = make_rocket();
let conn = rocket.get_one().await;
let config = rocket.config().await; // <-- The code I was afraid would not work, but is actually fine.
```

Since `get_one()` returns owned data, there is no conflict between calling `get_one` and other methods.

However, I see at least one place we would "lose" something: `on_launch` will no longer be able to call `get_one`. And `on_launch` shouldn't change to take `&mut Rocket`, because by that point it is too late to make many of the modifications that `&mut Rocket` allows.

core/lib/benches/simple-routing.rs
jebrosen: Ah right, I keep forgetting about the benches (and the test script doesn't include them). In one sense the answer is "yes, of course" - just write `block_on` somewhere. Doing it in such a way that the benchmarks actually measure what is supposed to be measured is the hard part - as I believe we have discussed, I expect these style of benchmarks will amplify the per-request overhead of `async` and fail to demonstrate the overall advantage provided by work-stealing.

How do you feel about deferring these fixes until an upcoming PR for a blocking API for `Client`, so we can compare some approaches?

examples/form_kitchen_sink/src/tests.rs
jebrosen: Very possible - https://paste.rs/9HQ.diff. However we do lose the documentation on what the types are supposed to be, and I think a blocking `Client` will end up being the nicer way to simplify this.

core/lib/benches/simple-routing.rs
SergioBenitez: Sounds good to me!

examples/form_kitchen_sink/src/tests.rs
SergioBenitez: I'm okay with either approach. This should _probably_ be a macro, however, so test failures point to the right line.

contrib/codegen/src/database.rs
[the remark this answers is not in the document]
SergioBenitez: That's a good point.

I suppose one idea, perhaps a bad one, is to use some form of interior mutability - say a reentrant lock - to make `inspect()` take an `&Rocket` instead of an `&mut`. I believe if we did that, however, we wouldn't need to expose `Manifest` at all, which is kind of interesting. In some ways, this feels right as `.inspect()` is logically read-only, and only takes an `&mut` due to its implementation.

What do you think about this? Are we gaining anything by separating `Rocket` and `Manifest`?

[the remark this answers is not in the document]
jebrosen: > say a reentrant lock

I think I can see it in principle, but it has some *really* awkward consequences:

```rust
let conf = rocket.config().await;                          // so far so good
let rocket = rocket.attach(AdHoc::on_attach(|rocket|
    Ok(rocket::ignite().manage(X))
));                                                        // this attach fairing is "queued"
let state = rocket.state::<X>().await;                     // this must run the attach fairing

conf.get_something();                                      // oops
```

Effectively, `inspect()` or functions that call it would now have to panic if outstanding borrows exist - this also includes calls made from `on_launch()`.

Issue #228, opened 2017-03-15
[Feature Request] Make `RocketConfig::parse` public


lawliet89:
... or something similar for the `rocket::config::Config` struct like:

 - Deriving/implementing `Deserialize` for `rocket::config::Config` 
 - Moving the parsing method to `rocket::config::Config` and making it public

## Why you believe this feature is necessary and use case

I am trying to implement a Rocket application with additional configuration options. Currently, the `Config` struct only supports a simple key-value store for extra configuration options. That is fine.

As an alternative, I am implementing a separate configuration structure that will be deserialized from another JSON/TOML file. 

I would like to allow users to specify the filename for this additional `JSON/TOML` file as one of the extra configuration options in `Rocket.toml`. This file can be, and should probably be, different for the three rocket environments.

Then, I would like to parse `Rocket.toml` with the same rules that Rocket does, and extract this extra value BEFORE I `launch` Rocket so that I can deserialize the JSON/TOML file and then add the struct into Rocket as to be `manage`d. 

This is not really do-able at the moment, or even to implement for myself because  `rocket::config::Config`  does not derive `Deserialize`. Alternative, I would have to essentially duplicate quite a bit of code to achieve this.

SergioBenitez:
I'm not sure which form this will come in, but I believe something like this should exist.

Slating for 0.4.

LukasKalbertodt:
To give another use case example:

In my **build script**, I compile SCSS files to CSS files. Depending on the Cargo profile (debug/release) I choose whether or not to minify the resulting CSS. It would be awesome to depend on the Rocket environment instead (or additionally). Something like "minify CSS in production". I imagine something like this:

```rust
// Creates the config just as rocket would normally do: `Rocket.toml`, env vars, ...
let rocket_config = rocket::Config::from_env();

if env::var("PROFILE").unwrap() == "release" && rocket_config.environment.is_prod() {
    // minify CSS
}
```

new-commer:
This issue seems has not been resolved in 0,4 , and i think this  issue  is necessary, just change  `parse` to public . We looking forward  it。

SergioBenitez:
Let's fix this in #852.

PR #1244, opened 2020-02-27
Add `RocketConfig::read_from()` that takes a file path

threecgreen:
This commit adds an alternative to `RocketConfig::read()`, that takes a path to the configuration file to read and parse. Being able to specify a configuration path programmatically gives users more flexibility on where they store configurations and in determining which file to read from.

Taking a single argument `path` seemed like a better interface than just [making `RocketConfig::parse` public](https://github.com/SergioBenitez/Rocket/issues/228).

### DIFF
--- a/core/lib/src/config/config.rs
+++ b/core/lib/src/config/config.rs
@@ -182,7 +182,7 @@ impl Config {
     }
 
     /// Returns the default configuration for the environment `env` given that
-    /// the configuration was stored at `config_file_path`.
+    /// the configuration was stored at `path`.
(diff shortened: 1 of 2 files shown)


SergioBenitez:
`RocketConfig` is not a public API, so this wouldn't affect any end-user. Nevertheless, I agree we should expose functionality similar to this, and I'm working on a PR that does just that.

SergioBenitez:
Added in a4301c0a9a407ca8a7c85260d045ca199f4a7254.

PR #1267, opened 2020-04-05
Implement UriDisplay and FromUriParam for Segments.

mplzik:
`Segments` were missing an implementation of UriDisplay and
FromUriParam, making it impossible to use `uri!` in cases where a route
handler had a `Segments argument.

Signed-off-by: Milan Plzik <milan.plzik@gmail.com>

### DIFF
--- a/core/http/src/uri/from_uri_param.rs
+++ b/core/http/src/uri/from_uri_param.rs
@@ -1,7 +1,7 @@
 use std::path::{Path, PathBuf};
 
 use RawStr;
-use uri::{self, UriPart, UriDisplay};
+use uri::{self, segments, UriPart, UriDisplay};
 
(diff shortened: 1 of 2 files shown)


mplzik:
Note: This uses unwrap() in FromUriParams implementation for PathBuf source; unfortunately I don't know how to deal with this (except for dropping this particular implementation)

SergioBenitez:
This seems plausible, though `master` has significantly diverged. My apologies for not getting to this.

I would be in favor of an implementation for the new `Segments`, but it would likely require `Segments` to additionally be aware of only the raw segments instead of the decoded version as it is now. Otherwise, we would need to _re_-decode to in the `UriDisplay` implementation. If you or anyone else would like to send a PR for review doing as such, I'd be happy to review. Closing this for now.

PR #1271, opened 2020-04-09
Mutual TLS support

MikailBag:
This is #657 (by @Kmoneal) rebased on master.
Currently, it compiles and passed `cargo test`.
As a next step, I will apply all suggestions from reviews on that PR.


MikailBag:
I don't know why mtls example is failing on Windows. It doesn't seem to contain platforms-specific code.

MikailBag:
Taking look on https://github.com/SergioBenitez/Rocket/blob/c0a0e46d7e9b007ce60d2e2ba818990771e3a0d8/core/lib/src/request/from_request.rs#L478
I am not sure this code is secure.
It seems to me that only validation in that code is [verify_is_valid_for_dns_name](https://github.com/SergioBenitez/Rocket/blob/c0a0e46d7e9b007ce60d2e2ba818990771e3a0d8/core/http/src/tls.rs#L15). According to [webpki docs](https://docs.rs/webpki/0.21.2/webpki/struct.EndEntityCert.html), `verify_is_valid_tls_client_cert` and `verify_signature` are required too.

It is also possible I am wrong, and `verify_is_valid_for_dns_name` is enough, so I would like to get some clarification on this.

Kmoneal:
I will need to look into this again. I don't see it hurting by adding the signature verification even if it is not needed.

Just a clarification of what was the original roadblock since I don't think this was every put on github. There was originally a dilemma of the cert list being used and not having a specific source of what DNS name the other server was providing. I don't know now, but that was a problem stemming from the library for certificates not specifying the DNS name and instead sending through a list of names provided in the certificate.

I have very limited time right now but would like to help as best I can.

SergioBenitez:
Closing due to inactivity, but I'd still love to see this implemented.

Issue #1224, opened 2020-02-08
Rocket should show 405 instead of 404 for method not allowed


silvioprog:
Hi.

Consider the following route:

```rust
#[post("/hello")]
pub fn hello() -> &'static str {
    "Hello"
}
```

now, if you call it with:

```bash
curl http://localhost:8000/hello
```

it returns:

```
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8" />
    <title>404 Not Found</title>
  </head>
  <body align="center">
    <div align="center">
      <h1>404: Not Found</h1>
      <p>The requested resource could not be found.</p>
      <hr />
      <small>Rocket</small>
    </div>
  </body>
</html>
```

however, since the passed URL is correct and only the method (verb) is wrong, according to [rfc7231#section-6.5.5](https://tools.ietf.org/html/rfc7231#section-6.5.5), it should return `405: Method Not Allowed` with an `Allow` header field in a `405` response containing a list of the target resource's currently supported methods.

So, if you don't agree with that, could you provide a global option to send `405` instead of `404` for handling `method not allowed` errors?

TIA

jebrosen:
From https://tools.ietf.org/html/rfc7231#section-4.1:

> When a request method is received that is known by an origin
   server but not allowed for the target resource, the origin server
   SHOULD respond with the 405 (Method Not Allowed) status code.

I sounds like a good idea to default to 405 instead of 404, then - and applications that want a 404 can add a 405 `catch`er that returns a 404.

bharathk005:
@SergioBenitez does #2616 fix this or do we still need a solution?

rubbieKelvin:
why's this still open?

igmrrf:
> why's this still open? it's been 5 years 😢

Sadly the PR is still open.

reneleonhardt:
> > why's this still open? it's been 5 years 😢
> 
> Sadly the PR is still open.

Sadly Rocket saw no release for a year... @silvioprog could you checkout, build and test https://github.com/rwf2/Rocket/pull/2616? 😅

@jespersm who can review and merge your branch, is it ready? 🙂

reneleonhardt:
@SergioBenitez could you review https://github.com/rwf2/Rocket/pull/2616?
When will Rocket be ready for a new release? 🚀

PR #1276, opened 2020-04-14
Implementing HTTP 405 if HTTP method doesn't match with required by the route

Drunpy:
This PR purpose is to meet [RFC 7231 sec6.6.5](https://tools.ietf.org/html/rfc7231#section-6.5.5) and #1224

### DIFF
--- a/core/codegen/tests/route-format.rs
+++ b/core/codegen/tests/route-format.rs
@@ -63,7 +63,7 @@ fn test_formats() {
     assert_eq!(response.body_string().unwrap(), "plain");
 
     let response = client.put("/").header(ContentType::HTML).dispatch();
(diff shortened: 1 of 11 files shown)


SergioBenitez:
It looks like the tests are failing due to a UI mismatch almost certainly caused by differing versions of rustc. Check that your nightly compiler is up to date as this is what the CI will use. Once it is, you should get the same failures locally.

jebrosen:
CI has been failing on `master` (probably for a while, sorry) and I only got around to fixing it yesterday.

@Drunpy, if you rebase your branch on top of the latest `master` the failing `ui-tests` should start passing again.

Drunpy:
Ready for review :heavy_check_mark:  @jebrosen

Issue #1279, opened 2020-04-14
Updating documentation under the overview markdown by removing unnecessary `#`


0x6f736f646f:
## Bug Reports

**Describe the bug**
When going throught the documentation under the overview markdown file there are some included `#` that makes the code to have comments hence not running

**To Reproduce**
1. Go to 'site/guide/3-overview.md'
2. Open it
3. Scroll down to code snippets
4. See errors

![Screenshot from 2020-04-15 01-32-52](https://user-images.githubusercontent.com/28790446/79281338-32885100-7ebb-11ea-9914-630730788d06.png)
![Screenshot from 2020-04-15 01-32-56](https://user-images.githubusercontent.com/28790446/79281357-387e3200-7ebb-11ea-9e0d-bb096e7c2117.png)
![Screenshot from 2020-04-15 01-33-02](https://user-images.githubusercontent.com/28790446/79281372-3e741300-7ebb-11ea-810b-59cfdd1dbf35.png)
![Screenshot from 2020-04-15 01-33-07](https://user-images.githubusercontent.com/28790446/79281383-45028a80-7ebb-11ea-82b7-0c59b9a5d9b2.png)
![Screenshot from 2020-04-15 01-33-14](https://user-images.githubusercontent.com/28790446/79281392-4b910200-7ebb-11ea-8ca2-d8768b5f4c82.png)

**Expected behavior**
The `#` shouldn't be there

![Screenshot from 2020-04-15 01-33-32](https://user-images.githubusercontent.com/28790446/79281462-6f544800-7ebb-11ea-83ec-d972a0efd349.png)
![Screenshot from 2020-04-15 01-33-35](https://user-images.githubusercontent.com/28790446/79281471-74b19280-7ebb-11ea-95f5-862fde521f39.png)
![Screenshot from 2020-04-15 01-33-38](https://user-images.githubusercontent.com/28790446/79281482-79764680-7ebb-11ea-8313-1eb298a35d52.png)
![Screenshot from 2020-04-15 01-33-46](https://user-images.githubusercontent.com/28790446/79281484-7b400a00-7ebb-11ea-96ed-fbef1744035a.png)

**Desktop (please complete the following information):**
 - OS: [e.g. Windows 10]
 - Browser [e.g. chrome, firefox]
 - Rocket Version [e.g. 0.4.4]

**Additional context**
No

0x6f736f646f:
I have worked on it and submitted a pull request

[...]

PR #1280, opened 2020-04-14
Removing unnecessary comments which are on code blocks

0x6f736f646f:
When going throught the documentation under the overview markdown file there are some included `#` that makes the code to have comments hence not running #1279


jebrosen:
This is a recent change (95c981d) that is working as intended: the "commented-out lines" are hidden by default in the guide, and the `if false` is used to only compile the code without running it. This even causes a CI failure, because one of the tests starts an HTTP server and never finishes!

PR #1311, opened 2020-05-26
Removed warnings 

komi1230:
When building example app (`hello_world`),  Nightly Rust warns in some lines.

I fixed these following cargo's guides.

You can see some these warnings below:

```
warning: use of deprecated item 'time::offset_date_time::OffsetDateTime::now': This function returns a value with an offset of UTC, which is not apparent from its name alone. You should use `OffsetDateTime::now_utc()` instead.
   --> core/http/src/cookies.rs:376:32
    |
376 |             cookie.set_expires(time::OffsetDateTime::now() + time::Duration::weeks(1));
    |                                ^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(deprecated)]` on by default

warning: 1 warning emitted

warning: `extern crate` is not idiomatic in the new edition
  --> core/codegen/src/lib.rs:62:1
   |
62 | extern crate proc_macro;
   | ^^^^^^^^^^^^^^^^^^^^^^^^ help: convert it to a `use`
   |
note: the lint level is defined here
  --> core/codegen/src/lib.rs:8:9
   |
8  | #![warn(rust_2018_idioms)]
   |         ^^^^^^^^^^^^^^^^
   = note: `#[warn(unused_extern_crates)]` implied by `#[warn(rust_2018_idioms)]`

warning: 1 warning emitted

warning: use of deprecated item 'time::offset_date_time::OffsetDateTime::now': This function returns a value with an offset of UTC, which is not apparent from its name alone. You should use `OffsetDateTime::now_utc()` instead.
   --> core/lib/src/local/request.rs:417:32
    |
417 |             let current_time = time::OffsetDateTime::now();
    |                                ^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(deprecated)]` on by default

warning: 1 warning emitted
```


core/codegen/src/lib.rs
```diff
@@ -59,7 +59,7 @@
 //! ```
 
 #[macro_use] extern crate quote;
-extern crate proc_macro;
+use proc_macro;
```
jebrosen, line 62: This change is not strictly necessary (since `extern crate` will continue to be allowed at least in the 2018 edition), but it *does* increase the minimum nightly `rustc`. For that reason I think it should be done in a separate commit if at all - for reference, it was done on the async branch in f0c3177 and ca65d88.

core/http/src/cookies.rs
```diff
@@ -373,7 +373,7 @@ impl Cookies<'_> {
         }
 
         if cookie.expires().is_none() {
-            cookie.set_expires(time::OffsetDateTime::now() + time::Duration::weeks(1));
+            cookie.set_expires(time::OffsetDateTime::now_utc() + time::Duration::weeks(1));
```
jebrosen, line 376: We should bump our time dependency in `Cargo.toml` to `0.2.11` along with this change: <https://github.com/time-rs/time/blob/master/CHANGELOG.md#0211-2020-04-27>

core/codegen/src/lib.rs
komi1230: Thanks for the explanation and the kind advice !
I've split commit.

core/http/src/cookies.rs
komi1230: Here is also done. Thanks a lot.

jebrosen:
I've merged the `time` change as a21221a, and left out the `extern crate` change for now.

PR #1313, opened 2020-05-28
Adding implementation of FromFormValue for Json type

bcecchinato:
This update is intended to use the ```Json``` type for request query parameters. One example could be this : ```GET /example?values=[0,1,2]``` which would be bound to the parameter ```Json<Vec<u8>>```.

On more complex cases it would be possible then to pass a full JSON object in a query param (not recommended but yet possible).

### DIFF
--- a/contrib/lib/src/json.rs
+++ b/contrib/lib/src/json.rs
@@ -18,23 +18,23 @@ use std::ops::{Deref, DerefMut};
 use std::io;
 use std::iter::FromIterator;
 
-use rocket::request::Request;
+use rocket::request::{Request, FromFormValue};
 use rocket::outcome::Outcome::*;
 use rocket::data::{Data, ByteUnit, Transform::*, Transformed};
 use rocket::data::{FromTransformedData, TransformFuture, FromDataFuture};
-use rocket::http::Status;
+use rocket::http::{Status, RawStr};
 use rocket::response::{self, Responder, content};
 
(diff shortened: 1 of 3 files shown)


jebrosen:
This looks correct, but before merging we should

1. Have tests for this everywhere it can be used (inside a form, as a query parameter, etc.) and
2. Be sure it makes sense to support `Json<T>` in all those places.

bcecchinato:
Yeah I agree, but I didn't found any tests on the Json type :laughing: Is there a reason on why there wasn't any ?

SergioBenitez:
@bcecchinato The `Json` type is tested via both doc-tests and the `json` example.

bcecchinato:
Thanks i'll try finish this by the end of week :)

bcecchinato:
I finally got the time to complete the tests. It's simple but should be enough. @SergioBenitez and @jebrosen i'll let you check if this is good for you, or if I need to add something else.

SergioBenitez:
This is neat, and looks good to me. Unfortunately, I'm about to break the entire implementation by pushing a completely revamped `Form` system. If it's okay with you, @bcecchinato, I'll take your changes here and rebase them myself, attributing the `Json` impl of `FromForm` to you.

bcecchinato:
Sure @SergioBenitez :)

bcecchinato:
Closig since this will be rebase outside of this PR :)

Issue #297, opened 2017-05-20
[Feature Request] Rocket auto-documentation route


TheDan64:
Hi, thanks for all your hard work. Rocket is awesome!

I think it would be an absolute killer feature if route/endpoint API documentation could somehow be generated right from Rocket. For example, in Python, Sphinx has the capability to pull routes' docstrings right from the endpoints and build HTML document(s) that get mounted to /docs and formats the docstrings via rst templates(for added customization). However, these docstrings are managed by the developers, so I'm wondering if there's the possibility for generating all (or almost all - maybe allowing for additional comments or descriptions of failure cases, etc) of the information straight from Rocket types. 

So I guess I'm imagining something like rustdoc, where the primary difference is a focus on routes rather than rust methods. So for example, the Rocket route:
```rust
/// This endpoint gets a user by their id
#[get("/user/<id>?<bar>")]
fn get_user(id: u64, bar: Option<String>) -> JSON {
    ...
}
```

Could be auto documented to something like the following, though prettier in HTML&CSS:
```
GET /user/<id>?<bar> -> JSON<User>
Path variables:
    id: int

Params:
    bar: string (optional)

This endpoint gets a user by their id
```
And maybe *User* could link to a definitions page/or have a hover display type of thing for reference:

```
User: {
    id: int,
    name: string,
}
```

I'm thinking enabling the autodoc would be easy to set up:

`rocket::ignite().mount("/", routes![get_user]).docs("/docs").launch();`

Or with a more rocket-themed name! (ground_control?!)

------

So, to respond to the required information more directly:

I think this feature is necessary because it would be a huge boon to developers. One team could quickly mock up a couple endpoint and data model (ie User) definitions and then point other consuming team(s) to their documentation almost immediately. 

I suppose this feature could exist outside of rocket as some sort of plugin, but I think auto documentation would be a common use case so it might make sense for it to be easily accessible right away when developing a Rocket app.

corbinu:
I think this is a great idea but this would be better served as a plugin. For example I want to have the API documented as swagger, but others might want another format. 

Currently in my node projects I use this: https://github.com/glennjones/hapi-swagger which I think is a great example of how this can be done super well.

bluejekyll:
FYI: I just played around with generating a quick Swagger/OpenAPI Spec: https://users.rust-lang.org/t/anyone-working-on-swagger-openapi-support-in-rocket/11231

TheDan64:
[...]

PR #1317, opened 2020-05-31
Add `#[doc(hidden)]` to more generated functions and structs.

jebrosen:
This PR hides items with names such as `static_rocket_route_info_for_index` from documentation. I *believe* this covers all generated top-level items that were not already `#[doc(hidden)]`.

It might be nicer to provide some sort of documentation (#297), but I think hiding it is preferable until something like that exists.

Hidden items can still be documented with the following command: `RUSTDOCFLAGS='-Z unstable-options --document-hidden-items' cargo doc [--open]`. It is expected that that command will be less cumbersome in the future (rust-lang/cargo#7766).


SergioBenitez:
Merged in 91fb0804bead41f723a72136c39723434a290b62.

PR #1318, opened 2020-05-31
[easy] remove tls warning in configuration guide

klnusbaum:
It looks like this warning about using TLS is stale. It was added 2 years ago according to the git blame. As far as I can tell, looking at the open github issues, there's currently nothing majorly wrong being tracked for TLS support. So it seemed reasonable to remove this somewhat scary warning. It's also possible I missed something and there is still TLS work that needs to be done.

### DIFF
--- a/site/guide/9-configuration.md
+++ b/site/guide/9-configuration.md
@@ -289,9 +289,6 @@ being ignored.
 
 ## Configuring TLS
 
-! warning: Rocket's built-in TLS is **not** considered ready for production use.
-  It is intended for development use _only_.
-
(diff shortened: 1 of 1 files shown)


SergioBenitez:
This still applies. As far as I know, `rustls` has not undergone any kind of security audit, nor is it clear where it stands performance wise. Until we have clear data on both of these points, the notice should and will remain. It is exactly intended to be scary! We _only_ want you to use the built-in TLS support during development and nothing more.

klnusbaum:
Rust TLS just got a formal audit. Do you think it's ok to remove this warning now?

https://github.com/ctz/rustls/blob/master/audit/TLS-01-report.pdf

PR #1324, opened 2020-06-03
Adding support for accepting Date / Time values in rocket_contrib

phayes:
This work-in-progress PR adds support for accepting ISO-8601 formatted strings as Params and Form Values in Rocket (rocket_contrib).  

Right now this PR only adds support for `chrono::NaiveDate` and `chrono::NaiveDateTime`, but if the maintainers think this is a good direction, I'll add support for `Date` and `DateTime`, which would mean that Rocket would have support for the all the major Date/Time formats.  

We could also talk about adding support for non-ISO8601 datetime formats such as RFC3339 or RFC2616, but that depends on how "batteries-included" we want to be here.  ISO-8601 is definitely the basic-one that should cover most use-cases.

### DIFF
--- a/contrib/lib/Cargo.toml
+++ b/contrib/lib/Cargo.toml
@@ -27,6 +27,7 @@ serve = []
 compression = ["brotli_compression", "gzip_compression"]
 brotli_compression = ["brotli"]
 gzip_compression = ["flate2"]
+datetime = ["chrono"]
 
 # The barage of user-facing database features.
 diesel_sqlite_pool = ["databases", "diesel/sqlite", "diesel/r2d2"]
@@ -59,6 +60,9 @@ notify = { version = "4.0.6" }
 
 # UUID dependencies.
 uuid = { version = ">=0.7.0, <0.9.0", optional = true }
+
+# Datetime dependencies.
+chrono = { version = "0.4.1", optional = true }
 
 # Database dependencies
 diesel = { version = "1.0", default-features = false, optional = true }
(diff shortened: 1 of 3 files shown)


jebrosen:
Dates are scary!

I do agree that some minimal date/time support should be provided in `rocket_contrib`, or possibly/eventually in `rocket`. At a minimum, I think it makes sense to have `FromFormValue`. I think we can support these, which all seem pretty straightforward:

* `<input type="date">`, defined as `chrono`'s format `%Y-%m-%d`, as a `chrono::NaiveDate`
* `<input type="time">`, defined as `chrono`'s format `%H:%M` or `%H:%M:%S`, as `chrono::NaiveTime`
* possibly `<input type="datetime-local">`, defined as `chrono`'s format `%Y-%m-%dT%H:%M`, as `chrono::NaiveDateTime`

It should definitely be stressed in documentation that these are timezone-less values and need to be handled accordingly.

I haven't thought as hard about `FromParam`, but it definitely seems less clear-cut to me. It seems like a nice use for const generics some day, but that day is not today:

```rust
#[get("/orders/<date>")]
fn list_orders(date: DateTime<"%d-%m-%Y">) {
}
```

SergioBenitez:
Note that we've effectively committed to using `time` instead of `chrono`. It would be ideal to use only one time crate in `rocket`.

migeruj:
Does this work with JSON?

jebrosen:
@migeruj This PR would not do anything for JSON. JSON doesn't even have direct date support (JS serializes them as strings, for example), and `rocket_contrib::json::Json` delegates entirely to `serde` and `serde_json` anyway. If you have JSON fields that are known to be dates, you can use date/time types from `time`, `chrono`, or another library with `serde` support, or an approach like <https://serde.rs/custom-date-format.html>.

SergioBenitez:
Checking in: what's the status?

SergioBenitez:
Rocket's form support includes support for dates and times now. We don't have `FromParam` impls for these, but we can easily add them.

PR #1255, opened 2020-03-08
Remove [target.'cfg(debug_assertions)'.dependencies] in contrib Cargo.toml

RalfJung:
This is a PR **against the v0.4 branch** to backport 4151cd46db9a871bd2088fbecfea145582edca9e as a fix for https://github.com/SergioBenitez/Rocket/issues/1251.

Cargo used to always add that dependency anyway, so this changes nothing for older versions of cargo, and fixes the build for newer versions.

PR #1328, opened 2020-06-06
Make 'notify' an optional dependency.

### LINKED ISSUE #1058: target.'cfg(debug_assertions)'.dependencies doesn't work
https://github.com/SergioBenitez/Rocket/blob/v0.4/contrib/lib/Cargo.toml#L79 contains a section `[target.'cfg(debug_assertions)'.dependencies]`. This doesn't work as intended: cargo always resolves the target cfgs by running `rustc --print=cfg`, regardless of which profile has been selected in the cargo command line, and regardless of which profile options have been set in `Cargo.toml`.

jebrosen:
It is now required only when the 'templates' feature is active.

Closes #1058.

* The original intent at the time (#537) was to have this dependency only when both `debug_assertions` and the `"templates"` feature were active. At the time I was able to determine that both conditions were not possible, so we prioritized `debug_assertions`.
* That never actually worked. It became a warning and later an error from `cargo`, and it was "upgraded" to a regular dependency in #1255.
* Since we can't condition it on `debug_assertions` anyway, we will now condition the dependency on the `"templates"` feature. See also https://github.com/SergioBenitez/Rocket/issues/1058#issuecomment-639896706.

### DIFF
--- a/contrib/lib/Cargo.toml
+++ b/contrib/lib/Cargo.toml
@@ -13,7 +13,7 @@ edition = "2018"
 
 [features]
 # Internal use only.
-templates = ["serde", "serde_json", "glob"]
+templates = ["serde", "serde_json", "glob", "notify"]
 databases = ["r2d2", "rocket_contrib_codegen/database_attribute"]
 
 # User-facing features.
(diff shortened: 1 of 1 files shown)


SergioBenitez:
Awesome. Merged in c19365dcbebce47a02c0cfa1de77fd278c4adb4e.

PR #1331, opened 2020-06-07
Update README.md

teymour-aldridge:


SergioBenitez:
Unfortunately, I don't quite see the advantage to this wording change. Thank you for the suggestion, however!

PR #1334, opened 2020-06-10
Relax lifetime bound for `&Request` in `respond_to` (async branch)

jebrosen:
This was reported in IRC: <https://freenode.logbot.info/rocket/20200525#c3970295> - see also GREsau/okapi#29.

Currently, `respond_to` is too restrictive: it requires an `&'r Request`, where `'r` is from the trait - so if you have an `R: Responder<'static>` (i.e. an owned `Responder`), you can only `respond_to` an `&'static Request`. This PR changes that.

Before merging, I would also want to find more concrete documentation of using `'async_trait` in this position. So far, the closest I have found is https://github.com/dtolnay/async-trait/issues/8#issuecomment-514812245.

### DIFF
--- a/core/lib/src/response/content.rs
+++ b/core/lib/src/response/content.rs
@@ -49,7 +49,7 @@ pub struct Content<R>(pub ContentType, pub R);
 #[crate::async_trait]
 impl<'r, R: Responder<'r> + Send + 'r> Responder<'r> for Content<R> {
     #[inline(always)]
-    async fn respond_to(self, req: &'r Request<'_>) -> response::Result<'r> {
+    async fn respond_to(self, req: &Request<'_>) -> response::Result<'r> where 'r: 'async_trait {
         Response::build()
             .merge(self.1.respond_to(req).await?)
             .header(self.0)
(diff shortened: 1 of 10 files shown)


SergioBenitez:
This is...weird. A magical lifetime that appears out of nowhere! I'm beginning to wonder if we should have our own `#[async_trait]` that's aware of trait names and "does the right thing" with lifetimes. Or perhaps using `#[async_trait]` is simply not a good idea, given just how many workaround we need.

jebrosen:
> Or perhaps using `#[async_trait]` is simply not a good idea, given just how many workaround we need.

I think that was my original conclusion at least for this trait, which is in the unique position of having a lifetime that applies to `self` and the return value but no other parameters. I was excited to see when you applied it to `Responder`, thinking the problems I ran into had been resolved - but it looks like those issues were only moved somewhere else in this case. For what it's worth, `rustc` might in principle be able to do this without the additional help; I believe `#[async_trait]` truly needs it here because the generated code for the `impl` block can't depend in any way on the original `trait` definition.

As far as I can tell the same concerns do not apply to `Fairing` and `FromRequest`, so it is still a net win to use `#[async_trait]` for those.

SergioBenitez:
Update: we've decided to make `Responder` synchronous with the rationale that:
  1. Constructors for types that implement `Responder` constructors can be `async` if they need be.
  2. We can make `set_sized_body` synchronous by differing the `Seek` operation to just before Rocket writes out the response.

I'll be assuming responsibility for these changes. I'll keep this PR open until I've pushed my own.

SergioBenitez:
Changes merged in https://github.com/SergioBenitez/Rocket/commit/4f792551d300e45738ddb2cfcd044b0f00379a40.

PR #1335, opened 2020-06-11
Implementing HTTP 405

Drunpy:
This PR is the same as #1276 but without formatting changes. Solves #1224


SergioBenitez:
Please see this part of my previous comment:

> My concern is the performance impact that this will have. It's very nice to be able to filter by method: it means we use it as a prefix. We _should not_ remove this from the router. At worst, we should simply iterate through all possible methods, checking to see if there's a match, in case no route was found. This means we take the performance hit _only if_ there's no match initially.

Drunpy:
Sure, I will implement this way.

Drunpy:
Hey @igalic thanks for your valuable comments, I've done some modifications based on these, thanks!

SergioBenitez:
@Drunpy Note that to get this PR accepted, in addition to the changes requested, you'll need to rebase on master, not merge it in.

SergioBenitez:
I'm still in favor of this, but simply closing due to inactivity.

PR #1340, opened 2020-06-13
Async: CI fixes

jebrosen:
This is a "forward-port" of a few commits from master that are still causing CI failures in `async`:

* f4bb8bb: We use the `nightly` feature of `parking_lot`, and something unrelated to our nightly usage broke in the older version. There is a [non-`nightly` function](https://docs.rs/parking_lot/0.10.2/parking_lot/fn.const_mutex.html) we can use now instead, but I don't want to introduce more conflicts between async and master than necessary
* 1ae3e68: 64-bit rustup installation
* 3abafaae: UI tests changed

If CI passes, I intend to merge directly.


jebrosen:
Merged as 74a70ee, f5ac9d35, bf76361e.

PR #1342, opened 2020-06-14
Async: on_attach fairings

jebrosen:
Supersedes #1242 - opening as a new PR so I can do a thorough review on a clean slate.
Reviews and comments on that PR have already been addressed and included; this is identical plus/minus commit reorganization and a few docs/strings.


core/lib/src/local/client.rs
```diff
@@ -114,11 +117,13 @@ impl Client {
     /// ```rust
     /// use rocket::local::Client;
     ///
-    /// let client = Client::new(rocket::ignite()).expect("valid rocket");
+    /// # let _ = async {
```
jebrosen, line 120: I think this pattern in some of these tests - checking for compilation only and not running any code - is my only misgiving. I did make a conscious effort to only do this when there were no `assert!`s made in the test, but it could still somewhat reduce coverage. @SergioBenitez would you prefer for all of these to use `rocket::async_test` instead of `let _ =`, to ensure the code actually runs as well?

SergioBenitez: Sure, if it's an easy change, why not?

PR #1343, opened 2020-06-15
async: Replace `#[database]` guard API with a `spawn_blocking` wrapper.

jebrosen:
The connection guard type generated by `#[database]` no longer implements `Deref` and `DerefMut`. Instead, it provides an `async fn run()` that gives access to the underlying connection on a closure run through `spawn_blocking()`.

## Background

All existing engines supported by `#[database]` are synchronous. Several of them, such as `diesel` and `sqlite`, do not have a well-supported `async` version or close alternative available. Therefore, I would consider it a non-starter to entirely drop support for them. However, using synchronous database libraries from `async` code has some issues:

## Motivation

Performing blocking operations while running on an asynchronous runtime is a footgun:
* Best case: the server does not experience enough load to exhibit any problems.
* Bad case: requests that are busy in a blocking operation will "artificially" prevent other requests from running or continuing, including any that are partway through I/O with the client, even if those other requests don't need to perform any blocking operation themselves.
* Worst case: deadlocks can occur. With database pools in particular, you could end up in a situation where all threads are waiting for a connection to be returned to the pool - but those connections are all held by tasks that cannot run, because the worker threads are all tied up waiting.

## Problems and Alternatives

* ~~`spawn_blocking` [panics when the blocking thread limit (currently 512) is reached](https://github.com/tokio-rs/tokio/issues/2309). In other words, there is no "queue" for spawned operations.~~ Apologies; I misread/misunderstood parts of the implementation of `spawn_blocking` *and* of the linked issue. There *is* a queue for spawned tasks, which is actually worse in a sense:
* `spawn_blocking` without any backpressure creates a ridiculously long queue of threads, which most of the time are just waiting for an open connection to be returned to the pool. These threads prevent the "return to the pool" operation from happening, which would cause a deadlock if it were not for r2d2's built-in timeouts.
  * A lightweight option that shows a lot of promise in my testing: put the connection acquisition operation inside behind an async `Semaphore` or `Mutex`. Limiting on acquisition should be sufficient, since the pool already limits the number of open connections - thereby limiting the number of active `spawn_blocking()` calls by construction.
  * Instead of `spawn_blocking`, we could use a work queue specifically for databases. This poses all kinds of new questions, like whether workers should be per-guard type or global and how many worker threads.
* It's inconvenient to do any asynchronous work inside the closure, so depending on the usage patterns `run()` might need to be called multiple times in between non-database work.
* `spawn_blocking`, and several other approaches such as a work queue, require the passed-in closure/data to be `'static`. This can become a headache in combination with any data that holds a borrow, such as `State<'_, T>` or an `&str`. I expect this to be the biggest problem with this approach in user code.
  * This drawback is *also* the most problematic one to "work around". In order to be sound, a wrapper that did *not* require a `'static` closure would need to either (synchronously!) wait on the spawned task's completion or abort the process in the event that the task was canceled before the blocking task completed.
* Instead of wrapping only database operations, we could wrap entire routes in `spawn_blocking`. I have avoided this approach, as I expect it to be *even more* problematic given the `'static` restrictions.

## Todo

* [x] Decide on the underlying approach and/or workarounds for any aforementioned problems with `spawn_blocking` - for now, treating this as "solved" by using a `Semaphore`. Further improvements are of course possible!
* [ ] Thoroughly review and fix all applicable documentation, including the guide.
* [x] Prevent the user from accessing the inner fields of the generated guard type and pool type by nesting them inside another module.
  * [x] ~~This turns out to be... tricky. We would be going from `struct A(B);` to `mod __inner { struct A(B); }` with (as far as I can tell) to reliable way to access `B` from `__inner`. One "awkward-but-works" option would be to generate `mod __inner { struct A<T>(T); } type A = __inner::A<B>;`.~~
* [ ] I now realize this will have a *lot* of conflicts with #1230 on `master`. Perhaps we can forward-port that commit to `async`.
* [x] error handling in the case that the database is unavailable.

### DIFF
--- a/contrib/lib/src/databases.rs
+++ b/contrib/lib/src/databases.rs
@@ -328,8 +328,8 @@
 //! }
 //!
 //! #[get("/")]
-//! fn my_handler(conn: MyDatabase) -> Data {
-//!     load_from_db(&*conn)
+//! async fn my_handler(mut conn: MyDatabase) -> Data {
+//!     conn.run(|c| load_from_db(c)).await
 //! }
 //! # }
 //! ```
--- a/contrib/lib/src/databases.rs
+++ b/contrib/lib/src/databases.rs
@@ -403,7 +403,7 @@ pub extern crate diesel;
 
 use std::collections::BTreeMap;
 use std::fmt::{self, Display, Formatter};
-use std::marker::{Send, Sized};
+use std::marker::PhantomData;
 use std::sync::Arc;
 
 use rocket::config::{self, Value};
@@ -847,12 +847,13 @@ impl Poolable for memcache::Client {
 ///
 /// This type is implemented here instead of in generated code to ensure all
 /// types are properly checked.
-pub struct ConnectionPool<C: Poolable> {
+pub struct ConnectionPool<K, C: Poolable> {
     pool: r2d2::Pool<C::Manager>,
     semaphore: Arc<Semaphore>,
+    _marker: PhantomData<fn() -> K>,
 }
 
-impl<C: Poolable> ConnectionPool<C> {
+impl<K: 'static, C: Poolable> ConnectionPool<K, C> {
     pub fn fairing(fairing_name: &'static str, config_name: &'static str) -> impl rocket::fairing::Fairing {
         rocket::fairing::AdHoc::on_attach(fairing_name, move |mut rocket| async move {
             let config = database_config(config_name, rocket.config().await);
@@ -860,9 +861,10 @@ impl<C: Poolable> ConnectionPool<C> {
 
             match pool {
                 Ok((size, Ok(p))) => {
-                    let managed = ConnectionPool::<C> {
+                    let managed = ConnectionPool::<K, C> {
                         pool: p,
                         semaphore: Arc::new(Semaphore::new(size as usize)),
+                        _marker: PhantomData,
                     };
                     Ok(rocket.manage(managed))
                 },
@@ -886,6 +888,7 @@ impl<C: Poolable> ConnectionPool<C> {
         cargo.state::<Self>().map(|c| Self {
             pool: c.pool.clone(),
             semaphore: c.semaphore.clone(),
+            _marker: PhantomData,
         })
     }
 
@@ -906,7 +909,7 @@ impl<C: Poolable> ConnectionPool<C> {
 }
 
 #[::rocket::async_trait]
-impl<'a, 'r, C: Poolable> rocket::request::FromRequest<'a, 'r> for ConnectionPool<C> {
+impl<'a, 'r, K: 'static, C: Poolable> rocket::request::FromRequest<'a, 'r> for ConnectionPool<K, C> {
     type Error = ();
 
     async fn from_request(request: &'a rocket::request::Request<'r>) -> rocket::request::Outcome<Self, ()> {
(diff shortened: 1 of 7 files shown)


examples/todo/src/tests.rs
```diff
@@ -40,7 +40,7 @@ fn test_insertion_deletion() {
             .dispatch().await;
 
         // Ensure we have one more task in the database.
-        let new_tasks = Task::all(&conn);
+        let new_tasks = conn.run(|c| Task::all(c)).await;
```
SergioBenitez, line 43: Does `conn.run(Task::all)` work?

jebrosen: No for two reasons: `run()` gives an `&mut PooledConnection<...>` (fixing now), but even after that `Task::all` wants an `&` but `run` provides an `&mut` and there's no automatic coercion here. Some databases *need* the `&mut` option, so we would have to provide two versions of `run` to get this convenience back for functions that want a shared reference instead of a mutable one.

contrib/lib/src/databases.rs
```diff
@@ -843,6 +846,82 @@ impl Poolable for memcache::Client {
...
+
+    pub async fn run<F, R>(&self, f: F) -> R
+    where
+        F: FnOnce(&mut r2d2::PooledConnection<C::Manager>) -> R + Send + 'static,
+        R: Send + 'static,
+    {
+        let _permit = self.semaphore.acquire().await;
+        let pool = self.pool.clone();
+        let ret = tokio::task::spawn_blocking(move || {
+            let mut conn = pool.get().expect("TODO");
```
jebrosen, line 904: This is... not great. By acquiring a connection in `run` instead of in `from_request`, `run()` will need to return a `Result<_, r2d2::Error>`. Database operations are also likely to return a `Result`, though, so `Result<Result<T, db::Error>, r2d2::Error>>` would be typical. Doing this would add a lot of error-handling boilerplate at usage sites that wasn't there before.

jebrosen:
We now have at least two APIs to consider:

* The original proposal, which called `spawn_blocking()` once each in `from_request`, `run()`, and `drop`. Connection errors were propagated from `from_request` as a `ServiceUnavailable` `Failure`.
* The current set of changes, which calls `spawn_blocking()` once in each `run()` call. Connection errors are returned directly by `run()`.

The trade-offs are roughly:
* Performance: single call to `spawn_blocking()` is faster. I expect this to be less noticeable the longer database queries take for a given application.
* Ergonomics: single call to `spawn_blocking()` is uglier to use, because the error handling is in the route code calling `run()` instead of in the request guard.
* Implementation: single call to `spawn_blocking()` is easier to implement and understand. Moving error handling to `run()` and spawning a blocking task in `drop()` lead to extra workarounds or inconveniences, such as extra `Option`s, having `run()` take `&mut self`, or introducing an async `Mutex`.

There is one particular ergonomic improvement left to do: `run()` could require the closure to return a result. This might allow us to "flatten" one layer of `Result`, and many interactions with the database will end up with a `Result` anyway so it shouldn't be much a burden.
  * The error type could be required to implement `From<r2d2::Error>`: `async fn run<F: FnOnce(&mut Connection) -> Result<T, E>, T, E: From<r2d2::Error>>(f: F) -> Result<T, E> { }`.
  * The two error paths could be conveyed through some sort of `Result`/`Error`/`Outcome` enum used specifically in `run()` calls: `enum DbError<E> { Error(E), PoolError(r2d2::Error) } async fn run<F: FnOnce(&mut Connection) -> Result<T, E>, T, E>(f: F) -> Result<T, DbError<E>> { }`. This is similar to the `run` API in `bb8` `0.3`.

PR #1345, opened 2020-06-15
Simplify ASCII checks for char and byte primitives

gcarq:
This PR simplifies some ASCII checks for byte and char types, it doesn't change behaviour.

### DIFF
--- a/examples/pastebin/src/paste_id.rs
+++ b/examples/pastebin/src/paste_id.rs
@@ -35,11 +35,7 @@ impl<'a> fmt::Display for PasteID<'a> {
 
 /// Returns `true` if `id` is a valid paste ID and `false` otherwise.
 fn valid_id(id: &str) -> bool {
-    id.chars().all(|c| {
-        (c >= 'a' && c <= 'z')
-            || (c >= 'A' && c <= 'Z')
-            || (c >= '0' && c <= '9')
-    })
+    id.chars().all(|c| c.is_ascii_alphanumeric())
 }
 
 /// Returns an instance of `PasteID` if the path segment is a valid ID.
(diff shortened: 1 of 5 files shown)


examples/pastebin/src/paste_id.rs
```diff
@@ -35,11 +35,7 @@ impl<'a> fmt::Display for PasteID<'a> {
 
 /// Returns `true` if `id` is a valid paste ID and `false` otherwise.
 fn valid_id(id: &str) -> bool {
-    id.chars().all(|c| {
-        (c >= 'a' && c <= 'z')
-            || (c >= 'A' && c <= 'Z')
-            || (c >= '0' && c <= '9')
-    })
+    id.chars().all(|c| c.is_ascii_alphanumeric())
```
jebrosen, line 38: Does `all(char::is_ascii_alphanumeric)` work here (and elsewhere in this changeset)?

gcarq: Unfortunately not, I think because the string is passed as reference.
As far as I can tell this is the only place where it would be applicable in the changeset.
```
error[E0631]: type mismatch in function arguments
  --> examples/pastebin/src/paste_id.rs:38:20
   |
38 |     id.chars().all(char::is_ascii_alphanumeric)
   |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |                    |
   |                    expected signature of `fn(char) -> _`
   |                    found signature of `for<'r> fn(&'r char) -> _`
```

SergioBenitez:
Merged in 86d0bdddb289c574c273c823f3e7c6be64202d57. Thank you!

Issue #1312, opened 2020-05-27
Clone implementation for LocalRequest is unsound


Qwaz:
## The Bug
LocalRequest is one of a few places that Rocket uses unsafe Rust. While auditing the code, I found that its Clone implementation is unsound.

First, let's look at the definition of LocalRequest, which implements a mutable Rc with a raw pointer:
https://github.com/SergioBenitez/Rocket/blob/ca4d1572d408fd8dd13db103926ad96da878126d/core/lib/src/local/request.rs#L68-L100

The comment justifies that it is safe to have LocalRequest and LocalResponse which share the same Request pointer because modification from LocalRequest is not observable from LocalResponse. However, LocalRequest's Clone implementation allows to create two LocalRequest that share the same Request raw pointer, which was not part of the justification:
https://github.com/SergioBenitez/Rocket/blob/ca4d1572d408fd8dd13db103926ad96da878126d/core/lib/src/local/request.rs#L477-L487

This ultimately permits the violation of aliasing rule with two LocalRequest instances that point to the same Request instance.

## Proof of Concept
PoC code:
```rust
use rocket::http::Header;
use rocket::local::Client;
use rocket::Request;

fn main() {
    let client = Client::new(rocket::ignite()).unwrap();

    // creates two LocalRequest instances that share the same Request pointer
    let request1 = client.get("/").header(Header::new("key", "val1"));
    let request2 = request1.clone();

    // sanity check
    assert_eq!(
        request1.inner() as *const Request<'_>,
        request2.inner() as *const Request<'_>
    );

    // save the iterator, which internally holds a slice
    let mut iter = request1.inner().headers().get("key");

    // insert headers to reallocate the header map
    request2
        .header(Header::new("1", "v1"))
        .header(Header::new("2", "v2"))
        .header(Header::new("3", "v3"))
        .header(Header::new("4", "v4"))
        .header(Header::new("5", "v5"))
        .header(Header::new("6", "v6"))
        .header(Header::new("key", "val2"));

    // heap massage
    let _: Vec<usize> = vec![0, 0xcafebabe, 31337, 0];

    // iter is dangling now!
    let s = iter.next().unwrap();

    // address and length controlled
    dbg!(s.as_ptr());
    dbg!(s.len());

    // segfaults
    println!("{}", s);
}
```

Output:
```
   Compiling rocket-unsound v0.1.0 (/home/yechan/rust/rocket-unsound)
[...]

PR #1346, opened 2020-06-16
async: blocking client API

jebrosen:
Adds a blocking variant of the APIs in `rocket::local::` which is more convenient to use in testing. Most tests are currently `async` only because the API is, and the use of `#[rocket::async_test]` and `.await`s sprinkled throughout is just ceremony.

An alternative approach could be to support only a blocking API and keep the `async` bits internal-only, but since requests can and will be executed concurrently in a live server I felt uncomfortable imposing this restriction. I believe that the `async`-capable test harness can be useful for simulating routes under load (e.g. interactions with databases or external APIs) without needing an HTTP client test harness.

## Implementation Overview

The blocking `Client` spawns a single background thread on construction. This thread runs a single-threaded `tokio` `Runtime`, and waits for a signal that the associated `Client` has been dropped. Methods on the blocking `Client`, `LocalRequest`, and `LocalResponse` use the runtime's `Handle::block_on` method to expose a synchronous API on top of the existing asynchronous `Client`. This is similar to the [blocking API for `reqwest`](https://docs.rs/reqwest/0.10.6/reqwest/blocking/index.html), but it takes advantage of the recently added `Handle::block_on` method in `tokio` to avoid an additional channel for communication to/from the runtime.

Method signatures and documentation are simply copied from the blocking to the non-blocking version, and method bodies have been replaced with `client.block_on()` as necessary. Due to the large amount of similarity, `diff -u client.rs blocking/client.rs` will be useful for reviewing this.

## Todo
* [x] Currently, all documentation is copied directly from the existing API. This raises two new issues:
  * [x] Where to highlight the differences, and where to leave documentation identical.
  * [x] How to keep these in sync in the future.
* [x] Add explanations and examples to the guide-level documentation.


core/lib/src/local/blocking/client.rs
```diff
@@ -0,0 +1,316 @@
...
+            .expect("create tokio runtime");
+
+        let (_shutdown_tx, shutdown_rx) = channel();
+        let handle = runtime.handle().clone();
+
+        // Initialize the Rocket instance
+        let inner = runtime.block_on(crate::local::Client::_new(rocket, tracked))?;
+
+        // Spawn a background thread to drive the single-threaded runtime
+        std::thread::spawn(move || {
```
SergioBenitez, line 34: Why is this necessary? See my top-level comment.

core/lib/src/local/blocking/request.rs
```diff
@@ -0,0 +1,384 @@
...
+    /// let client = Client::new(rocket::ignite()).expect("valid rocket");
+    /// let req = client.get("/");
+    /// let inner_req = req.inner();
+    /// ```
+    #[inline]
+    pub fn inner(&self) -> &Request<'c> {
+        self.inner.inner()
+    }
+
+    fn _map<F>(self, f: F) -> Self
```
SergioBenitez, line 99: This seems like a big dance to do. See my top-level comment.

SergioBenitez (changes requested):
Awesome.

My concerns can be summarized by this diff: ~~https://paste.rs/W1Y~~. ~~(https://paste.rs/W1Y.diff)~~

In short:
  * Why not store the `Runtime` directly behind a `Mutex`? This way, we don't need a special dance to keep the runtime on a `block_on`.
  * I'd rather have the `self.inner = foo; self` in each method than the `_map`. I don't care too much. In any case, the `_map` can be written more cleanly as: `self.inner = f(self.inner); self`.
  * I renamed `crate::local` as `aio` just for my own sanity. This isn't a request.

I have ideas on how to clean things up re: documentation/duplication/keeping things in sync, but let's agree on an implementation first.

core/lib/src/local/blocking/client.rs
jebrosen: Ah, I had this case in the back of my mind but did not actually write any example using it: suppose I have an `Arc<blocking::Client>`, and I want to run one request each in two separate threads? With the background thread doing the actual work, those requests will complete "as concurrently as possible". With the `Mutex`, the same requests will complete one at a time.

Since there's no I/O with the client, the only way way this makes a difference is for routes that use the database or timers... after reconsidering I would rather adopt the `Mutex` - "use the non-blocking `Client` if you want concurrency" seems a fair tradeoff. And I imagine the `Mutex` should be faster for the (more common) single-thread usage.

core/lib/src/local/blocking/request.rs
jebrosen: I think this is just an artifact of an older, much uglier, workaround to make `block_on` re-entrant and I'll remove it.

jebrosen:
@SergioBenitez were you planning on adding more methods to `Client` / or already have some locally, or shall I work on adding those?

SergioBenitez:
@jebrosen Yes. I'm finishing off the `Client` changes now. What's left after this is to do effectively the same thing for `LocalRequest` and `LocalResponse` as has been done with `Client` re:deduplication.

SergioBenitez:
@jebrosen Pushed! You'll see that I modified the deduplication macro to something that I believe is much cleaner and easier to work with. It also generates all of the examples, so no documentation is duplicated either. Finally, I added an `Rc` to `blocking::Client` to make it `!Send`; `RefCell` _is_ `Send`.

jebrosen:
> You'll see that I modified the deduplication macro to something that I believe is much cleaner and easier to work with. It also generates all of the examples, so no documentation is duplicated either, at the small cost of losing a bit of detail. 

Yeah, this does look much easier to work with!

> Finally, I added an `Rc` to `blocking::Client` to make it `!Send`

This begs the question: it obviously can't be `Sync` (we want `&self` -> `&mut Runtime` without wasting a `Mutex`), but is there any harm in it being `Send`? Are we future-proofing by not promising `Send` too eagerly? Whatever the case, a comment would be good since the main API of `Rc` (`Clone` and `Drop`) is never actually used.

SergioBenitez:
@jebrosen The main reason I'd like to be `Send` is so that the following is a compile-time error:

```rust
#[async_test]
async fn test() {
    let client = local::blocking:Client::new(rocket());
    let res = client.get("/").dispatch();
}
```

Otherwise, we'll be executing `block_on` in an `async` environment, which will invariably [result in a panic](https://docs.rs/tokio/0.2.21/tokio/runtime/struct.Runtime.html#method.block_on).

jebrosen:
Ah. That's not actually sufficient, since the `client` is not held across an `await`: <https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a0a4bbe10eab10576493ac31a5209cc1>

SergioBenitez:
@jebrosen Of course. Bummer. I suppose a `panic` isn't _too_ bad, given that these are tests, though it may result in false-positives for `#[should_panic]` tests.

One idea is to insert a dummy `.await` at the beginning of the user's function to _force_ the await point. Something like:


```rust
#[async_test]
async fn test() {
    let _rocket_inserted_await = dummy().await;
    let client = local::blocking:Client::new(rocket());
    let res = client.get("/").dispatch();
}
```

Would love any other ideas you might have, or perhaps you feel that that this isn't something to be too concerned about?

SergioBenitez:
Alright! I believe, barring a few doc fixes, this is complete! I rebased on `async` after the `Manifest` -> `Cargo` changes, which blew away commit history, but I'll restore things when I actually merge.

There are some sweeping changes here. Here's my attempt at iterating them:
  * `Request` is now `Clone`, but only privately. This is because we need to know that there are no outstanding borrows to internal structures to successfully clone everything. We know this in `LocalRequest`, so we use it there.
  * We use the `atomic` crate to atomically update `Method`. We _really_ don't want to not get the latest `Method` on clone. If we want to eschew the dependency, we can manually use an `AtomicUsize` and convert between it and `Method`.
  * `mut_dispatch` is gone. It causes issues and makes reasoning about the soundness of `unsafe` code more difficult. I've also figured out how to get rid of the one-line of `unsafe` code remaining, even though I'm convinced it's correct.
  * `LocalResponse` exposes `Response` as read-only. This makes sense to me; you have the `Response` already, what is there to modify? The body reader methods were changed to `into_string()` and `into_bytes()`. Perhaps we should add an `into_reader()` or something of the sort? Or `into_raw() -> ResponseBody`?
  * `Body` exposes a read-only mechanism to determine what the size of the body is, if it is known beforehand. A few tests make use of this.
  * `Response::body()` is now `Response::body_mut()`. `Response::body() -> &Body` was added.

@jebrosen Please do feel free to comment on any of this!

jebrosen:
This all looks really great!

* RE `Request` and `Clone` - I expect it could be safe and useful to clone a request that has never been dispatched (and that does not have any request-local state?). I believe the problems here are *mostly* related to request-local state, but `rocket.rs` does swap the method around a bit when auto-handling `HEAD` requests. This point is orthogonal to `unsafe`, although `unsafe` can "make things worse" as we saw in #1312.
* `into_reader` or `into_raw` may be helpful for testing "unending" responses, or responses that can be influenced by making other requests to the server. Such methods may actually be *necessary* to be able to test request flows such as "start request A, read start of response A" -> "make a second request B" -> "read more of response A" -> "make a third request C" -> "read rest of response A".

I *think* all that's left to do is docs:

* docs on structs (`Client`, `LocalRequest`, `LocalResponse`)
* docs on modules (`local`, `local::asynchronous`, `local::blocking`)
* guide-level docs
* ensuring all the docs are correct and useful. For example, I see one or two doc examples that are missing `.await` on `asynchronous::Client`.

I will start those by copying text from the 0.4 docs, adding in words such as "asynchronous" or "blocking" as necessary.

core/lib/src/local/client.rs
```diff
@@ -1,396 +1,238 @@
...
+        /// # Errors
+        ///
+        /// If launching the `Rocket` instance would fail, excepting network errors,
+        /// the `LaunchError` is returned.
+        ///
+        /// ```rust,no_run
+        #[doc = $import]
+        ///
+        /// let rocket = rocket::ignite();
+        /// let client = Client::new(rocket);
```
jebrosen, line 109: This line needs an `.await` in the asynchronous version. Is there a nice way to "paste" the `$(.$suffix)?` into the line instead of adding arguments like `$call_new:literal`? I can tell `#[doc = $(.$suffix)?]` won't work for several reasons. The same issue is present for the `async fn`s on `LocalRequest` and `LocalResponse`.

SergioBenitez:
@jebrosen Okay, I think I'm happy with this implementation + docs. I think the docs here strike a good balance between deduplication and having the relevant information in the right place.

Would love for you to take a glance at the implementation, @jebrosen, see if I missed anything. Otherwise, I propose that we migrate all of the examples that can use the `blocking` set of APIs to the blocking variant, after which I propose we merge.

jebrosen:
This looks great! I reviewed the documentation text and fixed one or two docs links, and I'm working on switching to `blocking` in examples. I'll also look for and possibly migrate documentation tests outside of `local::`. Would you prefer for tests in `core/lib/tests`, `contrib/lib/tests`, etc. to use the `asynchronous` or `blocking` APIs?

SergioBenitez:
> Would you prefer for tests in `core/lib/tests`, `contrib/lib/tests`, etc. to use the `asynchronous` or `blocking` APIs?

Yeah. Unless I'm missing something, I'd say if we _can_ use the blocking API, let's use it.

SergioBenitez:
Okay, minus the guide, I'm satisfied with merging this. If you'd like to take a crack at fixing up the guide for this, I'm happy to accept anything. I think we should effectively only discuss one of the two variants, ideally the blocking one, and then say, "You'll have noticed `blocking` up there. That's because ... Sometimes, you'll need `async`, such as when [...]. so there's `asynchronous`! It works the exact same way. See [here] for an example and [here] for more docs."

jebrosen:
Alright, I've added a note (hopefully formatted correctly). We don't currently have any examples or tests where asynchronous dispatching actually makes a difference, so the example is a bit thin.

SergioBenitez:
I added an example where `async` testing is _really_ required.

Funny enough, this revealed that Chrome doesn't concurrently dispatch two requests to the same server/path combo. Instead, it waits for the first to finish before dispatching the second. Thus, testing this example in the browser via two tabs to `/barrier` in Chrome gives the appearance that the server is blocking. Something to keep in mind for "bug reports" in the future.

jebrosen:
> Funny enough, this revealed that Chrome doesn't concurrently dispatch two requests to the same server/path combo. Instead, it waits for the first to finish before dispatching the second. Thus, testing this example in the browser via two tabs to /barrier in Chrome gives the appearance that the server is blocking. Something to keep in mind for "bug reports" in the future.

Very interesting. I replicated in Chromium, and Firefox sends the requests simultaneously. I expect that this behavior could depend on many factors in either browser, such as HTTP/2 and user configuration.

SergioBenitez:
Merged in 722497010b51ea29c816d839d8932eab9f34d1a9 ... 4419dd3b4623a5eb212ecfa4c339308e54c52b95. Woo!

Issue #998, opened 2019-05-09
`Uri` issues, summarized


jebrosen:
## Background

Rocket 0.4 [introduced Typed URIs](https://rocket.rs/v0.4/news/2018-12-08-version-0.4/#typed-uris) and a new [`Uri` type](https://api.rocket.rs/v0.4/rocket/http/uri/enum.Uri.html). The `uri!` macro and strict `Uri` validation is a valuable addition to Rocket's API and is a great example of Rocket's promise of safety and correctness, but there are a few problems with it due to deficiencies and/or bugs in Rocket and in browsers.

This is a summary of the following issues and PRs:

- [x] #842 - URI References /  `Uri` type does not support anchors.
- [x] #853 - Support for absolute `Uri`s / generate absolute URIs from `uri!()`.
- [x] #880/#882 - matching Url-Escaped Routes/match percent decoded urls
- [x] #924/#941 - Un-encoded curly braces fail to parse/Allow more unencoded characters in query strings
- [x] https://github.com/SergioBenitez/Rocket/issues/998#issuecomment-544169526
- #995 - (needs some confirmation/testing) Path segments must be valid UTF-8

## Bugs
* `Uri` does not appear to support fragments in any URI from
  * Workaround: Servers should only deal with fragments in URIs they generate, and most of those can be done via `format!`.
  * Possible solution: add fragment support, but disallow it or treat it as part of the path/query in incoming URIs 
* `Location` uses `Uri` and is therefore too strict about which kinds of URIs it accepts (no relative URI, no fragment)
  * Workaround: Manually set the header as a string.
  * Possible solution: Add support for relative URIs and URIs with fragments to `Location` and/or `Uri`
* `Uri` is strict about characters such as `{` needing to be percent-encoded
  * Several browsers send `{` and other characters unencoded, even when they ought to be encoded.
  * **There is currently no workaround**. If a user enters a `{` or certain other relatively innocent charactesr in a plain HTML form submitted with the GET method, Rocket may refuse to process the request entirely.
  * Possible solution: Allow `{` and some other commonly unencoded characters. Clearly document how this adheres or purposefully does not adhere to the relevant standard and why.
* (unconfirmed) `>` isn't even legal in the HTTP header, according to hyper 0.10? Worth double-checking what's going on here.
* Route matching to the request URI is sensitive to percent encoding (e.g. `$` does not match `%24`).
  * **There is currently no workaround**
[...]

PR #1349, opened 2020-06-18
Upgrade `percent-encoding` to 2.x and `base64` to 0.12.

jakubadamw:


jakubadamw:
I understand employing loops in `const fn` might be an overkill. 🙂 I am unsure of a better way. One way would be to convert the parser tables to use `AsciiSet`, as it is a rather neat structure, but there isn't even a way to create an empty `AsciiSet`, and so that route would require changes to `percent-encoding`.

jebrosen:
> I understand employing loops in the `const fn` context might be an overkill.

Since we are working towards compiling on stable (and nearly there!), I would reject a PR that fundamentally and unconditionally requires new `#![feature]` flags. But, this particular PR should be possible without a feature:

> One way would be to convert the parser tables to use AsciiSet, as it is a rather neat structure, but there isn't even a way to create an empty AsciiSet, and so that route would require changes to percent-encoding.

I agree using `AsciiSet` directly could be a cleaner approach; the one issue I know of up front is the differentiation of `<` and `>` for route specifiers only. The presets `CONTROLS` and `NON_ALPHANUMERIC` look like fine starting points - `AsciiSet` has both `add` and `remove`, so any set at all is possible although perhaps inconvenient. It should be possible to make a nice macro, even: `ascii_set!(PCHAR_SET: CONTROLS + b'X' - b'Y');

cc #998.

jakubadamw:
@jebrosen, `const_if_match` and `const_loop` have just recently been stabilized in the nightly Rust. Does that address your concern?

jebrosen:
Oh, that's awkward timing - it will hit stable in `1.46.0` (currently mislabeled as `1.45.0`), but I was hoping to only need to require `1.45.0` for 0.5. For that reason, I would still prefer to see a solution without the branching or loop. Since `AsciiSet` doesn't allow for introspection, It looks like we would have to either 1) duplicate the logic encoded in `is_pchar` in order to do this dependency upgrade without requiring `1.46.0`, or 2) have a macro that generates both the `PATH_CHARS` table and the equivalent `AsciiSet` at the same time.

Here's the hypothetical macro I mentioned earlier and some example usage; the actual sets are untested. This could be useful as a starting point for any approach:

```rust
use percent_encoding::AsciiSet;

macro_rules! ascii_set {
    ($e:expr;) => { $e };
    ($e:expr; - $sub:literal $($rest:tt)*) => { ascii_set!($e.remove($sub); $($rest)*) };
    ($e:expr; + $add:literal $($rest:tt)*) => { ascii_set!($e.add($add); $($rest)*) };
}

const PCHAR_SET: AsciiSet = ascii_set!(
    percent_encoding::NON_ALPHANUMERIC;
    -b'!' -b'$' -b'&' -b'\''
    -b'(' -b')' -b'*' -b'+' -b',' -b'-' -b'.' -b'/'
    -b':' -b';'
    -b'=' -b'@'
    -b'_'
    -b'~'
);

const QUERY_SET: AsciiSet = ascii_set!(PCHAR_SET;
    -b'?'
    +b'%'
    +b'+'
);
```

jebrosen:
@jakubadamw unless you have some unpushed work on this, I'm planning to rebase this onto `master` and finish it this week as part of some other dependency updates.

jakubadamw:
@jebrosen, I don't. Sure, sounds great, thanks!

PR #1353, opened 2020-06-23
Update 'toml' to '0.5'.

jebrosen:
I did this back in January, and simply forgot to make a PR!

### DIFF
--- a/core/lib/src/config/toml_ext.rs
+++ b/core/lib/src/config/toml_ext.rs
@@ -44,7 +44,8 @@ fn key_value<'a>(input: &mut &'a str) -> Result<(String, Value), &'a str> {
 
 #[parser]
 fn table<'a>(input: &mut &'a str) -> Result<Value, &'a str> {
-    Value::Table(collection('{', key_value, ',', '}')?)
+    let map: std::collections::BTreeMap<_, _> = collection('{', key_value, ',', '}')?;
+    Value::Table(map.into_iter().collect())
 }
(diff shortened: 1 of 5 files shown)


core/lib/src/config/toml_ext.rs
```diff
@@ -44,7 +44,8 @@ fn key_value<'a>(input: &mut &'a str) -> Result<(String, Value), &'a str> {
 
 #[parser]
 fn table<'a>(input: &mut &'a str) -> Result<Value, &'a str> {
-    Value::Table(collection('{', key_value, ',', '}')?)
+    let map: std::collections::BTreeMap<_, _> = collection('{', key_value, ',', '}')?;
```
jebrosen, line 47: This is the only thing I'm not super happy about. `pear::parsers::Collection` is not implemented for the new `toml::map::Map` type, which is either a `BTreeMap` or an `IndexMap` depending on a feature flag.

SergioBenitez: I've changed `Collection` in `pear` to be:

```rust

pub trait Collection<A>: Default + Extend<A> {
    fn push(&mut self, item: A) {
        self.extend(Some(item))
    }
}

impl<A, T: Default + Extend<A>> Collection<A> for T {  }
```

The line as before, and everything else, should work after this change. I'm working on a few other changes to pear before updating Rocket to use it and will ping you when it's committed.

Issue #1356, opened 2020-06-23
Minor improvement suggestions to the guide regarding cookies, json, errors and testing.


arctic-alpaca:
Hi,
while working though the guide, I found some minor things that could, in my opinion be improved.
I checked the master branch to see whether the mentioned sections might have changed but as far as I can tell, my suggestions should still be applicable.

- [x] https://rocket.rs/v0.4/guide/requests/#cookies
    This isn't related to Rocket specifically. Cookie paths weren't clear to me so I tried to set a cookie in "/cookies/set" and print it in "/cookies/get". This doesn't work because the path set by `Cookies::new()` in this case is "/cookies/set", so the cookie isn't accessible from "/cookies/get".
    Adding a hint or warning that the path you try to read the cookie from must start with the same path you set it from might be helpful for newcomers.

- [x] https://rocket.rs/v0.4/guide/responses/#wrapping
    ``` 
    #[get("/")] 
    fn json() -> content::Json<&'static str> { 
        content::Json("{ 'hi': 'world' }")  
    }
    ```
    Results in "SyntaxError: JSON.parse: expected property name or '}' at line 1 column 3 of the JSON data" in Firefox. JSON seems to require double quotes according to https://www.json.org/json-en.html.
    replacing 
        `"{ 'hi': 'world' }"`
    with
       ` r#"{ "hi": "world" }"#`
    or
        `"{ \"hi\": \"world\" }"`
    solved the issue for me.

- [x] https://rocket.rs/v0.4/guide/responses/#errors
    "Responders may fail; they need not always generate a response." 
    Is this correct? English isn't my first language. To me it seems like there something missing. 
    "Responders may fail; they don't always need to generate a response."
    This seems correct to me, but I'm not sure if some meaning is lost.
    

- [ ] https://rocket.rs/v0.4/guide/testing/#testing
    `let client = Client::new(rocket).expect("valid rocket instance");`
    Shouldn't an error description be used with expect? Something like
    `let client = Client::new(rocket).expect("Could not create valid rocket instance");`?

jebrosen:
> This isn't related to Rocket specifically. Cookie paths weren't clear to me so I tried to set a cookie in "/cookies/set" and print it in "/cookies/get". This doesn't work because the path set by Cookies::new() in this case is "/cookies/set", so the cookie isn't accessible from "/cookies/get".
> Adding a hint or warning that the path you try to read the cookie from must start with the same path you set it from might be helpful for newcomers.

[...]

PR #1359, opened 2020-06-27
Guide Json example correction

arctic-alpaca:
Corrects the Json example in the guide #1356

### DIFF
--- a/site/guide/5-responses.md
+++ b/site/guide/5-responses.md
@@ -59,7 +59,7 @@ use rocket::response::content;
 
 #[get("/")]
 fn json() -> content::Json<&'static str> {
-    content::Json("{ 'hi': 'world' }")
+    content::Json("{ \"hi\": \"world\" }")
 }
 ```
 


site/guide/5-responses.md
```diff
@@ -59,7 +59,7 @@ use rocket::response::content;
 
 #[get("/")]
 fn json() -> content::Json<&'static str> {
-    content::Json("{ 'hi': 'world' }")
+    content::Json("{ \"hi\": \"world\" }")
```
jebrosen, line 62:
```rust
"{ \"hi\": \"world\" }"
// vs
r#"{ "hi": "world" }"#
```

The second is shorter (and IMO a bit cleaner), but do you think the `r#` syntax could confuse readers?

arctic-alpaca: I agree with the second one being a generally nicer solution, but I think it could be confusing for people new to rust or programming in general. Escaped characters in a string are a concept that's covered in most beginners guides for most languages. People who know about the second variation are probably not confused by the first one, I'm not sure whether that's the case the other way round.

jebrosen:
Thanks! Merged as 1e623c8.

PR #1374, opened 2020-07-13
Graceful shutdown timeout

notriddle:
After some feedback, this PR also adds a graceful shutdown `wait()` function.


SergioBenitez:
This was discussed on Matrix long ago and we reached the conclusion that the API exposed here is not one we want to move forward with. Thank you for the contribution, however, @notriddle.

PR #1375, opened 2020-07-14
Replace `#[database]` guard API with a `spawn_blocking` wrapper.

jebrosen:
Replaces #1343. Most of this description is a verbatim copy from that PR, but "High-Level Implementation Notes" is new.

The connection guard type generated by `#[database]` no longer implements `Deref` and `DerefMut`. Instead, it provides an `async fn run()` that gives access to the underlying connection on a closure run through `spawn_blocking()`.

## Background

All existing engines supported by `#[database]` are synchronous. Several of them, such as `diesel` and `sqlite`, do not have a well-supported `async` version or close alternative available. Therefore, I would consider it a non-starter to entirely drop support for them. However, using synchronous database libraries from `async` code has some issues:

## Motivation

Performing blocking operations while running on an asynchronous runtime is a footgun:
* Best case: the server does not experience enough load to exhibit any problems.
* Bad case: requests that are busy in a blocking operation will "artificially" prevent other requests from running or continuing, including any that are partway through I/O with the client, even if those other requests don't need to perform any blocking operation themselves.
* Worst case: deadlocks can occur. With database pools in particular, you could end up in a situation where all threads are waiting for a connection to be returned to the pool - but those connections are all held by tasks that cannot run, because the worker threads are all tied up waiting.

## High-Level Implementation Notes
* Most code has been moved into library code instead of being generated. This isn't *supposed* to make a difference to consuming code, but it is revealed in an error message. It does break any code that relied on certain implementation details, such as `conn.0` being equivalent to `*conn` or the existence of `<db_type>Pool` inside managed state. Although these were not part of the documented interface, they might have been used on accident (e.g. an IDE or compiler suggestion).
* `FromRequest`, `run()`, and `Drop` perform operations inside `spawn_blocking()`. `FromRequest` and `run` await the operations, and `Drop` ignores the eventual result.
  * A previous version of this PR made a single `spawn_blocking` call in `run()`. This has a theoretical and not-measured performance advantage, but at *significant* cost to ergonomics: it moves the "connection unavailable" error condition into routes, instead of `FromRequest`.
* An async `Semaphore` limits the number of active guard objects to `pool_size`. This is necessary to prevent a lockup: the blocking task queue has capacity `N`, and if at least `N` blocking tasks from `FromRequest` are busy, the tasks spawned by `Drop`  never get to run. `r2d2` does have a few timeouts; without those this situtation would be a deadlock.

## Problems and Alternatives

* It's inconvenient to do any asynchronous work inside the closure, so depending on the usage patterns `run()` might need to be called multiple times in between non-database work.
* `spawn_blocking`, and several other approaches such as a work queue, require the passed-in closure/data to be `'static`. This can become a headache in combination with any data that holds a borrow, such as `State<'_, T>` or an `&str`. I expect this to be the biggest problem with this approach in user code.
  * This drawback is *also* the most problematic one to "work around". In order to be sound, a wrapper that did *not* require a `'static` closure would need to either (synchronously!) wait on the spawned task's completion or abort the process in the event that the parent task was canceled before the blocking task completed.
* Instead of wrapping only database operations, we could wrap entire routes in `spawn_blocking`. I have avoided this approach, as I expect it to be *even more* problematic given the `'static` restrictions.

## Todo

* [ ] Thoroughly review and fix all applicable documentation, including the guide -- especially code that is for demonstration purposes and is not actually checked or run.

### DIFF
--- a/contrib/lib/src/databases.rs
+++ b/contrib/lib/src/databases.rs
@@ -84,9 +84,9 @@
 //! # type Result<T> = std::result::Result<T, ()>;
 //! #
 //! #[get("/logs/<id>")]
-//! fn get_logs(conn: LogsDbConn, id: usize) -> Result<Logs> {
+//! async fn get_logs(conn: LogsDbConn, id: usize) -> Result<Logs> {
 //! # /*
-//!     Logs::by_id(&*conn, id)
+//!     conn.run(|c| Logs::by_id(c, id)).await
 //! # */
 //! # Ok(())
 //! }
@@ -222,11 +222,9 @@
 //! ```
 //!
 //! The macro generates a [`FromRequest`] implementation for the decorated type,
-//! allowing the type to be used as a request guard. This implementation
-//! retrieves a connection from the database pool or fails with a
-//! `Status::ServiceUnavailable` if no connections are available. The macro also
-//! generates an implementation of the [`Deref`](std::ops::Deref) trait with
-//! the internal `Poolable` type as the target.
+//! allowing the type to be used as a request guard. This implementation always
+//! succeeds; database availability is not actually checked until `run()` is
+//! called.
 //!
 //! The macro will also generate two inherent methods on the decorated type:
 //!
@@ -235,11 +233,10 @@
 //!      Returns a fairing that initializes the associated database connection
 //!      pool.
 //!
-//!   * `fn get_one(&Cargo) -> Option<Self>`
+//!   * `fn async get_one(&Cargo) -> Option<Self>`
 //!
-//!     Retrieves a connection from the configured pool. Returns `Some` as long
-//!     as `Self::fairing()` has been attached and there is at least one
-//!     connection in the pool.
+//!     Retrieves a connection wrapper from the configured pool. Returns `Some`
+//!     as long as `Self::fairing()` has been attached.
 //!
 //! The fairing returned from the generated `fairing()` method _must_ be
 //! attached for the request guard implementation to succeed. Putting the pieces
@@ -280,8 +277,8 @@
 //!
 //! ## Handlers
 //!
-//! Finally, simply use your type as a request guard in a handler to retrieve a
-//! connection to a given database:
+//! Finally, use your type as a request guard in a handler to retrieve a
+//! connection wrapper for the database:
 //!
 //! ```rust
 //! # #[macro_use] extern crate rocket;
@@ -300,8 +297,7 @@
 //! # }
 //! ```
 //!
-//! The generated `Deref` implementation allows easy access to the inner
-//! connection type:
+//! A connection can be retrieved and used with the `run()` method:
 //!
 //! ```rust
 //! # #[macro_use] extern crate rocket;
@@ -320,8 +316,8 @@
 //! }
 //!
 //! #[get("/")]
-//! fn my_handler(conn: MyDatabase) -> Data {
-//!     load_from_db(&*conn)
+//! async fn my_handler(mut conn: MyDatabase) -> Data {
+//!     conn.run(|c| load_from_db(c)).await
 //! }
 //! # }
 //! ```
@@ -389,9 +385,12 @@ pub use tokio::task::spawn_blocking;
 pub extern crate diesel;
 
 use std::fmt::{self, Display, Formatter};
-use std::marker::{Send, Sized};
+use std::marker::PhantomData;
+use std::sync::Arc;
 
 use rocket::config::{self, Value};
+
+use rocket::tokio::sync::{OwnedSemaphorePermit, Semaphore};
 
 use self::r2d2::ManageConnection;
 
@@ -688,7 +687,7 @@ pub trait Poolable: Send + Sized + 'static {
     type Manager: ManageConnection<Connection=Self>;
     /// The associated error type in the event that constructing the connection
     /// manager and/or the connection pool fails.
-    type Error;
+    type Error: std::fmt::Debug;
 
     /// Creates an `r2d2` connection pool for `Manager::Connection`, returning
     /// the pool on success.
@@ -777,6 +776,127 @@ impl Poolable for memcache::Client {
     fn pool(config: DatabaseConfig<'_>) -> Result<r2d2::Pool<Self::Manager>, Self::Error> {
         let manager = r2d2_memcache::MemcacheConnectionManager::new(config.url);
         r2d2::Pool::builder().max_size(config.pool_size).build(manager).map_err(DbError::PoolError)
+    }
+}
+
+#[doc(hidden)]
+/// Unstable internal details of generated code for the #[database] attribute.
+///
+/// This type is implemented here instead of in generated code to ensure all
+/// types are properly checked.
+pub struct ConnectionPool<K, C: Poolable> {
+    pool: r2d2::Pool<C::Manager>,
+    semaphore: Arc<Semaphore>,
+    _marker: PhantomData<fn() -> K>,
+}
+
+#[doc(hidden)]
+/// Unstable internal details of generated code for the #[database] attribute.
+///
+/// This type is implemented here instead of in generated code to ensure all
+/// types are properly checked.
+pub struct Connection<K, C: Poolable> {
+    connection: Option<r2d2::PooledConnection<C::Manager>>,
+    _permit: Option<OwnedSemaphorePermit>,
+    _marker: PhantomData<fn() -> K>,
+}
+
+impl<K: 'static, C: Poolable> ConnectionPool<K, C> {
+    #[inline]
+    pub fn fairing(fairing_name: &'static str, config_name: &'static str) -> impl rocket::fairing::Fairing {
+        rocket::fairing::AdHoc::on_attach(fairing_name, move |mut rocket| async move {
+            let config = database_config(config_name, rocket.config().await);
+            let pool = config.map(|c| (c.pool_size, C::pool(c)));
+
+            match pool {
+                Ok((size, Ok(p))) => {
+                    let managed = ConnectionPool::<K, C> {
+                        pool: p,
+                        semaphore: Arc::new(Semaphore::new(size as usize)),
+                        _marker: PhantomData,
+                    };
+                    Ok(rocket.manage(managed))
+                },
+                Err(config_error) => {
+                    ::rocket::logger::error(
+                        &format!("Database configuration failure: '{}'", config_name));
+                    ::rocket::logger::error_(&format!("{}", config_error));
+                    Err(rocket)
+                },
+                Ok((_, Err(pool_error))) => {
+                    ::rocket::logger::error(
+                        &format!("Failed to initialize pool for '{}'", config_name));
+                    ::rocket::logger::error_(&format!("{:?}", pool_error));
+                    Err(rocket)
+                },
+            }
+        })
+    }
+
+    async fn get(&self) -> Result<Connection<K, C>, ()> {
+        let permit = self.semaphore.clone().acquire_owned().await;
+        let pool = self.pool.clone();
+        match spawn_blocking(move || pool.get()).await.unwrap() {
+            Ok(c) => {
+                Ok(Connection {
+                    connection: Some(c),
+                    _permit: Some(permit),
+                    _marker: PhantomData,
+                })
+            }
+            Err(e) => {
+                error_!("Failed to get a database connection: {}", e);
+                Err(())
+            }
+        }
+    }
+
+    #[inline]
+    pub async fn get_one(cargo: &::rocket::Cargo) -> Option<Connection<K, C>> {
+        cargo.state::<Self>()?.get().await.ok()
+    }
+}
+
+impl<K, C: Poolable> Connection<K, C> {
+    #[inline]
+    pub async fn run<F, R>(&mut self, f: F) -> R
+    where
+        F: FnOnce(&mut C) -> R + Send + 'static,
+        R: Send + 'static,
+    {
+        let mut conn = self.connection.take()
+            .expect("Connection unexpectedly missing. Was run() called without being awaited?");
+
+        let (value, conn) = tokio::task::spawn_blocking(move || {
+            let value = f(&mut conn);
+            (value, conn)
+        }).await.expect("failed to spawn a blocking task to use a pooled connection");
+
+        self.connection = Some(conn);
+        value
+    }
+}
+
+impl<K, C: Poolable> Drop for Connection<K, C> {
+    fn drop(&mut self) {
+        if let Some(conn) = self.connection.take() {
+            tokio::task::spawn_blocking(|| drop(conn));
+        }
+    }
+}
+
+#[::rocket::async_trait]
+impl<'a, 'r, K: 'static, C: Poolable> rocket::request::FromRequest<'a, 'r> for Connection<K, C> {
+    type Error = ();
+
+    #[inline]
+    async fn from_request(request: &'a rocket::request::Request<'r>) -> rocket::request::Outcome<Self, ()> {
+        let inner = ::rocket::try_outcome!(request.guard::<::rocket::State<'_, ConnectionPool<K, C>>>().await);
+
+        match inner.get().await {
+            Ok(c) => rocket::Outcome::Success(c),
+            Err(()) => rocket::Outcome::Failure((rocket::http::Status::ServiceUnavailable, ())),
+        }
     }
 }
 
--- a/contrib/lib/src/databases.rs
+++ b/contrib/lib/src/databases.rs
@@ -322,6 +322,37 @@
 //! # }
 //! ```
 //!
+//! `run()` takes the connection by value. To make multiple calls to run,
+//! obtain a second connection first with `clone()`:
+//!
+//! ```rust
+//! # #[macro_use] extern crate rocket;
+//! # #[macro_use] extern crate rocket_contrib;
+//! #
+//! # #[cfg(feature = "diesel_sqlite_pool")]
+//! # mod test {
+//! # use rocket_contrib::databases::diesel;
+//! # type Data = ();
+//! #[database("my_db")]
+//! struct MyDatabase(diesel::SqliteConnection);
+//!
+//! fn load_from_db(conn: &diesel::SqliteConnection) -> Data {
+//!     // Do something with connection, return some data.
+//!     # ()
+//! }
+//!
+//! #[get("/")]
+//! async fn my_handler(mut conn: MyDatabase) -> Data {
+//!     let cloned = conn.clone().await.expect("");
+//!     cloned.run(|c| load_from_db(c)).await;
+//!
+//!     // Do something else
+//!     conn.run(|c| load_from_db(c)).await;
+//! }
+//! # }
+//! ```
+//!
+//!
 //! # Database Support
 //!
 //! Built-in support is provided for many popular databases and drivers. Support
@@ -799,6 +830,7 @@ pub struct ConnectionPool<K, C: Poolable> {
 /// types are properly checked.
 #[doc(hidden)]
 pub struct Connection<K, C: Poolable> {
+    pool: ConnectionPool<K, C>,
     connection: Option<r2d2::PooledConnection<C::Manager>>,
     _permit: Option<OwnedSemaphorePermit>,
     _marker: PhantomData<fn() -> K>,
@@ -842,6 +874,7 @@ impl<K: 'static, C: Poolable> ConnectionPool<K, C> {
         match spawn_blocking(move || pool.get()).await.unwrap() {
             Ok(c) => {
                 Ok(Connection {
+                    pool: self.clone(),
                     connection: Some(c),
                     _permit: Some(permit),
                     _marker: PhantomData,
@@ -866,22 +899,37 @@ impl<K: 'static, C: Poolable> ConnectionPool<K, C> {
     }
 }
 
-impl<K, C: Poolable> Connection<K, C> {
+impl<K, C: Poolable> Clone for ConnectionPool<K, C> {
+    fn clone(&self) -> Self {
+        Self {
+            pool: self.pool.clone(),
+            semaphore: self.semaphore.clone(),
+            _marker: PhantomData,
+        }
+    }
+}
+
+impl<K: 'static, C: Poolable> Connection<K, C> {
     #[inline]
-    pub async fn run<F, R>(&mut self, f: F) -> R
+    pub async fn run<F, R>(self, f: F) -> R
         where F: FnOnce(&mut C) -> R + Send + 'static,
               R: Send + 'static,
     {
-        let mut conn = self.connection.take()
-            .expect("Connection unexpectedly missing. Was run() called without being awaited?");
+        tokio::task::spawn_blocking(move || {
+            // 'self' contains a semaphore permit that should be held throughout
+            // this entire spawned task. Explicitly move it, so that a
+            // refactoring won't accidentally release the permit too early.
+            let mut this: Self = self;
 
-        let (value, conn) = tokio::task::spawn_blocking(move || {
-            let value = f(&mut conn);
-            (value, conn)
-        }).await.expect("failed to spawn a blocking task to use a pooled connection");
+            let mut conn = this.connection.take()
+                .expect("internal invariant broken: self.connection is Some");
+            f(&mut conn)
+        }).await.expect("failed to spawn a blocking task to use a pooled connection")
+    }
 
-        self.connection = Some(conn);
-        value
+    #[inline]
+    pub async fn clone(&mut self) -> Result<Self, ()> {
+        self.pool.get().await
     }
 }
 
--- a/contrib/lib/src/databases.rs
+++ b/contrib/lib/src/databases.rs
@@ -871,7 +871,16 @@ impl<K: 'static, C: Poolable> ConnectionPool<K, C> {
     async fn get(&self) -> Result<Connection<K, C>, ()> {
         let permit = self.semaphore.clone().acquire_owned().await;
         let pool = self.pool.clone();
-        match spawn_blocking(move || pool.get()).await.unwrap() {
+
+        let result = match spawn_blocking(move || pool.get()).await {
+            Ok(c) => c,
+            Err(e) => match e.try_into_panic() {
+                Ok(panic) => std::panic::resume_unwind(panic),
+                Err(_) => unreachable!("spawn_blocking tasks are never canceled"),
+            }
+        };
+
+        match result {
             Ok(c) => {
                 Ok(Connection {
                     pool: self.clone(),
@@ -915,7 +924,7 @@ impl<K: 'static, C: Poolable> Connection<K, C> {
         where F: FnOnce(&mut C) -> R + Send + 'static,
               R: Send + 'static,
     {
-        tokio::task::spawn_blocking(move || {
+        let spawn_result = tokio::task::spawn_blocking(move || {
             // 'self' contains a semaphore permit that should be held throughout
             // this entire spawned task. Explicitly move it, so that a
             // refactoring won't accidentally release the permit too early.
@@ -923,8 +932,20 @@ impl<K: 'static, C: Poolable> Connection<K, C> {
 
             let mut conn = this.connection.take()
                 .expect("internal invariant broken: self.connection is Some");
+
             f(&mut conn)
-        }).await.expect("failed to spawn a blocking task to use a pooled connection")
+        }).await;
+
+        // Propagate any panics upwards.
+        //
+        // TODO: If we like this, extract + dedupe with above copy
+        match spawn_result {
+            Ok(val) => val,
+            Err(e) => match e.try_into_panic() {
+                Ok(panic) => std::panic::resume_unwind(panic),
+                Err(_) => unreachable!("spawn_blocking tasks are never canceled"),
+            }
+        }
     }
 
     #[inline]
--- a/contrib/lib/src/databases.rs
+++ b/contrib/lib/src/databases.rs
@@ -389,6 +389,7 @@ use std::marker::PhantomData;
 use std::sync::Arc;
 
 use rocket::config::{self, Value};
+use rocket::fairing::{AdHoc, Fairing};
 
 use rocket::tokio::sync::{OwnedSemaphorePermit, Semaphore};
 
@@ -779,22 +780,22 @@ impl Poolable for memcache::Client {
     }
 }
 
-#[doc(hidden)]
 /// Unstable internal details of generated code for the #[database] attribute.
 ///
 /// This type is implemented here instead of in generated code to ensure all
 /// types are properly checked.
+#[doc(hidden)]
 pub struct ConnectionPool<K, C: Poolable> {
     pool: r2d2::Pool<C::Manager>,
     semaphore: Arc<Semaphore>,
     _marker: PhantomData<fn() -> K>,
 }
 
-#[doc(hidden)]
 /// Unstable internal details of generated code for the #[database] attribute.
 ///
 /// This type is implemented here instead of in generated code to ensure all
 /// types are properly checked.
+#[doc(hidden)]
 pub struct Connection<K, C: Poolable> {
     connection: Option<r2d2::PooledConnection<C::Manager>>,
     _permit: Option<OwnedSemaphorePermit>,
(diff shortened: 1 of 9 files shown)


contrib/lib/src/databases.rs
```diff
@@ -780,6 +779,127 @@ impl Poolable for memcache::Client {
...
+
+#[::rocket::async_trait]
+impl<'a, 'r, K: 'static, C: Poolable> rocket::request::FromRequest<'a, 'r> for Connection<K, C> {
+    type Error = ();
+
+    #[inline]
+    async fn from_request(request: &'a rocket::request::Request<'r>) -> rocket::request::Outcome<Self, ()> {
+        let inner = ::rocket::try_outcome!(request.guard::<::rocket::State<'_, ConnectionPool<K, C>>>().await);
+
+        match inner.get().await {
```
SergioBenitez, line 896: As far as I can tell, this can never happen.

contrib/lib/src/databases.rs
```diff
@@ -780,6 +779,127 @@ impl Poolable for memcache::Client {
...
+                    ::rocket::logger::error_(&format!("{:?}", pool_error));
+                    Err(rocket)
+                },
+            }
+        })
+    }
+
+    async fn get(&self) -> Result<Connection<K, C>, ()> {
+        let permit = self.semaphore.clone().acquire_owned().await;
+        let pool = self.pool.clone();
```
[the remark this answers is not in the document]
jebrosen: It's implemented as an `Arc` clone. Cheapness isn't explicitly guaranteed, but the language "referencing the same state as `self`" is a slight indication towards yes: <https://docs.rs/r2d2/0.8.9/r2d2/struct.Pool.html#impl-Clone>

contrib/lib/src/databases.rs
```diff
@@ -780,6 +779,127 @@ impl Poolable for memcache::Client {
...
+        match spawn_blocking(move || pool.get()).await.unwrap() {
```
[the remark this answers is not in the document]
jebrosen: This and the other `unwrap`/`expect` panic if `pool.get()` panics (or, more rarely, if the entire runtime is shut down). I will look more closely and see if a panic is possible and when. Furthermore, I think `unwrap` and `expect` would both swallow the panic, and should perhaps be handled specifically - see <https://docs.rs/tokio/0.2.16/src/tokio/runtime/task/error.rs.html#146>.

contrib/lib/src/databases.rs
```diff
@@ -780,6 +779,127 @@ impl Poolable for memcache::Client {
...
+                        &format!("Failed to initialize pool for '{}'", config_name));
+                    ::rocket::logger::error_(&format!("{:?}", pool_error));
+                    Err(rocket)
+                },
+            }
+        })
+    }
+
+    async fn get(&self) -> Result<Connection<K, C>, ()> {
+        let permit = self.semaphore.clone().acquire_owned().await;
```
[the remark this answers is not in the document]
jebrosen: The `Drop` implementations for `SemaphorePermit` and `OwnedSemaphorePermit` return the permits. There is actually a minor problem here: if the future is polled once and then dropped, the semaphore permit will be dropped even though the blocking task is still holding the connection in the background. Moving the `permit` into (and back out of the closure) should avoid that.

jebrosen:
> This is excellent. Aside from my comments inline, I have the following three bigger concerns:
> 
> 1. `take`ing the connection in `run` can fail
>    This is really unfortunate. It means that the following will `panic!`:
>    ```rust
>    let x = conn.run(|_| ());
>    let y = conn.run(|_| ());
>    ```
>    
>    
> To prevent this from happening, we'd need to take the connection from the pool in `run()` instead of in the request guard. This is, of course, similar to what the previous implementation did, and in general isn't possible here since we'd need to borrow `self` in the `spawn_blocking` closure. However, what if `self` was moved into `run()` and subsequently into the closure?

The given example shouldn't panic, because `run()` does nothing at all until the first `poll` or `await` and also because `run` takes `&mut self` (previnting further `run()` calls while `x` is live). This is an example that I would expect a possible panic in:

```rust
conn.run(|_| ()).timeout(Duration::from_secs(3)).await;
// If the timeout finishes first, the connection will still be in the `spawn_blocking`
// closure - but the code that would return it into `conn` will never run.
conn.run(|_| ()).timeout(Duration::from_secs(3)).await;
```

I find this even more distressing than the example you gave, because it's pretty nearly impossible for the panic message to be understandable and applicable to all possible situations.

So I'll work on the implementation with `self`-by-value and `Clone` and see how that goes. I think we would want `async fn clone(&self) -> Result<Self, Error>` if we want the new connection to have its own permit, or synchronous `Clone` with a `Mutex` or queue to serialize the `spawn_blocking` calls coming from the same original wrapper.

> Then, assuming a handle to the pool was stored in `self`, we could call `self.pool.get();`, which should succeed given the semaphore, _and_ drop the connection in one `spawn_blocking` call while not running the risk of an unnecessary `panic!()`. The only other modifications would be to make `Connection` `Clone` and modify the documentation to specify that you must `clone()` if you call `run()` more than once.

Agreed except for one point: holding the semaphore permit does *not* guarantee we can get a connection immediately. `r2d2` does connection closing and acquisition on a small thread pool, and errors are reported via an optional callback that defaults to `log::error!` -- we don't set one, but custom implementations of `Poolable` might. If a connection is detected as being "broken", it is discarded and a replacement will be acquired either immediately or in a later `get()` call. `get()` might be called before the pool is replenished, so it waits up until the timeout (default 30 seconds) before returning an error.

> 2. The todo example should move .run() calls into task.rs.
> The handler code should remain unchanged, save for new .await calls.

Will do!

> 3. Service unavailable is never returned.
> Previously, if the pool was empty, we'd return a "service unavailable" error, which was nice. As far as I can tell, this can't happen here. I _think_ what we want is to `select` on a configurable timeout and the semaphore; if the former returns, issue a service unavailable.

`ServiceUnavailable` was not returned immediately if the pool was empty; it was returned if `get()` timed out while waiting to establish a new connection. At least one of the existing documentation comments was wrong on this point. This PR still waits for r2d2 to time out before returning `ServiceUnavailable`, plus or minus the incorrect handling of `unwrap` and `expect` I mentioned above.

---

I'm still working through the other review comments and these larger points; certainly some of them can be done today. Heads up - after tonight, I might not be able to push any more changes up until Friday afternoon.

SergioBenitez:
> `ServiceUnavailable` was not returned immediately if the pool was empty; it was returned if `get()` timed out while waiting to establish a new connection. At least one of the existing documentation comments was wrong on this point. This PR still waits for r2d2 to time out before returning `ServiceUnavailable`, plus or minus the incorrect handling of `unwrap` and `expect` I mentioned above.

While this is true, it only happens _after_ the semaphore is acquired. That is, you wait until a connection should be available to try to get a connection. Previously, you also timed out if the "semaphore" failed to be acquired in some amount of time.

jebrosen:
I've addressed most of the last review in some form. Two things I deliberately left out:

* changing the `todo` example. I started this, and found it to be pretty cumbersome when cmobining the calls to `Context::new` with the new `run()` that takes `self` by value
* Putting a timeout on semaphore acquisition. I didn't look at this yet, and it should be mostly independent from these changes which are already done.

One particular difference:
* `clone()` instead of `impl Clone`, because it needs to be `await`ed.
* `clone()` takes `&mut self`, because it holds an `&<Connection>` across an await. Many `Connection`s are not `Sync`, so `&<Connection>` is not `Send`, so `clone` taking `&self` makes it not callable in most places.
  * Possible alternative: have `clone` take `&self`, and place the connection inside of a [SyncWrapper](https://docs.rs/sync_wrapper/0.1.0/sync_wrapper/struct.SyncWrapper.html)

I have one small misgiving that may be worth documenting explicitly: calling `clone()` repeatedly to get `max_size` connections in the same route will "plug up" the pool, wait for a timeout on the next `clone()`, and then (probably) cause a cascade of timeouts in other routes. Combined with `self`-by-value, multiple connections in one route seems more likely than before so this might occur by accident. Although to be honest I think (hope?) it's relatively unlikely that anyone would write that code.

SergioBenitez:
@jebrosen I'm okay with these trade-offs. My initial proposal for a re-do made `run` fallible by moving the `pool.get()` call into `run()` but afforded a `spawn_blocking()` call, requiring a single `spawn_blocking()` call in all uses of a `Connection`. Your implementation requires two uses of `spawn_blocking()` but returns the user's value directly from `run()` and preserves the `ServiceUnavailable` functionality. I think this is okay. Hopefully we can push folks towards async databases in `0.5`.

I'd really like the `todo` example to be cleaned up in the way I'd proposed. Barring this change and remediations of `TODO`s in the correct PR, I think we should merge this.

contrib/lib/src/databases.rs
```diff
@@ -780,6 +810,184 @@ impl Poolable for memcache::Client {
...
+            Err(_) => {
+                error_!("Failed to get a database connection within the timeout.");
+                return Err(());
+            }
+        };
+
+        let pool = self.pool.clone();
+
+        // TODO: timeout duration
+        match run_blocking(move || pool.get_timeout(std::time::Duration::from_secs(0))).await {
```
jebrosen, line 899: We have two opportunities to set a timeout here:
* When acquiring the semaphore. This waits on other connection handles to be released.
* When retrieving a connection from the pool. This should usually complete immediately, but it could need to wait on the network if connections are being / need to be replenished.

Previously there was only one timeout, that waited on other connections to be released. I tried to do the most straightforward conversion, but should we relax the timeout on checking out a connection to a few seconds?

SergioBenitez:
How about this? _I think_ gives us the best ergonomics with the minimal downside that we call `spawn_blocking` twice, always.

```patch
From 2842ef78021261d2a74c7ee33e897e6883321f32 Mon Sep 17 00:00:00 2001
From: Sergio Benitez <sb@sergio.bz>
Date: Thu, 3 Sep 2020 00:12:00 -0700
Subject: [PATCH 2/2] Borrow in run().

---
 contrib/codegen/src/database.rs |  10 +--
 contrib/lib/src/databases.rs    | 123 +++++++++++++++-----------------
 examples/todo/src/main.rs       |  39 +++++-----
 examples/todo/src/task.rs       |  10 +--
 examples/todo/src/tests.rs      |  21 +++---
 5 files changed, 96 insertions(+), 107 deletions(-)

diff --git a/contrib/codegen/src/database.rs b/contrib/codegen/src/database.rs
index cb906f94..d0f4ca62 100644
--- a/contrib/codegen/src/database.rs
+++ b/contrib/codegen/src/database.rs
@@ -97,7 +97,7 @@ pub fn database_attr(attr: TokenStream, input: TokenStream) -> Result<TokenStrea
             /// closure will be passed an `&mut r2d2::PooledConnection`.
             /// `.await`ing the return value of this function yields the value
             /// returned by the closure.
-            pub async fn run<F, R>(self, f: F) -> R
+            pub async fn run<F, R>(&self, f: F) -> R
             where
                 F: FnOnce(&mut #conn_type) -> R + Send + 'static,
                 R: Send + 'static,
@@ -105,10 +105,10 @@ pub fn database_attr(attr: TokenStream, input: TokenStream) -> Result<TokenStrea
                 self.0.run(f).await
             }

-            /// Asynchronously acquires another connection from the connection pool.
-            pub async fn clone(&mut self) -> ::std::result::Result<Self, ()> {
-                self.0.clone().await.map(Self)
-            }
+            // /// Asynchronously acquires another connection from the connection pool.
+            // pub async fn clone(&mut self) -> ::std::result::Result<Self, ()> {
+            //     self.0.clone().await.map(Self)
+            // }
         }

         #[::rocket::async_trait]
diff --git a/contrib/lib/src/databases.rs b/contrib/lib/src/databases.rs
index 039c4da6..40aa9a7b 100644
--- a/contrib/lib/src/databases.rs
+++ b/contrib/lib/src/databases.rs
@@ -322,36 +322,35 @@
 //! # }
 //! ```
 //!
-//! `run()` takes the connection by value. To make multiple calls to run,
-//! obtain a second connection first with `clone()`:
-//!
-//! ```rust
-//! # #[macro_use] extern crate rocket;
-//! # #[macro_use] extern crate rocket_contrib;
-//! #
-//! # #[cfg(feature = "diesel_sqlite_pool")]
-//! # mod test {
-//! # use rocket_contrib::databases::diesel;
-//! # type Data = ();
-//! #[database("my_db")]
-//! struct MyDatabase(diesel::SqliteConnection);
-//!
-//! fn load_from_db(conn: &diesel::SqliteConnection) -> Data {
-//!     // Do something with connection, return some data.
-//!     # ()
-//! }
-//!
-//! #[get("/")]
-//! async fn my_handler(mut conn: MyDatabase) -> Data {
-//!     let cloned = conn.clone().await.expect("");
-//!     cloned.run(|c| load_from_db(c)).await;
-//!
-//!     // Do something else
-//!     conn.run(|c| load_from_db(c)).await;
-//! }
-//! # }
-//! ```
-//!
+//! // `run()` serializes access to the connection. To create an independent
+//! // conection to the same database, use `clone()`.
+//!
+//! // ```rust
+//! // # #[macro_use] extern crate rocket;
+//! // # #[macro_use] extern crate rocket_contrib;
+//! // #
+//! // # #[cfg(feature = "diesel_sqlite_pool")]
+//! // # mod test {
+//! // # use rocket_contrib::databases::diesel;
+//! // # type Data = ();
+//! // #[database("my_db")]
+//! // struct MyDatabase(diesel::SqliteConnection);
+//!
+//! // fn load_from_db(conn: &diesel::SqliteConnection) -> Data {
+//! //     // Do something with connection, return some data.
+//! //     # ()
+//! // }
+//!
+//! // #[get("/")]
+//! // async fn my_handler(mut conn: MyDatabase) -> Data {
+//! //     let cloned = conn.clone().await.expect("");
+//! //     cloned.run(|c| load_from_db(c)).await;
+//!
+//! //     // Do something else
+//! //     conn.run(|c| load_from_db(c)).await;
+//! // }
+//! // # }
+//! // ```
 //!
 //! # Database Support
 //!
@@ -418,10 +417,11 @@ use std::sync::Arc;

 use rocket::config::{self, Value};
 use rocket::fairing::{AdHoc, Fairing};
+use rocket::request::{Request, Outcome, FromRequest};
+use rocket::outcome::IntoOutcome;
 use rocket::http::Status;
-use rocket::request::Outcome;

-use rocket::tokio::sync::{OwnedSemaphorePermit, Semaphore};
+use rocket::tokio::sync::{OwnedSemaphorePermit, Semaphore, Mutex};

 use self::r2d2::ManageConnection;

@@ -828,28 +828,25 @@ pub struct ConnectionPool<K, C: Poolable> {
 #[doc(hidden)]
 pub struct Connection<K, C: Poolable> {
     pool: ConnectionPool<K, C>,
-    connection: Option<r2d2::PooledConnection<C::Manager>>,
+    connection: Arc<Mutex<Option<r2d2::PooledConnection<C::Manager>>>>,
     _permit: Option<OwnedSemaphorePermit>,
     _marker: PhantomData<fn() -> K>,
 }

-// A wrapper around spawn_blocking that propagates panics to the calling code
+// A wrapper around spawn_blocking that propagates panics to the calling code.
 async fn run_blocking<F, R>(job: F) -> R
-where
-    F: FnOnce() -> R + Send + 'static,
-    R: Send + 'static,
+    where F: FnOnce() -> R + Send + 'static, R: Send + 'static,
 {
     match tokio::task::spawn_blocking(job).await {
         Ok(ret) => ret,
         Err(e) => match e.try_into_panic() {
             Ok(panic) => std::panic::resume_unwind(panic),
-            Err(_) => unreachable!("spawn_blocking tasks are never canceled"),
+            Err(_) => unreachable!("spawn_blocking tasks are never cancelled"),
         }
     }
 }

 impl<K: 'static, C: Poolable> ConnectionPool<K, C> {
-    #[inline]
     pub fn fairing(fairing_name: &'static str, config_name: &'static str) -> impl Fairing {
         AdHoc::on_attach(fairing_name, move |mut rocket| async move {
             let config = database_config(config_name, rocket.config().await);
@@ -900,7 +897,7 @@ impl<K: 'static, C: Poolable> ConnectionPool<K, C> {
             Ok(c) => {
                 Ok(Connection {
                     pool: self.clone(),
-                    connection: Some(c),
+                    connection: Arc::new(Mutex::new(Some(c))),
                     _permit: Some(permit),
                     _marker: PhantomData,
                 })
@@ -936,52 +933,48 @@ impl<K, C: Poolable> Clone for ConnectionPool<K, C> {

 impl<K: 'static, C: Poolable> Connection<K, C> {
     #[inline]
-    pub async fn run<F, R>(self, f: F) -> R
+    pub async fn run<F, R>(&self, f: F) -> R
         where F: FnOnce(&mut C) -> R + Send + 'static,
               R: Send + 'static,
     {
+        let mut connection = self.connection.clone().lock_owned().await;
         run_blocking(move || {
-            // 'self' contains a semaphore permit that should be held throughout
-            // this entire spawned task. Explicitly move it, so that a
-            // refactoring won't accidentally release the permit too early.
-            let mut this: Self = self;
-
-            let mut conn = this.connection.take()
+            let conn = connection.as_mut()
                 .expect("internal invariant broken: self.connection is Some");
-
-            f(&mut conn)
+            f(conn)
         }).await
     }

-    #[inline]
-    pub async fn clone(&mut self) -> Result<Self, ()> {
-        self.pool.get().await
-    }
+    // #[inline]
+    // pub async fn clone(&mut self) -> Result<Self, ()> {
+    //     self.pool.get().await
+    // }
 }

 impl<K, C: Poolable> Drop for Connection<K, C> {
     fn drop(&mut self) {
-        if let Some(conn) = self.connection.take() {
-            tokio::task::spawn_blocking(|| drop(conn));
-        }
+        let connection = self.connection.clone();
+        tokio::spawn(async move {
+            let mut connection = connection.lock_owned().await;
+            tokio::task::spawn_blocking(move || {
+                if let Some(conn) = connection.take() {
+                    drop(conn)
+                }
+            })
+        });
     }
 }

 #[rocket::async_trait]
-impl<'a, 'r, K: 'static, C: Poolable> rocket::request::FromRequest<'a, 'r> for Connection<K, C> {
+impl<'a, 'r, K: 'static, C: Poolable> FromRequest<'a, 'r> for Connection<K, C> {
     type Error = ();

     #[inline]
-    async fn from_request(request: &'a rocket::request::Request<'r>) -> Outcome<Self, ()> {
+    async fn from_request(request: &'a Request<'r>) -> Outcome<Self, ()> {
         match request.managed_state::<ConnectionPool<K, C>>() {
-            Some(inner) => {
-                match inner.get().await {
-                    Ok(c) => Outcome::Success(c),
-                    Err(()) => Outcome::Failure((Status::ServiceUnavailable, ())),
-                }
-            }
+            Some(c) => c.get().await.into_outcome(Status::ServiceUnavailable),
             None => {
-                error_!("Database fairing was not attached for {}", std::any::type_name::<K>());
+                error_!("Missing database fairing for `{}`", std::any::type_name::<K>());
                 Outcome::Failure((Status::InternalServerError, ()))
             }
         }
diff --git a/examples/todo/src/main.rs b/examples/todo/src/main.rs
index a37f51e6..36b9b5b3 100644
--- a/examples/todo/src/main.rs
+++ b/examples/todo/src/main.rs
@@ -7,6 +7,8 @@
 mod task;
 #[cfg(test)] mod tests;

+use std::fmt::Display;
+
 use rocket::Rocket;
 use rocket::fairing::AdHoc;
 use rocket::request::{Form, FlashMessage};
@@ -31,20 +33,20 @@ struct Context {
 }

 impl Context {
-    pub async fn err(conn: DbConn, msg: String) -> Context {
+    pub async fn err<M: Display>(conn: &DbConn, msg: M) -> Context {
         Context {
-            msg: Some(("error".to_string(), msg)),
+            msg: Some(("error".into(), msg.to_string())),
             tasks: Task::all(conn).await.unwrap_or_default()
         }
     }

-    pub async fn raw(conn: DbConn, msg: Option<(String, String)>) -> Context {
+    pub async fn raw(conn: &DbConn, msg: Option<(String, String)>) -> Context {
         match Task::all(conn).await {
             Ok(tasks) => Context { msg, tasks },
             Err(e) => {
                 error_!("DB Task::all() error: {}", e);
                 Context {
-                    msg: Some(("error".to_string(), "Couldn't access the task database.".to_string())),
+                    msg: Some(("error".into(), "Fail to access database.".into())),
                     tasks: vec![]
                 }
             }
@@ -57,7 +59,7 @@ async fn new(todo_form: Form<Todo>, conn: DbConn) -> Flash<Redirect> {
     let todo = todo_form.into_inner();
     if todo.description.is_empty() {
         Flash::error(Redirect::to("/"), "Description cannot be empty.")
-    } else if let Err(e) = Task::insert(todo, conn).await {
+    } else if let Err(e) = Task::insert(todo, &conn).await {
         error_!("DB insertion error: {}", e);
         Flash::error(Redirect::to("/"), "Todo could not be inserted due an internal error.")
     } else {
@@ -66,27 +68,23 @@ async fn new(todo_form: Form<Todo>, conn: DbConn) -> Flash<Redirect> {
 }

 #[put("/<id>")]
-async fn toggle(id: i32, mut conn: DbConn) -> Result<Redirect, Template> {
-    // TODO
-    let conn2 = conn.clone().await.unwrap();
-    match Task::toggle_with_id(id, conn).await {
+async fn toggle(id: i32, conn: DbConn) -> Result<Redirect, Template> {
+    match Task::toggle_with_id(id, &conn).await {
         Ok(_) => Ok(Redirect::to("/")),
         Err(e) => {
             error_!("DB toggle({}) error: {}", id, e);
-            Err(Template::render("index", Context::err(conn2, "Failed to toggle task.".to_string()).await))
+            Err(Template::render("index", Context::err(&conn, "Failed to toggle task.").await))
         }
     }
 }

 #[delete("/<id>")]
-async fn delete(id: i32, mut conn: DbConn) -> Result<Flash<Redirect>, Template> {
-    // TODO
-    let conn2 = conn.clone().await.unwrap();
-    match Task::delete_with_id(id, conn).await {
+async fn delete(id: i32, conn: DbConn) -> Result<Flash<Redirect>, Template> {
+    match Task::delete_with_id(id, &conn).await {
         Ok(_) => Ok(Flash::success(Redirect::to("/"), "Todo was deleted.")),
         Err(e) => {
             error_!("DB deletion({}) error: {}", id, e);
-            Err(Template::render("index", Context::err(conn2, "Failed to delete task.".to_string()).await))
+            Err(Template::render("index", Context::err(&conn, "Failed to delete task.").await))
         }
     }
 }
@@ -94,20 +92,19 @@ async fn delete(id: i32, mut conn: DbConn) -> Result<Flash<Redirect>, Template>
 #[get("/")]
 async fn index(msg: Option<FlashMessage<'_, '_>>, conn: DbConn) -> Template {
     let msg = msg.map(|m| (m.name().to_string(), m.msg().to_string()));
-    Template::render("index", Context::raw(conn, msg).await)
+    Template::render("index", Context::raw(&conn, msg).await)
 }

 async fn run_db_migrations(mut rocket: Rocket) -> Result<Rocket, Rocket> {
-    let conn = DbConn::get_one(rocket.inspect().await).await.expect("database connection");
-    conn.run(|c| {
-        match embedded_migrations::run(c) {
+    DbConn::get_one(rocket.inspect().await).await
+        .expect("database connection")
+        .run(|c| match embedded_migrations::run(c) {
             Ok(()) => Ok(rocket),
             Err(e) => {
                 error!("Failed to run database migrations: {:?}", e);
                 Err(rocket)
             }
-        }
-    }).await
+        }).await
 }

 #[launch]
diff --git a/examples/todo/src/task.rs b/examples/todo/src/task.rs
index 0e9ad8a8..98726819 100644
--- a/examples/todo/src/task.rs
+++ b/examples/todo/src/task.rs
@@ -29,14 +29,14 @@ pub struct Todo {
 }

 impl Task {
-    pub async fn all(conn: DbConn) -> QueryResult<Vec<Task>> {
+    pub async fn all(conn: &DbConn) -> QueryResult<Vec<Task>> {
         conn.run(|c| {
             all_tasks.order(tasks::id.desc()).load::<Task>(c)
         }).await
     }

     /// Returns the number of affected rows: 1.
-    pub async fn insert(todo: Todo, conn: DbConn) -> QueryResult<usize> {
+    pub async fn insert(todo: Todo, conn: &DbConn) -> QueryResult<usize> {
         conn.run(|c| {
             let t = Task { id: None, description: todo.description, completed: false };
             diesel::insert_into(tasks::table).values(&t).execute(c)
@@ -44,7 +44,7 @@ impl Task {
     }

     /// Returns the number of affected rows: 1.
-    pub async fn toggle_with_id(id: i32, conn: DbConn) -> QueryResult<usize> {
+    pub async fn toggle_with_id(id: i32, conn: &DbConn) -> QueryResult<usize> {
         conn.run(move |c| {
             let task = all_tasks.find(id).get_result::<Task>(c)?;
             let new_status = !task.completed;
@@ -54,13 +54,13 @@ impl Task {
     }

     /// Returns the number of affected rows: 1.
-    pub async fn delete_with_id(id: i32, conn: DbConn) -> QueryResult<usize> {
+    pub async fn delete_with_id(id: i32, conn: &DbConn) -> QueryResult<usize> {
         conn.run(move |c| diesel::delete(all_tasks.find(id)).execute(c)).await
     }

     /// Returns the number of affected rows.
     #[cfg(test)]
-    pub async fn delete_all(conn: DbConn) -> QueryResult<usize> {
+    pub async fn delete_all(conn: &DbConn) -> QueryResult<usize> {
         conn.run(|c| diesel::delete(all_tasks).execute(c)).await
     }
 }
diff --git a/examples/todo/src/tests.rs b/examples/todo/src/tests.rs
index 73930b80..a184945b 100644
--- a/examples/todo/src/tests.rs
+++ b/examples/todo/src/tests.rs
@@ -19,9 +19,8 @@ macro_rules! run_test {
             let rocket = super::rocket();
             let $client = Client::new(rocket).await.expect("Rocket client");
             let db = super::DbConn::get_one($client.cargo()).await;
-            let mut $conn = db.expect("failed to get database connection for testing");
-            let delete_conn = $conn.clone().await.expect("failed to get a second database connection for testing");
-            Task::delete_all(delete_conn).await.expect("failed to delete all tasks for testing");
+            let $conn = db.expect("failed to get database connection for testing");
+            Task::delete_all(&$conn).await.expect("failed to delete all tasks for testing");

             $block
         })
@@ -32,7 +31,7 @@ macro_rules! run_test {
 fn test_insertion_deletion() {
     run_test!(|client, conn| {
         // Get the tasks before making changes.
-        let init_tasks = Task::all(conn.clone().await.unwrap()).await.unwrap();
+        let init_tasks = Task::all(&conn).await.unwrap();

         // Issue a request to insert a new task.
         client.post("/todo")
@@ -42,7 +41,7 @@ fn test_insertion_deletion() {
             .await;

         // Ensure we have one more task in the database.
-        let new_tasks = Task::all(conn.clone().await.unwrap()).await.unwrap();
+        let new_tasks = Task::all(&conn).await.unwrap();
         assert_eq!(new_tasks.len(), init_tasks.len() + 1);

         // Ensure the task is what we expect.
@@ -54,7 +53,7 @@ fn test_insertion_deletion() {
         client.delete(format!("/todo/{}", id)).dispatch().await;

         // Ensure it's gone.
-        let final_tasks = Task::all(conn).await.unwrap();
+        let final_tasks = Task::all(&conn).await.unwrap();
         assert_eq!(final_tasks.len(), init_tasks.len());
         if final_tasks.len() > 0 {
             assert_ne!(final_tasks[0].description, "My first task");
@@ -72,16 +71,16 @@ fn test_toggle() {
             .dispatch()
             .await;

-        let task = Task::all(conn.clone().await.unwrap()).await.unwrap()[0].clone();
+        let task = Task::all(&conn).await.unwrap()[0].clone();
         assert_eq!(task.completed, false);

         // Issue a request to toggle the task; ensure it is completed.
         client.put(format!("/todo/{}", task.id.unwrap())).dispatch().await;
-        assert_eq!(Task::all(conn.clone().await.unwrap()).await.unwrap()[0].completed, true);
+        assert_eq!(Task::all(&conn).await.unwrap()[0].completed, true);

         // Issue a request to toggle the task; ensure it's not completed again.
         client.put(format!("/todo/{}", task.id.unwrap())).dispatch().await;
-        assert_eq!(Task::all(conn).await.unwrap()[0].completed, false);
+        assert_eq!(Task::all(&conn).await.unwrap()[0].completed, false);
     })
 }

@@ -91,7 +90,7 @@ fn test_many_insertions() {

     run_test!(|client, conn| {
         // Get the number of tasks initially.
-        let init_num = Task::all(conn.clone().await.unwrap()).await.unwrap().len();
+        let init_num = Task::all(&conn).await.unwrap().len();
         let mut descs = Vec::new();

         for i in 0..ITER {
@@ -107,7 +106,7 @@ fn test_many_insertions() {
             descs.insert(0, desc);

             // Ensure the task was inserted properly and all other tasks remain.
-            let tasks = Task::all(conn.clone().await.unwrap()).await.unwrap();
+            let tasks = Task::all(&conn).await.unwrap();
             assert_eq!(tasks.len(), init_num + i + 1);

             for j in 0..i {
--
2.28.0
```

SergioBenitez:
Oh: we need to move the `permit` into the `spawn_blocking()` call in `Drop` so that it's held until after the connection is dropped.

contrib/lib/src/databases.rs
SergioBenitez: > We have two opportunities to set a timeout here:
> 
> * When acquiring the semaphore. This waits on other connection handles to be released.
> * When retrieving a connection from the pool. This should usually complete immediately, but it could need to wait on the network if connections are being / need to be replenished.
> 
> Previously there was only one timeout, that waited on other connections to be released. I tried to do the most straightforward conversion, but should we relax the timeout on checking out a connection to a few seconds?

Ideally the latter would be configurable, with a default of `0`, and should be set on the `r2d2::Builder` so that `get()` automatically uses it. The config revamp will make doing this much easier. I propose we make a note of this and wait until I can get those changes in.

SergioBenitez:
For posterity, here are the UI test fixes:

```patch
From d90fca0ec13177449ab861d8802c89cf10fa31d0 Mon Sep 17 00:00:00 2001
From: Sergio Benitez <sb@sergio.bz>
Date: Wed, 2 Sep 2020 22:47:49 -0700
Subject: [PATCH 1/2] Fix UI tests for latest nightly.

---
 .../codegen/tests/ui-fail-nightly/database-types.stderr   | 8 ++++----
 1 file changed, 4 insertions(+), 4 deletions(-)

diff --git a/contrib/codegen/tests/ui-fail-nightly/database-types.stderr b/contrib/codegen/tests/ui-fail-nightly/database-types.stderr
index ce769c9f..643d93b6 100644
--- a/contrib/codegen/tests/ui-fail-nightly/database-types.stderr
+++ b/contrib/codegen/tests/ui-fail-nightly/database-types.stderr
@@ -4,9 +4,9 @@ error[E0277]: the trait bound `Unknown: rocket_contrib::databases::Poolable` is
 7   | struct A(Unknown);
     |          ^^^^^^^ the trait `rocket_contrib::databases::Poolable` is not implemented for `Unknown`
     |
-   ::: $WORKSPACE/contrib/lib/src/databases.rs:832:29
+   ::: $WORKSPACE/contrib/lib/src/databases.rs
     |
-832 | pub struct Connection<K, C: Poolable> {
+    | pub struct Connection<K, C: Poolable> {
     |                             -------- required by this bound in `rocket_contrib::databases::Connection`

 error[E0277]: the trait bound `std::vec::Vec<i32>: rocket_contrib::databases::Poolable` is not satisfied
@@ -15,7 +15,7 @@ error[E0277]: the trait bound `std::vec::Vec<i32>: rocket_contrib::databases::Po
 10  | struct B(Vec<i32>);
     |          ^^^^^^^^ the trait `rocket_contrib::databases::Poolable` is not implemented for `std::vec::Vec<i32>`
     |
-   ::: $WORKSPACE/contrib/lib/src/databases.rs:832:29
+   ::: $WORKSPACE/contrib/lib/src/databases.rs
     |
-832 | pub struct Connection<K, C: Poolable> {
+    | pub struct Connection<K, C: Poolable> {
     |                             -------- required by this bound in `rocket_contrib::databases::Connection`
--
2.28.0
```

contrib/lib/src/databases.rs
```diff
@@ -320,12 +316,42 @@
...
 //! #[get("/")]
-//! fn my_handler(conn: MyDatabase) -> Data {
-//!     load_from_db(&*conn)
+//! async fn my_handler(mut conn: MyDatabase) -> Data {
+//!     conn.run(|c| load_from_db(c)).await
 //! }
 //! # }
 //! ```
 //!
+//! // `run()` serializes access to the connection. To create an independent
```
igalic, line 325: why is this commented out instead of deleted?
or is this just a reminder that the documentation here is wrong and needs fixing?

igalic (commented):
🙋🏻‍♀️

contrib/lib/src/databases.rs
SergioBenitez: Right; I commented out to indicate this doc was no longer valid.

SergioBenitez (approved):

SergioBenitez:
Merged in https://github.com/SergioBenitez/Rocket/commit/bc8c5b9ee217e6acf16f9b6d855fb19ee1867a5d. Woo!

Issue #989, opened 2019-05-05
Static files should be sent with Last-Modified header


inferiorhumanorgans:
I realize that the intended use case for Rocket involves a proxy that would typically handle static content.  However as Rocket has a minimal NamedFile responder I think it would be a worthwhile and fairly lightweight add to both include the file's mtime (Last-Modified response header) and handle the If-Modified-Since request header.  This would allow browsers to cache static content as needed.

To that end I've made a [proof of concept](https://github.com/inferiorhumanorgans/Rocket/tree/iho/static-file-caching) that appears to work as intended with Firefox.

SergioBenitez:
I believe this is a dupe of #95, though this one has an implementation! Still, let's use the other one, for lineage's sake.

PR #1376, opened 2020-07-16
add support for Last-Modified and If-Modified-Since headers

qmatias:
#989, #95 

Adds support for Last-Modified and If-Modified-Since headers.

- If a `NamedFile` is requested with the If-Modified-Since header set and the file was last modified before the date in the header, the server instead responds with 304 Not Modified
- If that header is not present, the file is sent over with the Last-Modified header (on supported platforms/OS)

I'm pretty new to rocket so let me know if I'm doing anything wrong and if the style is ok! (no rustfmt... shame :stuck_out_tongue: )

### DIFF
--- a/core/lib/src/response/named_file.rs
+++ b/core/lib/src/response/named_file.rs
@@ -1,4 +1,4 @@
-use std::io;
+use std::{fs, io};
 use std::path::{Path, PathBuf};
 use std::ops::{Deref, DerefMut};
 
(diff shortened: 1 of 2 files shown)


jebrosen (commented):
Generally speaking, I'm in favor of having this or at least something similar. I have not audited the date comparison code, but that will *really* need to be correct. I do have two wider concerns:

* Reliability of timestamps from the filesystem is not necessarily great; there are many pitfalls here. `StaticFiles`/`NamedFile` seems like an okay place to do this with a few safeguards, such as ignoring filesystem dates that are from the future.
* Interactions with other preconditions. This PR seems to ignore most of <https://tools.ietf.org/html/rfc7232#section-6>. That might be okay! For example, maybe NamedFile specifically does not ever have to worry about whether `If-None-Match` is present, since `NamedFile` never generates an `ETag` header. But like the date comparisons, this is an easy thing to get subtly wrong and needs some scrutiny.
* And one very minor downside / bikeshed: with this implementation, we would effectively remove "File with extension-based MIME detection" and replace it with "File with extension-based MIME detection and filesystem-based time checking"; maybe this should be a builder or a separate type so that the two features aren't tied together.

SergioBenitez:
Checking in: what's the status here? We also have #1219; should one be closed in favor of another?

jebrosen:
I hadn't realized there were two PRs for this. I'm closing this one in favor of #1219, which is opt-in (resolving most of my concerns about reliability and composability) and more complete (including examples and tests).

PR #1378, opened 2020-07-18
Upgrade (almost) all dependencies

### LINKED ISSUE #1341: Update rusqlite
I am using rocket_contrib with the sqlite_pool feature. The current version of rusqlite is 0.23, but rocket uses 0.14. As you can see, there have been many changes since 0.14 https://github.com/rusqlite/rusqlite/releases.

jebrosen:
Includes a modified version of #1349.
Fixes #1341.
CC #1189.

This PR upgrades all direct dependencies except `toml` and `pear` (see #1353). It reduces the overall number of packages in `Cargo.lock` from 466 to 390 by my count.

* Removes `rusted_cypher`. It has a few old dependencies including hyper 0.10 and appears to be unmaintained for a few years now.
* Removes `mongodb`. The latest version of `mongodb` has `async`/`await` support and its own connection pooling, making it doubly unsuitable for the current implementation of `#[database]`.

### DIFF
--- a/core/http/src/parse/uri/tables.rs
+++ b/core/http/src/parse/uri/tables.rs
@@ -1,3 +1,39 @@
+use percent_encoding::AsciiSet;
+
+// Generates an AsciiSet from a full character table:
+//
+// table_to_ascci_set!( ExistingAsciiSet, 0, [ 0, 1, b'x', ... ] );
+//
+// A 0 or 1 token in the ith position indicates that the ASCII character i
+// should be included in the set. Any other literal in the ith position
+// indicates that the ASCII character with value i should not be included in the
+// set.
+//
+// The table's last index must be 127 or earlier. All values in the original
+// set, up to the end of the passed-in table, are overwritten.
+macro_rules! table_to_ascii_set {
+    ($base:expr, $i:expr, [ 0,           $($rest:tt,)* ]) => { table_to_ascii_set!($base.add($i),    $i+1, [ $($rest,)* ]); };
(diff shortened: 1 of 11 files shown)


core/http/src/parse/uri/tables.rs
```diff
@@ -1,3 +1,39 @@
...
+    (
+        const $name:ident: [u8; $size:expr] = [ $($block1:tt)* ] [ $($block2:tt)* ];
+        pub const $setname:ident: AsciiSet;
+    ) => {
+        const $name: [u8; $size] = [ $($block1)* $($block2)* ] ;
+        pub const $setname: AsciiSet = table_to_ascii_set!(percent_encoding::CONTROLS, 0, [ $($block1)* ]);
+    };
+}
+
+table_and_asciiset! {
```
jebrosen, line 36: I'm not *super* happy with this macro invocation (and `table_to_ascii_set` above). Most of the awkwardness is that `AsciiSet`'s public API only has `.add(u8)`, `.remove(u8)` and there is no public constructor for an "empty" `AsciiSet`. The other awkwardness is that you can't call `add` or `remove` with any `u8` value above `127` apparently due to `AsciiSet`'s bitpacking optimizations.

Some alternatives I thought of:

* The `const_if_match` feature as originally proposed in #1349, which is stabilized but in one `stable` release later than we want for 0.5.
* Have introspection APIs added to `AsciiSet`, which could make `PATH_CHARS` unnecessary since `AsciiSet` could be queried equivalently. `AsciiSet`'s packed representation might lose some performance difference compared to the `PATH_CHARS` array, however, and we would need to keep a macro like this one anyway.

SergioBenitez:
This is absolutely fantastic! Very clever, if unfortunately necessary, macro!

PR #1381, opened 2020-07-20
[Fix] common misspelling errors

Necmttn:
For reference: https://en.wikipedia.org/wiki/Wikipedia:Lists_of_common_misspellings/For_machines


core/codegen/src/bang/uri.rs
```diff
@@ -199,7 +199,7 @@ fn explode_query<'a, I: Iterator<Item = (&'a Ident, &'a Type, &'a ArgExpr)>>(
     Some(quote!(#uri_mod::UriArgumentsKind::Dynamic(&[#(#dyn_exprs),*])))
 }
 
-// Returns an Origin URI with the mount point and route path concatinated. The
+// Returns an Origin URI with the mount point and route path contaminated. The
```
jebrosen, line 202: This one should actually be `concatenated`.

SergioBenitez:
(Mostly) merged in fde6eda915460972a273763da28e6a5caaa3121d. Thank you!

PR #1384, opened 2020-07-22
cargo vendor fails due to missing file

pandaman64:
I'm using [Nix](https://nixos.org/) to build a package with Rocket, which depends on `cargo vendor` to copy packages.
However, it fails with the current master because `update-references.sh` is missing.
This PR readds the file. I Confirmed that `cargo vendor` worked with this change.

### DIFF
--- a/core/codegen/tests/ui-fail/update-references.sh
+++ b/core/codegen/tests/ui-fail/update-references.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+#
+# Copyright 2015 The Rust Project Developers. See the COPYRIGHT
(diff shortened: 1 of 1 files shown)


igalic:
when / why was that file removed?
maybe we should also remove the reference to it?

pandaman64:
https://github.com/SergioBenitez/Rocket/commit/185840320315a6803bf4e137b9372b5b0b44ba48 removes the referent and adds the references.

jebrosen:
That looks like a mistake; `update-references.sh` is superseded by `TRYBUILD=overwrite`.

Closing in favor of #1385, so we only have one issue open for this.

PR #1386, opened 2020-07-22
impl (de)serialize for UUID

### LINKED ISSUE #1370: rocket_contrib::uuid::Uuid missing serde Serialize/Deserialize?
Feature request: adding `serde::{Serialize, Deserialize}` impls for `rocket_contrib::uuid::Uuid`

## Feature Requests

It's probably expected functionality and can be very useful.

Serializing UUIDs to return it as a json document, for example

Implementing it yourself isn't super hard, but it would be nice if it was available (perhaps as a feature?) in `rocket_contrib`

meltinglava:
closes #1370

TODO: Think some discussion for the other options in the issue should be
discussed there before merging. This PR is mostly there for checking
what the implimitation could look like.

Also think we hit a limitation in cargo where we cant describe that we
want the feature rocket_conrtib::serde to be connected to the feature
flag of uuid::serde.

### DIFF
--- a/contrib/lib/src/uuid.rs
+++ b/contrib/lib/src/uuid.rs
@@ -64,6 +64,7 @@ type ParseError = <self::uuid_crate::Uuid as FromStr>::Err;
 ///     format!("User ID: {}", id)
 /// }
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Debug)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize), serde(transparent))]
 pub struct Uuid(uuid_crate::Uuid);
 
 impl Uuid {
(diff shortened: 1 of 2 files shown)


drusellers:
Is the goal of this feature to implement stand alone UUID from a request - like a path parameter or query string? Or is it also for the body of a request? Could they be different features?

The path param or query param or form param could be implemented by just parsing the uuid.

The request body would be a part of the serde-configuration yeah?

meltinglava:
I think [this comment of the issue](https://github.com/SergioBenitez/Rocket/issues/1370#issuecomment-708011976) describes it quite well.

SergioBenitez:
This LGTM. Can you rebase on `master`?

meltinglava:
Done

contrib/lib/src/uuid.rs
```diff
@@ -64,6 +64,7 @@ type ParseError = <self::uuid_crate::Uuid as FromStr>::Err;
 ///     format!("User ID: {}", id)
 /// }
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Debug)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize), serde(transparent))]
```
SergioBenitez, line 67: Given that the `serde` feature is enabled if this code is being compiled, is this `cfg_attr` a no-op?

meltinglava: We can not name it `uuid` as it would hit https://github.com/rust-lang/cargo/issues/1286#issue-57144439. What should we name it then? `uuid_integration` or `uuid_serde` or something else?

SergioBenitez: You can indeed name it `uuid` as long as the crate is named something else. See, for example, how `rocket_http` does this for its `serde` feature:

https://github.com/SergioBenitez/Rocket/blob/3f56490663a395d348447918ff09f20ba9c83efc/core/http/Cargo.toml#L17-L21

https://github.com/SergioBenitez/Rocket/blob/3f56490663a395d348447918ff09f20ba9c83efc/core/http/Cargo.toml#L48-L53

SergioBenitez:
Awesome. Merged in 165f4f59b3eedfdaa31024eb9dc24bc4ec4fc575. Thank you!

PR #1388, opened 2020-07-23
Remove superfluous fairing lifetimes

jcdickinson:


core/lib/src/fairing/mod.rs
```diff
@@ -453,7 +453,7 @@ pub trait Fairing: Send + Sync + 'static {
     ///
     /// The default implementation of this method does nothing.
     #[allow(unused_variables)]
-    async fn on_response<'a>(&'a self, req: &'a Request<'_>, res: &'a mut Response<'_>) {}
+    async fn on_response(&self, req: &Request<'_>, res: &mut Response<'_>) {}
```
SergioBenitez, line 456:
```suggestion
    async fn on_response<'r>(&self, req: &'r Request<'_>, res: &mut Response<'r>) {}
```

I think we want this, so that the `Response` can borrow from the `Request`.

jcdickinson: Awesome feedback, that's a great lesson in niche lifetime design.

core/lib/src/fairing/fairings.rs
```diff
@@ -66,7 +66,7 @@ impl Fairings {
     }
 
     #[inline(always)]
-    pub async fn handle_response<'r>(&self, request: &Request<'r>, response: &mut Response<'r>) {
+    pub async fn handle_response<'r>(&self, request: &'r Request<'_>, response: &mut Response<'r>) {
```
jcdickinson, line 69: This had to be adjusted, is it still good?

core/lib/src/fairing/mod.rs
```diff
@@ -452,8 +452,13 @@ pub trait Fairing: Send + Sync + 'static {
     /// ## Default Implementation
     ///
     /// The default implementation of this method does nothing.
+    ///
+    /// ## Lifetimes
+    ///
+    /// `on_response` has a single lifetime `'r`. This lifetime allows `res`
+    /// to borrow data from the `req`.
```
SergioBenitez, line 459: I don't think we need this documented; the lifetimes themselves document this.

SergioBenitez:
The last PR seems to include `rustfmt` changes. Can you revert those?

SergioBenitez:
Merged in f3beb68491d1fb2ab506271fdd6e3238f0bb7162. Thank you!

PR #1390, opened 2020-07-25
Remove appendix from Apache license

mikong:
From what I understand, the boilerplate notice in the Apache license's appendix is just meant to be part of the instruction on how to apply the license on a program file and not to actually be modified with the copyright year and owner of the open source project (as in ac4cca7). It should be kept as:

```
Copyright [yyyy] [name of copyright owner]
```

But people often and understandably get confused by it and think it needs to be updated. To avoid confusion, we can simply remove the appendix. The license doesn't actually require that section in the distribution, and other projects have removed it (e.g. see rust-lang/rust#67734). As highlighted in that Rust PR, it is safe to remove:

> "License" shall mean the terms and conditions for use, reproduction,
> and distribution as defined by Sections 1 through 9 of this document.

It was removed from the Rust project for a different reason — they don't practice putting a license header at the top of each file so they removed it to avoid that confusion.

Alternatively, we can keep the appendix and just revert the change in ac4cca7. Let me know if you'd like me to update this PR to do just that. Or, we can keep things as it is, and just close this PR and move on to more important issues.

### DIFF
--- a/LICENSE-APACHE
+++ b/LICENSE-APACHE
@@ -174,28 +174,3 @@ TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
    of your accepting any such warranty or additional liability.
 
 END OF TERMS AND CONDITIONS
-
-APPENDIX: How to apply the Apache License to your work.
-
-   To apply the Apache License to your work, attach the following
-   boilerplate notice, with the fields enclosed by brackets "[]"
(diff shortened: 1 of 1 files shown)


SergioBenitez:
I think you're totally right. Still, I'm not sure why, but it gives me pause to modify the license, even if it seems totally okay. I'll think I'll just leave this as is for no reason other than this unjustified pause. Thank you for bringing this to my attention.

PR #1392, opened 2020-07-27
feat: use <_> as path parameters to ignore them

timokoesters:
I have no idea if any of this is okay

This also works like <_..> for multiple segments


timokoesters:
This doesn't work at all, let me try harder...

timokoesters:
Okay, this works!

SergioBenitez:
Merged in 8e8fb4cae82d2c070ae0ce013a25e8d7c816fca2. Thanks!

PR #1398, opened 2020-07-31
Change the template customization callback to return a Result

### LINKED ISSUE #1064: Can Template::custom() take an FnOnce?
Rocket version: 0.4.2
Steps I've taken to answer the question myself: looked at the source and mentally examined the needs of such a closure.
Documentation I believe should include an answer: none.

Basically, see title. Is there a need for `Template::custom()` to take `Fn` instead of `FnOnce`?

I would like to initialise some values and do error handling outside of the closure, then move a value inside the closure and use it there. Unfortunately, this is not possible, since the type is not `Copy` nor `Clone`.

```rust
fn main() {
  // not copy or clone
  let something = do_some_initialisation();

  rocket::ignite()
    .attach(Template::custom(move |engines| {
      // move `something` in here
    })
    .launch();
}
```

This is currently impossible due to the constraints of `Fn`.

If `Template::custom()` can take an `FnOnce`, I'd be happy to PR such a change. :)

jebrosen:
The closure passed to `Template::custom` can now return an error.

* If the closure returns an error during launch, the fairing itself returns `Err(rocket)` and the launch is canceled.
* If the closure later returns an error during template reload (in debug mode), the error is logged and the existing templates are left alone (same as with an error in reloading itself).

Closes #1064.

## Unresolved
- [x] The error type is initially `Box<dyn Error + Send + Sync>`. Those `Send + Sync` bounds are probably unnecessary, actually. Or maybe it should be an entirely different trait, like `Box<dyn Debug>`?

### DIFF
--- a/contrib/lib/src/templates/engine.rs
+++ b/contrib/lib/src/templates/engine.rs
@@ -41,6 +41,7 @@ pub(crate) trait Engine: Send + Sync + 'static {
 ///         // ...
 ///         .attach(Template::custom(|engines: &mut Engines| {
 ///             engines.tera.register_filter("my_filter", my_filter);
+///             Ok(())
 ///         }))
 ///         // ...
 ///         # ;
--- a/contrib/lib/src/templates/fairing.rs
+++ b/contrib/lib/src/templates/fairing.rs
@@ -89,7 +89,9 @@ mod context {
         /// have been changes since the last reload, all templates are
         /// reinitialized from disk and the user's customization callback is run
         /// again.
-        pub fn reload_if_needed<F: Fn(&mut Engines)>(&self, custom_callback: F) {
+        pub fn reload_if_needed<F>(&self, custom_callback: F)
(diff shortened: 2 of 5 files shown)


SergioBenitez:
This looks great. Does `Box<dyn Error>` suffice/compile? If so, let's go with that.

We'll need some tests to ensure this works correctly, at least for the `launch() -> panic!()` version.

jebrosen:
As far as I can tell `Box<dyn Error>` is usable, and I can't think of a useful reason to actually require `Send`, much less `Sync` (probably force of habit).

I added that missing test for an error during the first run of the callback. The error case during reload only results in log messages at runtime, which (as far as I know) we don't test anywhere already and would need some sort of dedicated test setup for - so I haven't added that one.

SergioBenitez:
Merged in 2f98299272a2a196a3d053918ebd542e03abfc70.

PR #1402, opened 2020-08-03
Update pastebin document

howard0su:
After we change to async, PasteId has to postfix with an explicit lifetime parameter. Change the document to reflect it.

### DIFF
--- a/site/guide/10-pastebin.md
+++ b/site/guide/10-pastebin.md
@@ -410,12 +410,13 @@ the `retrieve` route, preventing attacks on the `retrieve` route:
 ```rust
 # #[macro_use] extern crate rocket;
 
+# use std::borrow::Cow;
 # use rocket::tokio::fs::File;
 
-# type PasteId = usize;
+# type PasteId<'a> = Cow<'a, str>;
 
 #[get("/<id>")]
-async fn retrieve(id: PasteId) -> Option<File> {
+async fn retrieve(id: PasteId<'_>) -> Option<File> {
     let filename = format!("upload/{id}", id = id);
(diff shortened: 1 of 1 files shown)


SergioBenitez:
This is not correct. What's more, the guide has already been updated and is tested on every commit.

jebrosen:
I think this is worth a second look, since following the guide as written does result in a compilation error:

```rust
/// A _probably_ unique paste ID.
pub struct PasteId<'a>(Cow<'a, str>);

...

#[get("/<id>")]
async fn retrieve(id: PasteId) -> Option<File> {
    //                ^ error[E0726]: implicit elided lifetime not allowed here
    //                ^- help: indicate the anonymous lifetime: `<'_>`
    let filename = format!("upload/{id}", id = id);
    File::open(&filename).await.ok()
}

```

(I'm not sure what the CI failure is about - it failed on the `pastebin` *example*, and only on Windows).

SergioBenitez:
My sincerest apologies, @howard0su. I misunderstood the purpose of your PR and the issue it was attempting to resolve. I realize now that the guide fakes the type of the parameter, and the stand-in type is no longer representative of the real type.￼ Thank you, @jebrosen, for pointing out my mistake.

howard0su:
Never mind. I can do better job to write detailed description.

howard0su:
what shall i do before this can be merged?

Voodlaz (approved):

jebrosen (approved):

jebrosen:
This has been merged in 2040693. Thanks!

Issue #1404, opened 2020-08-05
Stabilizing TLS Support


klnusbaum:
rusttls recently recieved a [formal security audit](https://github.com/ctz/rustls/blob/master/audit/TLS-01-report.pdf). The outcome of the audit seems pretty encouraging. Specifically:

> Cure53 was unable to uncover any application-breaking security flaws. After spending thirty days on the scope in late May and early June of 2020, the team of auditors considered the general code quality to be exceptional and can attest to a solid impression left consistently by all scope items

At this point, would it make sense to remove some of the [scary warnings regarding TLS support](https://rocket.rs/v0.4/guide/configuration/#configuring-tls)?

igalic:
even if rocket supports a high quality TLS library, it's still lacking protections against many other attack vectors, that a well established proxy server offers, as well as more in depth protection by "simply" loading a WAF like [ModSecurity](https://www.modsecurity.org/).

howard0su:
Agree, we should first target to remove the TLS warning. 

How to protect from other attack is another big topic but different than this issue.

igalic:
how about rewording this

> **Warning:** Rocket's built-in TLS is **not** considered ready for production use. It is intended for development use *only*.

to something like

> **Attention:** Even though Rocket uses an [excellent built-in TLS library](https://github.com/ctz/rustls/blob/master/audit/TLS-01-report.pdf), that alone may not be enough to fully protect an application in production.

klnusbaum:
> how about rewording this
> 
> > **Warning:** Rocket's built-in TLS is **not** considered ready for production use. It is intended for development use _only_.
> 
> to something like
> 
> > **Attention:** Even though Rocket uses an [excellent built-in TLS library](https://github.com/ctz/rustls/blob/master/audit/TLS-01-report.pdf), that alone may not be enough to fully protect an application in production.

What do other web frameworks that have well vetted TLS libraries say in their documentation? I've looked at the documentation for Rails, Gorilla, and Django. None of them have any warnings like this. I think adding this extra information is actually distracting and takes away from the documentation about TLS specifically. IMHO if we want to talk more about security in general, we should instead create a dedicated page called "Securing Rocket Applications".

My vote is to wholesale remove the current warning.

PR #1407, opened 2020-08-10
Remove TLS warning

howard0su:
based on the discussion here #1404, remove the warning of TLS from the document. Rocket is using a production quality of TLS lib.


blaggacao:
See #1449 &mdash; maybe that should be supported before removing?

SergioBenitez:
The warning has been changed in master.

PR #1410, opened 2020-08-11
Adopt Tracing

hawkw:
This branch begins the process of migrating Rocket from using the `log`
crate for diagnostics to using `tracing`. 

The motivation for switching from `log` to `tracing` was discussed in
detail on issue #21, in particular, in [this comment][1] and later. In
summary, `tracing` allows detailed tracking of contexts in programs,
including asynchronous ones. This is essentially a structured,
machine-readable form of the contexts that Rocket's current logging
system conveys through indentation. Additionally, `tracing` data can be
consumed to emit data in a wide variety of formats, and is
backward-compatible with libraries that use the `log` crate.

This branch begins the migration by doing the following:

- [x] Adding a `trace` module with re-exports of core `tracing` APIs
- [x] Adding a `tracing_subscriber::Layer` implementation that performs 
      "Rocket-style" log formatting for `tracing` spans and events, and
      setting it as the default unless a `tracing` subscriber is set by
      the user
- [x] Replacing all existing uses of the `log` macros with their `tracing`
      equivalents.

Furthermore, I've _begun_ the process of converting the existing logging
to more idiomatic `tracing`. I've added `tracing` spans to several
top-level scopes in Rocket, converted some log messages to structured
events, and replaced some existing uses of indentation with `tracing`
spans. Hopefully, this can serve as an example for others to continue
the migration and add tracing instrumentation in new code.

When possible, I've tried to keep the logging format as close to 
Rocket's existing logging. I've made a few tweaks to help communicate
more of the structured data that `tracing` diagnostics record, but 
the emoji and stuff are all still there. Of course, I am open to making
additional changes to the default format if desired.

I've also re-exported some core `tracing` APIs to try to provide a`
batteries-included experience, which seems in keeping with Rocket's
philosophy. If other folks like this, we can add more re-exports; if
we don't, it's fine to remove them.

Potential future work that this PR unlocks includes:

- [ ] Adding Rocket middleware for adding additional user-provided
     `tracing` instrumentation.
- [ ] Adding middleware for integrating with distributed tracing
     using [`tracing-opentelemetry`][2] and trace propagation 
     headers like `b3`.
- [ ] Adding middleware for emitting [`Server-Timing`] headers using
     `tracing` data (similar to [`tide-server-timing`][4]).
- [ ] Adding more detailed instrumentation to Rocket's internals.
- [ ] Integrating with error handling (using [`tracing-error`][5]).
- [ ] Ensuring that Rocket's code generation and `tracing`'s 
     `#[instrument]` attribute work together.

Enjoy, and please let me know if there are any questions!

[1]: https://github.com/SergioBenitez/Rocket/issues/21#issuecomment-632383369
[2]: https://docs.rs/tracing-opentelemetry/0.4.0/tracing_opentelemetry/
[3]: https://www.w3.org/TR/server-timing/
[4]: https://github.com/yoshuawuyts/tide-server-timing
[5]: https://crates.io/crates/tracing-error


igalic:
i only see four or five instances of using a `parent: &span`. can you briefly elaborate under which circumstances you decide to actually group tracing messages in such a way?

core/lib/src/rocket.rs
```diff
@@ -66,40 +67,38 @@ pub(crate) struct Token;
...
+        let _e = span.enter();
 
         for route in routes {
             let old_route = route.clone();
             let route = route.map_base(|old| format!("{}{}", base, old))
                 .unwrap_or_else(|e| {
-                    error_!("Route `{}` has a malformed URI.", old_route);
-                    error_!("{}", e);
+                    let span = tracing::error_span!("malformed_uri", "Route `{}` has a malformed URI.", old_route);
+                    error!(parent: &span, "{}", e);
```
AshtonKem, line 78: This doesn't look quite right. Don't you need to enter this span too?

core/lib/src/rocket.rs
```diff
@@ -66,40 +67,38 @@ pub(crate) struct Token;
...
                     panic!("Invalid route URI.");
                 });
 
-            info_!("{}", route);
+            tracing::info!(%route);
```
AshtonKem, line 82: Why `tracing::info!` here and `info!` elsewhere in the same file?

core/lib/src/rocket.rs
hawkw: using `parent: &span` in the event we record on the next line explicitly sets `span` as the event's parent, which is equivalent to
```rust 
let span = tracing::error_span!(...);
let _e = span.enter();
error!(...);
```

i used `parent` here, rather than entering the span, because the span is only used as a parent for that specific event, and it didn't involve adding another line of code, so i thought it was better to have a slightly smaller diff. it would be fine to enter the span, instead — perhaps i should have done that to make this clearer.

core/lib/src/rocket.rs
hawkw: my bad — i think that's just because i edited this file on multiple different days and may have forgotten that there was an import for the macros. i'll change this to be more consistent, thanks for catching that!

core/lib/src/rocket.rs
igalic: i find the other style more explicit
since it doesn't rely on side effect

hawkw:
> i only see four or five instances of using a `parent: &span`. can you briefly elaborate under which circumstances you decide to actually group tracing messages in such a way?

first, the parent span context of an event can actually be set in two ways: it can be explicitly passed in via `parent:`, or, if no value for `parent:` is provided, it's inferred from the _current_ span (the last span that was `enter`ed in the current scope). so,
```rust
let span = tracing::info_span!(...);
tracing::info!(parent: &span, ...);
```
and 
```rust
let span = tracing::info_span!(...);
let _entered = span.enter();
// ...
tracing::info!(...);
```
will result in equivalent formatting for the event. as i mentioned in https://github.com/SergioBenitez/Rocket/pull/1410#discussion_r470120719, i used a mix of `enter` and explicit parents to try and reduce the number of lines in the diff, but maybe that was not a great idea, since it introduces more complexity for reviewers — i'll make it more consistent.

second, the reason that parents are not set everywhere is just because i _started_ the process of replacing Rocket's current indentation-based logging contexts with spans, but didn't really have time to _finish_ it. i made the change to some of the primary files in core, and then (in the interests of getting a PR up sooner rather than later) did the rest of the change via find-and-replace so that the code would still compile. the hope was that the parts of the codebase i did get to could be used as a model for other Rocket contributors to make the same changes elsewhere. 

i'm happy to finish applying this transformation, but i got busy with other responsibilities and, after speaking with @SergioBenitez and @jebrosen, we decided that it was better to at least open a PR now, at least to start socializing the changes to other contributors. if we'd rather merge it in one go, i'm happy to keep the PR open while i update the rest of Rocket to use the same style.

hawkw:
> i'm happy to finish applying this transformation, but i got busy with other responsibilities and, after speaking with @SergioBenitez and @jebrosen, we decided that it was better to at least open a PR now, at least to start socializing the changes to other contributors. if we'd rather merge it in one go, i'm happy to keep the PR open while i update the rest of Rocket to use the same style.

okay, i went ahead and added `tracing` spans to everywhere else where indentation was previously used to indicate context (mainly in error messages).

i've also added generated spans for user-defined catcher and handler functions in the codegen. that way, user events in catchers or handlers will be annotated with the span context of that catcher/handler. the output for the `errors` example now looks like this (with the addition of a trace event in the catchers and handlers):
Screenshot_20200813_131318

core/lib/src/rocket.rs
AshtonKem: There's no difference in the time stamps that are emitted between an explicit `span.enter()` vs `parent: &span`?

hawkw:
Anything else I need to do to help move this along? It looks like it's gotten out of sync with the base branch...

jebrosen:
Oh geez! I actually did a high-level read through this a while ago but never posted anything. I expect a lot of the conflicts are with the recent configuration system changes in 1fb0614.

Overall, I am happy about all these changes (summarized from a few different comments):
* Rocket logs `tracing` events instead of `log` events, and some of these are "rich" events with structured data
* Rocket's "default logger" is now implemented as `tracing-subscriber` `Layers`, some of which are (or could be) public for enabling/disabling individual aspects of Rocket's default logging scheme
* Any libraries/applications that use the macros from `log` will "automatically" fit in and those will be converted to `tracing` events
* Any libraries/applications that were willfully using the undocumented macros (e.g. for indentation) will break, and should migrate to using `tracing` spans instead

I did have two general questions:
* What happens if an application sets up a `pretty_env_logger` first, which doesn't know about `tracing`? We should at least make a mention of this in the changelog / release notes.
* If the log is "busy", it can look fairly messy. Do you have any suggestions for a good "reliable line" to re-print (and how)? I'm not super happy with `Connection`, especially when the source address is always the same (e.g. behind a reverse proxy).

I am more than happy to continue this PR including rebasing, formatting tweaks, and other changes, although I see that #21 is targeted for **0.6.0**.

hawkw:
> Overall, I am happy about all these changes (summarized from a few different comments):
>   * Rocket logs `tracing` events instead of `log` events, and some of these are "rich" events with structured data
>   * Rocket's "default logger" is now implemented as `tracing-subscriber` `Layers`, some of which are (or could be) public for enabling/disabling individual aspects of Rocket's default logging scheme

Yup, that's right!

>   * Any libraries/applications that use the macros from `log` will "automatically" fit in and those will be converted to `tracing` events

This isn't _currently_ the case: in order to convert the records from `log`'s macros to `tracing` events, we need to set the `tracing-log` crate's [`LogTracer` type](https://docs.rs/tracing-log/0.1.1/tracing_log/#convert-log-records-to-tracing-events) as the default logger. I can update this PR to have Rocket do that automatically, in the case where no default logger was already set by the user.

>   * Any libraries/applications that were willfully using the undocumented macros (e.g. for indentation) will break, and should migrate to using `tracing` spans instead

This is correct.
 
> I did have two general questions:
> 
>     * What happens if an application sets up a `pretty_env_logger` first, which doesn't know about `tracing`? We should at least make a mention of this in the changelog / release notes.

Right now, the `pretty_env_logger` won't see the `tracing` spans and events. However, we could enable the `tracing` crate's `log` feature flag to cause it to emit `log` records as well. By default, `tracing` will only do this if no default `tracing` subscriber has been set. However, an additional feature flag causes it to always emit log records. If we want the `pretty_env_logger` set by the user to log all the tracing events that are emitted, we could do this, however, it would result in duplicate logging from the user's logger and Rocket's default `tracing` subscriber.

>     * If the log is "busy", it can look fairly messy. Do you have any suggestions for a good "reliable line" to re-print (and how)? I'm not super happy with `Connection`, especially when the source address is always the same (e.g. behind a reverse proxy).

Sorry, can you explain a bit more about what you mean here?

> I am more than happy to continue this PR including rebasing, formatting tweaks, and other changes, although I see that #21 is targeted for **0.6.0**.

Thanks! I'm also happy to continue working on this branch, but some of that work is probably better off done by someone more familiar with the rest of Rocket. :)

SergioBenitez:
@hawkw I'd love to see a rebase of this on `master`. What's more, I think we should provide a set of macros that users can use to log with by default.

@jebrosen I would _love_ to get this in by 0.5. I'm finalizing the last of the forms revamp now, which should allow for a 0.5-RC as early as next weekend. If you could work with @hawkw to get this in great shape by then, I'd love to this past the finish line for 0.5; I think Rocket's users would really appreciate it.

hawkw:
@SergioBenitez 
> @hawkw I'd love to see a rebase of this on `master`.

Can do! @jebrosen would you like to do that, since I think there are potentially a lot of conflicts with the recent config changes? Or shall I?

> What's more, I think we should provide a set of macros that users can use to log with by default.

This branch re-exports `tracing`'s macros from the `rocket::trace` module:
https://github.com/SergioBenitez/Rocket/blob/bc492ba4f3b34c0636483ab4ebd4078beb8f111b/core/lib/src/trace.rs#L208-L211
and provides some documentation on how to use them:
https://github.com/SergioBenitez/Rocket/blob/bc492ba4f3b34c0636483ab4ebd4078beb8f111b/core/lib/src/trace.rs#L23-L89

Does that work, or did you want to provide some form of Rocket-specific macros?

SergioBenitez:
> Does that work, or did you want to provide some form of Rocket-specific macros?

Yes, that's excellent. Perhaps, then, a quick note in the docs about how to integrate with `log` if so desired?

jebrosen:
I'm going to (attempt to!) rebase this tonight; I will be sure to document or ask about anything I get stuck on.

jebrosen:
I opted to do this as a manual squashed merge instead, which was *much* more manageable than a rebase given the merge commit in the middle. I was careful to keep changes from both sides where files had been reorganized and to re-apply them in the right places, and tonight I'll check everything again, fix any remaining errors, and ensure that any other new code from the configuration revamp also gets `tracing`-ified.

jebrosen:
I've pushed my tree here: <https://github.com/jebrosen/Rocket/tree/adopt-tracing>. `diff -u <(git diff master...hawkw/eliza/trace) <(git diff master...jebrosen/adopt-tracing)` works fairly well to review my changes, which are *mostly* merge resolutions / re-applying changes from both trees. If this looks good, I'd like to start from that branch either in this or a new PR.

I think I'm pretty happy with the scope of this PR and I found it easy to review; I just have some cleanups in mind before merging. I expect to get a lot of them done pretty quickly, but any advice or hints would be super helpful:

* [ ] perhaps one of the most difficult, deciding when to interpolate data in the message and when to use structured fields. (`error!("Something happened with {}", whatever)` vs `error!("Something happened", thing = %whatever)`)
* [ ] make sure we've used `_span` everywhere we want to.
* [ ] verify launch info printing - it looks to me like `filter_layer` addresses this; I just want to make sure it does (and how)
* [ ] keeping spans across more spawns. It looks like there are a few places we want to use something like `Instrumented`, but for `thread::spawn()` or `spawn_blocking()`
  * Is that achievable with a simple `span.clone().enter()` in the closure?
* [ ] document `log` interoperability.
  * [ ] we might have enough third-party integrations that emit `log` events that we ought to enable the log compatibility by default.
* [ ] rust-lang/rust#78227 landed recently, but I still see lots of log messages in tests even after upgrading my compiler. I didn't see this with the previous `RocketLogger`; I'm not sure yet exactly which layer of code is responsible for this change.
* [ ] a thorough review and understanding of the new code in `trace.rs`, particularly `EventFormat`
  * [ ] fix or explain the "workaround"s
  * [ ] add/remove from which APIs we reexport from `rocket::trace`
  * [ ] docs

hawkw:
@jebrosen great, thanks for taking care of the rebase conflicts. I can merge your branch and make the additional changes on top of that, if that works for you?

jebrosen:
Sure, that sounds great to me!

hawkw:
@jebrosen 
>  [rust-lang/rust#78227](https://github.com/rust-lang/rust/pull/78227) landed recently, but I still see lots of log messages in tests even after upgrading my compiler. I didn't see this with the previous `RocketLogger`; I'm not sure yet exactly which layer of code is responsible for this change.

This is because libtest only captures output from the `print!`/`println!` and `eprint!`/`eprintln!` macros, not from `write!(stdout().lock(), ...)` or similar. 9f86214 changes the default subscriber to use `println!`, so that logs will be captured when running tests.

> * we might have enough third-party integrations that emit `log` events that we ought to enable the log compatibility by default.

9776ca3 turns `log` compatibility on by default. It's feature-flagged so that users who want to avoid the overhead of `log` can turn it off, but it's on by default so that everything just works when dependencies are using `log`.

When Rocket's "log" feature flag is enabled, we will try to set up a `log` logger that outputs `log` records as `tracing` events, allowing users to collect logs from dependencies and third-party integrations that use `log`. This works with both the default Rocket tracing subscriber _or_ a user-provided one. If the user has already set up a `log` logger, such as `env_logger`, setting the `log` adapter as the logger will fail, and in that case we bail and don't set up the default `tracing` subscriber, and all the `tracing` events in Rocket will be collected by the user-provided `log` logger instead.

>  fix or explain the "workaround"s

Done, these were mainly just working around API fixes in `tracing` crates that have since been released.

>  keeping spans across more spawns. It looks like there are a few places we want to use something like `Instrumented`, but for `thread::spawn()` or `spawn_blocking()`
>   * Is that achievable with a simple `span.clone().enter()` in the closure?

Yup, that's right, you can just clone the span into the spawned closure. :)

jebrosen:
I just pushed a few minor changes I had been working on and a whitespace fix to run CI (I hope you don't mind!).

I was having trouble understanding some of `EventFormat` (although f15d432 seems to have changed or fixed some of what I was looking at). Consider this code:

```
                    let s = rocket::trace::info_span!("span1", "Span 1");
                    {
                        let _e = s.enter();
                        rocket::trace::info!("hello");
                    }
                    rocket::trace::info!("zonk!");
```

I'm getting log output like this:

```
📡 Connection, from: 127.0.0.1:56954
    => GET / text/html
    => Incoming request: GET / text/html
    => Changing method to `PUT`.
    => matched: PUT / (hello)
    => Outcome: Success
    => Rewriting response body.
📡 Connection, from: 127.0.0.1:56954
    => GET / text/html
    => Span 1
    => hello
📡 Connection, from: 127.0.0.1:56954
    => GET / text/html
    => zonk!
    => Response succeeded.
```

* So, I'm a bit perplexed by the output in general. `visit_spans` traverses "towards the root", but the output order looks like traversing "towards children". Am I missing something that's reversing the output?
* It's not super obvious that all of this was a single request. I believe this is because the parent spans are re-printed any time the "current" (leaf) span changes, even if it's in the same overall request scope.
  * This seems to mean that `span.enter()` vs `parent: &span` makes a difference to the indentation algorithm, which surprised me a bit.
  * (not pictured) `curl` said it reused a connection, but the log output printed "Connection" twice. My guess is the same cause.

I think a lot of these can be resolved neatly by choosing the remembered last-id differently, e.g. the span indicating the current connection (or request) or the "topmost span" instead of the current span. Presumably we could reach it either going from the bottom up (`visit_spans`), although top down seems like it should be a shorter search. Is there a function that does that?

SergioBenitez:
Hi all! Just checking in.

My ambitions for the forms revamp grew a bit outside of forms themselves, so it's a bit behind schedule. Nevertheless, it's nearing completion, and I'd love to see this PR be merged in time for a 0.5 RC. How close are we to that?

From what I can tell, here are the current issues:

 - [ ] Log output for a single request visually appears to be from separate requests due to indentation.
 - [ ] Use of `span` and entering a `span` is inconsistent.
 - [ ] Tests are currently failing.
 - [ ] Add a new log level, `support`, between `debug` and `normal`, for messages that could help resolve issues.

    I haven't mentioned this before, but I think this would resolve the tension between "always print launch information" and "just don't". There are a few other places where we could provide messages that would help resolve issues that are only visible at runtime that aren't quite warnings but also aren't just general info.

What else is missing?

jebrosen:
>  Log output for a single request visually appears to be from separate requests due to indentation.

I did two things to fix this. 

1) Change the span printing logic. Context is now printed if the top-level span has changed (i.e. a different connection/request), or when a higher depth has been reached than the previous message's. IMO this looks better but might still be slightly wrong; the only way I could think of to get 100% guaranteed nice output would be to store the entire chain of span IDs of the previous message, and start reprinting at the first different one. (Perhaps that could be simplified by implementing `on_enter` and `on_exit`  instead)
2) I deleted the "connection" span, so that "request" is the top-level span. We could also put the address on the request line instead to get that back.

> Use of span and entering a span is inconsistent.

Right. This can actually matter for the output formatting in some cases: `span.enter()` vs `parent: &span` is different to the indentation algorithm, since it goes off of the "stack of spans that have been entered" instead of the "parent-chain of the event to be printed". @hawkw, does that sound like a real distinction or did I misunderstand something here?

> Add a new log level, support, between debug and normal, for messages that could help resolve issues.

We can't easily add new log levels to `tracing` itself, but we could:

* log to a special target (e.g. "rocket::support")
* Add a new rule like `LoggingLevel::Support` => `...,rocket::support=info` similarly to how `LoggingLevel::Normal` disables hyper/rustls log messages.

I'll explore that path a bit further. Is there anything aside from launch info that should definitely go in this category right away?

> What else is missing?

I think it's mostly down to testing and tweaking at this point. I want to try mixing `log` and `tracing` events and heavy concurrent requests and make sure output always looks at least semi-reasonable. Also, I put in a quick but scary-looking hash function to give requests distinguishable but short IDs, since the bit patterns defeated most of my attempts at naively truncating or hashing them.

hawkw:
>     * log to a special target (e.g. "rocket::support")
> 
>     * Add a new rule like `LoggingLevel::Support` => `...,rocket::support=info` similarly to how `LoggingLevel::Normal` disables hyper/rustls log messages.

Another potential thing to note is that `tracing` has more log levels than Rocket's `LoggingLevel` type --- it has `ERROR`, `WARN`, `INFO`, `DEBUG`, and `TRACE`. We could, potentially, make Rocket's `LoggingLevel::Debug` map to `tracing`'s `Level::TRACE`, and `LoggingLevel::Support` map to `tracing`'s `Level::DEBUG`. 

We _may_ even want to consider changing some of the re-exports in the `rocket::trace` module to rename the `tracing` macros to use names that correspond with Rocket's `LoggingLevel`s, although I'm not sure whether or not this is actually a good idea. It may just make things more confusing for users, since the `tracing` crate's documentation would use different names (and, `tracing` uses the same level names as the `log` crate, so users familiar with `log` might also be confused by having to learn new names).

jebrosen:
`rocket::support=info` doesn't quite do what I wanted after all: I was trying to implement "show any messages at `warn` or above, and also any `rocket::support` messages at `info` or above", but it doesn't look like I can do that with `EnvFilter` since the global level `warn` takes precedence.

On the subject of logging levels, I would be slightly in favor of renaming some of `LogLevel` and/or changing the log level of some messages (e.g. `info` -> `debug`). It took me a while to get used to `Debug` also logging events at the `trace` level, and having no way to see events at the `debug` level without also getting `info` events.

---

I think the log output formatting at this point is pretty close to "good enough", and it's at least as good and in some ways better than current `master`. Here's one little wart that I'm still working on:

```
Warning: test warning 2, from log, log.target: "fairings", log.module_path: "fairings", log.file: "examples/fairings/src/main.rs", log.line: 63
```

* [ ] Skip printing the fields starting with `log.`, and interpret them in `WithFile` instead.

SergioBenitez:
> `rocket::support=info` doesn't quite do what I wanted after all: I was trying to implement "show any messages at `warn` or above, and also any `rocket::support` messages at `info` or above", but it doesn't look like I can do that with `EnvFilter` since the global level `warn` takes precedence.

What does this mean for the state of "support" messages in the current PR?

> On the subject of logging levels, I would be slightly in favor of renaming some of `LogLevel` and/or changing the log level of some messages (e.g. `info` -> `debug`). It took me a while to get used to `Debug` also logging events at the `trace` level, and having no way to see events at the `debug` level without also getting `info` events.

I agree. Perhaps we should simply use the names native to log/tracing.

hawkw:
> I think the log output formatting at this point is pretty close to "good enough", and it's at least as good and in some ways better than current `master`. Here's one little wart that I'm still working on:
> 
> ```
> Warning: test warning 2, from log, log.target: "fairings", log.module_path: "fairings", log.file: "examples/fairings/src/main.rs", log.line: 63
> ```
> 
>     * [ ]  Skip printing the fields starting with `log.`, and interpret them in `WithFile` instead.

There's existing machinery for normalizing the `log` crate's metadata into the same format used by `tracing` events, the default formatters in `tracing-subscriber` use this. I can push a commit to add it in Rocket's formatter, as well --- i forgot to when I wrote the initial implementation.

jebrosen:
> What does this mean for the state of "support" messages in the current PR?

I believe this is unchanged: "support" messages are not specially surfaced either before or after this PR. AFAICT the main obstacle is setting *only* support messages to log at a more verbose level than the global limit; neither `log` nor `tracing` seems to support this directly.

> I agree. Perhaps we should simply use the names native to log/tracing.

In that vein, does this look appropriate?
```
        LogLevel::Error => "error,hyper=off,rustls=off",
        LogLevel::Warn => "warn,hyper=off,rustls=off",
        LogLevel::Info => "info,hyper=off,rustls=off",
        LogLevel::Debug => "debug",
        LogLevel::Trace => "trace",
        LogLevel::Off => "off",
```

If we're concerned about compatibility, we could also accept `critical` as an alias for `warn`, and `normal` as an alias for `info`.

Then either in this PR or separately we could, if we wish, downgrade some `info` messages to `debug`, thereby making `info` more useful or maybe even identical to the desired "support" log level.

> Use of span and entering a span is inconsistent.

I haven't forgotten this!

hawkw:
> `rocket::support=info` doesn't quite do what I wanted after all: I was trying to implement "show any messages at `warn` or above, and also any `rocket::support` messages at `info` or above", but it doesn't look like I can do that with `EnvFilter` since the global level `warn` takes precedence.

Hmm, that _seems_ like a bug in the `EnvFilter` implementation? If you don't mind, can you make sure that Rocket depends on the latest `tracing` version, and if it isn't, open an issue against `tracing`? That _should_ work...

jebrosen:
The "wrong" behavior with `support` was actually my typo; it actually works exactly how I hoped it would. The CI failures are #1477.

hawkw:
> The "wrong" behavior with `support` was actually my typo; it actually works exactly how I hoped it would. The CI failures are #1477.

Oh, whoops! Glad it wasn't an upstream bug! :)

core/lib/src/trace.rs
```diff
@@ -224,6 +224,9 @@ pub mod prelude {
 pub enum LogLevel {
     /// Only shows errors and warnings: `"critical"`.
     Critical,
+    /// Shows errors, warnings, and some informational messages that are likely
+    /// to be relevant when troubleshooting (such as configuration).
```
hawkw, line 228: nit: looks like the other levels have docs that include the string name of that level in the config file, should that be included here?

jebrosen (commented):
I'm pretty happy with this. Here are the last few things I think this needs at this point:

* [ ] A rebase to get CI passing.
* [ ] Consider changing the hash used to disambiguate requests in the log, and/or the way it is formatted. Span `Id`s generated by `Registry` are a multi-level bit-packed generation index, and I threw something ad-hoc together that seemed to look nice on 64-bit systems.
* [ ] A review from @SergioBenitez
* [ ] Any remaining touchups on docs, and re-exports.
  * [ ] Potentially re-export the entire `tracing` and `tracing-subscriber` crates, like is done for `futures`, `tokio`, and `figment` today.

felipesere:
Anything that I could do to help this along? I've locally merged `master` into it and worked as I expected.

SergioBenitez:
@felipesere This is currently waiting on me to merge a big change before we rebase this on master.

SergioBenitez:
Okay, after _forever_, I am now ready to review this, pending a rebase on/fix-up for `master`, of course.

Really excited for this.

jebrosen:
I have *tentatively* pushed this to my fork to run CI and pick up any obvious mistakes (https://github.com/SergioBenitez/Rocket/compare/master...jebrosen:tracing-rebase-202103).

Due to merge commits and the recent changes in `master`, I decided again this time to manually re-apply all of the changes at once. So, this will need at least one more full look-through from me to make sure I did not lose anything important along the way. Before or after that I can either open a new PR or force-push over the branch - any preferences here, @hawkw @SergioBenitez?

SergioBenitez:
> Before or after that I can either open a new PR or force-push over the branch - any preferences here, @hawkw @SergioBenitez?

Whatever makes life easier for you two!

SergioBenitez:
@jebrosen Just checking in: should I be reviewing something somewhere?

SergioBenitez:
Moved to #1579.

PR #1413, opened 2020-08-16
Adds a default content-type to StaticFiles and NamedFile

### LINKED ISSUE #1397: Default content-type for StaticFiles
Rocket version:

```
version = "0.5.0-dev"
source = "git+https://github.com/SergioBenitez/rocket#ddfd73d6f33e8a8ef3f37be94bbbdcb840804326"
```

os version: OSX Catalina (10.15.6)

When using `StaticFiles` to serve content, it relies on `ContentType::from_extension` to generate the content-type header for a response. If a file does not have an extension, no content-type will be set. I'd like to know if there is a way to set a default content-type, or if that would be a useful feature to add. I'm willing to open a PR for the feature if there is a likely chance it will be accepted.

To reproduce: serve a file without an extension using `rocket_contrib::serve::StaticFiles`

Assuming there is no existing way to do this, I'd be curious about maintainers' ideas of how to solve this. I'd be happy defaulting to `text/plain`, and including it as an option in the existing `Options` bitmask. However, others might want the ability to set it to a specific content-type, like `text/html` or something else, and in that case I don't have a great idea for a design.

pwoolcoc:
Closes #1397

### DIFF
--- a/contrib/lib/src/serve.rs
+++ b/contrib/lib/src/serve.rs
@@ -17,7 +17,7 @@
 use std::path::{PathBuf, Path};
 
 use rocket::{Request, Data, Route};
-use rocket::http::{Method, uri::Segments, ext::IntoOwned};
+use rocket::http::{ContentType, Method, uri::Segments, ext::IntoOwned};
 use rocket::handler::{Handler, Outcome};
 use rocket::response::{NamedFile, Redirect};
 
@@ -207,6 +207,7 @@ pub struct StaticFiles {
     root: PathBuf,
     options: Options,
     rank: isize,
+    content_type: Option<ContentType>,
 }
 
 impl StaticFiles {
@@ -293,7 +294,7 @@ impl StaticFiles {
             panic!("refusing to continue due to invalid static files path");
         }
 
-        StaticFiles { root: path.into(), options, rank: Self::DEFAULT_RANK }
+        StaticFiles { root: path.into(), options, rank: Self::DEFAULT_RANK, content_type: None }
     }
 
     /// Sets the rank for generated routes to `rank`.
@@ -314,6 +315,15 @@ impl StaticFiles {
         self.rank = rank;
         self
     }
+
+    /// Sets the default content type
+    ///
+    /// If a file does not have an extension, this value will be used
+    /// for the content-type header
+    pub fn default_type(mut self, content_type: ContentType) -> Self {
+        self.content_type = Some(content_type);
+        self
+    }
 }
 
 impl Into<Vec<Route>> for StaticFiles {
@@ -331,7 +341,7 @@ impl Into<Vec<Route>> for StaticFiles {
     }
 }
 
-async fn handle_dir<'r, P>(opt: Options, r: &'r Request<'_>, d: Data, p: P) -> Outcome<'r>
+async fn handle_dir<'r, P>(opt: Options, r: &'r Request<'_>, d: Data, p: P, content_type: Option<&ContentType>) -> Outcome<'r>
     where P: AsRef<Path>
 {
     if opt.contains(Options::NormalizeDirs) && !r.uri().path().ends_with('/') {
@@ -346,7 +356,13 @@ async fn handle_dir<'r, P>(opt: Options, r: &'r Request<'_>, d: Data, p: P) -> O
         return Outcome::forward(d);
     }
 
-    let file = NamedFile::open(p.as_ref().join("index.html")).await.ok();
+    let file = {
+        let mut file = NamedFile::open(p.as_ref().join("index.html")).await.ok();
+        if let (Some(ref mut file), Some(content_type)) = (file.as_mut(), content_type) {
+            file.default_type(content_type.clone());
+        }
+        file
+    };
     Outcome::from_or_forward(r, d, file)
 }
 
@@ -358,7 +374,7 @@ impl Handler for StaticFiles {
         let current_route = req.route().expect("route while handling");
         let is_segments_route = current_route.uri.path().ends_with(">");
         if !is_segments_route {
-            return handle_dir(self.options, req, data, &self.root).await;
+            return handle_dir(self.options, req, data, &self.root, self.content_type.as_ref()).await;
         }
 
         // Otherwise, we're handling segments. Get the segments as a `PathBuf`,
@@ -370,8 +386,18 @@ impl Handler for StaticFiles {
             .map(|path| self.root.join(path));
 
         match path {
-            Some(p) if p.is_dir() => handle_dir(self.options, req, data, p).await,
-            Some(p) => Outcome::from_or_forward(req, data, NamedFile::open(p).await.ok()),
+            Some(p) if p.is_dir() => handle_dir(self.options, req, data, p, self.content_type.as_ref()).await,
+            Some(p) => {
+                let file = {
+                    let mut file = NamedFile::open(p).await.ok();
+                    if let (Some(ref mut file), Some(content_type)) = (file.as_mut(), self.content_type.as_ref()) {
+                        println!("setting default content-type");
+                        file.default_type(content_type.clone());
+                    }
+                    file
+                };
+                Outcome::from_or_forward(req, data, file)
+            },
             None => Outcome::forward(data),
         }
     }
--- a/contrib/lib/tests/static_files.rs
+++ b/contrib/lib/tests/static_files.rs
@@ -5,7 +5,7 @@ mod static_tests {
 
     use rocket::{self, Rocket, Route};
     use rocket_contrib::serve::{StaticFiles, Options, crate_relative};
-    use rocket::http::Status;
+    use rocket::http::{ContentType, Status};
     use rocket::local::blocking::Client;
 
     fn static_root() -> &'static Path {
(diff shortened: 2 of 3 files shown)


pwoolcoc:
SInce the content-type header is set in `NamedFile`s implementation of `Responder`, it was easiest to add the default content-type to that type as well

pwoolcoc:
@SergioBenitez any chance I could get a look at this?

core/lib/src/response/named_file.rs
```diff
@@ -78,6 +78,31 @@ impl NamedFile {
...
+    /// use rocket::http::ContentType;
+    ///
+    /// # #[allow(head_code)]
+    /// # async fn demo_default_type() -> io::Result<()> {
+    /// let mut file = NamedFile::open("foo.text").await?;
+    /// file.default_type(ContentType::Plain);
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn default_type(&mut self, content_type: ContentType) -> &mut Self {
```
SergioBenitez, line 102: I would rather this be a builder method, taking `self` by value.
```suggestion
    pub fn with_default(mut self, content_type: ContentType) -> Self {
```

Then the example can be:

```rust
#[get("/<path..>")]
fn file(path: PathBuf) -> Result<NamedFile, Debug<io::Error>> {
    Ok(NamedFile::open(path).await?.with_default(ContentType::Plain))
}
```

Which I think is much more illustrative.

core/lib/src/response/named_file.rs
```diff
@@ -9,9 +9,9 @@ use crate::response::{self, Responder};
 use crate::http::ContentType;
 
 /// A file with an associated name; responds with the Content-Type based on the
-/// file extension.
+/// file extension or with a default, if set.
 #[derive(Debug)]
-pub struct NamedFile(PathBuf, File);
+pub struct NamedFile(PathBuf, File, Option<ContentType>);
```
SergioBenitez, line 14: This should probably have named fields at this point:

```rust
pub struct NamedFile {
    path: PathBuf,
    file: File,
    default_ct: Option<ContentType>,
}
```

core/lib/src/response/named_file.rs
```diff
@@ -78,6 +78,31 @@ impl NamedFile {
     pub fn path(&self) -> &Path {
         self.0.as_path()
     }
+
+    /// Sets the default content type for the file
```
SergioBenitez, line 82:
```suggestion
    /// Sets the default Content-Type.
```

core/lib/src/response/named_file.rs
```diff
@@ -78,6 +78,31 @@ impl NamedFile {
...
+    ///
+    /// If the file does not have an extension, the content type will
+    /// be set to this value. If no default is set and the file does not
+    /// have an extension, no content-type header will be sent.
```
SergioBenitez, line 86: Perhaps:

Ordinarily, when responding, if the file path to `self` does not have an extension, or if the extension does not have a known Content-Type (see [`ContentType::from_extension()`]), no Content-Type is set in the response. By setting a default, the Content-Type of the response will be set to `content_type`.

core/lib/src/response/named_file.rs
```diff
@@ -91,7 +116,11 @@ impl<'r> Responder<'r, 'static> for NamedFile {
         if let Some(ext) = self.0.extension() {
             if let Some(ct) = ContentType::from_extension(&ext.to_string_lossy()) {
                 response.set_header(ct);
+            } else if let Some(ct) = self.2 {
+                response.set_header(ct);
             }
+        } else if let Some(ct) = self.2 {
+            response.set_header(ct);
         }
```
SergioBenitez, line 124:
```suggestion
        let content_type = self.path.extension()
            .and_then(|e| e.to_str())
            .and_then(ContentType::from_extension)
            .or(self.default_ct);
        
        if let Some(content_type) = content_type {
            response.set_header(content_type);
        }
```

contrib/lib/src/serve.rs
```diff
@@ -314,6 +315,15 @@ impl StaticFiles {
         self.rank = rank;
         self
     }
+
+    /// Sets the default content type
+    ///
+    /// If a file does not have an extension, this value will be used
+    /// for the content-type header
+    pub fn default_type(mut self, content_type: ContentType) -> Self {
```
SergioBenitez, line 323:
```suggestion
    pub fn with_default(mut self, content_type: ContentType) -> Self {
```

To match `with_default` in `NamedFile`.

contrib/lib/src/serve.rs
```diff
@@ -314,6 +315,15 @@ impl StaticFiles {
         self.rank = rank;
         self
     }
+
+    /// Sets the default content type
+    ///
+    /// If a file does not have an extension, this value will be used
+    /// for the content-type header
```
SergioBenitez, line 322: Let's also change this to be more inline with my suggestions for the `NamedFile::with_default()` docstring.

contrib/lib/src/serve.rs
```diff
@@ -346,7 +356,13 @@ async fn handle_dir<'r, P>(opt: Options, r: &'r Request<'_>, d: Data, p: P) -> O
...
     }
 
-    let file = NamedFile::open(p.as_ref().join("index.html")).await.ok();
+    let file = {
+        let mut file = NamedFile::open(p.as_ref().join("index.html")).await.ok();
+        if let (Some(ref mut file), Some(content_type)) = (file.as_mut(), content_type) {
+            file.default_type(content_type.clone());
+        }
+        file
+    };
```
SergioBenitez, line 365:
```suggestion
    let file = NamedFile::open(p.as_ref().join("index.html")).await
        .map(|file| match default_ct {
            Some(default) => file.with_default(default_ct.clone()),
            None => file
        }).ok();
```

contrib/lib/src/serve.rs
```diff
@@ -331,7 +341,7 @@ impl Into<Vec<Route>> for StaticFiles {
     }
 }
 
-async fn handle_dir<'r, P>(opt: Options, r: &'r Request<'_>, d: Data, p: P) -> Outcome<'r>
+async fn handle_dir<'r, P>(opt: Options, r: &'r Request<'_>, d: Data, p: P, content_type: Option<&ContentType>) -> Outcome<'r>
```
SergioBenitez, line 344:
```suggestion
async fn handle_dir<'r, P>(opt: Options, r: &'r Request<'_>, d: Data, p: P, default_ct: Option<&ContentType>) -> Outcome<'r>
```

contrib/lib/src/serve.rs
```diff
@@ -207,6 +207,7 @@ pub struct StaticFiles {
     root: PathBuf,
     options: Options,
     rank: isize,
+    content_type: Option<ContentType>,
```
SergioBenitez, line 210:
```suggestion
    default_ct: Option<ContentType>,
```

contrib/lib/src/serve.rs
```diff
@@ -370,8 +386,18 @@ impl Handler for StaticFiles {
...
 
         match path {
-            Some(p) if p.is_dir() => handle_dir(self.options, req, data, p).await,
-            Some(p) => Outcome::from_or_forward(req, data, NamedFile::open(p).await.ok()),
+            Some(p) if p.is_dir() => handle_dir(self.options, req, data, p, self.content_type.as_ref()).await,
+            Some(p) => {
+                let file = {
+                    let mut file = NamedFile::open(p).await.ok();
+                    if let (Some(ref mut file), Some(content_type)) = (file.as_mut(), self.content_type.as_ref()) {
+                        println!("setting default content-type");
```
SergioBenitez, line 394: Don't print this.

contrib/lib/src/serve.rs
```diff
@@ -370,8 +386,18 @@ impl Handler for StaticFiles {
             .map(|path| self.root.join(path));
 
         match path {
-            Some(p) if p.is_dir() => handle_dir(self.options, req, data, p).await,
-            Some(p) => Outcome::from_or_forward(req, data, NamedFile::open(p).await.ok()),
+            Some(p) if p.is_dir() => handle_dir(self.options, req, data, p, self.content_type.as_ref()).await,
+            Some(p) => {
+                let file = {
```
SergioBenitez, line 391: These are almost exactly the same lines as in `handle_dir`: dedup these.

contrib/lib/tests/static_files.rs
```diff
@@ -22,6 +22,8 @@ mod static_tests {
             .mount("/both", StaticFiles::new(&root, Options::DotFiles | Options::Index))
             .mount("/redir", StaticFiles::new(&root, Options::NormalizeDirs))
             .mount("/redir_index", StaticFiles::new(&root, Options::NormalizeDirs | Options::Index))
+            .mount("/no_extension_no_default", StaticFiles::new(&root, Options::None))
+            .mount("/no_extension_with_default", StaticFiles::new(&root, Options::None).default_type(ContentType::Plain))
```
SergioBenitez, line 26: Do we need `Options` at all here? Why not just `StaticFiles::from(&root).with_default(ContentType::Plain)`?

SergioBenitez (changes requested):

SergioBenitez:
@pwoolcoc Do you have interest in continuing this?

PR #1417, opened 2020-08-21
Ran cargo-diet to reduce download size from crates.io

Razican:
This makes sure to only add the needed files for cargo packages, so that they download faster. The gains come mostly from not including tests on every package:

contrib/codegen:
```
Saved 43% or 6.6 KB in 11 files (of 15.3 KB and 14 files in entire crate)
```

contrib/lib:
```
Saved 21% or 40.0 KB in 19 files (of 192.9 KB and 39 files in entire crate)
```

core/codegen:
```
Saved 65% or 228.9 KB in 104 files (of 354.3 KB and 123 files in entire crate)
```

core/lib:
```
Saved 7% or 43.4 KB in 30 files (of 665.1 KB and 97 files in entire crate)
```

### DIFF
--- a/contrib/codegen/Cargo.toml
+++ b/contrib/codegen/Cargo.toml
@@ -10,6 +10,7 @@ readme = "../../README.md"
 keywords = ["rocket", "contrib", "code", "generation", "proc-macro"]
 license = "MIT OR Apache-2.0"
 edition = "2018"
+include = ["src/**/*"]
 
 [features]
 database_attribute = []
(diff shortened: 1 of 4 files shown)


Voodlaz (approved):

SergioBenitez:
One significant downside from not including tests in the uploaded packages is that they won't be run by automated tools like crater. In the past, Rocket's tests have found regressions in `rustc` and have led to us being alerted of inadvertent breaking changes. For this reason alone, I'm hesitant to make this change.

Besides the obvious bandwidth savings, what other advantages are there to this this change?

Razican:
> Besides the obvious bandwidth savings, what other advantages are there to this this change?

As far as I can tell, the only advantage is the bandwidth savings, I hadn't thought about the tests in crater. Feel free to close it :)

SergioBenitez:
I think the downside here outweighs the upside, unfortunately. Thank you for the idea, however!

PR #1423, opened 2020-09-03
Implement FromRequest for IpAddr

### LINKED ISSUE #1414: Add `FromRequest for IpAddr` which maps to `Request.client_ip()`
## Feature Requests

> The `client_ip` is pretty useful for logging or subnet filtering as long as the `X-Real-IP` is trusted, and it's very straight to user about mapping `IpAddr` to `client_ip`.

> As we implemented, there are 2 addresses the server can get, `remote: SocketAddr` and `client_ip: IpAddr`, the former one already have a guard so I think provide another guard for the latter one looks good.

> Display user's IP at the footer of the page.

> It's not possible to implement `FromRequest for IpAddr` outside of Rocket.

Gerhut:
which maps to `request.client_ip()`

Closes #1414

### DIFF
--- a/core/lib/src/request/from_request.rs
+++ b/core/lib/src/request/from_request.rs
@@ -1,5 +1,5 @@
 use std::fmt::Debug;
-use std::net::SocketAddr;
+use std::net::{IpAddr, SocketAddr};
 
 use futures::future::BoxFuture;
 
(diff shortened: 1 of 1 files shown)


SergioBenitez:
Merged in 0673986c3222a44cdf7779552ec184a0491af9c9. Thank you!

PR #1426, opened 2020-09-07
Fix for forms that contain structures affected by encoding (such as SocketAddrV4) failing to parse

### LINKED ISSUE #1425: Forms that contain structures affected by encoding (such as SocketAddrV4) fail to parse
## Bug Report

Version 0.4.5
OS: Debian 10 (Linux 4.19.0-10-cloud-amd64 #1 SMP Debian 4.19.132-1 (2020-07-24) x86_64 GNU/Linux)

A Form containing a SocketAddrV4 field is failing to parse input such as 127.0.0.1:80. I'm under the impression that this is because FromFormValue::from_form_value() and subsequently SocketAddrV4::from_str() is being called with the encoded address as a parameter (e.g. 127.0.0.1%3A80) instead of 127.0.0.1:80.

```rust
#[derive(FromForm)]
struct Input {
    ip: SocketAddrV4,
}

#[post("/test", data = "<input>")]
fn test(input: Form<Input>) {}
```

The above code will not parse input such as 127.0.0.1:80 because the colon ':' is encoded. This results in 127.0.0.1%3A80 being passed into FromFormValue::from_form_value() which will fail.

lewisclark:
Create new macro impl_with_fromstr_encoded to implement FromFormValue for structures that contain characters affected by encoding (especially colons ':'), such as ipv6 addresses and ipv4 + port combos

Fixes #1425

### DIFF
--- a/core/lib/src/request/form/from_form_value.rs
+++ b/core/lib/src/request/form/from_form_value.rs
@@ -243,7 +243,24 @@ impl_with_fromstr!(
     f32, f64, isize, i8, i16, i32, i64, i128, usize, u8, u16, u32, u64, u128,
     NonZeroI8, NonZeroI16, NonZeroI32, NonZeroI64, NonZeroI128, NonZeroIsize,
     NonZeroU8, NonZeroU16, NonZeroU32, NonZeroU64, NonZeroU128, NonZeroUsize,
-    IpAddr, Ipv4Addr, Ipv6Addr, SocketAddrV4, SocketAddrV6, SocketAddr
+    Ipv4Addr
+);
+
+macro_rules! impl_with_fromstr_encoded {
+    ($($T:ident),+) => ($(
+        impl<'v> FromFormValue<'v> for $T {
+            type Error = &'v RawStr;
+
+            #[inline(always)]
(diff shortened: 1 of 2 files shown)


SergioBenitez:
Excellent. Could you add some tests to confirm this works as expected? This should be added as a regression test in the core/tests directory using the naming convention in place. Do we need to add the float types to this new list, too?

lewisclark:
Definitely. I'll write up some tests today. 

Regarding floats, I believe all representations should parse correctly without decoding, given the only representations are:
- '3.14'
- '-3.14'
- '2.5E10', or equivalently, '2.5e10'
- '2.5E-10'
- '5.'
- '.5', or, equivalently, '0.5'
- 'inf', '-inf', 'NaN'
There aren't any reserved characters in there that would be encoded (list taken from https://doc.rust-lang.org/std/primitive.f64.html#impl-FromStr)

SergioBenitez:
Thank you! Excellent work. Merged in https://github.com/SergioBenitez/Rocket/commit/53d13a309b14fe427560a1cb965c9423d85a9cf9.

PR #1428, opened 2020-09-11
replace lazy_static with serial_test

komi1230:
Hi, thanks for such a awesome project !
I sincerely appreciate your efforts for maintenance.

I found `lazy_static` is used just to run tests in serial.
To do it, `serial_test` is better.

`serial_test` includes `lazy_static` as dependency, but `serial_test` can make codes cleaner.

### DIFF
--- a/core/lib/src/config/mod.rs
+++ b/core/lib/src/config/mod.rs
@@ -459,7 +459,6 @@ impl FullConfig {
 #[cfg(test)]
 mod test {
     use std::env;
-    use std::sync::Mutex;
 
     use super::{Config, FullConfig, ConfigError, ConfigBuilder};
     use super::{Environment, GLOBAL_ENV_NAME};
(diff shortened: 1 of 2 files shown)


core/lib/src/config/mod.rs
```diff
@@ -459,23 +459,18 @@ impl FullConfig {
...
     use super::environment::CONFIG_ENV;
     use super::Environment::*;
     use super::Result;
 
+    use serial_test::serial;
     use crate::logger::LoggingLevel;
 
     const TEST_CONFIG_FILENAME: &'static str = "/tmp/testing/Rocket.toml";
 
-    // TODO: It's a shame we have to depend on lazy_static just for this.
```
jebrosen, line 474: I think the spirit of this comment was "why do we need to depend on `lazy_static` at all". Therefore in my eyes `serial_test` is a worse situation by being a procedural macro *and* depending on `lazy_static` (and `parking_lot`).

I would personally be much happier to see an approach based on something like <https://docs.rs/parking_lot/0.11.0/parking_lot/fn.const_mutex.html>, which was added some time within the last year or so. Ironically, `tokio` (and/or enabled optional features) depends on `lazy_static` but not `parking_lot`, which makes this idea inconvenient for a different reason.

SergioBenitez:
We no longer use `lazy_static`, instead using Figment's `Jail`. Thanks!

PR #1429, opened 2020-09-11
(v0.4) Fix versions of libraries in database documentation to match Cargo.toml

jebrosen:
`master` had more recent versions of the databases than `v0.4` at the time the table was backported (in 225218d). I believe these are all matched up properly now, but I'd appreciate if anyone could give it a second look.


SergioBenitez:
Merged in b109d151a687c5ef0d82a23f6bab3e44cb4d8ce3. Thanks for catching this!

Issue #1117, opened 2019-09-04
Tracking Issue for Asynchronous Database Pool Support


githubaccount624:
## Why?

Creating a database connection can be an expensive operation. Connection pools can alleviate this and serve as an important performance optimization for web servers. The current connection pool in Rocket is synchronous. It would be desirable to have support for an asynchronous connection pool in the up-and-coming async Rocket.

## How?

jebrosen elaborates in [this issue reply](https://github.com/SergioBenitez/Rocket/issues/1065#issuecomment-518480660). One of the main blockers is that Rocket and the database engine should first run within the same runtime (tokio 0.1 or tokio 0.2). This seems partially resolved as the async branch of [tokio-postgres](https://github.com/sfackler/rust-postgres/tree/std-futures/tokio-postgres) and [Rocket](https://github.com/SergioBenitez/Rocket/tree/async) are both currently using `tokio=0.2.0-alpha.4`.

Jeb's second point is that `connect` may be an asynchronous operation and whether or not `FromRequest` should return a `Future`. Connecting to a database can sometimes take 50+ milliseconds and `tokio-postgres` has already made `connect` an asynchronous operation, so it seems like a safe bet to accommodate this fact.

I'll just paste the rest of Jeb's minimum required tasks for reference:

> * Create a connection pool with tokio-postgres and/or bb8 and put it in managed state. This should tell us if on_attach and on_launch need to be async-capable.
> * Create a request guard type wrapping a connection (or a future that will resolve to a connection). This and the previous bullet point together are more or less what the guide demonstrated for databases in Rocket 0.3 before we had rocket_contrib::databases.
> * Use said connection via the request guard within a route to query/alter the database. This should work fine, because routes can now be async fn.
> * Test, figure out, and document cross-runtime compatibility (or lack thereof). That is, a definitive statement in the form of "Rocket uses tokio 0.2, so your database futures have to as well" if that's the case. Alternatively, maybe we could find a proof of concept of running two runtimes - tokio 0.2 (for rocket) alongside tokio 0.1 (for tokio-postgres) with some kind of message passing between them. I leave this intentionally vague because it's an annoying solution to an annoying problem and someone other than I will probably think of a better idea.

jebrosen:
Here's an updated list of things I think we would need to address in order to fully support async database pools. Importantly, I think it's not yet possible after all to set up a pool "the manual way" because some of Rocket's traits--such as `FromRequest`--are still synchronous.

[...]

PR #1433, opened 2020-09-14
Add ArangoDB to rocket_contrib databases

Weasy666:
This PR adds support for [ArangoDB](https://www.arangodb.com/) by implementing `Poolable` for [arangors::Connection](https://github.com/fMeow/arangors/blob/master/src/connection/mod.rs#L145) in conjunction with [r2d2_arangors](https://github.com/inzanez/r2d2-arangors).

`arangors` expects a url, username and password for its `Connection::establish_...()` functions, so i had to make it mandatory, to include them into the Rocket.toml file. I hope that is ok, if not...another idea could be to use a `http://root:root@localhost:8529` url and extract the user and password from it.

### DIFF
--- a/contrib/lib/Cargo.toml
+++ b/contrib/lib/Cargo.toml
@@ -39,6 +39,7 @@ postgres_pool = ["databases", "postgres", "r2d2_postgres"]
 sqlite_pool = ["databases", "rusqlite", "r2d2_sqlite"]
 memcache_pool = ["databases", "memcache", "r2d2-memcache"]
 diesel_mysql_pool = ["databases", "diesel/mysql", "diesel/r2d2"]
+arango_pool = ["databases", "arangors", "r2d2_arangors"]
 
 [dependencies]
 # Global dependencies.
@@ -68,6 +69,9 @@ rusqlite = { version = "0.24", optional = true }
 r2d2_sqlite = { version = "0.17", optional = true }
 memcache = { version = "0.15", optional = true }
 r2d2-memcache = { version = "0.6", optional = true }
+arangors = { version = "0.5", default-features = false, features = ["reqwest_blocking", "rocksdb"], optional = true }
+#r2d2_arangors = { version = "0.2", optional = true }
+r2d2_arangors = { git = "https://github.com/Weasy666/r2d2-arangors", branch = "develop", optional = true }
 
 # SpaceHelmet dependencies
 time = { version = "0.2.9", optional = true }
--- a/contrib/lib/src/databases/mod.rs
+++ b/contrib/lib/src/databases/mod.rs
@@ -316,7 +316,7 @@
 //! The list below includes all presently supported database adapters and their
 //! corresponding [`Poolable`] type.
 //!
-// Note: Keep this table in sync with site/guite/6-state.md
+// Note: Keep this table in sync with site/guide/6-state.md
 //! | Kind     | Driver                | Version   | `Poolable` Type                | Feature                |
 //! |----------|-----------------------|-----------|--------------------------------|------------------------|
 //! | MySQL    | [Diesel]              | `1`       | [`diesel::MysqlConnection`]    | `diesel_mysql_pool`    |
@@ -325,6 +325,7 @@
 //! | Sqlite   | [Diesel]              | `1`       | [`diesel::SqliteConnection`]   | `diesel_sqlite_pool`   |
 //! | Sqlite   | [`Rusqlite`]          | `0.24`    | [`rusqlite::Connection`]       | `sqlite_pool`          |
 //! | Memcache | [`memcache`]          | `0.15`    | [`memcache::Client`]           | `memcache_pool`        |
+//! | ArangoDB | [`arangors`]          | `0.4`     | [`arangors::Connection`]       | `arango_pool`          |
 //!
 //! [Diesel]: https://diesel.rs
 //! [`rusqlite::Connection`]: https://docs.rs/rusqlite/0.23.0/rusqlite/struct.Connection.html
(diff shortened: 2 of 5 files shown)


inzanez:
@Weasy666 wouldn't make sense to integrate using `mobc-arangors` as this would provide an `async` connection pool including `async` database driver connection?

Weasy666:
I went with `r2d2-arangors`, because `r2d2` is already used as pool for the other databases and i am not sure if @SergioBenitez and @jebrosen would want to integrate a different one.

inzanez:
@Weasy666 Oh, that makes sense I guess. Well, I will wait what happens on that end. If required, I will upgrade r2d2-arangors to allow serving async driver connections.

jebrosen:
The database pooling support implemented in `rocket_contrib` in the `master` branch is for synchronous databases, with a wrapper for using them safely in `async` code (#1375); this is partly because several of the existing supported drivers from 0.4 have no good `async` replacements. There is no async database pool support yet; there's some more discussion of that in #1117.

To be usable with the current (synchronous) implementation of `#[database]`, this PR would need to use `r2d2_arangors` and the blocking mode of `arangors` - which it looks like it does.

One concern I have with this approach is that it could make it difficult to add an integration with the async version of `arangors` until this one is removed (i.e. in a new major version of `rocket_contrib`). This is because you can't use the async and sync versions of the `arangors` crate simultaneously within one workspace, so we would need to clearly document that and make sure nothing (including e.g. rocket's own CI) ends up in that position.

inzanez:
@jebrosen couldn't both be integrated using features? Sync would be using `r2d2_arangors` which should work even if `rocket` is async, async would be using `mobc` or whatever will find it's way into async `rocket`.

jebrosen:
@inzanez Feature flags for dependencies are combined, so `arangors` will either be `async` or not-`async` for the whole project. If the synchronous version is used (e.g. for `r2d2-arangors`), then any code that uses the async API for `arangors` (such as `mobc-arangors`) won't compile. This would be a problem if `rocket_contrib` wants to add an async integration later, because you would have to make sure to never try to compile both at the same time.

inzanez:
@jebrosen I see. That makes totally sense. Maybe it would be better to wait for version `0.5` to be released and integrate the `async` version? In the meantime, @Weasy666 could be using `r2d2-arangors` with the pool as managed state? I currently have it integrated in a struct called `AppConfig` that is contained in an `Arc`. It carries application configuration data (non-rocket related) and the database pool from `r2d2`. Works pretty well, happy to share some code.

Weasy666:
@inzanez thanks for the offer. I'm already working with my fork and i think will continue doing that until one of the different options finds its way inside `rocket_contrib`. I'll come back to you offer, when i'm getting too annoyed with rebasing to keep up with Rockets.

@jebrosen That is a really annoying "feature" of cargo, with which i fought a few time in the past. I can understand your concerns about this problem, but doesn't this also already apply to the currently integrated databases? Or is `arangors` somewhat special in this regard because of its `async` and `blocking` features?

jebrosen:
@Weasy666 This situation is unique to `arangors` at least compared to `rocket_contrib`'s existing database integrations, because it has feature flags that *change* code instead of adding it. The existing databases provide only blocking APIs, sometimes with an async API in an entirely separate crate. Another approach that does not have this issue is to provide both async and blocking APIs simultaneously but in different modules, similarly to `reqwest` vs `reqwest::blocking`.

I personally don't have any huge issues with integrating `arangors` as either async or blocking; I only want to avoid the situation where we *want* to add it as async later and discover we can't.

Weasy666:
Ok, that makes sense...then i will try and ask the `arangors` maintainer, about their opinion to maybe go the `reqwest` way and split up `sync` and `async` in a breaking change.

SergioBenitez:
Indeed, it looks like the `arangors` project is misusing feature flags. This needs to be resolved before this PR can be merged.

contrib/lib/src/databases.rs
```diff
@@ -700,6 +707,27 @@ impl Poolable for memcache::Client {
     }
 }
 
+#[cfg(feature = "arango_pool")]
+impl Poolable for arangors::Connection {
+    type Manager = r2d2_arangors::pool::ArangoDBConnectionManager;
+    type Error = r2d2::Error;
```
SergioBenitez, line 713: This should be `std::convert::Infallible`.

contrib/lib/src/databases.rs
```diff
@@ -700,6 +707,27 @@ impl Poolable for memcache::Client {
...
+        #[derive(Deserialize, Debug)]
+        struct ArangoConfig {
+            username: String,
+            password: String,
+            use_jwt: bool,
+        }
+
+        let config = Config::from(db_name, rocket)?;
+        let extra = Figment::from(rocket.figment())
+            .extract_inner::<ArangoConfig>(&format!("databases.{}", db_name))?;
```
SergioBenitez, line 725: These can be on one line. Also, `Figment::from(rocket.figment()` is redundant; `rocket.figment()` is already a `Figment`.

contrib/lib/src/databases.rs
```diff
@@ -700,6 +707,27 @@ impl Poolable for memcache::Client {
...
 }
 
+#[cfg(feature = "arango_pool")]
+impl Poolable for arangors::Connection {
+    type Manager = r2d2_arangors::pool::ArangoDBConnectionManager;
+    type Error = r2d2::Error;
+
+    fn pool(db_name: &str, rocket: &rocket::Rocket) -> PoolResult<Self> {
+        #[derive(Deserialize, Debug)]
+        struct ArangoConfig {
```
SergioBenitez, line 717: Where are these configuration options documented? We'll need them somewhere, otherwise it'll be impossible for someone to properly configure this database and thus use it at all.

SergioBenitez (changes requested):

contrib/lib/src/databases.rs
Weasy666: Ah...i oriented myself on what is done in `Config::from()`, but that is to please the borrow checker?

contrib/lib/src/databases.rs
Weasy666: That's right, i will write something about that. Is it ok to add it here, right above the `## Extended`?
https://github.com/SergioBenitez/Rocket/blob/77faddfc8b458f5b23135f86fa857573cc18d81c/contrib/lib/src/databases.rs#L352-L353

contrib/lib/src/databases.rs
SergioBenitez: That one calls `.merge()` on the `Figment`, which requires a move of `Figment`, but `rocket.figment()` gives you an `&Figment`. Thus, we either `.clone()` or create a new one from the borrow; I chose the latter for no particular reason. Here, you simply extract from the `Figment`, so no move is needed and you can use the borrow directly.

contrib/lib/src/databases.rs
Weasy666: I've added docs at that position. Feel free to move them around if that is the wrong place 😅

SergioBenitez:
Marking as a draft until we have upstream movement.

SergioBenitez:
I am closing this given that there is no progress possible until upstream resolves this issue. Furthermore, we will start advocating for using the `async` database pools in #1696 as soon as it's ready, eventually deprecating `sync_db_pools`. If there is an `async` ArangoDB driver, we should seek to integrate it into #1696.

Weasy666:
@SergioBenitez @jebrosen 
Understandable, seeing the lack of progress/activity upstream.
`arangors` is not only a `sync` but also an `async` driver, which was the whole problem here, or better its handling of both features. If it is not a problem for you that, when integrating `async` arangors, a user can not use `sync` arangors in rocket while using the rocket provided `async` integration of arangors, then i can take a look and try to integrate it after #1696 was merged.

SergioBenitez:
I think that would be a better, if imperfect, compromise.

PR #1439, opened 2020-09-26
Close code block in docs for into_inner()

olback:


jebrosen (approved):
Interestingly enough `rustdoc` doesn't seem to care. But it's a good idea anyway!

jebrosen:
Thanks! Merged as 29f4726.

PR #1441, opened 2020-09-28
websocket support based on tungstenite

markazmierczak:
This is my attempt on #90 WebSocket support in Rocket.


SergioBenitez:
Thank you for this! I think this serves as a wonderful starting point. More efforts are currently taking place in #1639. Also see my comment in https://github.com/SergioBenitez/Rocket/issues/90#issuecomment-841866637 for what a "full" implementation would look like.

Issue #1411, opened 2020-08-12
State should implement Clone+Copy


timokoesters:
State is just a reference to a type. There's no reason why it can't be copied or clones (to create another reference to that type).
Use-case: One of my endpoints calls another endpoint method internally, supplying a State<>. Because this consumes the state, I can't do anything else with it afterwards.

igalic:
usually the solution is to put an `Arc` into `State` which can very cheaply be cloned

Since that is such a common pattern, we might as well implement `State` as `Arc`…

timokoesters:
Even if the type in State is an Arc, the State struct does not implement Clone or Copy

jebrosen:
Yeah, `State<'_, T>` should be constructible from some kind of pointer-to-T for easier reuse. Conservatively, I see no problem with implementing `Clone` but I am less sure about `Copy` (which is more restrictive on the kinds of changes we could make to `State` in the future).

igalic:
does this now mean we can use `State` without an `Arc`?

SergioBenitez:
@igalic You can put whatever you'd like in `State` as long as it's `Send + Sync + 'static`; there was never any requirement to use an `Arc`. You'd only need an `Arc` prior if you were sharing ownership of the value inside of the `State` - this remains true today.

The latest change simply implements `Clone` by copying the internal reference to the managed state. It doesn't provide shared ownership, nor does it copy or clone the internal value.

PR #1442, opened 2020-10-02
Implement Clone on State

jontymorris:
Hello, this is my first Rocket contribution, so please bear with me :)
I'm going after the low-hanging fruit described in issue #1411

The issue originally asked for Clone and Copy to be implemented on the State struct, and although @jebrosen was fine with implementing Clone, he wasn't particularly fond of Copy as he thought we might end up shooting ourselves in the foot.

Please let me know what you think :+1:

### DIFF
--- a/core/lib/src/request/tests.rs
+++ b/core/lib/src/request/tests.rs
@@ -1,8 +1,8 @@
-use std::net::{IpAddr, Ipv4Addr, SocketAddr};
 use std::collections::HashMap;
+use std::net::{IpAddr, Ipv4Addr, SocketAddr};
 
-use crate::{Rocket, Request, Config, State};
 use crate::http::hyper;
+use crate::{Config, Request, Rocket, State};
 
 macro_rules! assert_headers {
     ($($key:expr => [$($value:expr),+]),+) => ({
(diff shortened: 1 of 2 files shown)


SergioBenitez:
Will the derive generate the implementation we want (a copy) without adding a `T: Clone` bounds? We should add a test to confirm; we should be able to clone a `State<T>` for any `T`.

SergioBenitez:
@jontymorris Ping!

jontymorris:
@SergioBenitez Pong!

I've added a test and found the cloning will only work if their underlying config struct also implements the Clone trait. So if people want to be able to clone their States then they will just need to derive Clone on their config struct. Does this sound acceptable to you?

SergioBenitez:
@jontymorris That's only a consequence of deriving `Clone`. You can simply implement `Clone` manually as:

```rust
impl<'r, T: Send + Sync + 'static> State<'r, R> {
    fn clone(&self) -> State<'r, T> {
        State(self.0)
    }
}
```

jontymorris:
@SergioBenitez Nice, looks like I still have a lot to learn about Rust. Thanks for sharing the knowledge! What do you think of the PR now?

jontymorris:
Note: The State's config struct no longer has to derive Copy

core/lib/src/request/tests.rs
```diff
@@ -47,3 +47,23 @@ fn test_multiple_headers_merge_into_one_from_hyp() {
     assert_headers!("friend" => ["alice"], "friend" => ["bob"], "friend" => ["carol"]);
     assert_headers!("friend" => ["alice"], "friend" => ["bob"], "enemy" => ["carol"]);
 }
+
+struct MyConfig {
+    user_val: String,
+}
+
+#[tokio::test]
```
jebrosen, line 55: Other async tests use `#[rocket::async_test]` (which defaults to the single-threaded runtime instead of the multi-threaded runtime), which I would prefer for consistency with other tests.

jebrosen (approved):
One really minor thing, but otherwise LGTM!

SergioBenitez:
Merged with a few changes in b7565172eb07e64e35e3cbb3d762281e001dc77e. Thanks!

PR #1445, opened 2020-10-10
Added DerefMut impl to Form and LenientForm

ThouCheese:
I ran into a limitation with `Form` where I needed to modify the inner data, but the `DerefMut` implementation was missing. I have also added `DerefMut` to `LenientForm` for consistency.

### DIFF
--- a/core/lib/src/request/form/form.rs
+++ b/core/lib/src/request/form/form.rs
@@ -1,4 +1,4 @@
-use std::ops::Deref;
+use std::ops::{Deref, DerefMut};
 
 use crate::outcome::Outcome::*;
 use crate::request::{Request, form::{FromForm, FormItems, FormDataError}};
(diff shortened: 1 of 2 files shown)


jebrosen:
From the matrix chat:
> ThouCheese: Is there a reason why Form<T> and LenientForm<T> implement Deref<Target=T>, but not DerefMut?
> Sergio: Don't think so. There used to be, but I can't think of a reason for why now.

Prior to ec4cc3a `Form` was self-referential, so mutable access would have been quite dangerous. Now that `Form<T>` is just a wrapper around `T`, and #810 even made the field public, I can't think of a good reason not to add this.

jebrosen (changes requested):
The docs for `Form<T>` say ``automatically dereferences into an `&T` ``. Let's add `&mut T` as well: ``dereferences into an `&T` or `&mut T` ``.

ThouCheese:
@jebrosen fixed! I cant' figure out why the CI on Linux doesn't work (or how this PR could possibly break anything :sweat:)

jebrosen (approved):
LGTM. 83a7fc4 in `master` seems to have fixed the CI failure in this PR.

SergioBenitez:
Merged in 080d586a356b7ffc3617b98ecbc57457f1d9acaa. Thank you!

PR #1453, opened 2020-10-21
Change `Poolable::pool()` arguments to `&Cargo` and `name: &str`

Weasy666:
This change allows custom implementations of `Poolable` to access all config parameters instead of only the ones specified by `database::Config`.

A few things that i am not so sure about:
- I have added a new variant to [DbError](https://github.com/Weasy666/Rocket/blob/bc85af98cac0684752c5e33fcf9a78a3ec192504/contrib/lib/src/databases.rs#L507-L508). But as a consequence i also had to change errors from `r2d2::Error` to `DbError<r2d2::Error>` and remove the last doc block of `Poolable`, the one starting with "In the event that a connection manager isn't fallible...".
- Another consequence of the new error variant is, that i don't know if [this](https://github.com/Weasy666/Rocket/blob/bc85af98cac0684752c5e33fcf9a78a3ec192504/contrib/lib/src/databases.rs#L762-L769) should stay as is, or if the `config: Config` field in `ConnectionPool` should be replaced with a `cargo: &Cargo` field. But we then we maybe end up in a lifetime hell 🤔. We could also replace it with a `timeout` field, because the `config` field is only used to access the timeout [here](https://github.com/Weasy666/Rocket/blob/bc85af98cac0684752c5e33fcf9a78a3ec192504/contrib/lib/src/databases.rs#L792) (at least as far as i could see).
- And...i am also not sure how to handle the error [here](https://github.com/Weasy666/Rocket/blob/bc85af98cac0684752c5e33fcf9a78a3ec192504/contrib/lib/src/databases.rs#L782). I would really like to match for `DbError::Config` to be able to use a different error message, but associated type defaults on a trait are unstable i don't know how to match this otherwise.

### DIFF
--- a/contrib/lib/src/databases.rs
+++ b/contrib/lib/src/databases.rs
@@ -402,23 +402,22 @@ use self::r2d2::ManageConnection;
 /// [global.databases.my_database]
 /// url = "postgres://root:root@localhost/my_database"
 /// pool_size = 10
-/// certs = "sample_cert.pem"
-/// key = "key.pem"
+/// timeout = 5
 /// ```
 ///
-/// The following structure would be generated after calling
-/// [`database_config`]`("my_database", &config)`:
+/// would generate the following struct:
 ///
 /// ```rust,ignore
 /// Config {
-///     url: "dummy_db.sqlite",
+///     url: "postgres://root:root@localhost/my_database",
 ///     pool_size: 10,
-///     extras: {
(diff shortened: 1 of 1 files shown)


SergioBenitez:
Merged in f5fd1007e2e7d241c643321a3c7ba2948b4c699b. Thank you!

Weasy666:
You are welcome.

PR #1459, opened 2020-10-27
Add description to mimic Rocket's config behaviour

Weasy666:
When using Rocket with the new config system, i had to wrestle a little bit with it.
I followed the guide and API docs, i used Rockets default configs, merged them with my own config defaults and then pointed them to my `TOML` file and....was wondering why the `port` was still Rocket's default. It took me a little while to understand, that i had to also replicate the profiles and merge with `.nested()`. it makes absolute sense if you have a certain basic knowledge about `Figment`, but i think adding that to the guide and rustdocs is helpful for people that are just starting with it.

### DIFF
--- a/core/lib/src/config/mod.rs
+++ b/core/lib/src/config/mod.rs
@@ -92,10 +92,21 @@
 //! }
 //! ```
 //!
+//! If the application's config should behave like the configuration of Rocket,
+//! i.e. use the same [`Profile`]s and behaviour, you need to instead merge them as follows:
+//!
+//! ```rust,ignore
+//! let figment = Figment::from(rocket::Config::default())
+//!     .merge(Serialized::defaults(Config::default()))
+//!     .merge(Toml::file(Env::var_or("APP_CONFIG", "App.toml")).nested())
+//!     .merge(Env::prefixed("APP_").global());
+//! ```
+//!
 //! [`rocket::custom()`]: crate::custom()
 //! [`rocket::ignite()`]: crate::ignite()
 //! [`Toml`]: figment::providers::Toml
 //! [`Env`]: figment::providers::Env
+//! [`Profile`]: https://docs.rs/figment/0.9/figment/struct.Profile.html
 
 mod secret_key;
 mod config;
--- a/site/guide/9-configuration.md
+++ b/site/guide/9-configuration.md
@@ -340,6 +340,16 @@ Rocket will extract it's configuration from the configured provider. This means
 that if values like `port` and `address` are configured in `Config`, `App.toml`
 or `APP_` environment variables, Rocket will make use of them. The application
 can also extract its configuration, done here via the `Adhoc::config()` fairing.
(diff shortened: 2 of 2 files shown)


SergioBenitez:
I'd like to understand better where you feel you could have used more/better docs.

> I followed the guide and API docs, i used Rockets default configs, merged them with my own config defaults and then pointed them to my `TOML` file and....was wondering why the `port` was still Rocket's default.

It sounds like you're saying you used Rocket's default config; is this `Config::default()` or `Config::figment()`? And, if you did use some of Rocket's defaults, why were you surprised that Rocket's default `port` was being used? Were you setting a `port` in your own config without it being used?

> It took me a little while to understand, that i had to also replicate the profiles and merge with `.nested()`. it makes absolute sense if you have a certain basic knowledge about `Figment`, but i think adding that to the guide and rustdocs is helpful for people that are just starting with it.

You don't need to use `nested()` at all; `nested()` only determines how values from the config source will be merged/joined into the `Figment`.

Weasy666:
Ok i will explain it on a simplified example. I wanted to use a `App.toml` to configure my application including Rocket. The file had content like this
```toml
[global.limits]
forms = "64 kB"
json = "1 MiB"

[debug]
domain = "http://localhost:2020"
address = "127.0.0.1"
port = 2020

[debug.mailer]
from = "info@example.com"
smtp_server = "smtp.example.com:587"
smtp_user = "john"
smtp_password = "super_secret"

[release]
address = "0.0.0.0"
port = 2020
```

I then initially loaded the config like shown in the guide
```rust
let figment = Figment::from(rocket::Config::default())
        .merge(Serialized::defaults(AppConfig::default()))
        .merge(Toml::file("App.toml"))
        .merge(Env::prefixed("APP_"));
```

This resulted in Rocket still using its default port, because the port in the debug profile was not overwritten. I understand now why this was so, and like i said, it makes sense. But if you are not familiar with `Figment`, then you scratch your head. Especially, because i interpreted the text in the guide, such that the code snipped lets me simply use a toml file with a different name instead of `Rocket.toml`. So i thought it would be helpful for other to explicitely mention that. I hope this answers you question?

SergioBenitez:
> This resulted in Rocket still using its default port, because the port in the debug profile was not overwritten. I understand now why this was so, and like i said, it makes sense. But if you are not familiar with `Figment`, then you scratch your head. Especially, because i interpreted the text in the guide, such that the code snipped lets me simply use a toml file with a different name instead of `Rocket.toml`. So i thought it would be helpful for others to explicitely mention that. I hope this answers you question?

Got it. If this is the case, then I think simply changing the example to use `.nested()` (and `.global()` on `Env` if we're really mimicking Rocket) and specifying what it does should suffice. I've pushed a commit (edc91f65b77dde09e85605c1ff4c1418a3254257) that does just this. If you feel that this is still insufficient, please do let me know.

Weasy666:
No, i think this is sufficient. Thank you!

PR #1490, opened 2020-12-11
Implement FromData for String and Vec<u8> in release builds.

toxeus:
These trait implementations are useful and I was surprised
when the release build suddenly failed.


toxeus:
I've noticed today that keeping the implementation debug-only is intended behavior and documented [here](https://api.rocket.rs/master/rocket/data/trait.FromTransformedData.html?search=#provided-implementations).

PR #1497, opened 2020-12-29
Allow clippy::unit_arg lint in generated code

### LINKED ISSUE #1495: clippy: "warning: passing a unit value to a function"
## description:
When a route returns a unit (`()`), clippy emits this warning:
```
warning: passing a unit value to a function
  --> src/rest/api/auth/mod.rs:26:8
   |
26 | pub fn logout(ctx: &App, jwt: &JWT) {
   |        ^^^^^^
   |
   = note: `#[warn(clippy::unit_arg)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unit_arg
help: move the expression in front of the call...
   |
26 | pub fn logout;
   |        ^^^^^^^
help: ...and use a unit literal instead
   |
26 | pub fn ()(ctx: &App, jwt: &JWT) {
   |    
```
adding a return type resolves this warning.
The snipped which provoked this warning is this:
```rust
#[post("/logout")]
pub fn logout(ctx: &App, jwt: &JWT) {
    ctx.jwt.invalidated_users.invalidate(jwt.user_id);
}
```

jebrosen:
Fixes #1495.

This PR can be verified with `cargo clippy -p request_local_state -- --no-deps`, which shows two `unit_arg` warnings in the example before this PR and no warnings after.

This is tricky to test in CI in the current state, since we have some failing clippy lints in `rocket`, `rocket_http`, etc (some of which are "errors" in the default configuration).

### DIFF
--- a/core/codegen/src/attribute/route.rs
+++ b/core/codegen/src/attribute/route.rs
@@ -431,6 +431,7 @@ fn codegen_route(route: Route) -> Result<TokenStream> {
         /// Rocket code generated proxy static conversion implementation.
         impl From<#user_handler_fn_name> for #StaticRouteInfo {
             fn from(_: #user_handler_fn_name) -> #StaticRouteInfo {
+                #[allow(clippy::unit_arg)]
                 fn monomorphized_function<'_b>(
                     #req: &'_b #Request,
                     #data: #Data


SergioBenitez:
Can we simply allow _all_ `clippy` lints for generated code? We could gate this in such a way that we can selectively enable clippy linting if required.

jebrosen:
I believe we could insert `#[allow(clippy::all)]` in a few places - certainly for the autogenerated `#[route]` and `#[catch]` code which is the most likely to run into lints. I'm not sure if lint attributes are allowed in every single place we generate code (in particular I'm worried about `uri!`), but let's see how far we can get!

> We could gate this in such a way that we can selectively enable clippy linting if required.

Would this part be needed right away? We could do another environment variable e.g. `ROCKET_CODEGEN_UNMASK_CLIPPY=1`

SergioBenitez:
> Would this part be needed right away? We could do another environment variable e.g. `ROCKET_CODEGEN_UNMASK_CLIPPY=1`

I was intending as a `cfg`, to the effect of the existing `rocket_unsafe_secret_key`, so that we can enable it in scripts when needed, perhaps even as a alllow-fail CI task.

SergioBenitez:
This appears to no longer be required: https://github.com/SergioBenitez/Rocket/issues/1495

PR #1499, opened 2020-12-30
Implement HTTP 405

### LINKED ISSUE #1224: Rocket should show 405 instead of 404 for method not allowed

peng1999:
This is basically a squash merge of #1335. Original changes are preserved as possible, and I added some minor fixes and improvements.

This should fix #1224.

### DIFF
--- a/core/codegen/tests/route-format.rs
+++ b/core/codegen/tests/route-format.rs
@@ -61,7 +61,7 @@ fn test_formats() {
     assert_eq!(response.into_string().unwrap(), "plain");
 
     let response = client.put("/").header(ContentType::HTML).dispatch();
-    assert_eq!(response.status(), Status::NotFound);
+    assert_eq!(response.status(), Status::MethodNotAllowed);
 }
 
 // Test custom formats.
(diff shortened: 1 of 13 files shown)


SergioBenitez:
This approach unfortunately has significant performance impacts. To properly support 405 without a large performance impact, we would need to redesign our routing strategy entirely, perhaps to one based automata. Until then, we cannot implement this well.

PR #1500, opened 2020-12-30
Update dependencies for tokio 1.0

### LINKED ISSUE #1451: Update to tokio 1.0
New version of `tokio` is released
rocket vesrion is `0.2` [Cargo.toml](https://github.com/SergioBenitez/Rocket/blob/master/core/http/Cargo.toml#L32)

Chance to make breaking changes while `rocket = "0.5"` not released yet

jebrosen:
Closes #1451.

It might be nice to switch some code to use [`tokio_util::io::ReaderStream`](https://docs.rs/tokio-util/0.6.0/tokio_util/io/struct.ReaderStream.html) and [`tokio_util::io::StreamReader`](https://docs.rs/tokio-util/0.6.0/tokio_util/io/struct.StreamReader.html) now instead of the implementations here, but I would rather do that part separately.

Possible blocker: support in multipart library

### DIFF
--- a/contrib/lib/Cargo.toml
+++ b/contrib/lib/Cargo.toml
@@ -15,7 +15,7 @@ edition = "2018"
 # Internal use only.
 templates = ["serde", "serde_json", "glob", "notify"]
 databases = [
-    "serde", "r2d2", "tokio/blocking", "tokio/rt-threaded",
+    "serde", "r2d2", "tokio/rt", "tokio/rt-multi-thread",
     "rocket_contrib_codegen/database_attribute"
 ]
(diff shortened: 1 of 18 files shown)


SergioBenitez:
> Possible blocker: support in multipart library

On it.

SergioBenitez:
@jebrosen I'm happy to merge this either with or without the use of the new stream adapters: your call!

jebrosen:
Unfortunately those stream adapters aren't quite drop-in replacements due to some API mismatches, so I'd rather merge this without.

PR #1507, opened 2021-01-08
Add setter and getter for request_local

hei-pa:
In some situations (in a fairing - on_request) you only want to set a value to the local cache. There is no need to retrieve a value in this situation.

And if you are only interessted in getting a value when present a getter is sufficient.

### DIFF
--- a/core/lib/src/request/request.rs
+++ b/core/lib/src/request/request.rs
@@ -580,6 +580,12 @@ impl<'r> Request<'r> {
         self.state.cache.try_get()
     }
 
+    pub fn local_cache_set<T>(&self, d: T)
+    where T: Send + Sync + 'static
+    {
+        self.state.cache.set(d);
+    }
+
     /// Retrieves the cached value for type `T` from the request-local cached
     /// state of `self`. If no such value has previously been cached for this
     /// request, `fut` is `await`ed to produce the value which is subsequently
--- a/core/lib/src/request/request.rs
+++ b/core/lib/src/request/request.rs
@@ -574,6 +574,12 @@ impl<'r> Request<'r> {
             })
     }
 
+    pub fn local_cache_get<T>(&self) -> Option<&T>
+    where T: Send + Sync + 'static
+    {
+        self.state.cache.try_get()
+    }
+
     /// Retrieves the cached value for type `T` from the request-local cached
     /// state of `self`. If no such value has previously been cached for this
     /// request, `fut` is `await`ed to produce the value which is subsequently
--- a/core/lib/src/request/request.rs
+++ b/core/lib/src/request/request.rs
@@ -574,12 +574,39 @@ impl<'r> Request<'r> {
             })
     }
 
+    /// Retrieves the cached value for type `T` from the request-local cached
+    /// state of `self`. If no such value has previously been cached for this
+    /// request, `None` is returned.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// async fn on_response<'r>(&self, req: &'r Request<'_>, res: &mut Response<'r>) {
+    ///     if let Some(session) = req.local_cache_get::<Session>() {
+    ///         debug!("{:?}", session);
+    ///         res.set_header(Cookie::new("rocket-session", String::from(&session.id)));
+    ///     }
+    /// }
+    /// ```
     pub fn local_cache_get<T>(&self) -> Option<&T>
     where T: Send + Sync + 'static
     {
         self.state.cache.try_get()
     }
 
+    /// Sets the given value to the request-local cached
+    /// state of `self`.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// async fn on_request(&self, req: &mut Request<'_>, _data: &mut Data) {
+    ///    if let Some(session_cookie) = req.cookies().get("rocket-session") {
+    ///       let session_id = session_cookie.value().to_string();
+    ///       req.local_cache_set(session_id);
+    ///    }
+    /// }
+    /// ```
     pub fn local_cache_set<T>(&self, d: T)
     where T: Send + Sync + 'static
     {


jebrosen:
The CI failure is <https://gitlab.com/antonok/enum_dispatch/-/issues/37> via `memcache`. I will try upgrading `memcache` in a separate PR and see if that works as a fix.

jebrosen (commented):
This API looks fine to me on its own, but I think it was not done this way intentionally: <https://github.com/SergioBenitez/Rocket/issues/654#issuecomment-395124033>.

---

There are a few different patterns I can think of for request-local state:

* A cached value, where the computed value would be the same every time. This is useful for "expensive" `FromRequest` implementations e.g. database access.
* A value that is set only once in a request. This could be useful for storing a timestamp or ID.
* A value that can be mutated during the request.

All of these can be accomplished with either API, but some patterns are easier than others:

| operation | getter/setter API | cache API \* |
| --- | --- | --- |
| cached value | `try_get`; `set` + `get` if `None` | `local_cache(compute_value())` |
| set a value once | `set` | `local_cache(value)` or `local_cache(Some(value))` |
| set a value that can be modified later | `set` | `local_cache(Mutex::new(value))` or `local_cache(Some(Mutex::new(value))` |
| get a value | `try_get` | `local_cache(default_value)` or `local_cache(None).unwrap()` |

\* These should be `local_cache(|| ...)` instead of `local_cache(...)`, but it breaks GitHub's table formatting.

SergioBenitez:
I am generally not in favor of this change, but I could be convinced. This introduces the possibility of failure where the current API has none. Furthermore, I do not see the ergonomic advantages of the `set` API:

```diff
- req.local_cache_set(value);
+ req.local_cache(|| value);
```

The `get` API, on the other hand, allows for the possibility of setting values that are only available once, for whatever reason. For instance, it allows for the following:

```rust
req.local_cache(move || once_value);
// ... somewhere else, without `once_value`.
let val = req.locally_cached::<T>()?;
```

If there is a compelling example requiring the latter, I would be in favor of this change.

hei-pa:
If I am not interessted in a value, eg. during a `fairing.on_request` it feels weird to use `local_cache(|| value)` and ignore the returned result. Why do you think it produced a possibility of failure? It is using the same setter which is used in the `local_cache(..)` when nothing is found.

And in a `fairing.on_response` I am interessted in getting a value from the cache if present. Yes using `local_cache(my_type)` would be possible but I have to wrap my value in a `Option<>` just to make it possible providing a "default" value, because my type does not have `Default` implemented. So using `local_cache_get(..)` returns an `Option<>` by default and there is no need to wrap my type in an extra `Option<>`.

SergioBenitez:
> If I am not interessted in a value, eg. during a `fairing.on_request` it feels weird to use `local_cache(|| value)` and ignore the returned result. Why do you think it produced a possibility of failure? It is using the same setter which is used in the `local_cache(..)` when nothing is found.

I don't think this is weird at all. There are several APIs in `stdlib` where this occurs all the time, for instance, [`HashMap::insert()`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.insert).

> And in a `fairing.on_response` I am interessted in getting a value from the cache if present. Yes using `local_cache(my_type)` would be possible but I have to wrap my value in a `Option<>` just to make it possible providing a "default" value, because my type does not have `Default` implemented. So using `local_cache_get(..)` returns an `Option<>` by default and there is no need to wrap my type in an extra `Option<>`.

I don't follow. Why do you need to wrap it in `Option` if you have the value? Perhaps concrete code will help.

hei-pa:
I am implementing a Session Fairing

```rust
fn query_session(&self, id: &String) -> Session {
 ...
}

impl Fairing for SessionFairing {
    fn info(&self) -> Info {
        Info {
            name: "Session Fairing",
            kind: Kind::Request | Kind::Response,
        }
    }
    async fn on_request(&self, req: &mut Request<'_>, _data: &mut Data) {
        if let Some(session_cookie) = req.cookies().get("rocket-session") {
            let session_id = session_cookie.value().to_string();
            req.local_cache_set(self.query_session(&session_id)); // here i want to set only ... there's nothing in the cache ever
        } else {
            req.local_cache_set(self.create_session());
        }
    }
    async fn on_response<'r>(&self, req: &'r Request<'_>, res: &mut Response<'r>) {
        if let Some(session) = req.local_cache_get::<Session>() { // here i need to get a value ... don't want to set ever
            res.set_header(Cookie::new("rocket-session", String::from(&session.id)));
        }
        /* using the local_cache(...)
        if let Some(session) = req.local_cache::<Session>(|| what goes here ? dummy 'Session' ?) {
        if let Some(session) = req.local_cache::<Option<Session>>(|| None) { // possible but I have to wrap my Session in an Option
            ...
        }
        */
    }
}
```

SergioBenitez:
I'd write your example as follows:

```rust
impl SessionFairing {
    fn set_and_get_session<'a>(&self, req: &'a Request<'_>) -> &'a Session {
        req.local_cache(|| {
            req.cookies()
                .get("rocket-session")
                .map(|c| self.query_session(c.value()))
                .unwrap_or_else(|| self.create_session())
        })
    }
}

impl Fairing for SessionFairing {
    fn info(&self) -> Info {
        Info {
            name: "Session Fairing",
            kind: Kind::Request | Kind::Response,
        }
    }

    async fn on_request(&self, req: &mut Request<'_>, _data: &mut Data) {
        self.set_and_get_session();
    }

    async fn on_response<'r>(&self, req: &'r Request<'_>, res: &mut Response<'r>) {
        res.set_header(Cookie::new("rocket-session", self.set_and_get_session().id));
    }
}
```

As you can see, this removes the need to check for `Option` in `on_response`. Do you have an objection to this example?

hei-pa:
Yes I agree to your example and will rewrite my code ;-)

SergioBenitez:
Glad to hear! Closing. :)

hei-pa:
Hey @SergioBenitez how to solve the need of a 'getter' here?

```
#[rocket::async_trait]
impl<'a, 'r> FromRequest<'a, 'r> for &'a Session {
    type Error = std::convert::Infallible;

    async fn from_request(request: &'a Request<'r>) -> request::Outcome<Self, Self::Error> {
        Success(
            request
                .local_cache_get::<Session>()
                .expect("Failed to get session"),
        )
    }
}
```

PR #1419, opened 2020-09-01
[WIP] Add MTLS support

howard0su:
Without adding new dependency, tries to add MTLS support.

Not finished yet. Open PR to get earlier review as I am not a experienced rust developer.

PR #1531, opened 2021-02-06
Once again, TLS client auth

MikailBag:
Based on #1419 by @howard0su.
TODO: add tests for the `tls` example. (done)
# Description
Server configuration can enable client authentication. Rocket will
verify that presented client certificate is signed with a certificate
authority, specified in the configuration. There is also mode when
clients without certificate are not allowed to connect.
    
Also, request handler can fetch certificate information using
special request guard. That way the handler can verify e.g.
CN or SAN fields values.


SergioBenitez:
This contains a bunch of style changes making it difficult to review. Can you remove any and all style changes?

MikailBag:
I hope all formatting changes are gone now.

SergioBenitez:
@jebrosen Could you take a look at this when you have a chance?

core/http/src/tls.rs
```diff
@@ -125,4 +151,25 @@ impl Connection for TlsStream<TcpStream> {
...
+        (self.get_ref().1).get_peer_certificates()
+    }
+}
+
+/// Mutual TLS information.
+/// This struct contains client certificate itself and
+/// its trust chain. Additionally, several useful certificate fields
+/// are exposed for your convenience.
+#[derive(Debug)]
+pub struct MutualTlsInfo {
```
jebrosen, line 165: This is a lot of public fields; I would be happier exposing functions that return `&[u8]` or `Vec<u8>`.

core/lib/src/request/from_request.rs
```diff
@@ -501,3 +503,56 @@ impl<'a, 'r, T: FromRequest<'a, 'r> + 'a> FromRequest<'a, 'r> for Option<T> {
...
+        _ => None,
+    }
+}
+
+#[cfg(feature = "tls")]
+#[crate::async_trait]
+impl<'a, 'r> FromRequest<'a, 'r> for MutualTlsInfo {
+    type Error = ();
+
+    async fn from_request(request: &'a Request<'r>) -> Outcome<Self, Self::Error> {
```
jebrosen, line 524: This looks relatively expensive to do in `from_request`, which might be called multiple times on the way to one route handler or will only have one or two of these fields used. Can we move the parsing code (and the `x509-parser` dependency) onto methods of the `MutualTlsInfo` struct in `rocket_http`, and construct `MutualTlsInfo` from only the underlying certificate bytes?

jebrosen (changes requested):
This looks like a great start; it looks like roughly the minimum amount of code to support mTLS yet still useful, and in more or less the right places.

My biggest design question / concern is the API of `MutualTlsInfo`. From cursory research the `rustls` stack seems to not reexport any certificate type in public APIs and generally passes certificates around as `&[u8]`s. We could do that, but the full API of e.g. `x509-parser` is certainly a bit nicer.

MikailBag:
I've performed quick some research.

I have not found validation code in rustls.
Certificate validation is done by `webpki` [here](https://github.com/briansmith/webpki/blob/5fd730eb1844248451a330f669249c27e553b9a1/src/end_entity.rs#L144)
- webpki does not seem to validate/parse CN at all
- when parsing SANs, it skips all unsupported/non-`DNSName` SANs

It means, both CNs and SANs can contain some strange thing, and Rocket mustn't panic in this case.

I am going to implement your suggestion and move the certificate parsing code to methods.

MikailBag:
Failures don't seem to be related.
Should I rebase onto master?

jebrosen:
I recently pushed a commit to `master` to fix the errors, so they should go away after a rebase now.

SergioBenitez:
Checking in. Is this ready for another review?

MikailBag:
I think I've addressed all concerns raised so far.

However, there are now some merge conflicts I need to resolve.

MikailBag:
I hope I haven't messed up anything during rebasing.
At least unit tests seem to pass.

examples/tls/src/tests.rs
```diff
@@ -1,8 +1,40 @@
...
+    let response = client.get("/secret").dispatch();
+    assert_eq!(response.status().code, 404);
+}
+
+fn get_cert() -> Certificate {
+    let sans = ["mary".to_string()];
+    let cert = rcgen::generate_simple_self_signed(sans.to_vec()).unwrap();
+    let cert = cert.serialize_der().unwrap();
+    Certificate(cert)
+}
```
SergioBenitez, line 22: We shouldn't ask users to figure out how to generate a simple one of these. We should provide methods on `Certificate` to do this much more easily.

examples/tls/src/tests.rs
```diff
@@ -1,8 +1,40 @@
-use rocket::local::blocking::Client;
+use rocket::{http::tls::Certificate, local::blocking::Client};
```
SergioBenitez, line 1: This is a private type!

core/http/src/tls.rs
```diff
@@ -125,4 +151,81 @@ impl Connection for TlsStream<TcpStream> {
...
 }
+
+impl MutualTls {
+    /// Parses presented client certificate.
+    ///
+    /// This function ignores all data it can not understand.
+    ///
+    /// For more advanced scenarios consider using crates such as
+    /// `x509-parser` or `openssl`.
+    pub fn parse(&self) -> CertificateFields {
```
SergioBenitez, line 204: Why is this method on `MutualTls` and not on `Certificate`?

core/http/src/tls.rs
```diff
@@ -125,4 +151,81 @@ impl Connection for TlsStream<TcpStream> {
...
+        // TODO: can this panic? Maybe return empty fields instead?
+        let crt = x509_parser::parse_x509_certificate(&self.end_entity.0)
+            .expect("invalid certificate was accepted by rustls")
```
SergioBenitez, line 207: We can't do this.

MikailBag:
Thanks for your review!

I'm going to make the following changes to address some problems raised:
1) Provide our own `Certificate` type (instead of rustls re-export)
  - `parse` can be moved to `Certificate`
  - `rustls` is not visible in public API
2) Add `Certificate::generate` on top of the `rcgen`.
3) Change `Option<Vec<Certificate>>` to `Option<(Certificate, Vec<Certificate>)>` for better type safety. 
4) According to [this](https://security.stackexchange.com/questions/210669/is-it-possible-to-setup-1-way-tls-client-authentication-without-server-certifica) stackexchange Q&A client authentication is completely independent of server authentication. I'm going to implement it (with some naming changes, e.g. `MutuatTls` -> `ClientTls`). 

Regarding some other comments:
1) `parse` function loses very much information (it only extracts CN and SANs and skips all names it cannot parse). The most simple way to fix it is to just return `x509_parser::Certificate` instead of `CertificateFields`. However, it will make `x509-parser` a public dependency. Is it OK?
2) Potential parse errors. Unfortunately, it seems that `webpki` implements its own certificate parser, and we can't reuse that. It is always possible that `x509-parser` does not accept a certificate that was accepted by `webpki`, so panicking is not a good option. Do you think we should return `Result/Option` in this case?
3) I don't think the `Connection` trait should have a certificate as an associated type, because we can't make `MutualTls` generic over the underlying connection. And if `MutualTls` contains some concrete certificate type, then all connections have to return this concrete type.
4) IIUC Certificates, configured on LocalRequest, completely bypass the transport layer, so they are not validated against the CA, nor they are checked to be valid. I think we should just document this behavior.

SergioBenitez:
> * `parse` function loses very much information (it only extracts CN and SANs and skips all names it cannot parse). The most simple way to fix it is to just return `x509_parser::Certificate` instead of `CertificateFields`. However, it will make `x509-parser` a public dependency. Is it OK?

Why don't we copy the relevant fields to our own `Certificate` type or embed the `x509_parser::Certificate` in our `Certificate` and  expose data via methods?

> * Potential parse errors. Unfortunately, it seems that `webpki` implements its own certificate parser, and we can't reuse that. It is always possible that `x509-parser` does not accept a certificate that was accepted by `webpki`, so panicking is not a good option. Do you think we should return `Result/Option` in this case?

Yes. We cannot panic.

> * I don't think the `Connection` trait should have a certificate as an associated type, because we can't make `MutualTls` generic over the underlying connection. And if `MutualTls` contains some concrete certificate type, then all connections have to return this concrete type.

That's fine.

> 4\. IIUC Certificates, configured on LocalRequest, completely bypass the transport layer, so they are not validated against the CA, nor they are checked to be valid. I think we should just document this behavior.

That's right.

On a related note: is it possible, at the moment, to distinguish between a bad certificate and a missing certificate? What if I want to issue a 403 in one case and redirect in the latter? Related to testing, we should be able to "generate" certificates that let us test both cases as well as the "good" certificate case.

MikailBag:
> Is it possible, at the moment, to distinguish between a bad certificate and a missing certificate? 

My understanding is as follows:

If client authentication is `required`, both bad and missing certificate will be rejected by webpki during establishing TLS connection, so neither Rocket nor application will not observe request at all

If client authentication is `optional`, bad (syntactically invalid, or not matching configured CAs, or expired, etc) again will be rejected at the transport level. But if the certificate is not provided at all, the request will be accepted and handled (but with an empty certificate chain). The application can then distinguish authenticated and unauthenticated clients using the `Option<MutualTls>` request guard. If it is Some, then the client had presented a valid, trusted certificate.

If we want to e.g. silently ignore bad certificates, or reject them at some later stage, I think we have to provide manual  [ClientCertVerifier](https://docs.rs/rustls/0.19.0/rustls/trait.ClientCertVerifier.html) implementation.

SergioBenitez:
> If we want to e.g. silently ignore bad certificates, or reject them at some later stage, I think we have to provide manual [ClientCertVerifier](https://docs.rs/rustls/0.19.0/rustls/trait.ClientCertVerifier.html) implementation.

I think this would be idea. Perhaps we can leave this as `TODO`, open an issue with examples of what _would_ be possible, and wait for feedback.

MikailBag:
Sorry for the long delay!

I have rebased my branch onto master to resolve merge conflicts.

Additionally, I've added a comment regarding broken certificate handling, and it seems that this was the last remaining piece of feedback.

So it seems that this PR is ready for re-review.

SergioBenitez:
Please see the build failures. Please run `./scripts/test.sh` locally.

MikailBag:
Tests pass locally now

scripts/test.sh
```diff
@@ -54,6 +54,8 @@ function check_style() {
   if ! [ -z "${matches}" ]; then
     echo "Trailing whitespace was found in the following:"
     echo "${matches}"
+    echo "help: Run following command from repository root:"
+    echo 'git grep -E -I -n "\s+$" .'
```
SergioBenitez, line 58: Please remove this. The script is already printing this information.

MikailBag: Yeah, it seems that the test script now prints not only filenames but also offending lines (IIRC it didn't do this when I created this PR...). Anyway, this change is not related to this PR goal.

MikailBag:
IIUC, there is only one big design question left. The question is what certificate information Rocket should expose besides raw bytes.
1. Nothing. User has to parse certificate themselves. Pros: minimal API surface, cons: user has to make choices (e.g. select parser) even for simple purposes.
2. Bless a couple of parsers (maybe one, maybe more) and add simple methods that parse certificates using them (e.g. `parse_x509_parser`, `parse_openssl`, etc). This option is currently implemented in this PR. Pros: user does not have to decide on anything. Cons: data exposed (e.g. currently [this](https://docs.rs/x509-parser/0.9.2/x509_parser/certificate/struct.X509Certificate.html) type is returned) is not too friendly for simple use cases, and underlying parser becomes a public dependency.
3. Create a struct that holds some basic metadata of the certificate and add a method that returns it. This option was implemented previously. Pros: probably the best option for simple purposes, cons: hopefully no.

What's your opinion? Should we return to option 3? If yes, should we also implement option 2?

By the way, I hope PR is ready for review. CI in my forked repository is green.

SergioBenitez:
> What's your opinion? Should we return to option 3? If yes, should we also implement option 2?

Yes, I believe option 3 is what I've advocated for in prior comments. Please review those. 

> By the way, I hope PR is ready for review. CI in my forked repository is green.

This looks good. I will prioritize this as soon as we have an RC for the next major release.

SergioBenitez:
@MikailBag Has the previous review been addressed? If so, I am happy to review once a rebase on `master` is complete.

MikailBag:
I have resolved merge conflicts, and CI is green.

I _think_ the only unreviewed part of the PR is the `Certificate::parse` method. Currently, it implements option 3 (i.e. completely hides underlying x509 parser). I don't have good ideas of which parts of the certificate should be fetched by this function and which should be ignored. The current choice (common name and organization unit) is just the first one that came into my head.

Also, if it is easier for you, I can exclude `Certificate::parse` from this PR and submit it as a follow-up PR (smaller diff will be likely easier to review). The same applies to the `Certificate::generate`.

SergioBenitez:
Unfortunately I've had to re-implement the majority of this PR due to issues too numerous and cross-cutting for the GitHub review process. Nevertheless, your implementation @MikailBag and the original one by @howard0su made the task significantly easier. I've split the implementation into two commits. The first, bbc36ba27fea494cbab20eaaff73a204aac93c12, is one co-authored by all three of us, and the second 7ffe3a73601aabdfe3190a809e660538acd5bdce largely contains the changes I've made since then.

PR #1532, opened 2021-02-08
Disable rustfmt

MikailBag:
Rocket does not use rustfmt because it still has some bugs and sometimes
produces badly formatted code. However, some IDEs and editors
automatically run rustfmt in background, polluting PRs with unrelated
formatting changes.

This patch avoids that specific problem by adding syntactically invalid
rustfmt config. That way rustfmt will immediately crash and will not
modify sources.

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,3 @@
+# rustfmt produces badly formatted code so it is
+# disabled
+ignore = ["/"]


SergioBenitez:
That's pretty interesting. How do editors respond to this? Will folks receive an error message, or is this silent?

It would be ideal if rustfmt itself had a way to say, "hey, don't touch this!"

MikailBag:
I have only vscode+rust-analyzer installed, and it seems that the formatting request gets silently ignored.

It's not ideal behavior, but I think it is still better than the current situation when unrelated code is changed.

> It would be ideal if rustfmt itself had a way to say, "hey, don't touch this!"

I think that's a good idea (at least it will be less hacky than creating broken toml), I will create rustfmt issue too.

MikailBag:
Filed https://github.com/rust-lang/rustfmt/issues/4691

jebrosen:
I believe the last several times this (or similar) were proposed, `ignore` was an unstable option. Now that it's stable, it does seem like the right way to opt out.

MikailBag:
Turns out there is a non-hacky solution. I changed PR to use it.

Unfortunately, rust-analyzer does not respect `ignore` option, but it seems this should be fixed on rust-analyzer side.

calebcartwright:
@MikailBag - apologies as I forgot that folks wouldn't be working exclusively with nightly. Somewhat ironically, the unstable/nightly-only `ignore` option in the config would be ignored when on stable and formatting would proceed anyway.

If your immediate goal is to prevent editor format-on-save, including on stable, then your original approach of an invalid config file would be your best bet.

There's some outstanding issues with `ignore` that make near-term stabilization unlikely, but I'll look into seeing if I can stabilize [`disable_all_formatting`](https://rust-lang.github.io/rustfmt/?version=v1.4.35&search=#disable_all_formatting) within the next release or two because I know folks that have configured their editors to format-on-save can have frustrating experiences when working on projects that don't use the formatter. Will note for transparency though that our core goal on the rustfmt team is ensuring code is reformatted according to the [Style Guide](https://github.com/rust-dev-tools/fmt-rfcs), whereas helping folks avoid rustfmt is not exactly the highest of priorities for us.

SergioBenitez:
There does not appear to be a stable solution in `rustfmt`. It seems `disable_all_formatting` was stabilized and then unstabilized? The issue history there is a bit perplexing.

Nevertheless, I would be in favor of either disabling `rustfmt` for the project or finding options that get very close to what the codebase does now. In other words, options for which running `cargo fmt` results in minimal churn today, except for improving consistency and likely reordering imports and such. As this PR doesn't quite do either, I'm closing it out.

Issue #1498, opened 2020-12-29
Consider using `http::HeaderMap` internally


SergioBenitez:
At present, Rocket spends significantly more time than necessary converting an [`http::HeaderMap`](https://docs.rs/http/0.2.2/http/header/struct.HeaderMap.html) and its internal values into a [`rocket::http::HeaderMap`](https://api.rocket.rs/v0.4/rocket/http/struct.HeaderMap.html). More importantly, the conversion is directly in the request processing path, so it effects the performance of every request:

https://github.com/SergioBenitez/Rocket/blob/9671115796e42865eaebd020ac27542802027b02/core/lib/src/request/request.rs#L887-L893

Instead of our entirely custom `HeaderMap` type, we should consider leveraging the work in `http` by using an `http::HeaderMap` internally, wrapping it in a custom type to preserve our documentation and interoperability with the rest of Rocket's types. The `http::HeaderMap` allows customization of the internal value; for `hyper`, this is [`http::header::HeaderValue`](https://docs.rs/http/0.2.2/http/header/struct.HeaderValue.html). Unfortunately, unlike Rocket's header map, this does not allow borrowing header values of a particular lifetime; all values are heap-allocated. The same issue occurs with header names. What's more, because values are stored as bytes in `HeaderMap` (a good thing), retrieving them as strings requires a UTF-8 check (a bad thing). The map also does not preserve case which is known to cause issues downstream.

In summary, there are _at least_ the following road-blocks to using an `http::HeaderMap` internally:

  * Header names and values cannot borrow.
  * Retrieving a value as a string requires a UTF-8 check every time.
  * The map does not preserve header name casing; it should still be case-insensitive during comparison.

jebrosen:
> Header names and values cannot borrow.

We *might* actually be able to do this already, by having `Request::from_hyp` take an `&'r HeaderMap`! This still allocates when `String::from_utf8_lossy` does.

SergioBenitez:
> > Header names and values cannot borrow.
> 
> We _might_ actually be able to do this already, by having `Request::from_hyp` take an `&'r HeaderMap`! This still allocates when `String::from_utf8_lossy` does.

[...]

PR #1535, opened 2021-02-10
Fix #1067 by importing typed headers from hyperx

### LINKED ISSUE #1067: (async) Remove or replace of typed headers

jespersm:
First attempt patch for #1067.

Imports all typed headers from `hyperx` except "Set-Cookie" since it requires multiline format, and thus can't do simple to_string().

Probably not a big deal since Rocket handles cookies through CookieJar mechanism.

### DIFF
--- a/core/codegen/tests/ui-fail-nightly/responder-types.stderr
+++ b/core/codegen/tests/ui-fail-nightly/responder-types.stderr
@@ -6,15 +6,18 @@ error[E0277]: the trait bound `u8: Responder<'_, '_>` is not satisfied
   |
   = note: required by `respond_to`
 
-error[E0277]: the trait bound `Header<'_>: From<u8>` is not satisfied
+error[E0277]: the trait bound `Header<'_>: std::convert::From<u8>` is not satisfied
   --> $DIR/responder-types.rs:11:5
    |
 11 |     other: u8,
-   |     ^^^^^^^^^ the trait `From<u8>` is not implemented for `Header<'_>`
+   |     ^^^^^^^^^ the trait `std::convert::From<u8>` is not implemented for `Header<'_>`
    |
    = help: the following implementations were found:
-             <Header<'static> as From<&Cookie<'_>>>
-             <Header<'static> as From<Cookie<'_>>>
+             <Header<'static> as std::convert::From<&rocket::http::Cookie<'_>>>
+             <Header<'static> as std::convert::From<AcceptCharset>>
+             <Header<'static> as std::convert::From<AcceptEncoding>>
+             <Header<'static> as std::convert::From<AcceptLanguage>>
+           and 55 others
    = note: required because of the requirements on the impl of `Into<Header<'_>>` for `u8`
 
 error[E0277]: the trait bound `u8: Responder<'_, '_>` is not satisfied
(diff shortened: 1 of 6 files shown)


jespersm:
I think this PR is ready for review now, @jebrosen :-)

I'm not entirely sure that the reexport of HyperxHeaderTrait is elegant enough, but there's a "Header" confusion to take into account.

jebrosen (changes requested):
I do like this, but I think we can organize it a bit more nicely.

One thing I did realize while reviewing this is that `hyperx` does not implement some of traits `hyper` uses it for headers, cc #1498. That's one point in favor of `headers` instead of `hyperx` that I did not consider, depending on how we prioritize #1498.

jespersm:
I’ll take another look at #1498 to see how this interacts.

jespersm:
I've looked further at `headers`, which appears to use a `HeaderValue` struct from `headers-core`, which is yet another one,  different one from the `HeaderValue` in `hyper`. That doesn't seem to line up with the change proposed in #1498, but I might not be Rust-y enough to know if there's a simple way to do that without cloning the underlying bytes.

jebrosen:
> I've looked further at `headers`, which appears to use a `HeaderValue` struct from `headers-core`, which is yet another one, different one from the `HeaderValue` in `hyper`. That doesn't seem to line up with the change proposed in #1498, but I might not be Rust-y enough to know if there's a simple way to do that without cloning the underlying bytes.

Both `headers-core`'s and `hyper`'s `HeaderValue` types are actually re-exports of `http::headers::HeaderValue` (<https://docs.rs/headers-core/0.2.0/src/headers_core/lib.rs.html#13>, <https://docs.rs/hyper/0.14.4/src/hyper/lib.rs.html#73>).

@SergioBenitez is #1498 enough of a concern that we should avoid using `hyperx` to provide typed headers, which currently has non-zero overhead to be converted to the `HeaderValue` type used by `hyper`?

jespersm:
> Both `headers-core`'s and `hyper`'s `HeaderValue` types are actually re-exports of `http::headers::HeaderValue` (https://docs.rs/headers-core/0.2.0/src/headers_core/lib.rs.html#13, https://docs.rs/hyper/0.14.4/src/hyper/lib.rs.html#73).

Ah, I was confused by a layer of turtles.

Anyway, I'm giving #1498 a shot to see how big an impact it would have.

SergioBenitez:
> @SergioBenitez is #1498 enough of a concern that we should avoid using `hyperx` to provide typed headers, which currently has non-zero overhead to be converted to the `HeaderValue` type used by `hyper`?

I believe this is the single largest source of slowdown in request handling. We should do nothing to exacerbate the issue or further prolong its mending.

SergioBenitez:
> @SergioBenitez #1498 addressed handling of headers in the request, but didn't touch `Response`s which still unconditionally copies data as it did in 0.4 (https://github.com/SergioBenitez/Rocket/blob/master/core/lib/src/server.rs#L130). What needs to happen to implement this with the desired performance characteristics?

I don't think there's anything we can reasonable do about the response, unfortunately. See https://github.com/hyperium/http/pull/467. Hopefully we get an upstream fix.

SergioBenitez:
I am closing this due to lack of progress. Happy to revisit if all of the points raised above are addressed.

Issue #1283, opened 2020-04-16
New Module For Retrieving Headers


gsquire:
## Idea
While working through a small application, I wanted to fetch a request header to do some inspection and respond accordingly. I looked through existing issues and comments in this repository saying to use the `FromRequest` trait for the desired header. This was pretty straightforward and I was wondering if it would be useful to have a new module in `contrib` that implemented this as a macro for various common headers? This way, developers could use them as request guards to fetch headers that they want to use.

[Here](https://github.com/gsquire/rust_bin/blob/1d04e9354e25ad9e7ceca8397cb6b94034efeb11/src/main.rs#L43-L68) is how I used it to fetch the user agent. In this case, I failed if the header wasn't there but in the implementation, we can use an `Option` and always succeed.

## Implementation Idea

This is roughly what I think it could look like. Keep in mind this won't compile as is.

```rust
// Perhaps have a macro to make types too.
#[derive(Debug)]
struct $ty {
    value: String,
}

// This could be inside of a macro.
impl<'a, 'r> FromRequest<'a, 'r> for $ty {
    type Error = ();

    fn from_request(request: &'a Request<'r>) -> request::Outcome<Option<Self>, Self::Error> {
        const HEADER: &str = $k; // The macro would expand the key 'k'.

        let value = request.headers().get_one(HEADER);
        match value {
            Some(v) => Outcome::Success(Some($ty {
                value: v.to_string(),
            })),
            None => Outcome::Success(None),
        }
    }
}
```

jebrosen:
Although this example is useful as-is and makes sense in a utility module or crate, I think for inclusion in `rocket_contrib` I would rather see a solution using a typed header API. Note also that in 0.4 we do have the reverse (typed header from hyper -> rocket `Header`), but in the `async` branch we have lost typed headers altogether (see #1067).

---

A feature like this could also be a really nice application of const generics (still quite a ways away from being stable, I fear):

```rust
const CONTENT_LENGTH: &str = "Content-Length";
const CONTENT_TYPE: &str = "Content-Type";

struct Header<'a, const NAME: &str>(value: &'a str);

impl<'a, 'r, const NAME> FromRequest<'a, 'r> for Header<'a, NAME> {
    type Error = ();
    fn from_request(request: &'a Request<'r>) -> request::Outcome<Self, Self::Error> {
        request.headers().get_one(NAME).expect("TODO: example")
    }
}

#[get("/")]
[...]

PR #1538, opened 2021-02-11
Implement #1283 idea using typed headers

jespersm:
This PR implements FromRequest for a number of the typed headers from the `hyperx`crate.
It depends on PR #1535.

It allows for constructs such as: 

```rust
use rocket::http::hyper::header::AcceptLanguage;

#[get("/poem")]
fn poem(lang: Option<AcceptLanguage>) -> String {
    if let Some(lang) = lang {
        format!("TODO: Write a poem in '{}'", lang)
    } else {
        format!("A poem in any other language would sound as sweet!")
    }
}
```

The list of headers may need to be looked at since a number of the supported typed headers are only used in responses.


SergioBenitez:
I am closing this as an implication of closing #1535.

PR #1552, opened 2021-02-27
Drop the 'local::asynchronous::Client' inside 'local::blocking::Client' on the runtime

jebrosen:
Currently the default `Drop` impl drops the `Client` "off-runtime", when other operations were done "on-runtime". This affects e.g. `#![database]`, which attempts to call `spawn_blocking` to safely drop the database pool.

This could probably be fixed or worked around in `#[database]` instead, in which case any other fairings or managed state that did blocking I/O on drop would need to implement a similar workaround separately.

### DIFF
--- a/core/lib/src/local/blocking/client.rs
+++ b/core/lib/src/local/blocking/client.rs
@@ -26,7 +26,7 @@ use crate::http::Method;
 ///     .dispatch();
 /// ```
 pub struct Client {
-    pub(crate) inner: asynchronous::Client,
+    pub(crate) inner: Option<asynchronous::Client>,
     runtime: RefCell<tokio::runtime::Runtime>,
 }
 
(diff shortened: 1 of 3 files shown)


core/lib/src/local/blocking/client.rs
```diff
@@ -88,6 +88,13 @@ impl Client {
     pub_client_impl!("use rocket::local::blocking::Client;");
 }
 
+impl Drop for Client {
+    fn drop(&mut self) {
+        let client = self.inner.take();
```
SergioBenitez, line 93: Instead of making this an `Option`, so we can `take()`, can we simply `swap` in a new `asynchronous::Client` and drop the old one in `self.block_on()`? This would avoid all of the `as_ref().unwrap()` noise above.

SergioBenitez (changes requested):

core/lib/src/local/blocking/client.rs
jebrosen: That was my first idea, but swapping in a new `asynchronous::Client` first requires constructing an entire `Rocket` instance.

SergioBenitez:
Merged in d778c2cb! :)

PR #1563, opened 2021-03-07
Support selecting TLS cipher suites and protocol versions

BinAlyan:
**This my first pull request ever, sorry for any mistakes.**

**To summarize the major changes:**

- I have created a public enum 'Version" to indicate TLS protocol version which contains only 1.2 and 1.3
  -  rustls only support version 1.2 and 1.3 however their enum [ProtocolVersion](https://docs.rs/rustls/0.19.0/rustls/enum.ProtocolVersion.html) includes all TLS and SSL versions.
- I have created a public enum 'CipherSuite' to indicate supported cipher suites.
  - rustls currently support 9 cipher, they are stored in a static array [here](https://docs.rs/rustls/0.19.0/rustls/static.ALL_CIPHERSUITES.html)
- I have added two more members in TlsConfig which are both Vec of the previous enums. They can be set via the following parameters tls.ciphersuites and tls.versions.

I am not sure where to add tests. I have added doc examples for the added functions, changed an existing doc example and modified the TLS example to set "tls.ciphersuites" and "tls.versions" from Rocket.toml.

### DIFF
--- a/core/http/src/tls.rs
+++ b/core/http/src/tls.rs
@@ -5,6 +5,8 @@ use std::pin::Pin;
 use std::sync::Arc;
 use std::task::{Context, Poll};
 
+pub use rustls::{ProtocolVersion, SupportedCipherSuite, ciphersuite};
+
 use rustls::internal::pemfile;
 use rustls::{Certificate, PrivateKey, ServerConfig};
 use tokio::net::{TcpListener, TcpStream};
@@ -95,6 +97,9 @@ pub async fn bind_tls<C: io::BufRead + Send, K: io::BufRead + Send>(
     address: SocketAddr,
     mut cert_chain: C,
     mut private_key: K,
+    ciphersuites: Vec<&'static SupportedCipherSuite>,
+    versions: Vec<ProtocolVersion>,
+    prefer_server_ciphers_order: bool,
 ) -> io::Result<TlsListener> {
     let cert_chain = load_certs(&mut cert_chain).map_err(|e| {
         let msg = format!("malformed TLS certificate chain: {}", e);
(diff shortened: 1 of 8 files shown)


SergioBenitez:
> **This my first pull request ever, sorry for any mistakes.**

Amazing! Welcome to open source! :)

> I am not sure where to add tests. I have added doc examples for the added functions, changed an existing doc example and modified the TLS example to set "tls.ciphersuites" and "tls.versions" from Rocket.toml.

We want config tests in `core/lib/config/mod.rs`. We should check that we can parse all possible values and that we _don't_ parse what we don't expect. See the existing tests there for pointers.

It also seems, in this case, we'd want to test that the actual server comes up/doesn't come up with valid/invalid combinations (more on this in my review). For these, we'll need to start up the actual server and then shut it down. This should be done as an integration test. See the `on_launch_fairing_can_inspect_port` test in `core/lib/tests` for how to do this.

SergioBenitez (changes requested):
There a few big concerns:

  1. Nowhere is it documented what these options do, precisely. The docstring in the `TlsConfig` structure for these new values needs to be more precise, and details needs to be added to the config guide. In particular, we need to be very explicit that the server will pick the first ciphersuite in the array that the client allows. Or will it? These seems to depend on whether we set `ignore_client_order` or not, which isn't mentioned here not does it appear to be configurable by the changes in this PR.
  2. What happens if I only enable TLS 1.2 but don't provide a TLS 1.2 ciphersuite? Do I get an error somewhere? Does rustls just happily move along without every negotiating a TLS connection? Ideally, this would be structurally impossible.

In short, what these options do need to be as obvious as possible and as error-prone-free as possible. I'm not sure we're quite there yet. Here are some suggestions to get us there:
  * Set V1.2 and V1.3 ciphersuites separately.
  * Either enable/disable v1.2/v1.3 based on whether any ciphersuites for the version are set or fail if the config doesn't make sense otherwise.
  * Allowing setting `ignore_client_order`, perhaps as a positive variant: `prefer_server_ciphers`.

My suggestion is to add `prefer_server_ciphers`, `v12.ciphers` and `v13.ciphers` to the `tls` config:

```rust
struct TlsConfig {
    prefer_server_ciphers: bool,
    v12: V12Config,
    v13: V13Config,
}

struct V12Config {
    ciphers: Vec<V12Ciphers>,
}

struct V13Config {
    ciphers: Vec<V13Ciphers>,
}
```

You can also flatten the physical structure with a custom `deserialize` implementation. Then, if they are non-empty, that version is enabled.

Alternatively, keep the current structure but check that there is at least one cipher for every configured version.

SergioBenitez:
We'll also need this rebased on master.

BinAlyan:
Hopefully I resolved all the previous comments, I feel the previous code was very different so most comments were outdated thus I marked them as resolved. Probably I would have missed some, misunderstood others, made new mistakes.

Issue #1566, opened 2021-03-08
TLS example test is not correct


BinAlyan:
It seems the test does not depend on Rocket.toml so even if there is an invalid key the test will be successful.

Demonstration:

```
$ cd examples/tls
$ rm private/key.pem
$ cargo test
   Compiling tls v0.0.0 (/home/abdullah/Workspace/forks/Rocket/examples/tls)
    Finished test [unoptimized + debuginfo] target(s) in 2.35s
     Running /home/abdullah/Workspace/forks/Rocket/target/debug/deps/tls-1eac40ebf78ab934

running 1 test
test tests::hello_world ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
$ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.07s
     Running `/home/abdullah/Workspace/forks/Rocket/target/debug/tls`
🔧 Configured for debug.
    => address: 127.0.0.1
    => port: 8000
    => workers: 12
    => keep-alive: 5s
    => limits: bytes = 8KiB, data-form = 2MiB, file = 1MiB, form = 32KiB, string = 8KiB
    => tls: enabled
    => secret key: [zero]
    => temp dir: /tmp
    => log level: normal
    => cli colors: true
🛰  Mounting /:
    => (hello) GET /
Error: Rocket failed to launch due to an I/O error.
    => error reading TLS file `private/key.pem`: No such file or directory (os error 2)
thread 'main' panicked at 'aborting due to i/o error', core/lib/src/error.rs:178:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

```

SergioBenitez:
TLS keys are only read/parsed at launch-time, when the server listens for TLS connections. Because the local client (as used in the example tests) cannot open a TCP/TLS connection, the TLS keys are never actually checked for validity in tests.

In general, in the config revamp, we'd erred towards only checking the semantic validity of configuration values when they are actually used. For example, a `secret_key` is only validated when the `secrets` feature is enabled; `tls` is only validated when TLS is enabled and in-use. This is safe and secure and reduces the need to configure properties that have no impact on the server.

[...]

PR #1449, opened 2020-10-12
Implement ecdsa keys

blaggacao:
towards #1448 (long term)

- [x] I have no real idea what I'm doing (first time touching rust code)
- [x] I'm remastering based on my common sense
- [x] I'm [blocked by this](https://github.com/ContainerSolutions/trow/issues/194), so it looked like the only option at hand.
- [x] `./scripts/test.sh --core` are green :tada: &mdash; by pure observation, my hands did the right thing without much of my brain involved.

This is intended to be a hand over, since I'm really not capable of addressing anything that goes much further than what's already in the PR.

PR #1567, opened 2021-03-08
Support more private keys types on PKCS8 format

BinAlyan:
Based on the previous pull request [here](https://github.com/SergioBenitez/Rocket/pull/1449). rustls already support different private key types, however they need to be in PKCS8 format [Check [here](https://github.com/ctz/rustls/pull/409#issuecomment-782300121)]. Adding support to them from Rocket is trivial.

I have added a tls_ecdsa example which use certifcate with signature scheme ECDSA_NISTP256_SHA256. The example includes a test, however it is copied from the TLS example so similarly it have this [issue](https://github.com/SergioBenitez/Rocket/issues/1566). The example ran successfully.


SergioBenitez:
Instead of creating a new example, can we simply set the keys as part of a new config profile in the existing `TLS` example? We should add an internal test for this in `http` so we don't depend on the example code.

BinAlyan:
In this pull request, I have added three new certificates with their keys and their gen scripts in the tls example. They can be run via Rocket profiles. They represent the currently supported signature schemes by rustls. They are as follows:
- ecdsa_nistp256_sha256
- ecdsa_nistp384_sha384
- ed2551

I have added two test functions which load these keys and certs. These two tests are unit tests for the functions load_private_key() and load_certs(). I am not sure if these two unit tests are enough or an integration test is required.

I load the keys and the certs from the tls example folder, I don't know if this is appropriate or kinda hacky. I am not sure of the macro I wrote because I never did Rust macros before.

SergioBenitez (commented):
Similarly with the generation scripts: why don't we have _one_ script that takes an argument controlling the kind of cert to generate?

BinAlyan:
> Similarly with the generation scripts: why don't we have _one_ script that takes an argument controlling the kind of cert to generate?

I am starting to make a single bash script generate all the certificates, however the bash script feel hacky and not that readable (to me), can I write the generation code using a rust code that invoke openssl via std::process::Command? I feel that would be simpler.

SergioBenitez:
Rust would be overkill for this. Just something like the following would be great:

```sh
function gen_ed25519() {
  openssl genpkey -algorithm ED25519 > ed25519_key.pem

  openssl req -new -key ed25519_key.pem -subj "${SUBJECT}" -out server.csr
  openssl x509 -req -extfile <(printf "subjectAltName=${ALT}") -days 3650 \
      -CA ca_cert.pem -CAkey ca_key.pem -CAcreateserial \
      -in server.csr -out ed25519_cert.pem

  rm ca_cert.srl server.csr
}

function gen_rsa_sha256() {
  openssl req -newkey rsa:4096 -nodes -sha256 \
    -keyout rsa_sha256_key.pem -subj "${SUBJECT}" -out server.csr

  openssl x509 -req -sha256 -extfile <(printf "subjectAltName=${ALT}") -days 3650 \
      -CA ca_cert.pem -CAkey ca_key.pem -CAcreateserial \
      -in server.csr -out rsa_sha256_cert.pem

  rm ca_cert.srl server.csr
}

case $1 in
  ed25519) gen_ed25519 ;;
  rsa_sha256) gen_rsa_sha256 ;;
  *)
    gen_ed25519
    gen_rsa_sha256
    ;;
esac
```

BinAlyan:
> Rust would be overkill for this. Just something like the following would be great:
> 
> ```shell
> function gen_ed25519() {
>   openssl genpkey -algorithm ED25519 > ed25519_key.pem
> 
>   openssl req -new -key ed25519_key.pem -subj "${SUBJECT}" -out server.csr
>   openssl x509 -req -extfile <(printf "subjectAltName=${ALT}") -days 3650 \
>       -CA ca_cert.pem -CAkey ca_key.pem -CAcreateserial \
>       -in server.csr -out ed25519_cert.pem
> 
>   rm ca_cert.srl server.csr
> }
> 
> function gen_rsa_sha256() {
>   openssl req -newkey rsa:4096 -nodes -sha256 \
>     -keyout rsa_sha256_key.pem -subj "${SUBJECT}" -out server.csr
> 
>   openssl x509 -req -sha256 -extfile <(printf "subjectAltName=${ALT}") -days 3650 \
>       -CA ca_cert.pem -CAkey ca_key.pem -CAcreateserial \
>       -in server.csr -out rsa_sha256_cert.pem
> 
>   rm ca_cert.srl server.csr
> }
> 
> case $1 in
>   ed25519) gen_ed25519 ;;
>   rsa_sha256) gen_rsa_sha256 ;;
>   *)
>     gen_ed25519
>     gen_rsa_sha256
>     ;;
> esac
> ```

I did not that was possible in bash. Thank you for your help. I have done the required changes.

core/http/src/tls.rs
```diff
@@ -126,3 +126,45 @@ impl Connection for TlsStream<TcpStream> {
...
+        let ecdsa_nistp256_sha256_cert = include_from_tls_example!("ecdsa_nistp256_sha256_cert.pem");
+        let ecdsa_nistp384_sha384_cert = include_from_tls_example!("ecdsa_nistp384_sha384_cert.pem");
+        let ed2551_cert = include_from_tls_example!("ed25519_cert.pem");
+
+        load_certs(&mut Cursor::new(rsa_sha256_cert))?;
+        load_certs(&mut Cursor::new(ecdsa_nistp256_sha256_cert))?;
+        load_certs(&mut Cursor::new(ecdsa_nistp384_sha384_cert))?;
+        load_certs(&mut Cursor::new(ed2551_cert))?;
+
+
```
SergioBenitez, line 167:
```suggestion

```

examples/tls/Rocket.toml
```diff
@@ -1,9 +1,26 @@
...
+# these certificate/key pairs. They are here for DEMONSTRATION PURPOSES ONLY.
+[default.tls]
+certs = "private/rsa_sha256_cert.pem"
+key = "private/rsa_sha256_key.pem"
+
+[rsa_sha256.tls]
+certs = "private/rsa_sha256_cert.pem"
+key = "private/rsa_sha256_key.pem"
+
+
```
SergioBenitez, line 15:
```suggestion

```

examples/tls/private/gen_certs.sh
```diff
@@ -0,0 +1,80 @@
+#! /bin/bash
+
+# Should use gen_ca.sh first to generate the CA.
```
SergioBenitez, line 3: Why don't we simply generate it here?

SergioBenitez (changes requested):

examples/tls/private/gen_certs.sh
BinAlyan: I separated the script for the case where the script user want to generate each certificate type separately. If the CA is generated each time, each CA generated will overwrite the previous CA. Should I add code that generates the CA only if it does not exist?

SergioBenitez: I think so! We could add a "ca" case to the script that overwrites an existing CA cert (IE, does what the gen_ca.sh script does) and otherwise write out the CA cert if it doesn't exist.

SergioBenitez:
I've finished off the PR and will merge as soon as the CI passes.

SergioBenitez:
Merged in af48d1f2e64ac353511c079855951c9cdc51177a!

BinAlyan:
Thank you for completing the pull request and merging it!  Sorry for being late in completing this simple commit.

In core/http/src/tls.rs tests, I don't think the name of the macro **tls_example_key** is appropriate as it also was used for certs.

I think there is a typo in examples/tls/Rocket.toml:

> The certificate/private key pairs used here **where** generated via openssl using

Issue #1137, opened 2019-09-15
Feature request: implementing IntoIterator on http::header::HeaderMap


atroche:
Hello! First some context:

I'd like to log request headers as JSON. I can't implement Serialize directly on HeaderMap, but I know if I can turn it into a `HashMap` I'll be fine. `HashMap` implements `FromIterator`, which means I can construct it easily from anything which implements `IntoIterator`.

`HeaderMap`'s underlying `IndexMap` implements it  `IntoIterator`, but is private. Meanwhile, `HeaderMap` has its own `into_iter` method, but not an implementation of `IntoIterator`.

So it's not like this is blocking me, it's just a little harder than it could be. Are you open to my making a PR with an `IntoIterator` implementation for `HeaderMap`?

Cheers.

edit: Hah! Just found [this TODO](https://github.com/SergioBenitez/Rocket/blob/master/core/http/src/header.rs#L609) in the code ;)

jebrosen:
Every `Iterator` implements `IntoIterator`, so I believe something like `headers.iter().collect::<HashMap<_, _>>()` should work already. Implementing `IntoIterator` on `HeaderMap` itself wouldn't help much because none of Rocket's APIs ever give out a `HeaderMap` by value, but we could implement `IntoIterator` for `&HeaderMap`.

SergioBenitez:
Agreed with @jebrosen. Happy to take a PR, but I don't think there's a need to track this.

PR #1574, opened 2021-03-13
Fix all Clippy errors and warnings

dallaires:
Good day,

I was trying Rocket's master branch because I want to use the stable compiler. I was not able to run Clippy on my project because of errors in Rocket so I've fixed all errors and warnings (I don't like having warning leaching into my projects).

I've made sure not add breaking changes. Add tests passed without modifications. I had to had some `#[allow(clippy::{lint name})]` to achieve this goal, but I've commented why I've allowed some lints.

I've also implemented the `IntoIterator` trait for the `HeaderMap` and removed the associated TODO. I've also updated the examples of `HeaderMap::into_iter()` and `HeaderMap::iter()` to make the test more relevant and making the example a bit more pedagogical. This change will also fix the [issue 1137](https://github.com/SergioBenitez/Rocket/issues/1137).

Now that all Clippy error and warnings are fixed, I would recommend to also ask for a clean Clippy output before accepting a pull request. Otherwise the amount of warnings and errors will only grow.

I hope you'll find this pull request useful.
Simon


SergioBenitez:
Please split off any functionality changes to a different PR!

contrib/lib/src/databases.rs
```diff
@@ -494,12 +494,12 @@ pub enum Error<T> {
     /// An error occurred while initializing an `r2d2` pool.
     Pool(r2d2::Error),
     /// An error occurred while extracting a `figment` configuration.
-    Config(figment::Error),
+    Config(Box<figment::Error>),
```
SergioBenitez, line 497: This is unnecessary. I don't think we should box here. Sure, we'll copy a bit more, but that's fine.

contrib/lib/src/databases.rs
```diff
@@ -848,7 +848,7 @@ impl<K, C: Poolable> Drop for Connection<K, C> {
                 // Explicitly dropping the permit here so that it's only
                 // released after the connection is.
                 drop(permit);
-            })
+            }).await
```
SergioBenitez, line 851: This...is interesting. I don't believe this does anything. @jebrosen: do you have any additional insight?

core/codegen/src/attribute/param/mod.rs
```diff
@@ -11,7 +11,7 @@ pub enum Parameter {
     Static(Name),
     Ignored(Dynamic),
     Dynamic(Dynamic),
-    Guard(Guard),
+    Guard(Box<Guard>),
```
SergioBenitez, line 14: I don't think we should do this.

core/codegen/src/attribute/route/parse.rs
```diff
@@ -104,12 +104,12 @@ impl FromMeta for RouteUri {
...
 impl Route {
     pub fn upgrade_param(param: Parameter, args: &Arguments) -> Result<Parameter> {
-        if !param.dynamic().is_some() {
+        if param.dynamic().is_none() {
             return Ok(param);
         }
 
         let param = param.take_dynamic().expect("dynamic() => take_dynamic()");
-        Route::upgrade_dynamic(param, args).map(Parameter::Guard)
+        Route::upgrade_dynamic(param, args).map(|gard| Parameter::Guard(Box::new(gard)))
```
SergioBenitez, line 112: We should be able to remove this change if we remove the `Box`.

core/codegen/src/http_codegen.rs
```diff
@@ -22,7 +22,7 @@ pub struct Optional<T>(pub Option<T>);
 impl FromMeta for Status {
     fn from_meta(meta: &MetaItem) -> Result<Self> {
         let num = usize::from_meta(meta)?;
-        if num < 100 || num >= 600 {
+        if !(100..600).contains(&num) {
```
SergioBenitez, line 25: This is much less obvious than the prior version. Let's revert this.

core/http/src/header/accept.rs
```diff
@@ -146,24 +146,22 @@ impl Accept {
         let mut all = self.iter();
         let mut preferred = all.next().unwrap_or(&ANY);
         for media_type in all {
-            if media_type.weight().is_none() && preferred.weight().is_some() {
```
SergioBenitez, line 149: Not only is this less legible than the previous, but it is also different logically as the previous version short-circuited and this one does not. Let's revert this.

core/http/src/header/header_impl.rs
```diff
@@ -535,29 +535,28 @@ impl<'h> HeaderMap<'h> {
     /// use rocket::http::{HeaderMap, Header};
     ///
     /// // The headers we'll be storing.
-    /// let all_headers = vec![
-    ///     Header::new("X-Custom", "value_1"),
-    ///     Header::new("X-Other", "other"),
-    ///     Header::new("X-Third", "third"),
+    /// let mut all_headers = vec![
```
SergioBenitez, line 538: Let's make this changes in a different PR.

core/http/src/header/mod.rs
```diff
@@ -3,11 +3,11 @@ mod known_media_types;
 mod media_type;
 mod content_type;
 mod accept;
-mod header;
+mod header_impl;
```
SergioBenitez, line 6: Why? Let's revert this.

core/http/src/listener.rs
```diff
@@ -136,12 +136,9 @@ impl<L: Listener + Unpin> Accept for Incoming<L> {
...
 fn is_connection_error(e: &io::Error) -> bool {
-    match e.kind() {
-        io::ErrorKind::ConnectionRefused |
-        io::ErrorKind::ConnectionAborted |
-        io::ErrorKind::ConnectionReset => true,
-        _ => false,
-    }
+    matches!(e.kind(), io::ErrorKind::ConnectionRefused |
+                       io::ErrorKind::ConnectionAborted |
+                       io::ErrorKind::ConnectionReset)
```
SergioBenitez, line 141:
```suggestion
    use io::ErrorKind::*;

    matches!(e.kind(), ConnectionRefused | ConnectionAborted | ConnectionReset)
```

core/http/src/parse/checkers.rs
```diff
@@ -5,9 +5,6 @@ pub fn is_whitespace(&byte: &char) -> bool {
 
 #[inline]
 pub fn is_valid_token(&c: &char) -> bool {
-    match c {
-        '0'..='9' | 'A'..='Z' | '^'..='~' | '#'..='\''
-            | '!' | '*' | '+' | '-' | '.'  => true,
-        _ => false
-    }
+    matches!(c, '0'..='9' | 'A'..='Z' | '^'..='~' | '#'..='\''
+                | '!' | '*' | '+' | '-' | '.')
```
SergioBenitez, line 9: A single line should be fine, I think.

core/http/src/raw_str.rs
```diff
@@ -759,7 +759,7 @@ macro_rules! impl_partial {
     )
 }
 
-impl_partial!(RawStr : &RawStr);
+// impl_partial!(RawStr : &RawStr); -> Using the default implementation to match Hash's default implementation
```
SergioBenitez, line 762: I don't understand; why are we removing this?

core/http/src/uri/mod.rs
```diff
@@ -1,6 +1,6 @@
 //! Types for URIs and traits for rendering URI components.
 
-mod uri;
+mod uri_impl;
```
SergioBenitez, line 3: I think this is unnecessary. Let's revert this.

core/lib/build.rs
```diff
@@ -3,19 +3,19 @@
...
-    const MIN_VERSION: &'static str = "1.46.0";
+    const MIN_VERSION: &str = "1.46.0";
 
     if let Some(version) = version_check::Version::read() {
         if !version.at_least(MIN_VERSION) {
             let msg = "Rocket requires a more recent version of rustc.";
             eprintln!("{} {}", Red.paint("Error:").bold(), Paint::new(msg).bold());
             eprintln!("Installed version: {}", Yellow.paint(format!("{}", version)));
-            eprintln!("Minimum required:  {}", Yellow.paint(format!("{}", MIN_VERSION)));
+            eprintln!("Minimum required:  {}", Yellow.paint(MIN_VERSION.to_string()));
```
SergioBenitez, line 13: I don't think we need `to_string()` at all, actually. Same with the `format!` above.

core/lib/src/config/mod.rs
```diff
@@ -111,15 +111,15 @@
 //! [`Toml`]: figment::providers::Toml
 //! [`Env`]: figment::providers::Env
 
-mod config;
+mod config_impl;
```
SergioBenitez, line 114: Let's revert this.

core/lib/src/data/data_stream.rs
```diff
@@ -59,7 +59,7 @@ enum State {
 /// The kinds of streams we accept as `Data`.
 enum StreamKind {
     Body(hyper::Body),
-    Multipart(multer::Field)
+    Multipart(Box<multer::Field>)
```
SergioBenitez, line 62: This seems unnecessary.

core/lib/src/data/mod.rs
```diff
@@ -2,13 +2,13 @@
 
 #[macro_use]
 mod capped;
-mod data;
+mod data_impl;
```
SergioBenitez, line 5: Let's revert this.

core/lib/src/data/temp_file.rs
```diff
@@ -184,6 +184,27 @@ impl<'v> TempFile<'v> {
         Ok(())
     }
 
+    /// Returns true if the file is empty.
```
SergioBenitez, line 187: Let's put this in another PR.

core/lib/src/fairing/ad_hoc.rs
```diff
@@ -61,22 +61,34 @@ pub struct AdHoc {
 //     );
 // }
 
+/// The attach fairing's function type
```
SergioBenitez, line 64: I don't think this is worth the noise. If we ever get trait aliases. then we can do this in a much nicer way that gets more reuseability.

core/lib/src/form/error.rs
```diff
@@ -804,9 +804,9 @@ impl fmt::Display for ErrorKind<'_> {
                 }
             }
             ErrorKind::InvalidChoice { choices } => {
-                match choices.as_ref() {
-                    &[] => write!(f, "invalid choice")?,
-                    &[ref choice] => write!(f, "expected {}", choice)?,
+                match *choices.as_ref() {
+                    [] => write!(f, "invalid choice")?,
+                    [ref choice] => write!(f, "expected {}", choice)?,
```
SergioBenitez, line 809: I find the previous version easier to read. Let's use that one.

core/lib/src/form/from_form.rs
```diff
@@ -415,10 +415,7 @@ impl<'v, T: FromForm<'v>> VecContext<'v, T> {
 
     fn context(&mut self, name: &NameView<'v>) -> &mut T::Context {
         let this_key = name.key();
-        let keys_match = match (self.last_key, this_key) {
-            (Some(k1), Some(k2)) if k1 == k2 => true,
-            _ => false
-        };
+        let keys_match = matches!((self.last_key, this_key), (Some(k1), Some(k2)) if k1 == k2);
```
SergioBenitez, line 418: This one is too complex, I think, for a single liner. Let's revert.

core/lib/src/form/from_form_field.rs
```diff
@@ -226,11 +226,11 @@ use crate::form::prelude::*;
 #[crate::async_trait]
 pub trait FromFormField<'v>: Send + Sized {
     fn from_value(field: ValueField<'v>) -> Result<'v, Self> {
-        Err(field.unexpected())?
+        Err(field.unexpected().into())
```
SergioBenitez, line 229: I think this and the rest in this file are fine. Let's revert them all.

core/lib/src/form/mod.rs
```diff
@@ -450,7 +450,7 @@ mod field;
 mod options;
 mod from_form;
 mod from_form_field;
-mod form;
+mod form_impl;
```
SergioBenitez, line 453: No need; revert.

core/lib/src/form/options.rs
```diff
@@ -1,6 +1,9 @@
 /// Form guard options.
 ///
 /// See [`Form#leniency`](crate::form::Form#leniency) for details.
+#[non_exhaustive] // TODO Remove when more than one option is available.
+                  // This attribute was added to avoid Clippy's "needless_update" warning inside
+                  // the function strict::init.
```
SergioBenitez, line 6: This is fine, actually, no need for a `TODO`.

core/lib/src/form/validate.rs
```diff
@@ -78,6 +78,9 @@
 //! }
 //! ```
 
+// The use of Result<()> in this file is part of the public API, fixing it will break everyone's code.
+#![allow(clippy::result_unit_err)]
+
```
SergioBenitez, line 83: This is a clippy bug. We shouldn't have this.

core/lib/src/form/validate.rs
```diff
@@ -228,6 +231,9 @@ pub fn neq<'v, A, B>(a: &A, b: B) -> Result<'v, ()>
 ///
 /// [`form::Result<'_, T>`]: crate::form::Result
 pub trait Len<L> {
+    /// Check if the value is empty in constant time.
+    fn is_empty(&self) -> bool;
```
SergioBenitez, line 235: Why are we adding this?

core/lib/src/form/validate.rs
```diff
@@ -596,7 +607,7 @@ pub fn range<'v, V, R>(value: &V, range: R) -> Result<'v, ()>
     };
 
 
-    Err((start, end))?
+    Err((start, end).into())
```
SergioBenitez, line 610: At the end of the function, I'm totally okay with an `Err(..)?`. Please revert this and other similar invocations.

core/lib/src/request/mod.rs
```diff
@@ -1,13 +1,13 @@
 //! Types and traits for request parsing and handling.
 
-mod request;
+mod request_impl;
```
SergioBenitez, line 3: Please revert.

core/lib/src/request/request_impl.rs
```diff
@@ -741,9 +741,8 @@ impl<'r> Request<'r> {
     pub fn query_value<'a, T>(&'a self, name: &str) -> Option<form::Result<'a, T>>
         where T: FromForm<'a>
     {
-        if self.query_fields().find(|f| f.name == name).is_none() {
-            return None;
-        }
+        // No need to continue if the name is not found.
+        self.query_fields().find(|f| f.name == name)?;
```
SergioBenitez, line 745: I don't know...I'd prefer the obvious version above, even if they are identical.

core/lib/src/response/mod.rs
```diff
@@ -27,7 +27,7 @@ mod responder;
 mod redirect;
 mod named_file;
 mod stream;
-mod response;
+mod response_impl;
```
SergioBenitez, line 30: Please revert.

SergioBenitez (changes requested):
The majority of these are fantastic! Please revert the indicates changes, however. And thank you for this! This is really great.

contrib/lib/src/databases.rs
dallaires: Replaced by allowing the lint with the associated comment :
```
#[allow(clippy::large_enum_variant)] // Config is only 210 bytes longer than the second longest
                                     // variant (Pool). This difference is acceptable.
```

contrib/lib/src/databases.rs
dallaires: This change was made to fix the error [async_yields_async](https://rust-lang.github.io/rust-clippy/master/index.html#async_yields_async). It think this lint is to avoid code like this one :
```
let x = async {
    other_async_fn()
  };
x.await
  .await
```

core/codegen/src/attribute/param/mod.rs
dallaires: I don't understand why it should not be boxed, but I suggest replacing it by :
```
#[allow(clippy::large_enum_variant)] // Gard should not be boxed even if it's 344 bytes longer than
                                     // Dynamic.
```
I don't like that comment, feel free to suggest me a better one : I try to explain why the lint shall be disabled because it should be the exception rather than the rule.

core/codegen/src/http_codegen.rs
dallaires: Clippy recommends it because it's less error prone (see [manual_range_contains](https://rust-lang.github.io/rust-clippy/master/index.html#manual_range_contains)). I've replace this line with :
```
#[allow(clippy::manual_range_contains)]  // Manual ranges are more obvious.
```
Even if I'm not used to that kind of syntax, I would recommend against reverting this change and allowing the lint because it is more idiomatic. I do believe that kind of syntax will become clearer when we get use to it.

core/http/src/header/accept.rs
dallaires: The logic did not change because the content of the if is the same (see [if_same_then_else](https://rust-lang.github.io/rust-clippy/master/index.html#if_same_then_else)). Here's the allow I've added :
```
    #[allow(clippy::if_same_then_else)] // For this particular case, we need to keep the if/else
                                        // branching for legibility.
```
If it is not normal to have the same content inside theses if, we will need to create an issue to fix this function properly.

core/http/src/raw_str.rs
dallaires: This one was to fix Clippy's error [derive_ord_xor_partial_ord](https://rust-lang.github.io/rust-clippy/master/index.html#derive_ord_xor_partial_ord). When I've removed `PartialOrd` and `Ord`, I've got the error [derive_hash_xor_eq](https://rust-lang.github.io/rust-clippy/master/index.html#derive_hash_xor_eq). I'm not fluent enough in Rust to know what is the underlying problem (this is my first pull request in Rust), but my guess is we should do one of the following :
- Comment the impl_partial that override the default implementation (which I did because it is easier, I've tried to explain why inside a comment because I knew it would not be obvious).
- Create an issue to fix `impl_partial` properly.
If you prefer the latter, I will add a TODO on this line.

core/lib/build.rs
dallaires: Good catch!

core/lib/src/fairing/ad_hoc.rs
dallaires: I've added this allow :
`#[allow(clippy::type_complexity)] // Private enum, the types will be simplified if needed.`

core/lib/src/form/error.rs
dallaires: I've added the following allow :
`#[allow(clippy::match_ref_pats)] // It's more readable to repeat the reference.`
As I said earlier, it will be fine once we get use this way of writing code. So I would recommend respecting Clippy's style warnings because they are present to make the code more standard between projects.

core/lib/src/form/from_form_field.rs
dallaires: I've notice that you prefer the question mark way I've allowed the lint [try_err](https://rust-lang.github.io/rust-clippy/master/index.html#try_err) for each traits with the line :
`#[allow(clippy::try_err)] // Returning Err(...)? is more readable than Err(...into())`
Feel free to allow the lint for the whole project, but as always, I would recommend respecting Clippy's way to get a standard style between projects.

contrib/lib/src/databases.rs
jebrosen: The goal of this awkwardness is to run `async` and blocking code in the appropriate contexts; it doesn't actually need to be waited on. If `;` silences the lint that might even be better!

Adding this `.await` will likely keep one task alive a little longer than necessary on every `Drop`.

core/codegen/src/http_codegen.rs
SergioBenitez: The issue is not one of unfamiliarity with the syntax but rather with the application in this context. You must understand that clippy is an automated tool; aside from heuristics, it cannot determine whether a style choice will _always_ be the right one. In this case, we went from a positive conditional check with two simple clauses to a negative conditional with an object literal, a method call, and a borrow. The latter is more complex in just about every way and obscures the intended logic of this function.

core/codegen/src/attribute/param/mod.rs
SergioBenitez: It's simply a choice: do you pay the cost for boxing the value in exchange for cheaper copies of the box later, or do you eschew the cost of boxing but have more
expensive copies of the value later? In this case, we don't copy the structure around much, so I suspect we won't see a difference either way, and adding the Box here removes symmetry, so I consider it the worse option.

core/http/src/header/accept.rs
SergioBenitez: By "logically" I mean that they compute differently. The version you've written computes every component of the `||` on every loop whereas the previous version only computes up to the first that matches. This is fundamentally different.

There is no such thing as "normal". The code was written the original way for a reason: to more clearly explain what it is doing. There is nothing to fix. The version you've written makes it much more difficult, in my opinion, to see this, and has the side effect of being more expensive to compute in expectation.

core/lib/src/form/error.rs
SergioBenitez: I'm happily used to both versions but prefer the former. This one adds a deref of a type that is almost always behind a reference, which is odd. Let's stick to the previous version.

core/lib/src/form/from_form.rs
SergioBenitez: No, please use the existing version.

core/lib/src/form/from_form_field.rs
SergioBenitez: I believe my preference is: if it is the return expression of the function, `Err(e)?` is okay. Otherwise if it's an early return, prefer `return Err(e.into())`.

core/codegen/src/http_codegen.rs
dallaires: I do understand that Clippy is an automated tool. This is why I add a comment to explain why the lint is allowed for this case (and it will help other developers to make decisions when they get the same warning). I'll allow these cases because, even if it is not perfect, Clippy does help us to find mistakes when they are not drowned in inopportune warnings.

core/http/src/header/accept.rs
dallaires: Thank you Sergio for the explanation. I'll update the comment to better explain the initial intent:
```
    #[allow(clippy::if_same_then_else)] // For this particular case, we need to keep the if/else
                                        // branching for legibility of comments and to avoid
                                        // unnecessary computation.
```

core/lib/src/form/error.rs
dallaires: I'm ok with that. I think this comment will be more inline with the given explanation :
```
#[allow(clippy::match_ref_pats)] // Cow is almost always behind a reference, we'll keep it that way for consistency.
```

core/lib/src/form/from_form.rs
dallaires: I think I've did not give this lint the proper attention because there were lots of them to fix. I suggest two solution :
- Fixing the lint like this :
```
match (self.last_key, this_key) {
    (Some(k1), Some(k2)) if k1 == k2 => (),
    _ => {
        self.shift();
        self.current = Some(T::init(self.opts));
    }
};
```
- Allowing the lint and renaming `keys_match` to make the intent clearer :
```
#[allow(clippy::match_like_matches_macro)] // Using the macro will make the match more complex.
let is_new_key = match (self.last_key, this_key) {
    (Some(k1), Some(k2)) if k1 == k2 => false,
    _ => true
};
if is_new_key {
    self.shift();
    self.current = Some(T::init(self.opts));
}
```
Which one you prefer?

contrib/lib/src/databases.rs
SergioBenitez: I don't like the idea of having these `allow(..)` spread throughout the codebase. Can we centralize them somewhere? I also don't think there's a need to justify with comments.

core/lib/src/form/from_form_field.rs
dallaires: Thank you, I'll allow those cases with the comment explaining your preference :
```
#[allow(clippy::try_err)] // For return values, please use Err(...)? instead of .into().
```
I think it will help contributors to maintain that rule.

core/http/src/raw_str.rs
SergioBenitez: All of the existing code is correct. The lint appears to be wrong; we _do_ derive both `Ord` and `PartialOrd`; these implementations are for different types which we can't derive implementations for. What's more, they do the same thing. So, this one is a false positive.

core/lib/src/form/from_form.rs
SergioBenitez: I prefer the original version.

core/lib/src/form/from_form_field.rs
dallaires: Thank you, I'll allow those cases with the comment explaining your preference :
```
#[allow(clippy::try_err)] // For return values, use : "Err(...)?" For early return, use : "return Err(...into());".
```
I think writing that rule will help new contributors to respect it.

SergioBenitez:
> I was trying Rocket's master branch because I want to use the stable compiler. I was not able to run Clippy on my project because of errors in Rocket so I've fixed all errors and warnings (I don't like having warning leaching into my projects).

I presume this is only because you're depending on Rocket from GitHub; this shouldn't be the case otherwise, is that right?

contrib/lib/src/databases.rs
dallaires: Good call @jebrosen! The semicolon was missing, not the await.

Clippy's lint and their explanations are not always good, but they have the merit to make us think and to make us learn new stuff. Thanks for the explanation. :)

core/http/src/raw_str.rs
dallaires: Since the macro `impl_partial` is and will remain private, it is fine to allow this lint. So I'll do it with these lines :
```
#[allow(clippy::derive_ord_xor_partial_ord)] // The default implementation of Eq, Ord and Hash
#[allow(clippy::derive_hash_xor_eq)]         // works well with the custom ones.
```

core/lib/src/form/from_form.rs
dallaires: Ok thanks, I'll use the original code with this allow :
```
#[allow(clippy::match_like_matches_macro)] // In this context, match makes the intent clearer.
```

core/lib/src/form/validate.rs
dallaires: It is to fix the warning [len_without_is_empty](https://rust-lang.github.io/rust-clippy/master/index.html#len_without_is_empty). They say that, for some containers, `is_empty` is much faster because it is done in constant time instead of `len` which will be in linear time.

To be honest, I don't know if it is useful in this context because unlike C++'s STL, Rust API don't tell the complexity of accessing a container. Moreover I did not check if I could replace some `len` with `is_empty`.

For now, all the lints in this commit will be allowed and I'll fix them in another PR. I will check whether this function will be used or not. If it is useful, I'll replace `len() == 0` with `is_empty()`, otherwise I will allow this lint. If you prefer not implementing this function at all, please tell it to me : I don't want to waste my time on things that won't be merged.

contrib/lib/src/databases.rs
dallaires: Just to be clear, the goal of this PR is not to decide which `allow(...)` should be on what level. I just want Clippy's output to be clean so I can lint my project with Rocket source code in local (without telling Clippy not to lint dependencies). For the justifications, I've added them because they are very strict on this at my day job and I don't know how much comments you want. Since you're the project owner, I'll implement your preference.

To answer your question, there is a [Clippy issue](https://github.com/rust-lang/cargo/issues/5034) to add global configurations for lints. Since it is still open, the only way to centralize `alow(...)` will be to add them on top of the file. It is also possible to add the to mod.rs and lib.rs files and it will apply recursively to the submodules.

Based on your last comments, I will not justify `allow(...)` and I will add them based on this rule :
- if there is only one lint to allow in the file, I'll add it to the function;
- otherwise I'll add it at the top of the file.
Seeing how `allow(..)` are spreed will help you decide which lint should be allowed on what level. I don't think there will be that much `allow(...)` with this rule because they are mostly in the same files. I will only allow the lint [module_inception](https://rust-lang.github.io/rust-clippy/master/index.html#module_inception) in the crate's lib.rs because it is the reason why I've renamed files.

Does that way to centralize `allow(...)` suit you?

core/http/src/header/mod.rs
dallaires: All the module rename was to fix the warning [module_inception](https://rust-lang.github.io/rust-clippy/master/index.html#module_inception). I did it because all of their content are exported by the parent module so renaming them is a minor change. I would allow the lint if it had changed the public API or more than one file. I knew you might not like that change, this is why I've put it in a separate commit (I did forget to move some non-trivial fixes in this commit because there were lots of fixes).

The goal of this pull request is to suggest solutions to fix the lint instead of drowning the source code in `allow(...)`. To put you in context, one of my employers asked us to enable all GCC's warnings and to consider them as errors -- including the pedantic ones. All the code looked like this one  :
```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "{Some lint}"
 // ...
#pragma GCC diagnostic pop
```
This is why I did my best to offer solutions that don't need to add `allow(...)` to fix warnings.

Since you own this repository, I don't mind adding `allow(...)` for the solution you don't like. As I mention earlier, I'll only add one per ignored lint per file to keep the code clean (without the justifying comment). I'll let you decide which `allow(..)` should be applied on what level.

core/lib/src/form/options.rs
dallaires: Ok thanks, I'll remove the TODO.

I've added it because, since I didn't have context, I did not know if I had to remove `..opts` and rename the parameter `_opts` in this function :
```
fn init(opts: Options) -> Self::Context {
        T::init(Options { strict: true, ..opts })
    }
```
But the Clippy's page [needless_update](https://rust-lang.github.io/rust-clippy/master/index.html#needless_update) says It is ignored with `non_exhaustive` types because they are work in progress. Both solutions have implications at compile time. But I've felt like `Options` was work in progress in this case.

Somme warning like this one are a bit hard to fix without context. Moreover fixing hundreds of them in one PR is time consuming and can be very error prone. Which is why I've suggested in this PR's intro to ask people to fix their own warning.

dallaires:
> > I was trying Rocket's master branch because I want to use the stable compiler. I was not able to run Clippy on my project because of errors in Rocket so I've fixed all errors and warnings (I don't like having warning leaching into my projects).
> 
> I presume this is only because you're depending on Rocket from GitHub; this shouldn't be the case otherwise, is that right?

Your half right. If I've used GitHub repository, Clippy would have ignored Rocket (I think it ignores everything in `~/.cargo` directory). I've figured this out after spending two weeks fixing warnings... so I've decided to share the work I've done.

The real reason is because I've cloned the repository in local to try the examples. Since I did not know the tooling well enough, I've set the `Cargo.toml` using the relative path like the example you provided. But that's fine : I also wanted to learn how Rocket works. I was also thinking about lending you a hand to help you release 0.5.0. Then, it will depend on how much time I have.

contrib/lib/tests/templates.rs
```diff
@@ -33,11 +33,12 @@ mod templates_tests {
     }
 
     #[test]
+    #[allow(clippy::try_err)]
```
dallaires, line 36: There is [an issue](https://github.com/rust-lang/rust-clippy/issues/5757) to disable this lint for try blocks. Since the [try_err](https://rust-lang.github.io/rust-clippy/master/index.html#try_err) lint purpose is to avoid this style for early return, I think we can ask them to ignore function return for this lint.

What do you think?

core/lib/src/fairing/ad_hoc.rs
```diff
@@ -60,35 +60,23 @@ pub struct AdHoc {
...
-
-/// The request fairing's function type
-type FnRequest = dyn for<'a> Fn(&'a mut Request<'_>, &'a Data)
-    -> BoxFuture<'a, ()> + Send + Sync + 'static;
-
-/// The response fairing's function type
-type FnResponse = dyn for<'a> Fn(&'a Request<'_>, &'a mut Response<'_>)
-    -> BoxFuture<'a, ()> + Send + Sync + 'static;
-
+#[allow(clippy::type_complexity)]
```
dallaires, line 63: The [type_complexity](https://rust-lang.github.io/rust-clippy/master/index.html#type_complexity) is configurable inside [clippy.toml](https://github.com/rust-lang/rust-clippy#configuration). You could put a very high number in the configuration file if you want do disable it.

dallaires:
Since it is my first pull request on GitHub, I was wondering if it is better to rebase my branch before pushing new commits. I did not do it this time because I did not want to mess the comments history, but I think it would have been better to rebase the branch and fix conflicts.

SergioBenitez:
Yes, please rebase.

dallaires:
I've rebased the branch and fixed the few new warnings except the three new ones that need to change a function definition. I've also ran all tests before pushing the result.

SergioBenitez:
https://github.com/SergioBenitez/Rocket/pull/1671 was merged which handled the majority of these. Thank you for the effort!

PR #1578, opened 2021-03-15
Add 'RouteUri' struct and use instead of base/path/uri on 'Route'

### LINKED ISSUE #1575: Add Origin::as_str method
It would be great to be able to access the entire URI of a route in a non-allocating way.

My current use case is [while creating a suitable label for endpoints](https://github.com/sd2k/rocket_prometheus/blob/6ae16098189033531a69226b9248375528fa4f3d/src/lib.rs#L327) in the [rocket-prometheus](https://github.com/sd2k/rocket_prometheus) crate: currently I call `request.route().unwrap().uri.to_string()` in an `on_response` fairing, which returns something like `"/hello/<name>?<caps>"` (i.e. the URI of the route). It feels like this should be doable without requiring an allocation on every request though, since we already have the URI defined statically.

The `path` and `query` methods already exist to get those two respective elements, but the only way to get the full URI is using the `Display` impl.

I'd be happy to have a go at this and submit a PR but might need a bit of guidance; the `origin.rs` module seems to contain quite a few optimizations!

[Discussed on Matrix here.](https://matrix.to/#/!kDIcCXWSVfdahNCJWq:mozilla.org/$Rfy_1G9Tp979Pkt3LdISApkF-M1vYOtaFT71iWpzcQo?via=matrix.org&via=mozilla.org)

sd2k:
This PR adds a new `RouteUri` struct (based on the suggestions in #1575) which consolidates the various `base`, `path`, and 'complete' `uri` subsets of a route's URI, and also provides access to the 'complete' URI as a '&str' by storing the full URI as a `String` on the `RouteUri` struct.

Questions:

- do you agree with the changes to 'base', 'path' and 'uri'? This does change the API somewhat, requiring users to go through the `RouteUri` struct to get hold of these things. I can add back the old `base` and `path` methods on `Route` if desired.
- should `RouteUri` be exported as `rocket::RouteUri`? or just `rocket::routes::RouteUri`?
- I'm not sure about the naming of the 'uri' and 'origin' fields/methods - any ideas?

TODO:

- [ ] Add docstring to `RouteUri`
- [ ] Add better docstring to `RouteUri::new`

Closes #1575.

### DIFF
--- a/core/lib/benches/simple-routing.rs
+++ b/core/lib/benches/simple-routing.rs
@@ -95,7 +95,7 @@ fn bench_simple_routing(b: &mut Bencher) {
     // Hold all of the requests we're going to make during the benchmark.
     let mut requests = vec![];
     for route in client.rocket().routes() {
-        let request = client.req(route.method, route.uri.path().as_str());
+        let request = client.req(route.method, route.uri.path());
         requests.push(request);
     }
 
--- a/core/lib/src/error.rs
+++ b/core/lib/src/error.rs
@@ -212,30 +212,3 @@ impl Drop for Error {
         }
     }
 }
-
-use crate::http::uri;
-use crate::http::ext::IntoOwned;
-
-/// Error returned by [`Route::map_base()`] on invalid URIs.
-#[derive(Debug)]
-pub enum RouteUriError {
-    /// The route URI is not a valid URI.
-    Uri(uri::Error<'static>),
-    /// The base (mount point) contains dynamic segments.
-    DynamicBase,
-}
-
-impl<'a> From<uri::Error<'a>> for RouteUriError {
-    fn from(error: uri::Error<'a>) -> Self {
-        RouteUriError::Uri(error.into_owned())
-    }
-}
-
-impl fmt::Display for RouteUriError {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match self {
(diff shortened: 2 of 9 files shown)


core/lib/src/lib.rs
```diff
@@ -153,7 +153,7 @@ mod cookies;
 #[doc(inline)] pub use crate::data::Data;
 #[doc(inline)] pub use crate::config::Config;
 #[doc(inline)] pub use crate::catcher::Catcher;
-pub use crate::router::Route;
+pub use crate::router::{Route, RouteUri};
```
SergioBenitez, line 156: This is unfortunate. Perhaps it's time for a `router` module.

SergioBenitez (changes requested):
I see now that this is much more involved than I anticipated. Instead of leaving a length review, I went ahead and made changes to the code directly. Can you finish off the PR by fixing doctests and filling out any missing docs?

sd2k:
I've updated the docstrings and docs. Since the `router` module isn't public, `RouteUri` isn't accessible in the docs. Should it be publicly exported somewhere to enable this?

SergioBenitez:
Merged in 2463637d518f98cf97223e6afde90277af3fcb74, 97acd8d9a9845562e7cc26d259f9790da63a6cc2. I took care of exposing `RouteUri` and adding a docstring for it. Thank you!

Issue #982, opened 2019-04-29
Tracing Middleware (opentracing)


nicholasamorim:
Is there any opentracing middleware available for Rocket?

I know that middlewares have been added since #55 has been closed but before putting in some work create this middleware, I'd like to understand if there's any implementation available or `contrib` repository of sorts.

jebrosen:
Glancing at https://crates.io/search?q=opentracing I don't see any opentracing middleware for Rocket, and the Rust libraries for opentracing itself look stale.

If you or anyone else were to implement it in a resuable way, it might be accepted into `rocket_contrib`. But it will likely be easier for it get more traction as a separate crate first, similarly to `rocket_cors`.

SergioBenitez:
Exactly as @jebrosen states. Happy to accept a PR, but it's not something the project itself should take on. As such, closing out.

oren0e:
Did anyone do something? I'm in a desperate need of something that will attach requestIDs to rocket status responses etc.

somehowchris:
@oren0e I'm not an expert on that but since Sergio is off due to personal matters let me try to help out.

[Fairings](https://rocket.rs/v0.5-rc/guide/fairings/) are kind of middleware steps. You can mutate responses and look into requests. Tough I am not aware of any rewuest-ids stored in rocket.

oren0e:
Thanks. about the request ids, they will probably need to be created.
But I am not sure - will the middleware should replace the native logging sort of like in tracing-actix-web crate for actix-web?

somehowchris:
I never used actix, so what you mean with `replaces the native logging`?

oren0e:
I never used actix as well, but I know it's pretty similar to rocket so I figured that I might take a look at the library in order to get some ideas for how to implement a similar middleware for rocket.
From the crate, `middleware.rs` doc strings:
```rust
/// `TracingLogger` is a middleware to capture structured diagnostic when processing an HTTP request.
/// Check the crate-level documentation for an in-depth introduction.
///
/// `TracingLogger` is designed as a drop-in replacement of [`actix-web`]'s [`Logger`].
```

oren0e:
Bottom line of what I'm trying to say is that if someone from the rocket team is willing in helping me writing this fairing then I'll give it a try.

oren0e:
[...]

PR #1579, opened 2021-03-15
Use 'tracing' library instead of 'log'.

jebrosen:
Supersedes #1410.

I likely missed a few spots and will need to make a few minor changes in follow-on commits, but this basically "ready to review".

TODOs:

* [x] `examples/todo` should use `rocket::tracing`, not `log`
* [x] `trace_env_logger` example: `hi` should take and return an `&str`, not `String`. And every macro except `trace` is used via `rocket::`; should we just import them?
* [x] Some config values on launch are colored white and others aren't.
* [ ] Change the hash used to disambiguate requests in the log, and/or the way it is formatted. Span `Id`s generated by `Registry` are a multi-level bit-packed generation index, and I threw something together that seemed to look nice on 64-bit systems.
* [x] Lots of lines now exceed 100 characters, largely those that name the trace/span. We should avoid this.
* [ ] Remove `in_scope` calls
  * [x] especially the large ones.
* [ ] Benchmark compile and run time with and without tracing's [compile time filters](https://docs.rs/tracing/0.1.25/tracing/level_filters/index.html#compile-time-filters) to estimate overhead
* [ ] Integrate additions and changes made in `logger.rs` that have not been rebased into `trace.rs`
* [ ] Review every log message and make sure it's been updated if necessary
    * [ ] name spans consistently, to avoid confusion with message
    * [ ] Use span fields in more places. e.g.: config errors
* [ ] API organization, re-exports, and docs
  * [ ] Potentially re-export the entire `tracing` and `tracing-subscriber` crates, like is done for `futures`, `tokio`, and `figment` today.


core/lib/src/server.rs
```diff
@@ -94,14 +99,20 @@ async fn hyper_service_fn(
...
-    });
+            // Dispatch the request to get a response, then write that response out.
+            let token = rocket.preprocess_request(&mut req, &mut data).await;
+            let r = rocket.dispatch(token, &mut req, data).await;
+            rocket.send_response(r, tx).await;
+        }
+            .instrument(span)
+            .await
+
+    }.in_current_span());
```
SergioBenitez, line 115: Do we need both of these?

core/lib/src/server.rs
```diff
@@ -351,28 +360,30 @@ impl Rocket {
     //
     // On catcher failure, the 500 error catcher is attempted. If _that_ fails,
     // the (infallible) default 500 error cather is used.
-    pub(crate) async fn handle_error<'s, 'r: 's>(
+    pub(crate) fn handle_error<'s, 'r: 's>(
```
SergioBenitez, line 363: Why this change? It seems unnecessary? The code also got harder to read.

In general, these tail `instrument` calls are hard to follow.

core/lib/src/server.rs
```diff
@@ -396,7 +407,8 @@ impl Rocket {
         self.fairings.pretty_print_counts();
         self.fairings.handle_launch(&self);
 
-        launch_info!("{}{} {}{}",
+        trace::info!(target: "launch",
```
SergioBenitez, line 410: This one is `launch` but the config one is `rocket::support`. Is this intended? Why are these different?

core/lib/src/server.rs
```diff
@@ -418,14 +430,26 @@ impl Rocket {
...
                     hyper_service_fn(rocket.clone(), remote, req)
-                }))
+                });
+                Ok::<_, std::convert::Infallible>(service)
             }
         });
 
-        // NOTE: `hyper` uses `tokio::spawn()` as the default executor.
+        #[derive(Clone)]
+        struct InCurrentSpanExecutor;
```
SergioBenitez, line 441: Why is this here? Isn't this what `.instrument()` and `.in_current_span()` are for?

SergioBenitez:
I haven't read through `trace.rs` itself - I'll hold of on this until the rest of the code is good - but I've left some comments on the rest of the changes. Some higher-level comments:
  * The `rocket` library feature should be called `tracing` or similar.
  * Is the feature intended to be disabled? If so, it should be documented in
    `lib.rs`, be added as a tested feature in `scripts/test.sh`, and explicitly
    enabled in examples that make use of it. If not, why is it a feature at all?
    At the moment, codegen seem to depend on this being enabled, so either this
    should be fixed or it shouldn't be a feature.
  * If codegen depends on this, why are the --core tests passing?
  * What is the overhead this is adding? This is now in the critical path. If we
    could really disable the feature, resulting in trace macro calls compiling
    to nothing, we should be able to measure this accurately.
  * `examples/todo` should use `rocket/tracing`, not `log`.
  * The `trace_env_logger` example needs some motivation. Is it just there to
    check that `log` integration "works"? In any case, `hi` should take and
    return an `&str`, not `String`. What's more, every macro except `trace` is
    used via `rocket::`; should we just import them?
  * Some config values on launch are colored white and others aren't.
  * In some places, tracing fields come before the message and sometimes after.
    We should be consistent. `trace!("msg", field = foo);`
  * Sometimes span fields are used, sometimes they aren't. For example, when
    logging config errors, why aren't all of those `{}` fields?
  * Why add a name to some spans and not others? What is the utility? Can we
    just use `module_name!()` or something to that effect?
  * Lots of lines now exceed 100 characters, largely those that name the
    trace/span. We should avoid this.
  * The `config` changes are particularly difficult to read.
  * Running arbitrary, non-trace related code `in_scope()` feels quite wrong to
    me. There must be a better way.
  * It seems like we really want a second macro, to avoid all of the `.in_scope(|| { })`:

    ```rust
    // the `, $expr` should be optional.
    error_span!("Some message", {
        info!("inside the span's scope");
    });
    ```

SergioBenitez (changes requested):
I haven't read through `trace.rs` itself - I'll hold of on this until the rest of the code is good - but I've left some comments on the rest of the changes. Some higher-level comments:
  * The `rocket` library feature should be called `tracing` or similar.
  * Is the feature intended to be disabled? If so, it should be documented in
    `lib.rs`, be added as a tested feature in `scripts/test.sh`, and explicitly
    enabled in examples that make use of it. If not, why is it a feature at all?
    At the moment, codegen seem to depend on this being enabled, so either this
    should be fixed or it shouldn't be a feature.
  * If codegen depends on this, why are the --core tests passing?
  * What is the overhead this is adding? This is now in the critical path. If we
    could really disable the feature, resulting in trace macro calls compiling
    to nothing, we should be able to measure this accurately.
  * `examples/todo` should use `rocket/tracing`, not `log`.
  * The `trace_env_logger` example needs some motivation. Is it just there to
    check that `log` integration "works"? In any case, `hi` should take and
    return an `&str`, not `String`. What's more, every macro except `trace` is
    used via `rocket::`; should we just import them?
  * Some config values on launch are colored white and others aren't.
  * In some places, tracing fields come before the message and sometimes after.
    We should be consistent. `trace!("msg", field = foo);`
  * Sometimes span fields are used, sometimes they aren't. For example, when
    logging config errors, why aren't all of those `{}` fields?
  * Why add a name to some spans and not others? What is the utility? Can we
    just use `module_name!()` or something to that effect?
  * Lots of lines now exceed 100 characters, largely those that name the
    trace/span. We should avoid this.
  * The `config` changes are particularly difficult to read.
  * Running arbitrary, non-trace related code `in_scope()` feels quite wrong to
    me. There must be a better way.
  * It seems like we really want a second macro, to avoid all of the `.in_scope(|| { })`:

    ```rust
    // the `, $expr` should be optional.
    error_span!("Some message", {
        info!("inside the span's scope");
    });
    ```

jebrosen:
I added most of those comments to the top-level comment as additional TODOs, and will keep working through them along with the other inline comments.


>     * The `trace_env_logger` example needs some motivation. Is it just there to
>       check that `log` integration "works"?

I believe it was, yes. That, and/or as a demonstration to users how Rocket's and users's `tracing` messages look as viewed by a `log`-based logger. @hawkw ?

>     * In some places, tracing fields come before the message and sometimes after.
>       We should be consistent. `trace!("msg", field = foo);`

The order is not flexible in that way: `info!(field = value ..., "format", arguments)` or `info_span!("span_name", field = value ..., "format", arguments)`. I think this is a actually a consequence of:

>     * Why add a name to some spans and not others? What is the utility? Can we
>       just use `module_name!()` or something to that effect?

*Every* span has a name. I mixed several of these up very early on, which now look like they have only a message, and never cleaned them all up.

>     * Running arbitrary, non-trace related code `in_scope()` feels quite wrong to
>       me. There must be a better way.

Earlier versions of this PR did use the [`Span::enter`](https://docs.rs/tracing/0.1.25/tracing/span/struct.Span.html#method.enter) API in more places, which returns a guard that exits the span when `Drop`ped. However, the API depends on having some sort of task-local or thread-local state, which hasn't been nicely solved for `async/await` and is easy to misuse accidentally (see caveats in documentation). `in_scope()` is much more difficult to use in an incorrect way.

I do agree that some of the `in_scope` calls would be better as `enter()`, so that's also on the top-level todo list now.

contrib/lib/src/lib.rs
```diff
@@ -40,7 +40,7 @@
 //! This crate is expected to grow with time, bringing in outside crates to be
 //! officially supported by Rocket.
 
-#[allow(unused_imports)] #[macro_use] extern crate log;
+#[allow(unused_imports)] #[macro_use] extern crate tracing;
```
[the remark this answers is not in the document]
jebrosen: What do you suggest changing here? I assume `#[macro_use]` was used to avoid needing to add `use` statements to each file, and I'm not aware of another way to do this (aside from making all of the different logging macros top-level exports from `rocket`, and `#[macro_use] extern crate rocket;`)

core/lib/src/fairing/fairings.rs
```diff
@@ -103,15 +103,16 @@ impl Fairings {
 
             let num = names.len();
             let joined = names.join(", ");
-            info_!("{} {}: {}", Paint::default(num).bold(), prefix, Paint::default(joined).bold());
+            info!("{} {}: {}", Paint::default(num).bold(), prefix, Paint::default(&joined).bold());
```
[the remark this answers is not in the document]
jebrosen: It appears that `info!` expands the expression twice (once to emit a `log` event; once to emit a `tracing` event) - the call to `Paint::default()` moves the value, so it can only be done twice via a reference.

core/lib/src/rocket.rs
```diff
@@ -349,7 +343,7 @@ impl Rocket {
             let mut fairings = std::mem::replace(&mut self.fairings, Fairings::new());
             let rocket = fairings.attach(fairing, self).await;
             (rocket, fairings)
-        };
+        }.in_current_span();
```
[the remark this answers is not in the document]
jebrosen: Without this, the state that tracks the "current" chain of spans won't be reset when tasks are switched between, and log messages may sometimes appear to be e.g. nested under the wrong request.

core/lib/src/rocket.rs
```diff
@@ -542,13 +536,14 @@ impl Rocket {
                 self.config.secret_key = crate::config::SecretKey::generate()
                     .unwrap_or(crate::config::SecretKey::zero());
 
-                warn!("secrets enabled without a stable `secret_key`");
-                info_!("disable `secrets` feature or configure a `secret_key`");
-                info_!("this becomes an {} in non-debug profiles", Paint::red("error"));
+                warn_span!("unconfigured_secret_key", "secrets enabled without a stable `secret_key`").in_scope(|| {
```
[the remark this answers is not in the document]
jebrosen: I don't think I agree on this particular usage. The more distance, `if` statements, etc. between calling `enter()` and the end of the scope, the harder it is to tell which messages are still "indented"; that's why I have (so far) used `in_scope` here. There is also the alternative of `info!(parent: span, ..)`, although it is more repetitive than the other options.

core/lib/src/server.rs
jebrosen: The `instrument` call is what set ups the `Request:` line in the log (and indents everything under it).

The `in_current_span` is there for the same reason as the other call to it.

core/lib/src/server.rs
jebrosen: This is added so that hyper's calls to `tokio::spawn` will "inherit" the active spans (e.g. the current connection, request, etc.), instead of being logged at the root level.

hawkw:
Answering some of the questions from above:

@jebrosen 
> I believe it was, yes. That, and/or as a demonstration to users how Rocket's and users's `tracing` messages look as viewed by a `log`-based logger. @hawkw ?

Yeah, that's correct; the example was supposed to demonstrate how to use Rocket with the `log` ecosystem. I thought this was valuable to provide for users who might already have a preferred `log` logger that they want to continue to use. It's probably not hugely important to provide an example of this, but I thought it's better to have more examples when possible?

@SergioBenitez 
>   * Running arbitrary, non-trace related code `in_scope()` feels quite wrong to
>     me. There must be a better way.

Running "arbitrary, non-trace related code" in `in_scope` is actually an intended use of the API; see [the documentation](https://docs.rs/tracing/0.1.25/tracing/span/index.html#entering-a-span) for details. In general, correct use of `tracing` places any code that's part of the logical unit of work or context represented by a span _inside_ that span --- this allows spans to be used not only for providing context to log messages, but also for timing code sections or for distributed tracing. 

@jebrosen 
> >
> > * In some places, tracing fields come before the message and sometimes after.
> >   We should be consistent. `trace!("msg", field = foo);`
> >
> 
> The order is not flexible in that way: `info!(field = value ..., "format", arguments)` or `info_span!("span_name", field = value ..., "format", arguments)`. 

Yeah, this is correct.  Format-string-based messages must always come last in the `tracing` macros, as the macros must parse them as any number of arbitrary token trees in order to pass them to `format_args!`. Spans always have names, which are string literals that come first in the macro invocation, and may optionally have format string messages at the *end* of the macro invocation.

@SergioBenitez 
> * What is the overhead this is adding? This is now in the critical path. If we
>   could really disable the feature, resulting in trace macro calls compiling
>   to nothing, we should be able to measure this accurately.

See [here](https://docs.rs/tracing/0.1.25/tracing/level_filters/index.html#compile-time-filters) for details on statically disabling specific `tracing` levels, or _all_ traces, at compile-time. :)

Since the compile-time filtering is based on feature flags, which are always additive, a library dependency like Rocket should generally not enable them, since this prevents the user application from being able to configure them. However, we might want to document the existence of these feature flags in Rocket's docs so that users are aware of them, and they can be used for benchmarking.

core/lib/src/server.rs
hawkw: This could alternatively remain an `async fn` and use the `#[instrument]` attribute, like:
```rust
#[tracing::instrument(
    level = "warn",
    skip(self, status, req),
    fields(message = %format_args!("Responding with {} catcher.", Paint::red(&status)),
)]
pub(crate) async fn handle_error<'s, 'r: 's>(
    // ...
```

hawkw:
> * ~The `rocket` library feature should be called `tracing` or similar.~ I see now that the `log` feature is for enabling compatibility with the `log` crate. It does not appear that we can disable tracing at all. What kind of compile-time overhead are we adding? What kind of runtime overhead are we adding?
> 
> * ~Is the feature intended to be disabled? If so,~ it should be documented in
>   `lib.rs`, ~be added as a tested feature in `scripts/test.sh`, and explicitly
>   enabled in examples that make use of it. If not, why is it a feature at all?
>   At the moment, codegen seem to depend on this being enabled, so either this
>   should be fixed or it shouldn't be a feature.~

It would also definitely be possible to make the `tracing` dependency itself optional. To do this, we would need to wrap the macros with Rocket's own macros that expand to nothing when the feature flag is disabled. I didn't do this since `tracing`'s compile-time filtering is sufficient to avoid runtime overhead, and it was a lot of extra code. However, this *would* allow us to avoid downloading and compiling the `tracing` dependency when it's not in use.

oren0e:
Hi, any news about when should we expect the move to `tracing` to be completed? I was just trying to do something related in https://github.com/SergioBenitez/Rocket/issues/982, i.e., insert a request_id to monitor the full cycle of a request-response. Currently I have to setup a custom subscriber and formatting layer like:
```rust
    LogTracer::init().expect("Unable to setup log tracer!");

    let (non_blocking_writer, _guard) = tracing_appender::non_blocking(std::io::stdout());
    let formatting_layer = fmt::layer().json().with_writer(non_blocking_writer);
    let subscriber = Registry::default()
        .with(EnvFilter::new(env!("LOG_LEVEL")))
        .with(formatting_layer);
    tracing::subscriber::set_global_default(subscriber).expect("Failed to set global subscriber");

    rocket_builder().launch();
```
Which is not ideal because it interferes with rocket's logging configuration and then I end up getting ugly jsons with terminal color codes in raw form, which look like the following:
```json
{"timestamp":"2022-01-06T10:23:04.323575Z","level":"INFO","fields":{"message":"\u001b[1;49;39mOutcome:\u001b[0m \u001b[49;32mSuccess\u001b[0m","log.target":"_","log.module_path":"rocket::rocket","log.file":"/Users/me/.cargo/registry/src/github.com-1ecc6299db9ec823/rocket-0.4.10/src/rocket.rs","log.line":303},"target":"_"}
```
And notice that I don't have a request_id here as well for some reason...

SohumB:
Hi there,

I notice this PR has stalled. Is there anything a third party (like me) can do to help push it along? I'm eagerly looking forward to using `tracing` in Rocket and would absolutely be willing to put work into making it happen.

Thanks!

SergioBenitez:
I'm closing this as Jeb has indicated that he won't be able to push it forward. This is something I will personally prioritize for Rocket 0.6.

SergioBenitez:
By the way: if anyone is interested in making this happen, I would love to actively mentor that person.

oren0e:
I might be interested, let me check availability over the weekend and I'll get back to you with an answer.

oren0e:
ah what is the Matrix channel? You have a discord?

SergioBenitez:
It's in the README and the website. Here's a direct link: https://chat.mozilla.org/#/room/#rocket:mozilla.org

wrenix:
any news @oren0e or @SergioBenitez  or newer state of work?

maybe i would like to take a look

oren0e:
You're welcome to work on this, I will try to help with this as much as I can but I think I can't be the owner if this migration currently. I initially thought that I will have much more time to devote to this task where in reality I have something like only 10% of my initial estimation. I'm sorry that I inform of this only now.

contrib/lib/src/databases/connection.rs
```diff
@@ -62,8 +63,8 @@ async fn run_blocking<F, R>(job: F) -> R
 
 macro_rules! dberr {
     ($msg:literal, $db_name:expr, $efmt:literal, $error:expr, $rocket:expr) => ({
-        rocket::error!(concat!("database ", $msg, " error for pool named `{}`"), $db_name);
-        error_!($efmt, $error);
+        error!(concat!("database ", $msg, " error for pool named `{}`"), $db_name);
```
wrenix, line 66: please just one log per message / error + please keep logmessage clean from values (it is easier to indexing, search and anonymize in log systems, if all values are in defined files):
```suggestion
        error!(
            database.name = $msg,
            database.pool = $db_name,
            error = $error,
            $efmt
        );
```

artob:
Worth mentioning here that this did happen: https://github.com/rwf2/Rocket/issues/21#issuecomment-2095020409

PR #1582, opened 2021-03-17
Make push_error() and push_errors() publicly accessible on a form context

fstephany:
Switch the `pub(crate)` visibility to `pub`.  
It can be useful when performing additional validations to add errors into the form context (e.g., validation against a database or remote state).


SergioBenitez:
These need docstrings, examples, and tests.

PR #1589, opened 2021-03-23
use tera include in example

blackghost1987:
This is my first PR in the project, so sorry for any issues in advance.
I've made a small improvement to the Tera template example: I've added an "include" statement, because I had some issues with figuring out the correct path to use in my own project, and I wanted to save others the hassle by adding it to the example. The official Tera website uses examples like `{% include "items.html" %}` but for Rocket it should be simply `{% include "items" %}`, even though the template file is named `items.html.tera`.
I've also added some extra tests to check if the context is properly passed between template and then displayed correctly. I know the "expected" value assert already checks the whole rendered HTML, but I presumed it's better to explicitly check if the context values are present. Let me know if you think it's too superfluous and I'll remove it.

### DIFF
--- a/examples/templating/templates/tera/base.html.tera
+++ b/examples/templating/templates/tera/base.html.tera
@@ -5,7 +5,7 @@
     <title>Tera Demo - {{ title }}</title>
   </head>
   <body>
-    <a href="/tera/hello/Unknown">Hello</a> | <a href="/tera/about">About</a>
+    {% include "tera/nav" %}
 
     {% block content %}{% endblock content %}
 
--- a/examples/templating/templates/tera/nav.html.tera
+++ b/examples/templating/templates/tera/nav.html.tera
@@ -1,0 +1,1 @@
+<a href="/tera/hello/Unknown">Hello</a> | <a href="/tera/about">About</a>


SergioBenitez:
Instead of including this `items`, maybe we can move the navbar into `nav.html` and include that instead in `base.html`. This is what the handlebars variant does for illustration purposes, so it would be nice for parity.

examples/tera_templates/src/tests.rs
```diff
@@ -49,7 +49,13 @@ fn test_name() {
...
         let expected = Template::show(client.rocket(), "index", &context).unwrap();
         assert_eq!(response.status(), Status::Ok);
-        assert_eq!(response.into_string(), Some(expected));
+
+        let response_str = response.into_string().unwrap_or_default();
+        assert_eq!(response_str, expected);
+        assert!(response_str.contains("Jack"));
+        assert!(response_str.contains("One"));
+        assert!(response_str.contains("Two"));
+        assert!(response_str.contains("Three"));
```
SergioBenitez, line 58: This is equivalent to what we had before. Please revert.

examples/tera_templates/templates/index.html.tera
```diff
@@ -2,12 +2,7 @@
 
 {% block content %}
     <h1>Hi {{name}}</h1>
-    <h3>Here are your items:</h3>
-    <ul>
-        {% for s in items %}
-            <li>{{ s }}</li>
-        {% endfor %}
-    </ul>
+    {% include "items" %}
```
SergioBenitez, line 5: See my comment about using a `nav.html` as an include instead.

SergioBenitez (changes requested):

examples/tera_templates/src/tests.rs
blackghost1987: It's actually a bit more than before. The original code tests if the response contents are the same as the rendered template, but it doesn't verify that the values from the context are displayed in that rendering. You could comment out the whole include line from the template and the tests would still pass.
So I still think it's a useful test, but I agree, it's a bit weird here. The tera repository probably would be better place for a test like this, because it's not really testing any Rocket features. I'll revert it.

examples/tera_templates/templates/index.html.tera
blackghost1987: Should I add an "about" page similar to the handlebars example, so the nav bar makes some sense? I don't really see a use for a nav bar with only one route.

SergioBenitez: Sure! Let's have full parity, why not.

blackghost1987: Sorry I haven't realized there were other changes to the templating examples on master, an about page and a nav template already exists in the latest version, your comment makes much more sense know

blackghost1987:
This became much more simple when I merged the other changes from master (I was not aware of those). So this is finished, ready for review!

examples/tera_templates/src/tests.rs
SergioBenitez: You're right, there's nothing actually testing that we have those items in there. It's a bit of a weird case because, in reality, these items wouldn't be static, and you'd want to check that the rendered template contains the dynamic items. So it's a bit weird to test them in this case. If we extend this in the future to make the items dynamic, then I think it would make sense to test this. Thanks for your comment.

SergioBenitez:
Sweet, thank you!

SergioBenitez (approved):

SergioBenitez:
Merged in https://github.com/SergioBenitez/Rocket/commit/469f3942eb6567c2fdad7308350ee23dade030e9.

PR #1595, opened 2021-03-27
Allow configuring a default value for form fields

ThouCheese:
I have implemented setting a default value when deriving `FromForm`, and updated the docs accordingly.

### DIFF
--- a/core/codegen/src/derive/form_field.rs
+++ b/core/codegen/src/derive/form_field.rs
@@ -13,7 +13,7 @@ pub enum FieldName {
     Uncased(Name),
 }
 
-#[derive(Debug, FromMeta)]
+#[derive(FromMeta)]
 pub struct FieldAttr {
     pub name: Option<FieldName>,
     pub validate: Option<SpanWrapped<syn::Expr>>,
@@ -326,35 +326,40 @@ pub fn validators<'v>(
     Ok(exprs)
 }
 
-pub fn default<'v>(
-    field: Field<'v>,
-) -> Result<Option<syn::Expr>> {
+pub fn default<'v>(field: Field<'v>) -> Result<Option<syn::Expr>> {
     let mut exprs = FieldAttr::from_attrs(FieldAttr::NAME, &field.attrs)?
         .into_iter()
-        .map(|a| a.default)
-        .flat_map(move |expr| {
-            let field_span = field.ident().span()
-                .join(field.ty.span())
-                .unwrap_or(field.ty.span());
-
-            expr.map(|expr| {
-                let span = expr.key_span.unwrap_or(field_span);
-                define_spanned_export!(span => _form);
-                syn::parse2(quote_spanned!(span => {
-                    let __default: Option<_> = { #expr }.into();
-                    __default
-                })).unwrap()
-            })
+        .filter_map(|a| a.default)
+        .map(move |expr| {
+            use syn::{Expr, Lit, ExprLit};
+            // As a result of calling `#expr.into()`, type inference fails for
+            // two common expressions: integer literals and the bare `None`. As
+            // a result, we cheat: if the syntax matches either of these two
+            // conditions, we provide the field type as a hint.
+            let is_int_lit = matches!(*expr, Expr::Lit(ExprLit { lit: Lit::Int(_), .. }));
+            let is_none = matches!(*expr, Expr::Path(ref e) if e.path.is_ident("None"));
+            let ty = field.stripped_ty();
+            let ty_hint = (is_int_lit || is_none)
+                .then(|| quote!(#ty))
+                .unwrap_or_else(|| quote!(_));
+            let opt_expr = if is_none {
+                quote_spanned!(expr.span => None)
+            } else if is_int_lit {
+                quote_spanned!(expr.span => Some(#expr))
+            } else {
+                quote_spanned!(expr.span => Some({ #expr }.into()))
+            };
+            syn::parse2(quote_spanned!(expr.span => {
+                let __default: Option<#ty_hint> = #opt_expr;
+                __default
+            })).unwrap()
         });
-    let first = exprs.next();
-    if first.is_none() {
-        return Ok(None);
-    }
-    let second: Option<syn::Expr> = exprs.next();
-    if let Some(expr) = second {
+
+    let first: Option<syn::Expr> = exprs.next();
+    if let Some(expr) = exprs.next() {
         return Err(expr.span()
-            .error("unexpected second default")
-            .note("only one default is allowed in this context"));
+            .error("duplicate `default` form field attribute")
+            .help("form fields can have at most one `default`"));
     }
 
     Ok(first)
--- a/core/codegen/src/derive/form_field.rs
+++ b/core/codegen/src/derive/form_field.rs
@@ -326,11 +326,10 @@ pub fn validators<'v>(
     Ok(exprs)
 }
 
-pub fn defaults<'v>(
+pub fn default<'v>(
     field: Field<'v>,
-) -> Result<syn::Expr> {
-    let ty = &field.inner.ty;
-    let expr = FieldAttr::from_attrs(FieldAttr::NAME, &field.attrs)?
+) -> Result<Option<syn::Expr>> {
+    let mut exprs = FieldAttr::from_attrs(FieldAttr::NAME, &field.attrs)?
         .into_iter()
         .map(|a| a.default)
         .flat_map(move |expr| {
@@ -342,22 +341,23 @@ pub fn defaults<'v>(
                 let span = expr.key_span.unwrap_or(field_span);
                 define_spanned_export!(span => _form);
                 syn::parse2(quote_spanned!(span => {
-                    {
-                        let __default: Option<#ty> = Some({ #expr }.into());
-                        __default
-                    }
+                    let __default: Option<_> = { #expr }.into();
+                    __default
                 })).unwrap()
             })
-            
-        })
-        .next()
-        .unwrap_or_else(|| {
-            syn::parse2(quote!({
-                { None }
-            })).unwrap()
         });
+    let first = exprs.next();
+    if first.is_none() {
+        return Ok(None);
+    }
+    let second: Option<syn::Expr> = exprs.next();
+    if let Some(expr) = second {
+        return Err(expr.span()
+            .error("unexpected second default")
+            .note("only one default is allowed in this context"));
+    }
 
-    Ok(expr)
+    Ok(first)
 }
 
 pub fn first_duplicate<K: Spanned, V: PartialEq + Spanned>(
--- a/core/codegen/src/derive/form_field.rs
+++ b/core/codegen/src/derive/form_field.rs
@@ -13,10 +13,11 @@ pub enum FieldName {
     Uncased(Name),
 }
 
-#[derive(FromMeta)]
+#[derive(Debug, FromMeta)]
 pub struct FieldAttr {
     pub name: Option<FieldName>,
     pub validate: Option<SpanWrapped<syn::Expr>>,
+    pub default: Option<SpanWrapped<syn::Expr>>,
 }
 
 impl FieldAttr {
@@ -322,7 +323,41 @@ pub fn validators<'v>(
             })).unwrap()
         });
 
-        Ok(exprs)
+    Ok(exprs)
+}
+
+pub fn defaults<'v>(
+    field: Field<'v>,
+) -> Result<syn::Expr> {
+    let ty = &field.inner.ty;
+    let expr = FieldAttr::from_attrs(FieldAttr::NAME, &field.attrs)?
+        .into_iter()
+        .map(|a| a.default)
+        .flat_map(move |expr| {
+            let field_span = field.ident().span()
+                .join(field.ty.span())
+                .unwrap_or(field.ty.span());
+
+            expr.map(|expr| {
+                let span = expr.key_span.unwrap_or(field_span);
+                define_spanned_export!(span => _form);
+                syn::parse2(quote_spanned!(span => {
+                    {
+                        let __default: Option<#ty> = Some({ #expr }.into());
+                        __default
+                    }
+                })).unwrap()
+            })
+            
+        })
+        .next()
+        .unwrap_or_else(|| {
+            syn::parse2(quote!({
+                { None }
+            })).unwrap()
+        });
+
+    Ok(expr)
 }
 
 pub fn first_duplicate<K: Spanned, V: PartialEq + Spanned>(
--- a/core/codegen/src/derive/from_form.rs
+++ b/core/codegen/src/derive/from_form.rs
@@ -231,26 +231,22 @@ pub fn derive_from_form(input: proc_macro::TokenStream) -> TokenStream {
                 let (ident, ty, name_view) = (f.ident(), f.stripped_ty(), f.name_view()?);
                 let validator = validators(f, &ident, true)?;
                 let user_default = default(f)?;
-                let fallback = match user_default {
-                    Some(expr) => quote_spanned! { ty.span() =>
-                        #expr
-                            .or_else(|| <#ty as #_form::FromForm<'__f>>::default(_opts))
+                let default = user_default
+                    .map(|expr| quote_spanned!(ty.span() =>
+                        #expr.or_else(|| <#ty as #_form::FromForm<'__f>>::default(_opts))
                             .ok_or_else(|| #_form::ErrorKind::Missing.into())
-                    },
-                    None => quote_spanned! { ty.span() =>
+                    ))
+                    .unwrap_or_else(|| quote_spanned!(ty.span() =>
                         <#ty as #_form::FromForm<'__f>>::default(_opts)
                             .ok_or_else(|| #_form::ErrorKind::Missing.into())
-                    },
-                };
+                    ));
                 let _err = _Err;
                 Ok(quote_spanned! { ty.span() => {
                     let _name = #name_view;
                     let _opts = __c.__opts;
                     __c.#ident
                         .map(<#ty as #_form::FromForm<'__f>>::finalize)
-                        .unwrap_or_else(|| {
-                            #fallback
-                        })
+                        .unwrap_or_else(|| #default)
                         .and_then(|#ident| {
                             let mut _es = #_form::Errors::new();
                             #(if let #_err(_e) = #validator { _es.extend(_e); })*
--- a/core/codegen/src/derive/from_form.rs
+++ b/core/codegen/src/derive/from_form.rs
@@ -230,6 +230,7 @@ pub fn derive_from_form(input: proc_macro::TokenStream) -> TokenStream {
             .try_field_map(|_, f| {
                 let (ident, ty, name_view) = (f.ident(), f.stripped_ty(), f.name_view()?);
                 let validator = validators(f, &ident, true)?;
+                let default = defaults(f)?;
                 let _err = _Err;
                 Ok(quote_spanned! { ty.span() => {
                     let _name = #name_view;
@@ -237,7 +238,8 @@ pub fn derive_from_form(input: proc_macro::TokenStream) -> TokenStream {
                     __c.#ident
                         .map(<#ty as #_form::FromForm<'__f>>::finalize)
                         .unwrap_or_else(|| {
-                            <#ty as #_form::FromForm<'__f>>::default(_opts)
+                            #default
+                                .or_else(|| <#ty as #_form::FromForm<'__f>>::default(_opts))
                                 .ok_or_else(|| #_form::ErrorKind::Missing.into())
                         })
                         .and_then(|#ident| {
--- a/core/codegen/src/derive/from_form.rs
+++ b/core/codegen/src/derive/from_form.rs
@@ -230,7 +230,18 @@ pub fn derive_from_form(input: proc_macro::TokenStream) -> TokenStream {
             .try_field_map(|_, f| {
                 let (ident, ty, name_view) = (f.ident(), f.stripped_ty(), f.name_view()?);
                 let validator = validators(f, &ident, true)?;
-                let default = defaults(f)?;
+                let user_default = default(f)?;
+                let fallback = match user_default {
+                    Some(expr) => quote_spanned! { ty.span() =>
+                        #expr
+                            .or_else(|| <#ty as #_form::FromForm<'__f>>::default(_opts))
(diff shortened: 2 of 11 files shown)


core/codegen/src/derive/form_field.rs
```diff
@@ -291,5 +292,39 @@ pub fn validators<'v>(
             })).unwrap()
         });
 
-        Ok(exprs)
+    Ok(exprs)
+}
+
+pub fn defaults<'v>(
+    field: Field<'v>,
+) -> Result<syn::Expr> {
```
SergioBenitez, line 300: This should be called `default`; there can only be one default. It should error if there is more than one default specified.

core/codegen/src/derive/form_field.rs
```diff
@@ -291,5 +292,39 @@ pub fn validators<'v>(
...
+                .unwrap_or(field.ty.span());
+
+            expr.map(|expr| {
+                let span = expr.key_span.unwrap_or(field_span);
+                define_spanned_export!(span => _form);
+                syn::parse2(quote_spanned!(span => {
+                    {
+                        let __default: Option<#ty> = Some({ #expr }.into());
+                        __default
+                    }
```
SergioBenitez, line 317: Why the double wrapping in `{ }`?

core/codegen/src/derive/from_form.rs
```diff
@@ -244,14 +244,16 @@ pub fn derive_from_form(input: proc_macro::TokenStream) -> TokenStream {
             .try_field_map(|_, f| {
                 let (ident, ty, name_view) = (f.ident(), f.stripped_ty(), f.name_view()?);
                 let validator = validators(f, &ident, true)?;
+                let default = defaults(f)?;
```
SergioBenitez, line 247: Let's call this `user_default` or something to that effect, to disambiguate from the form's default.

core/codegen/src/derive/form_field.rs
```diff
@@ -291,5 +292,39 @@ pub fn validators<'v>(
...
+                })).unwrap()
+            })
+            
+        })
+        .next()
+        .unwrap_or_else(|| {
+            syn::parse2(quote!({
+                { None }
+            })).unwrap()
+        });
```
SergioBenitez, line 327: Can't we do this in the `flat_map`?

core/codegen/src/derive/form_field.rs
SergioBenitez, line 300: I think it would be nicer/cleaner if this returned a `Result<Option<syn::Expr>>` which is `Err` when the syntax is wrong, `Ok(Some)` when there's a default, and `Ok(None)` when there is no default. Then we can do the `or_else()` at compile-time, using the Rocket-default when there is no default specified.

This should also let you get away from the `flat_map().next()` complications, closer to how `one_field_name` is written.

SergioBenitez (changes requested):
There are a few issues:

  * This seems to accept more than one `default` per field, but that doesn't make sense. We should error out if more than one default is provided. See how `Field::one_field_name()` errors.
  * There are no UI tests for codegen errors. See `core/codegen/tests/ui-fail`. The tests go in `ui-fail` and are symlinked into the nightly and stable directories.
  * There are no tests for this in `core/codegen/tests`. We should check that the default actually works, and that Rocket's default still get used if the user doesn't specify one.
  * The EBNF of the `FromForm` derive needs to be updated.
  * The `FromForm` derive docs need to be updated.
  * The form sections of the guide needs to be updated.
  * See the CI errors. Try running `./scripts/test.sh` locally. If you have enough memory, run `./scripts/test.sh --all`.

core/codegen/src/derive/form_field.rs
ThouCheese: If this returns an `Option`, then I need to check whether a default was provided in `from_form.rs`, and emit different code when the default is there. I'll change it around a bit and we can see which approach we like better :+1:

ThouCheese:
I have update the PR, taking into account all your comments! Let me know what is right or wrong right now and I'll get to it a bit faster than last time :)

core/codegen/src/derive/form_field.rs
```diff
@@ -13,10 +13,11 @@ pub enum FieldName {
     Uncased(Name),
 }
 
-#[derive(FromMeta)]
+#[derive(Debug, FromMeta)]
```
SergioBenitez, line 16:
```suggestion
#[derive(FromMeta)]
```

core/codegen/src/derive/form_field.rs
```diff
@@ -291,5 +292,39 @@ pub fn validators<'v>(
...
+    }
+    let second: Option<syn::Expr> = exprs.next();
+    if let Some(expr) = second {
+        return Err(expr.span()
+            .error("unexpected second default")
+            .note("only one default is allowed in this context"));
+    }
+
+    Ok(first)
 }
```
SergioBenitez, line 330:
```suggestion
pub fn default<'v>(
    field: Field<'v>,
) -> Result<Option<syn::Expr>> {
    let mut exprs = FieldAttr::from_attrs(FieldAttr::NAME, &field.attrs)?
        .into_iter()
        .filter_map(|a| a.default)
        .map(move |expr| {
            let ty = &field.ty;
            define_spanned_export!(expr.span => _form);
            syn::parse2(quote_spanned!(expr.span => {
                let __default: Option<#ty> = { #expr }.into();
                __default
            })).unwrap()
        });

    let first: Option<syn::Expr> = exprs.next();
    if let Some(expr) = exprs.next() {
        return Err(expr.span()
            .error("duplicate `default` form field attribute")
            .help("form fields may have at-most one `default`"));
    }

    Ok(first)
}
```

A few notes:

  1. Use `filter_map` to avoid the `flat_map`.
  2. No need to check if `first` is `None` and return `None`; just return it.
  3. We want to use the span of the `expr` as the user wrote.
  4. Use the known type in `Option<#ty>`, otherwise inference will surely fail in some cases.

core/codegen/src/derive/from_form.rs
```diff
@@ -244,15 +244,26 @@ pub fn derive_from_form(input: proc_macro::TokenStream) -> TokenStream {
...
                 Ok(quote_spanned! { ty.span() => {
                     let _name = #name_view;
                     let _opts = __c.__opts;
                     __c.#ident
                         .map(<#ty as #_form::FromForm<'__f>>::finalize)
                         .unwrap_or_else(|| {
-                            <#ty as #_form::FromForm<'__f>>::default(_opts)
-                                .ok_or_else(|| #_form::ErrorKind::Missing.into())
+                            #fallback
                         })
```
SergioBenitez, line 267: We can just do: `.unwrap_or_else(|| #fallback)`.

core/codegen/tests/ui-fail/from_form.rs
```diff
@@ -163,4 +163,38 @@ struct Validate3 {
...
+
+struct NotCopy {}
+
+static NOT_COPY: NotCopy = NotCopy {};
+
+#[derive(FromForm)]
+struct Default3 {
+    #[field(default = NOT_COPY)]
+    first: NotCopy,
+}
```
SergioBenitez, line 192: The `stderr` for this test is missing, so surely codegen tests will fail.

But this is an interesting example. This should pass, should it not? Otherwise how can we ever set defaults for something like `String`? It would be good to have more passing test cases.

core/codegen/tests/from_form.rs
```diff
@@ -618,3 +618,33 @@ fn test_nested_multi() {
...
+#[test]
+fn test_defaults() {
+    #[derive(FromForm, PartialEq, Debug)]
+    struct FormWithDefaults {
+        #[field(default = 100)]
+        field1: i32,
+        field2: i32,
+
+        #[field(default = true)]
+        field3: bool,
```
SergioBenitez, line 631:
```suggestion
        field2: i32,
        #[field(default = true)]
        field3: bool,
```

core/codegen/tests/from_form.rs
```diff
@@ -618,3 +618,33 @@ fn test_nested_multi() {
...
+    assert_eq!(form, Some(FormWithDefaults {
+        field1: 101,
+        field2: 102,
+        field3: true,
+        field4: false,
+        field5: false,
+    }));
+    let form: Option<FormWithDefaults> = strict(&form_string).ok();
+    assert!(form.is_none());
+
```
SergioBenitez, line 649: Add a test that specifies all of the fields and ensure that the strict parse works.

core/codegen/tests/from_form.rs
```diff
@@ -618,3 +618,33 @@ fn test_nested_multi() {
 //     println!("String: {}", string);
 //     panic!(string);
 // }
+
+#[test]
+fn test_defaults() {
```
SergioBenitez, line 623: Can we test more types? Ideally every "unique-ish" type we implement `FromForm` for (one or so per entry in https://api.rocket.rs/master/rocket/form/trait.FromForm.html#provided-implementations). The following should work, for example:

```rust
#[form(default = vec![1, 2, 3])]
numbers: Vec<usize>,
```

core/codegen/tests/from_form.rs
```diff
@@ -618,3 +618,33 @@ fn test_nested_multi() {
...
+    #[derive(FromForm, PartialEq, Debug)]
+    struct FormWithDefaults {
+        #[field(default = 100)]
+        field1: i32,
+        field2: i32,
```
SergioBenitez, line 628: I suspect you're using `i32` here because inference failed for `usize` or another integer. You should call this out next time. If you try something first and it doesn't work but you expected to, we should figure out how to make it work instead of changing the test or interface to meet our defied expectations.

SergioBenitez (changes requested):
A bit of cleanup needed and a _bunch_ more tests, a rebase, and we're done.

core/codegen/tests/from_form.rs
ThouCheese: I don't know if the experience is different for you here but the type inference works for all types here so there wasn't anything to call out. The reason I used `i32` is simply that it is the number type that I most usually reach for. I have changed these types to  `usize` and `i128` now, to make sure that this case is tested :+1:

core/codegen/tests/ui-fail/from_form.rs
ThouCheese: Yes this is a mistake. The test should fail because this type is not `Clone`, which should definitely fail. When the type is clone we may want this to work, but in that case a call to `.clone()` needs to be emitted. I don't know how to detect whether a default is a literal/function call, or when the default is a static that needs to be cloned however

ThouCheese:
I am having some trouble running the UI tests (previously I had ran into an error with mysql missing, and I figured that since there were no other errors and I didn't touch any database stuff it should be fine). The current issue is that I am running into ui errors with diffs like this:
```txt
137c137
<    | |     ^- help: consider using a semicolon here: `;`
---
>    | |     ^- help: try adding a semicolon: `;
```
Should I switch to a specific version of rustc or set some flags somewhere?

SergioBenitez:
The tests are always run against the latest versions of nightly and stable, so you'll need those. Then, make sure you set the TRYBUILD=overwrite env variable to automatically get stderr written out. You'll need to do this for both nightly and stable. Oh, if you haven't already, symlink the test to the adjacent nightly and stable directories.

Clearly I should write this down somewhere.

core/codegen/src/derive/form_field.rs
```diff
@@ -300,30 +300,21 @@ pub fn default<'v>(
...
-                    let __default: Option<_> = { #expr }.into();
-                    __default
-                })).unwrap()
-            })
+        .filter_map(|a| a.default)
+        .map(move |expr| {
+            let ty = &field.ty;
+            define_spanned_export!(expr.span => _form);
+            syn::parse2(quote_spanned!(expr.span => {
+                let __default: Option<#ty> = { #expr }.into();
```
ThouCheese, line 308: So there is a problem with this line. The `into` is used now to convert expr into an Option, so it only works if `#expr` is already of the type `#ty`. We wanted to have the into conversion, so it think this should be `let __default: Option<_> = Some({ #expr }.into());`

SergioBenitez:
@ThouCheese What is the status of this? We discussed on Matrix an implementation approach for `.into()`; has it been implemented?

ThouCheese:
I have implemented that changes as we discussed, added all types that implement FromForm, added some extra forms to the tests that do things like `#[field(default = 1 + 3usize)]`, I added time as a dependency of `core/codegen/tests`, to test the `FromForm`'s for the types there. I also rebased, updated to the latest stable and nightly and blessed the output of the failures with `TRYBUILD=overwrite`. Let me know what you think so far!

core/codegen/src/derive/form_field.rs
```diff
@@ -322,7 +323,64 @@ pub fn validators<'v>(
...
+                    let __default: Option<#ty> = __expr.map(|v| v.into());
+                    __default
+                })).unwrap()
+            } else {
+                syn::parse2(quote_spanned!(expr.span => {
+                    let __default: Option<#ty> = Some({ #expr }.into());
+                    __default
+                })).unwrap()
+            }
+        });
```
SergioBenitez, line 374:
```suggestion
            use syn::{Expr, Lit, ExprLit};

            // As a result of calling `#expr.into()`, type inference fails for
            // two common expressions: integer literals and the bare `None`. As
            // a result, we cheat: if the syntax matches either of these two
            // conditions, we provide the field type as a hint.
            let is_int_lit = matches!(*expr, Expr::Lit(ExprLit { lit: Lit::Int(_), .. }));
            let is_none = matches!(*expr, Expr::Path(ref e) if e.path.is_ident("None"));
            let ty = field.stripped_ty();
            let ty_hint = (is_int_lit || is_none)
                .then(|| quote!(#ty))
                .unwrap_or_else(|| quote!(_));

            syn::parse2(quote_spanned!(expr.span => {
                let __expr: Option<#ty_hint> = #expr.into();
                let __default: Option<#ty> = __expr.map(|v| v.into());
                __default
            })).unwrap()
        });
```

core/codegen/src/derive/form_field.rs
```diff
@@ -322,7 +323,64 @@ pub fn validators<'v>(
...
+
+    let first: Option<syn::Expr> = exprs.next();
+    if let Some(expr) = exprs.next() {
+        return Err(expr.span()
+            .error("duplicate `default` form field attribute")
+            .help("form fields may have at-most one `default`"));
```
SergioBenitez, line 380:
```suggestion
            .help("form fields can have at most one `default`"));
```

core/codegen/src/derive/form_field.rs
```diff
@@ -322,7 +323,64 @@ pub fn validators<'v>(
             })).unwrap()
         });
 
-        Ok(exprs)
+    Ok(exprs)
+}
+
+pub fn default<'v>(
+    field: Field<'v>,
+) -> Result<Option<syn::Expr>> {
```
SergioBenitez, line 331:
```suggestion
pub fn default<'v>(field: Field<'v>) -> Result<Option<syn::Expr>> {
```

core/codegen/src/derive/from_form.rs
```diff
@@ -230,16 +230,25 @@ pub fn derive_from_form(input: proc_macro::TokenStream) -> TokenStream {
...
+                    Some(expr) => quote_spanned! { ty.span() =>
+                        #expr
+                            .or_else(|| <#ty as #_form::FromForm<'__f>>::default(_opts))
+                            .ok_or_else(|| #_form::ErrorKind::Missing.into())
+                    },
+                    None => quote_spanned! { ty.span() =>
+                        <#ty as #_form::FromForm<'__f>>::default(_opts)
+                            .ok_or_else(|| #_form::ErrorKind::Missing.into())
+                    },
+                };
```
SergioBenitez, line 244:
```suggestion
                let default = user_default
                    .map(|expr| quote_spanned!(ty.span() =>
                        #expr.or_else(|| <#ty as #_form::FromForm<'__f>>::default(_opts))
                            .ok_or_else(|| #_form::ErrorKind::Missing.into())
                    ))
                    .unwrap_or_else(|| quote_spanned!(ty.span() =>
                        <#ty as #_form::FromForm<'__f>>::default(_opts)
                            .ok_or_else(|| #_form::ErrorKind::Missing.into())
                    ));
```

core/codegen/src/derive/from_form.rs
```diff
@@ -230,16 +230,25 @@ pub fn derive_from_form(input: proc_macro::TokenStream) -> TokenStream {
...
                 Ok(quote_spanned! { ty.span() => {
                     let _name = #name_view;
                     let _opts = __c.__opts;
                     __c.#ident
                         .map(<#ty as #_form::FromForm<'__f>>::finalize)
-                        .unwrap_or_else(|| {
-                            <#ty as #_form::FromForm<'__f>>::default(_opts)
-                                .ok_or_else(|| #_form::ErrorKind::Missing.into())
-                        })
+                        .unwrap_or_else(|| #fallback)
```
SergioBenitez, line 251:
```suggestion
                        .unwrap_or_else(|| #default)
```

core/codegen/tests/from_form.rs
```diff
@@ -1,4 +1,6 @@
-use rocket::form::{Form, Strict, FromForm, FromFormField, Errors};
+use std::{collections::{BTreeMap, HashMap}, net::{IpAddr, SocketAddr}, num::NonZeroI32};
+
+use rocket::{form::{self, Form, Strict, FromForm, FromFormField, Errors}};
```
SergioBenitez, line 3:
```suggestion
use rocket::form::{self, Form, Strict, FromForm, FromFormField, Errors};
```

core/codegen/tests/from_form.rs
```diff
@@ -612,3 +614,242 @@ fn test_multipart() {
...
+    map
+}
+
+fn test_btreemap() -> BTreeMap<&'static str, &'static str> {
+    let mut map = BTreeMap::new();
+    map.insert("key", "value");
+    map
+}
+
+mod default_macro_tests {
```
SergioBenitez, line 630: We should move the unique cases in this `mod` into the test and check that the default is actually used. An important missing one is to check that `None` works as expected, especially for `bool`.

SergioBenitez (changes requested):
The UI tests aren't quite in the right place. If you look at the diff, you can see that the previous errors are no longer appearing in `stderr`; this is bad. What's happening is you've mixed UI tests that result in type errors with UI tests that result in syntax errors. `rustc` will not typecheck if there are syntax errors, so can test both at the same time.

You should likely move the syntax errors into a new `form_form_syntax.rs` file in `ui-fail`. There's a new `synchronize.sh` script there that will automatically create the right symlinks. You can tell if you've done this right if 1) the diff for the form_form.stderr is empty, and 2) you get two new `.stderr` files (for nightly and stable) with the errors you expect.

core/codegen/src/derive/form_field.rs
ThouCheese: This change causes some type inference edge cases to fail again:

```rust
#[derive(rocket::FromForm)]
struct Form {
    #[field(default = 10 + 2)]
    field2: i8,
}
```
Will currently expand into:
```rust
.unwrap_or_else(|| {
    {
        let __default: Option<i8> = Some({ 10 + 2 }.into());
        __default
    }
    .or_else(|| <i8 as ::rocket::form::prelude::FromForm<'__f>>::default(_opts))
    .ok_or_else(|| ::rocket::form::prelude::ErrorKind::Missing.into())
})
```
With this change it expands into:
```rust
.unwrap_or_else(|| {
    {
        let __expr: Option<_> = 10 + 2.into();
        let __default: Option<i8> = __expr.map(|v| v.into());
        __default
    }
    .or_else(|| <i8 as ::rocket::form::prelude::FromForm<'__f>>::default(_opts))
    .ok_or_else(|| ::rocket::form::prelude::ErrorKind::Missing.into())
})
```
which fails to compile with:
```
error[E0282]: type annotations needed
   --> src/main.rs:122:49
    |
122 |         let __default: Option<i8> = __expr.map(|v| v.into());
    |                                                 ^ consider giving this closure parameter a type
    |
    = note: type must be known at this point
```

ThouCheese: Hmm second issue:
```rust
#[derive(rocket::FromForm)]
struct Form {
    #[field(default = Some(true))]
    field2: Option<bool>,
}
```
Used to get expanded to:
```rust
.unwrap_or_else(|| {
    {
        let __default: Option<Option<bool>> = Some({ Some(true) }.into());
        __default
    }
    .or_else(|| {
        <Option<bool> as ::rocket::form::prelude::FromForm<'__f>>::default(
            _opts,
        )
    })
    .ok_or_else(|| ::rocket::form::prelude::ErrorKind::Missing.into())
})
```
whereas now it expands to:
```rust
.unwrap_or_else(|| {
    {
        let __expr: Option<_> = { Some(true) }.into();
        let __default: Option<Option<bool>> = __expr.map(|v| v.into());
        __default
    }
    .or_else(|| {
        <Option<bool> as ::rocket::form::prelude::FromForm<'__f>>::default(
            _opts,
        )
    })
    .ok_or_else(|| ::rocket::form::prelude::ErrorKind::Missing.into())
})
```
This fails to compile with:
```
error[E0282]: type annotations needed
   --> src/main.rs:116:59
    |
116 |         let __default: Option<Option<bool>> = __expr.map(|v| v.into());
    |                                                           ^ consider giving this closure parameter a type
    |
    = note: type must be known at this point
```
I am not sure how you'd like to see this resolved though.

ThouCheese:
I looked at the output of the tests, and I made a mistake that caused the syntax error. I can fix this by patching the test, rather than creating the syntax test file. If you think of some critical untested cases we might need this file after all.

core/lib/src/form/form.rs
```diff
@@ -56,6 +56,23 @@ use crate::form::prelude::*;
...
+/// # #[macro_use] extern crate rocket;
+/// use rocket::form::Form;
+///
+/// #[derive(FromForm)]
+/// struct UserInput<'r> {
+///     #[field(default = "anonymous")]
+///     value: &'r str
+/// }
+/// ```
+///
```
SergioBenitez, line 75: We should clarify that the expression is only evaluated if the field doesn't parse.

SergioBenitez:
Merged in https://github.com/SergioBenitez/Rocket/commit/ebb9f3cfddc809502cee0a34a0b62112eda8b016 with fixes in https://github.com/SergioBenitez/Rocket/commit/ebb9f3cfddc809502cee0a34a0b62112eda8b016. Thanks so much, @ThouCheese! :)

PR #1597, opened 2021-03-29
RFC: Change DataField API to allow accesing the raw file name

### LINKED ISSUE #1594: Expose the original file name including the extension in the multipart DataField
### Why do I believe the feature is necessary

In cases were the client is trusted, the file extension can contain useful information.

### Use case

In my use-case  the client is trusted, and the server not publicly accessible. The client is uploading debian package build artifacts, which are output in a specific naming from a build tool.
There is a sort of manifest file, which lists the other files that belong to the build.

After uploading all those files including the manifest as a multipart request, the manifest file is passed to a tool that manages the repository, and it imports all related files automatically. This depends on the files to be named the same as output by the build tool.
Reconstructing the file extension from the content type is hard and probably error prone.

### Why not outside rocket 

As far as I understand the needed information is not exposed by Rocket, so this is not possible to implement as an additional crate

## General comments

@SergioBenitez mentioned on Matrix that APIs tend to be used the wrong way if the wrong way is as easy to use and find as the proper way to do things.

In occasions were the client is not trusted, the file extension should not be used, as it might be wrong. The content type instead tells the real type of data.

One way to solve this is hiding features that depend on a trusted client behind a feature flag, so they require concious enabling.

JBBgameich:
My attempt at an API for accessing the raw file name. Let me know if you have ideas for improving it.

Fixes #1594

### DIFF
--- a/core/lib/Cargo.toml
+++ b/core/lib/Cargo.toml
@@ -22,6 +22,7 @@ all-features = true
 default = []
 tls = ["rocket_http/tls"]
 secrets = ["rocket_http/private-cookies"]
+trusted_client = []
 
 [dependencies]
 futures = "0.3.0"
--- a/core/lib/src/data/temp_file.rs
+++ b/core/lib/src/data/temp_file.rs
@@ -335,7 +335,7 @@ impl<'v> FromFormField<'v> for Capped<TempFile<'v>> {
     async fn from_data(
         f: DataField<'v, '_>
     ) -> Result<Self, Errors<'v>> {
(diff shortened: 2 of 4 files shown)


SergioBenitez:
Merged in https://github.com/SergioBenitez/Rocket/commit/fa1b75ba74c5c7ae844c416637dac82e31581fb3 with changes in https://github.com/SergioBenitez/Rocket/commit/579508d58f0c8b240da0ea4f7cb74eb954dc4250.

Issue #1591, opened 2021-03-25
Ease Testing Forms w/`LocalForm`


SergioBenitez:
Testing forms in Rocket is currently an entirely manual process:

```rust
let client = Client::tracked(rocket).unwrap();

let response = client.post("/")
    .header(ContentType::Form)
    .body("field=value&is+it=a+cat%3F")
    .dispatch();
```

With the addition of multipart form support, form testing verbosity is even further exacerbated:

```rust
let client = Client::tracked(rocket).unwrap();

let ct = "multipart/form-data; boundary=X-BOUNDARY"
    .parse::<ContentType>()
    .unwrap();

let body = &[
    "--X-BOUNDARY",
    r#"Content-Disposition: form-data; name="names[]""#,
    "",
    "abcd",
    "--X-BOUNDARY",
    r#"Content-Disposition: form-data; name="names[]""#,
    "",
    "123",
    "--X-BOUNDARY",
    r#"Content-Disposition: form-data; name="file"; filename="foo.txt""#,
    "Content-Type: text/plain",
    "",
    "hi there",
    "--X-BOUNDARY--",
    "",
].join("\r\n");

let response = client.post("/")
    .header(ct)
    .body(body)
    .dispatch();
```

Rocket should make testing applications with forms a _much_ simpler task.

Here's how we might hope testing the above forms would look:

```rust
let client = Client::tracked(rocket).unwrap();
let response = client.post("/")
    .form(&[("field", "value"), ("is it", "a cat?")])
    .dispatch();

let client = Client::tracked(rocket).unwrap();
let response = client.post("/")
    .form(LocalForm::new()
        .field("names[]", "abcd")
        .field("names[]", "123")
        .file("foo.txt", ContentType::Plain, "hi there"))
    .dispatch();
```

An API enabling this might resemble:

```rust
impl Client {
    pub fn form<F: Into<LocalForm>>(&mut self, form: F) -> &mut Self {
        let form = form.into();
        self.set_header(form.content_type());
        self.set_body(form.body_data());
        self
    }
}

struct LocalForm { /* .. */ }

impl LocalForm {
    pub fn new() -> Self;

    /// A percent-decoded `name` and `value`.
    pub fn field<'v, N, V>(mut self, name: N, value: V) -> Self
        where N: Into<NameBuf<'v>>, V: AsRef<str>;

    /// Adds all of the `fields` to `self`.
    pub fn fields<'v, I, F>(mut self, fields: I) -> Self
[...]

PR #1605, opened 2021-04-11
Add `LocalForm` to make testing with form data more ergonomic

ELD:
This adds the testing improvement described in #1591.

The transformation of the form fields into the resulting body data is a bit clumsy and could probably use some improvement/cleanup. I believe the multipart form translation is correct, but there's only one, narrow test case that confirms it works as expected.

Additionally, I was unable to get something like this to work without ugly ergonomics:
```rust
let client = Client::tracked(rocket).unwrap();
let response = client.post("/")
    .form(&[("field", "value"), ("is it", "a cat?")])
    .dispatch();
```

I might be missing something, but attempting to do something like this:
```rust
impl<'v, F: Into<ValueField<'v>>, I: Iterator<Item = F>> From<I> for LocalForm { /* .. */ }
```

requires the caller to call the method in a way similar to this:

```rust
client.post("/")
    .form(&[("field", "value"), ("is it", "a cat?")].iter().cloned());
```

due to a slice reference not being coerced into an iterator and when it is iterated over, it returns `&(&str, &str)`. `ValueField` doesn't have a present `From<..>` impl for `&(&str, &str)`, so this is not possible.

Also, docs are likely not complete/elaborate enough at this stage, but an attempt at documenting relevant methods/structs was made in this first cut.

Please pick it apart so I can make the implementation better! :)


core/lib/src/local/form.rs
```diff
@@ -0,0 +1,222 @@
...
+            ct,
+            data.as_ref().into(),
+        ));
+        self
+    }
+
+    /// The full content-type for this form.
+    pub fn content_type(&self) -> ContentType {
+        if self.contains_data_field() {
+            return "multipart/form-data; boundary=X-BOUNDARY"
```
SergioBenitez, line 114: use `ContentType::with_params()`.

core/lib/src/local/form.rs
```diff
@@ -0,0 +1,222 @@
...
+        } else {
+            self.format_simple()
+        }
+    }
+
+    fn contains_data_field(&self) -> bool {
+        self
+            .0
+            .iter()
+            .any(|field| matches!(field, LocalField::Data(..)))
```
SergioBenitez, line 135:
```suggestion
        self.0.iter().any(|field| matches!(field, LocalField::Data(..)))
```

core/lib/src/local/form.rs
```diff
@@ -0,0 +1,222 @@
...
+    /// The full body data for this form.
+    pub fn body_data(&self) -> Vec<u8> {
+        if self.contains_data_field() {
+            self.format_multipart()
+        } else {
+            self.format_simple()
+        }
+    }
+
+    fn contains_data_field(&self) -> bool {
```
SergioBenitez, line 131: I don't think this needs to be a standalone function.

core/lib/src/local/form.rs
```diff
@@ -0,0 +1,222 @@
...
+            self.0.push(LocalField::Value(
+                NameBuf::from(field.name.as_name()),
+                field.value.to_string(),
+            ));
+        });
+        self
+    }
+
+    /// A percent-encoded `name` and `value`.
+    pub fn raw_field(mut self, name: &'v RawStr, value: &'v RawStr) -> Self {
```
SergioBenitez, line 73: Take an `Into<Cow<'_, RawStr>>`.

core/lib/src/local/form.rs
```diff
@@ -0,0 +1,222 @@
...
+
+    /// Adds a field for the file with the name `file_name`, content-type `ct`, and
+    /// file contents `data`.
+    pub fn file<N, V>(mut self, file_name: N, ct: ContentType, data: V) -> Self
+    where
+        N: Into<Option<&'v str>>,
+        V: AsRef<[u8]>,
+    {
+        self.0.push(LocalField::Data(
+            NameBuf::from("file"),
```
SergioBenitez, line 89: I didn't have it in my original API, but this field name needs to be a parameter to the function. We can't just set it to `file`.

core/lib/src/local/form.rs
```diff
@@ -0,0 +1,222 @@
...
+        self
+    }
+
+    /// Add a data field with a content-type `ct` and binary `data`.
+    pub fn data<V>(mut self, ct: ContentType, data: V) -> Self
+    where
+        V: AsRef<[u8]>,
+    {
+        self.0.push(LocalField::Data(
+            NameBuf::from("file"),
```
SergioBenitez, line 103: Same here.

core/lib/src/local/form.rs
```diff
@@ -0,0 +1,222 @@
...
+    fn format_simple(&self) -> Vec<u8> {
+        self.0.iter().fold(Vec::new(), |mut acc, field| {
+            match field {
+                LocalField::Value(name, value) => {
+                    acc.push(format!("{}={}", Uri::percent_encode(&format!("{}", name)), Uri::percent_encode(value)));
+                    return acc
+                },
+                _ => acc,
+            }
+        }).join("&").as_bytes().to_vec()
```
SergioBenitez, line 147: Ideally we allocate a total of 1 `Vec<u8>` here, pass a pointer it around, and use `write!` to fill it up.

Here's an example of how to do that:

```rust
use std::io;

fn f(buf: &mut dyn io::Write) -> io::Result<()> {
    write!(buf, "hello, {}", "there")?;
    Ok(())
}

pub fn main() {
    let mut buf = vec![];
    let _ = f(&mut buf);
}
```

I'd like the this code to have _zero_ calls to `format!`, or `to_string()`, or similar.

core/lib/src/local/form.rs
```diff
@@ -0,0 +1,222 @@
...
+                    acc.push(format!("Content-Disposition: form-data; name=\"{}\"", name));
+                    acc.push("".to_string());
+                    acc.push(format!("{}", value));
+                },
+                LocalField::Data(name, file_name, content_type, data) => {
+                    acc.push("--X-BOUNDARY".to_string());
+                    acc.push(format!("Content-Disposition: form-data; name=\"{}\"; filename=\"{}\"", name, file_name.unwrap_or("")));
+                    acc.push(format!("Content-Type: {}", content_type));
+                    acc.push("".to_string());
+                    acc.push(format!("{}", String::from_utf8_lossy(data)));
```
SergioBenitez, line 164: This is wrong; `data` may not be UTF-8, and this will incorrectly format the data. This shouldn't be a `String`, but part of the byte buffer.

SergioBenitez (changes requested):
The code is handling percent-encoding incorrectly. It's decoding when it shouldn't be and failing to encode when it should be. Here are the conditions:

  * If a provided string _is_ percent-encoded, it should be left alone.
  * If a provided string _is not_ percent-encoded, it should be percent-encoded.

Specifically, note that for some `s`, `encode(decode(s)) != s` (for example, `%`).

Note that we have the `UriDisplay<Query>` trait that already knows how to do this for a bunch of types. We should probably use it here. Perhaps better type signatures are:

```rust
pub fn field<N, V>(mut self, name: N, value: V) -> Self
    where N: Into<NameBuf<'v>>, V: UriDisplay<Query>,
```

Let's start by fixing the incorrect code and style and we'll go from there.

ELD:
@SergioBenitez This should be ready for another pass.

I updated the style to better match Rocket's. I think it's mostly there, but there might be some minor style issues remaining.

I changed the writing of the body data to a single `Vec<u8>` with the `write!` macro. There's a couple places where I call `format!`. There's probably a better way than what I did, so any help there would be great!

Error handling is a little weird as it currently silently outputs an empty buffer if the writing failed for some reason. This probably isn't a good user experience because an error would result in empty body data for the form request. Any pointers on how to do better error handling would be great!

slyons:
Has anyone had a chance to look at this yet?

PR #1606, opened 2021-04-11
XML support in contrib using quick-xml

romatthe:
This PR should add XML support to `contrib` along the lines of the already existing `contrib::json` and `contrib::msgpack` modules, using [quick-xml](https://crates.io/crates/quick-xml).

One thing to consider is that `quick-xml` only implements `DeserializeOwned` and not `Deserialize<'de>`. [xml-rs](https://crates.io/crates/xml-rs) does, but is much less performant than `quick-xml`.

Note that, of course, this adds a new dependency to the project.

(This is my first PR, if it is not of sufficient quality, feel free to reject it, I implemented this quickly because I'm communicating with an XML client in a recent project.)

### DIFF
--- a/contrib/lib/Cargo.toml
+++ b/contrib/lib/Cargo.toml
@@ -23,6 +23,7 @@ databases = [
 default = ["json", "serve"]
 json = ["serde", "serde_json", "tokio/io-util"]
 msgpack = ["serde", "rmp-serde", "tokio/io-util"]
+xml = ["serde", "quick-xml", "tokio/io-util"]
 tera_templates = ["tera", "templates"]
 handlebars_templates = ["handlebars", "templates"]
 helmet = ["time"]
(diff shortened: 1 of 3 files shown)


SergioBenitez:
> One thing to consider is that `quick-xml` only implements `DeserializeOwned` and not `Deserialize<'de>`. [xml-rs](https://crates.io/crates/xml-rs) does, but is much less performant than `quick-xml`.

I suspect that this negates most of the performance gains in many cases. Unfortunately, not allowing `Deserialize` isn't an option as it removes parity with `json` and `msgpack` and would force the user to allocate where they would otherwise not need to. It seems there's a PR open in `quick-xml` that seeks to implement `Deserialize`, though it has not seen movement in a year. 

I'd be happy to have built-in support for XML in Rocket should there be a performant implementation that does not force owned structures. As that does not appear to be the case now, I'm closing out this issue. Thank you for bringing the current state of affairs to my attention!

Issue #1225, opened 2020-02-09
Dev version of Rocket can't find marco routes


Voodlaz:
I was having a problem with Tera Templates. I'm using windows, and i found [this](https://github.com/SergioBenitez/Rocket/issues/169) issue(and it solved my broblem). I solved the problem just by using the unrealeased version. So, rust is saying that he can't find marco routes!.

main.rs code:
```
extern crate rocket;
extern crate rocket_contrib;

use std::collections::HashMap;

use rocket_contrib::templates::Template;

fn index() -> Template {
    let context: HashMap<&str, &str> = HashMap::new();
    Template::render("home", &context)
}

fn main() {
    rocket::ignite().attach(Template::fairing()).mount("/", routes![index]).launch();
}
```

cargo.toml code:
```
[dependencies]`
rocket = { git = "https://github.com/SergioBenitez/Rocket" }
rocket_codegen = { git = "https://github.com/SergioBenitez/Rocket" }

[dependencies.rocket_contrib]
git = "https://github.com/SergioBenitez/Rocket"
default-features = false
features = ["handlebars_templates","tera_templates"]
```

P.S. Sorry for my English. I wasn't born with this language.

jebrosen:
That issue should have been a fixed a long time ago; if you are using 0.4 I would expect that your problem with tera is different from that one.

As for this issue, you haven't imported the `routes!` macro anywhere. You need either a `#[macro_use] extern crate rocket;` in `main.rs` to import all macros, or a `use rocket::routes;` to import that specific macro.

Issue #169, opened 2017-02-03
Render a view file without any context


dumindu:
**Is there any way to render a tera view file without any context?**
```
#[get("/hello_template/")]
fn hello_template() -> Template {
    Template::render("main")
}
```
Get ```Template::render expected 2 parameters```

----

```
#[get("/hello_template/")]
fn hello_template() -> Template {
    Template::render("main", ())
}
```
Get ```expected type `&_` found type `()` ```

----

```
#[get("/hello_template/")]
fn hello_template() -> Template {
    Template::render("main", &())
}
```

No error when ```cargo build``` but when runnning ```cargo run```

```
=> GET /hello_template
Warning: GET /hello_template and GET /hello_template collide!
Warning: Route collisions detected!
Rocket has launched from http://localhost:8000...

GET /hello_template:
    => Matched: GET /hello_template
    => Error: Tera template 'main.tera' does not exist.
    => Outcome: Failure
    => Warning: Responding with 500 Internal Server Error catcher.
    => Response succeeded.
```

Even I had templates/main.tera

mkroman:
The error `Error: Tera template 'main.tera' does not exist.` is unrelated to the context.

However, `Template::render("main", &())` will give the error `Error rendering Tera template 'main': Failed to value_render 'main.tera': context isn't an object`

I suspect this might be because [tera expects a valid json object](https://docs.rs/tera/0.7.0/src/tera/tera.rs.html#204-205) and most things can't be represented as a JSON object.

dumindu:
something I have missed? 
https://pastebin.mozilla.org/8973142

SergioBenitez:
Which version of Rocket are you using? Please make sure you're on the latest one: `v0.1.6`.

dumindu:
I used v0.1.5. Also updated and checked but get the same error. Have I missed something on Cargo.toml ?

```
[dependencies]
rocket = "0.1.6"
rocket_codegen = "0.1.6"

[dependencies.rocket_contrib]
version = "*"
default-features = false
features = ["tera_templates"]
```

```
$ rustc -V
rustc 1.16.0-nightly (eedaa94e3 2017-02-02)
```
using rustup on Windows.

SergioBenitez:
@dumindu Please see my updated comment above.

dumindu:
Got it. Thanks @mkroman @SergioBenitez  for the effort and time to check this.

PR #1613, opened 2021-04-22
Implement Template::render_without_context for a simple interface, 

shinichi-takayanagi:
If we don't have a `context` and want to use `Template` for `rendering` , 
we need to write the following described in #169, #1225

```rust
...
fn index() -> Template {
    let context: HashMap<&str, &str> = HashMap::new();
    Template::render("home", &context)
}
```

This one seems to be a little bit redundant, so I implemented `redner_without_context`.
Using this, we can write as follows, which I think that it will be quite simple.

```rust
fn index() -> Template {
    Template::render_without_context("home")
}
```

### DIFF
--- a/contrib/lib/src/templates/mod.rs
+++ b/contrib/lib/src/templates/mod.rs
@@ -355,7 +355,7 @@ impl Template {
         S: Into<Cow<'static, str>>,
     {
         let context: HashMap<&str, &str> = HashMap::new();
-        render("index", &context)
+        Template::render(name, &context)
     }
 
     /// Render the template named `name` with the context `context` into a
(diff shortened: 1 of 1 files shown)


jebrosen:
I'm not really convinced this deserves a separate method: one of the main points of using templates is to provide a rendering context at runtime, so this method should be very rarely used. Neither `tera` nor `handlebars` seems to have it, either. Templates that will never have context should probably be served as [StaticFiles](https://api.rocket.rs/v0.4/rocket_contrib/serve/struct.StaticFiles.html) instead. Finally, using `render_without_context` adds friction when adding a context later since the method call needs to be adjusted, and (at least in my experience) I have found myself needing to add some data to an empty context much more often than the other way around.

If something is to be fixed, I would prefer documentation that encourages a simpler way to get an empty context that can still be easily extended, such as:

```rust
use serde_json::json; // TODO: exact path is or will be re-exported somewhere nicer within `rocket::`
Template::render("home", json!({}))
```

PR #1614, opened 2021-04-22
Derive Clone for Template

rotoclone:
I was trying to set up an in-memory cache for rendered `Template`s, but found that this is made much more difficult by `Template` not implementing `Clone`. So I made it do dat.

### DIFF
--- a/contrib/lib/src/templates/mod.rs
+++ b/contrib/lib/src/templates/mod.rs
@@ -197,7 +197,7 @@ const DEFAULT_TEMPLATE_DIR: &str = "templates";
 /// You may use the [`Template::custom()`] method to construct a fairing with
 /// customized templating engines. Among other things, this method allows you to
 /// register template helpers and register templates from strings.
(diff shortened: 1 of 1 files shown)


rotoclone:
I've done a little more digging and it turns out a `Template` doesn't actually contain the rendered template; it still has to hit the filesystem to be turned into a string. So this change isn't as useful as I thought. (Though I'm not sure there's a downside to adding it, other than maybe implying that cloning a `Template` is something someone might want to do, which I'm now doubting.)

SergioBenitez:
> other than maybe implying that cloning a `Template` is something someone might want to do, which I'm now doubting

Indeed, this is my concern. Perhaps we can add a simple note to the `Template` docs indicating _why_ `Template` isn't `Clone`? In particular, that it just holds metadata and _not_ the rendered template,

SergioBenitez:
Closing this particular PR out, but I'd welcome one that adds the doc note specified above. Thank you!

PR #1618, opened 2021-04-26
Initialize synchronous database pools inside of `spawn_blocking`.

### LINKED ISSUE #1610: Sometimes cargo test fails with async requests (master)
**Version:** latest master
**OS:** Ubuntu 20.10
**Description:** Following code works fine in normal conditions, but when running with `cargo test` sometimes it fails with double panic. Couldn't reproduce it with one test.

**Code:**
```rust
#[macro_use] extern crate rocket_contrib;

use rocket_contrib::databases::postgres;
use rocket::{Rocket, Build};


#[database("storage")]
pub struct ApiDatabase(postgres::Client);


#[get("/")]
async fn get_entries(conn: ApiDatabase) -> &'static str {
    println!("Running database task..");

    let _some_result = conn.run(|c| c.query("SELECT * FROM entries", &[])).await;

    println!("Ran database task!");
    "Works!?"
}


#[launch]
fn rocket() -> Rocket<Build> {
    rocket::build()
        .mount("/entries", routes![get_entries])
        .attach(ApiDatabase::fairing())
}


#[cfg(test)]
mod test {
    use super::rocket;

    #[rocket::async_test]
    async fn test_get_entries() {
        use rocket::local::asynchronous::Client;

        let client = Client::tracked(rocket()).await.unwrap();
        // The code below breaks as well, but it's much harder to reproduce with the debugger
        //let _resp  = client.get("/entries").dispatch().await;
        let (_r1, _r2, _r3, _r4) = rocket::tokio::join!(
            client.get("/entries").dispatch(),
            client.get("/entries").dispatch(),
            client.get("/entries").dispatch(),
            client.get("/entries").dispatch()
        );
    }

    #[rocket::async_test]
    async fn test_get_entries_2() {
        use rocket::local::asynchronous::Client;

        let client = Client::tracked(rocket()).await.unwrap();
        // The code below breaks as well, but it's much harder to reproduce with the debugger
        //let _resp  = client.get("/entries").dispatch().await;
        let (_r1, _r2, _r3, _r4) = rocket::tokio::join!(
            client.get("/entries").dispatch(),
            client.get("/entries").dispatch(),
            client.get("/entries").dispatch(),
            client.get("/entries").dispatch()
        );
    }
}
```

**Issue:**
```bash
$ cargo test
    Finished test [unoptimized + debuginfo] target(s) in 0.07s
     Running unittests (target/debug/deps/rocket_bug_repro-5b440ae7b98fd091)

running 2 tests
test test::test_get_entries_2 ... ok
thread panicked while panicking. aborting.
error: test failed, to rerun pass '--bin rocket-bug-repro'

Caused by:
  process didn't exit successfully: `/home/kittyandrew/tmp/rocket-bug-repro/target/debug/deps/rocket_bug_repro-5b440ae7b98fd091` (signal: 4, SIGILL: illegal instruction)
```

**From the debugger:**
msg:
```
"Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks."
```

jebrosen:
One situation where this is noticeable is when initialization failure
leads to `Drop`; unlike in a successful initialization, `postgres`
detects and panics when `Drop` is called from within asynchronous code.
Several other database pools do not panic in this same situation, but
would still block the current thread.

Also sets the minimum version of `tokio` to 1.4 in `rocket_contrib`,
which is the version where `Handle::block_on` (used in `Drop` impls) was
introduced.

Fixes #1610.

---

This should be reviewed closely, since there have been several bugs and regressions in this area. The goal is thus:

Synchronous database pool and connection usage, *including* creation, usage, and `Drop`, should always be done inside some kind of "blocking-safe" environment. Since #1375 databases use `spawn_blocking`; alternatives include `block_in_place` (risky) or a dedicated thread with message passing to communicate between the async and sync code (safer, but probably not better than the status quo in terms of complexity to implement maintain or troubleshoot and likely still pretty bad ergonomically).

This problem is now addressed by starting the database pool inside a `spawn_blocking` closure. Usage of the database pool (i.e. acquiring connections) was already wrapped, along with usage of the connections themselves. This leaves `Drop`, which was also already wrapped properly -- aside from a race condition between server shutdown and handlers that were still running, which was recently fixed in e1307dd.

### DIFF
--- a/contrib/lib/Cargo.toml
+++ b/contrib/lib/Cargo.toml
@@ -42,7 +42,7 @@ diesel_mysql_pool = ["databases", "diesel/mysql", "diesel/r2d2"]
 
 [dependencies]
 # Global dependencies.
-tokio = { version = "1.0", optional = true }
+tokio = { version = "1.4", optional = true }
 rocket_contrib_codegen = { version = "0.5.0-dev", path = "../codegen", optional = true }
 rocket = { version = "0.5.0-dev", path = "../../core/lib/", default-features = false }
 log = "0.4"
--- a/contrib/lib/src/databases/connection.rs
+++ b/contrib/lib/src/databases/connection.rs
@@ -71,23 +71,25 @@ macro_rules! dberr {
 impl<K: 'static, C: Poolable> ConnectionPool<K, C> {
     pub fn fairing(fairing_name: &'static str, db: &'static str) -> impl Fairing {
         AdHoc::try_on_ignite(fairing_name, move |rocket| async move {
-            let config = match Config::from(db, &rocket) {
-                Ok(config) => config,
-                Err(e) => dberr!("config", db, "{}", e, rocket),
-            };
+            run_blocking(move || {
+                let config = match Config::from(db, &rocket) {
+                    Ok(config) => config,
+                    Err(e) => dberr!("config", db, "{}", e, rocket),
+                };
 
-            let pool_size = config.pool_size;
-            match C::pool(db, &rocket) {
-                Ok(pool) => Ok(rocket.manage(ConnectionPool::<K, C> {
-                    config,
-                    pool: Some(pool),
-                    semaphore: Arc::new(Semaphore::new(pool_size as usize)),
-                    _marker: PhantomData,
-                })),
-                Err(Error::Config(e)) => dberr!("config", db, "{}", e, rocket),
-                Err(Error::Pool(e)) => dberr!("pool init", db, "{}", e, rocket),
-                Err(Error::Custom(e)) => dberr!("pool manager", db, "{:?}", e, rocket),
-            }
+                let pool_size = config.pool_size;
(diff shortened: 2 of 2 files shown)


SergioBenitez (commented):
GitHub won't let me comment on lines that aren't modified, so here are my notes:

  * Can `self.pool.clone()` on line 32 block? If so, that needs a `spawn_blocking()` too.
  * Same with the `cloned()` in `get()`.

Otherwise, everything appears to be wrapped in `spawn_blocking()`, so I don't see where else an issue can be introduced.

jebrosen:
`self.pool` is an `r2d2::Pool`, which contains only an `Arc`. And that `cloned()` call is actually `clone()`ing the exact same thing, so that sounds good to me.

Issue #1598, opened 2021-03-30
Unused routes don't generate unused warnings


SergioBenitez:
Functions decorated with a Rocket route attribute do not generate a warning, even if they are otherwise unused:

```rust
#[rocket::get("/")]
fn foo() { }
```

This is a "regression" of sorts introduced by the mechanism in 092e03f7207941f9585db4a2d04dbc08c731e5db. It would be more than ideal if we could emit code that results in `rustc` generating the appropriate warning.

the10thWiz:
I'm interested in working on this issue. I'm not super familiar with proc macros, or rocket's internals, but it looks to me like the fact that the function is used within the implementation of `From<route> for ::StaticInfo` prevents the compiler from providing an unused warning.

One way to avoid this is to implement the methods on the struct itself, rather than as a trait. This does solve this issue (with a reasonably small patch to the codegen), but I suspect it might cause other issues with more complex codebases. It does expose 1 or 2 methods to the user, in a way that doesn't feel like it should make sense, but I think this might be a reasonable trade-off.

Can you provide some guidance on what issues this patch might cause?

I will be submitting the patch as a pull request shortly.

PR #1620, opened 2021-04-27
Change From<> to a normal method

the10thWiz:
Changes the codegen for routes to use a method on the proxy object,
rather than implementing a trait on the proxy object. This helps rustc
identify unused routes, although it may break compatibility for some
specific use cases.

See https://github.com/SergioBenitez/Rocket/issues/1598 for more info.

### DIFF
--- a/core/codegen/src/attribute/route/mod.rs
+++ b/core/codegen/src/attribute/route/mod.rs
@@ -336,7 +336,7 @@ fn codegen_route(route: Route) -> Result<TokenStream> {
         #vis struct #handler_fn_name {  }
 
         /// Rocket code generated proxy static conversion implementation.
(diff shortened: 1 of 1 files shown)


SergioBenitez:
Can you check out the build failure so the CI can run?

the10thWiz:
It looks like you fixed the CI error. Please let me know if this change would cause any compatibility issues with existing code.

As an addition, it may be a good idea to rename the methods to something like `as_route` and `as_static_info`, although this would require changes to the `routes!` macros (and potentially others), but I don't think it's nessecary for this specific features.

SergioBenitez:
I don't foresee any compatibility issues! Merged in d4466b73af747dfaec0880d85eb9a210fc8e3686.

Issue #1593, opened 2021-03-26
Add Serialize/Deserialize support to rocket::http::uri::Uri enum


DanielJoyce:
This would make loading of Uris from config files, etc nicer.

I looked at the Url crate, but it's api is a mess and it has some very unintutiive behaviors.

DanielJoyce:
I looked at the Url crate, but it's api is a mess and it has some very unintutiive behaviors.

SergioBenitez:
We should absolutely do this. Note that we _don't_ use the `Url` crate; we have our own implementation of URIs.

I'm happy to accept a PR for this. Implementation should be rather straightforward to/from a string.

Some notes for those wishing to submit a PR:
  * The changes should be reflected in `rocket_http`.
  * The changes should be gated by the `http` crate's `serde` feature.
  * The implementation should likely be a manual implementation of `(De)Serialize` to/from string.
  * All of the `Uri` variants should support (de)serialization.
  * Tests should be added.
  * The URI docs should be updated to mention support for (de)serialization of `Uri` and friends.

the10thWiz:
I'm interested in working on this issue. I have taken a look at the rocket_http crate, and it is fairly straight forward to add the serde traits to `Uri`. I have not touched the http crate outside of Rocket, since it doesn't appear to have a serde feature.

Implementing Serialize and Deserialize is easy, especially since a Uri is always serialized as a string. This is especially easy since I am relying on the existing `TryFrom` and `Display` impletations to do the heavy lifting. I have added serde_json as a dev-dependency to make testing the Serialize and Deserialize implementations easier.

I have discovered that the existing `Uri` parsing doesn't seem to be as stable as it should be. There are a number of edge cases my testing has found, see [https://github.com/the10thWiz/Rocket/blob/2b82ec9bd1aaeed75f02ec78d791f69fe1cf69b4/core/http/src/uri/uri.rs] for the full tests. The current implementation accepts '&', '.', and a range of other odd uris as valid, as well as a curious set of empty parts. For example, 'example.com/test?' is not a valid Uri, but '/test?' is. You can take a look at `serde_invalid_uri` and `serde_empty_parts` to see which Uris cause errors, and which don't.

There are a few other quirks of the Uri parsing I discovered, that are documented in the `serde_quirks` test.

PR #1621, opened 2021-04-27
Uri serde

the10thWiz:
PR to address https://github.com/SergioBenitez/Rocket/issues/1593

This is not ready to be merged, although it ticks most of the boxes. It looks like the existing Uri parsing functions may need to be modified.

### DIFF
--- a/core/http/Cargo.toml
+++ b/core/http/Cargo.toml
@@ -58,4 +58,4 @@ features = ["std"]
 
 [dev-dependencies]
 rocket = { version = "0.5.0-dev", path = "../lib" }
-serde_json = "*"
+serde_test = "1.0"
--- a/core/http/Cargo.toml
+++ b/core/http/Cargo.toml
@@ -58,3 +58,4 @@ features = ["std"]
 
 [dev-dependencies]
 rocket = { version = "0.5.0-dev", path = "../lib" }
+serde_json = "*"
--- a/core/http/src/uri/uri.rs
+++ b/core/http/src/uri/uri.rs
@@ -60,6 +60,15 @@ use crate::uri::encoding::{percent_encode, DEFAULT_ENCODE_SET};
 /// [`Uri::percent_encode()`]: crate::uri::Uri::percent_encode()
 /// [`Uri::percent_decode()`]: crate::uri::Uri::percent_decode()
 /// [`Uri::percent_decode_lossy()`]: crate::uri::Uri::percent_decode_lossy()
+///
+/// ## Serde
+///
+/// When the `serde` feature is enabled, the Uri type implements both serialize and
+/// deserialize, through the appropriate Serde traits. This allows the use of Uri
+/// in configuration documents that are parsed via serde, with little to no issue.
+/// 
+/// The implementation currently relies on the Uri's display and parse functions,
+/// and may not be compatible across mutiple Rocket versions
 #[derive(Debug, PartialEq, Clone)]
 pub enum Uri<'a> {
     /// An origin URI.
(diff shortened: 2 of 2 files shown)


SergioBenitez:
Thank you for this!

Instead of using `serde_json` to test, use https://docs.serde.rs/serde_test/, which is designed exactly for this usecase, and will remove all of the `\" .. \"` in the tests.

core/http/src/uri/uri.rs
```diff
@@ -316,3 +325,264 @@ macro_rules! impl_uri_from {
...
+        }
+        impl From<serde_json::Error> for TestError {
+            fn from(e: serde_json::Error) -> Self {
+                Self::JsonError(e)
+            }
+        }
+
+        // Test Helper function
+        // 'a sould almost always be 'static, but it may be useful to not require that
+        fn test_helper<'a>(raw: &str, json: &'a str, uri_obj: Uri<'a>) -> Result<(), TestError> {
```
SergioBenitez, line 406: Using `serde_test`, we should be able to get ride of the second parameter. Or even without `serde_test`, the second parameter is just `format!("\"{}\"", raw);`.

Nevertheless, this should either be `#[track_caller]` or a macro so that test failures point to the correct line.

core/http/src/uri/uri.rs
```diff
@@ -316,3 +325,264 @@ macro_rules! impl_uri_from {
...
+
+            assert_eq!(serde_json::to_string(&uri)?, json, "Serialization did not produce the expected Json string");
+            assert_eq!(serde_json::from_str::<Uri<'a>>(json)?, uri, "Deserialization from a `&str` did not produce the expected Uri");
+            Ok(())
+        }
+        
+        /// /path?query
+        fn origin<'a>(path: &'a str, query: Option<&'a str>) -> Uri<'a> {
+            Uri::Origin(Origin::new(path, query))
+        }
```
SergioBenitez, line 418: Do we need this? `Uri` implements `PartialEq<Origin>`, and this is otherwise `Origin::new(path, query)`.

core/http/src/uri/uri.rs
```diff
@@ -316,3 +325,264 @@ macro_rules! impl_uri_from {
...
+            test_helper("/a/nice/url", "\"/a/nice/url\"", origin("/a/nice/url", None))?;
+
+            test_helper("/a?param", "\"/a?param\"", origin("/a", Some("param")))?;
+            test_helper("/a?param=value", "\"/a?param=value\"", origin("/a", Some("param=value")))
+        }
+
+        /// username:password@some.host:8088
+        fn authority<'a>(user: Option<&'a str>, host: &'a str, port: Option<u16>) -> Uri<'a> {
+            Uri::Authority(Authority::new(user, Host::Raw(host), port))
+        }
```
SergioBenitez, line 435: Same here as with `origin()`.

SergioBenitez (changes requested):
The majority of these seem to actually be testing parsing rather than the ser(de) implementation. These should all be moved to `http/src/parse/uri/tests.rs` where there exists better infrastructure to test these.

Can you summarize, as a list, the strings that you believe fail to parse correctly and how you believe they should parse?

core/http/src/uri/uri.rs
```diff
@@ -316,3 +325,264 @@ macro_rules! impl_uri_from {
...
+        fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
+        where
+                E: de::Error, {
+            Uri::try_from(v.to_string()).map_err(de::Error::custom)
+        }
+        fn visit_string<E>(self, v: String) -> Result<Self::Value, E>
+        where
+                E: de::Error, {
+            Uri::try_from(v).map_err(de::Error::custom)
+        }
```
SergioBenitez, line 371: Please follow the style in the codebase. Namely:

```rust
fn f<E>() -> T
    where E: Foo
{
    ...
}
```

or

```rust
fn f<E: Foo>() -> T {
    ...
}
```

SergioBenitez: As well as a single space between items.

core/http/src/uri/uri.rs
```diff
@@ -316,3 +325,264 @@ macro_rules! impl_uri_from {
...
+        impl<'a> From<uri::Error<'a>> for TestError {
+            fn from(e: uri::Error<'a>) -> Self {
+                Self::UriError(Box::new(e.into_owned()))
+            }
+        }
+        impl From<serde_json::Error> for TestError {
+            fn from(e: serde_json::Error) -> Self {
+                Self::JsonError(e)
+            }
+        }
```
SergioBenitez, line 402: We need a space between items.

core/http/src/uri/uri.rs
```diff
@@ -316,3 +325,264 @@ macro_rules! impl_uri_from {
...
+            test_helper("http:///index.html?query", "\"http:///index.html?query\"", absolute("http", None, Some(""), None, Some("/index.html"), Some("query")))?;
+            test_helper("http:///path/with/segments/?query=value", "\"http:///path/with/segments/?query=value\"", absolute("http", None, Some(""), None, Some("/path/with/segments/"), Some("query=value")))
+        }
+        
+        #[test]
+        fn serde_asterisk_form() -> Result<(), TestError> {
+            test_helper("*", "\"*\"", Uri::Asterisk)
+        }
+        
+
```
SergioBenitez, line 526:
```suggestion

```

core/http/src/uri/uri.rs
```diff
@@ -316,3 +325,264 @@ macro_rules! impl_uri_from {
...
+            // These are valid, albiet unusual Uris
+            // `http:/` => scheme: `http`, origin: `/`, authority: None
+            test_helper("http:/", "\"http:/\"", absolute("http", None, None, None, Some("/"), None))?;
+            // `http:` => scheme: `http`, origin: ``, authority: None
+            test_helper("http:", "\"http:\"", absolute("http", None, None, None, Some(""), None))?;
+            Ok(())
+        }
+    }
+}
+
```
SergioBenitez, line 588:
```suggestion
```

core/http/src/uri/uri.rs
```diff
@@ -316,3 +325,264 @@ macro_rules! impl_uri_from {
...
+    use _serde::{Deserialize, Deserializer, Serialize, Serializer, de::{self, Visitor}};
+
+    // The serialize implementation depends on the fmt::Display implementation. However, this
+    // should be okay, and is likely the best option, since allocating a string is unavoidable, and
+    // this method should allocate exactly one string
+    impl<'a> Serialize for Uri<'a> {
+        fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+        where
+                S: Serializer {
+            serializer.serialize_str(&format!("{}", self))
```
SergioBenitez, line 345: Style.

core/http/src/uri/uri.rs
```diff
@@ -316,3 +325,264 @@ macro_rules! impl_uri_from {
 impl_uri_from!(Origin);
 impl_uri_from!(Authority);
 impl_uri_from!(Absolute);
+
+mod uri_serde {
+    #![cfg(feature = "serde")]
+
```
SergioBenitez, line 331:
```suggestion
#[cfg(feature = "serde")]
mod uri_serde {
```

Place the attr on the `mod`.

core/http/Cargo.toml
```diff
@@ -58,3 +58,4 @@ features = ["std"]
 
 [dev-dependencies]
 rocket = { version = "0.5.0-dev", path = "../lib" }
+serde_json = "*"
```
SergioBenitez, line 61: Never use `*` dependencies. We _must_ use a specific version. Otherwise, our crate can be broken by upstream changes.

SergioBenitez (changes requested):

the10thWiz:
Thanks for the advice. I have been updating the PR to use serde_test, and I will be making the changes requested. I had not been aware of serde_test until now, so thanks for letting me know.

The odd formatting is likely due to my workflow, since I use rust analyser as my language server, and the function signatures in the traits were generated by it, thanks for pointing out the odd formatting.

The actual implementation for Serialize and Deserialize is entirely handled by the parse and display methods already implemented, so there really shouldn't be any issues.

Based on my testing, the following Uris are parsed as valid, but do not appear to be.

1 `://@example.com/test`
1 `://example.com:/test`
1 `://@example.com:/test`
1 `://example.com/test?`
1 `://example.com:/test?`
1 `://@example.com/test?`
1 `://@example.com:/test?`
1 `/http?`
1 `http:?`

The empty scheme in the first 7 is probably fine, but without the empty scheme most of them are rejected as invalid. I believe all of these should be rejected as invalid, but I'm not 100% sure. I will be adding the tests to the Uri parsing code in a seperate PR. I do need looking into what exactly the parser is outputing, since some of them might still be okay, even if they aren't exactly valid.

core/http/src/uri/uri.rs
the10thWiz: It actually is nessecary, at least with serde_test. The type passed to `assert_tokens` is the type it serializes to and deserialized from.

core/http/src/uri/uri.rs
the10thWiz: I will be making it `#[track_caller]`, thanks.

core/http/Cargo.toml
the10thWiz: I have changed it to 1.0, to match serde

the10thWiz:
I'm moving my Uri parsing thoughts over to #998, I think it's a better place to air them.

SergioBenitez:
> 8.  `/http?`

This is an origin URI with 1 segment `http` and an empty query.

> 9.  `http:?`

This is an absolute URI with scheme `http`, an empty `hier-part`, and an empty `query`.

> 1. `://@example.com/test`
> 2.  `://example.com:/test`
> 3.  `://@example.com:/test`
> 4.  `://example.com/test?`
> 5.  `://example.com:/test?`
> 6.  `://@example.com/test?`
> 7.  `://@example.com:/test?`

These are all invalid. Excellent find! A URI starting with `:` must be followed by a port, according to the spec. I even wrote this down:

https://github.com/SergioBenitez/Rocket/blob/078cf1725f5c7e50b45beeb1ec3e5e0e3a2cb0dc/core/http/src/parse/uri/spec.txt#L23-L25

Pushing a fix shortly.

SergioBenitez:
Please let me know when this is a ready for another review.

the10thWiz:
It should be ready for another review now.

core/http/src/uri/uri.rs
SergioBenitez: Why does that matter? All of the URI variants implement `Into<Uri>`. You could simply make `test_helper` generic on a `T: Into<Uri>` and pass in an `Origin` or `Authority` directly.

core/http/src/uri/uri.rs
```diff
@@ -316,3 +325,200 @@ macro_rules! impl_uri_from {
...
+        #[test]
+        fn serde_asterisk_form() {
+            test_helper("*", Uri::Asterisk);
+        }
+
+        // Commented out tests currently fail
+        #[test]
+        fn serde_invalid_uri() {
+            assert_de_tokens_error::<Uri<'static>>(&[Token::Str("@")], "unexpected EOF: expected some token at index 0");
```
SergioBenitez, line 505: I don't think we want to match on the exact error message. We already have tests for this in `uri.rs`. Simply check that parsing failed.

SergioBenitez (changes requested):
As I said before, the majority of these tests are checking parsing, not serde functionality, and they should be moved to the parsing tests file and use the existing infrastructure there. As you mentioned, we don't need to test much of the serde implementation: test that serialization serializes into a string, deserialization deserializes from a string, and that deserialization fails when we expect it to. That's about it.

core/http/src/uri/uri.rs
the10thWiz: I would have implemented it like that, however serde_test does require it, and I don't currently see a way around it. Please let me know if there is something I missed, but I don't think there is a way around this. I will be adding a comment to explain why and how the errors where created, but I don't think there is a way to avoid specifying them.

the10thWiz:
I have made a number of changes, primarily reducing testing. I did not feel that I should add the tests to the URI testing in this pull request, so I did not. If there is a way to assert that the URI parsing fails in serde_test without specifying the exact error message let me know, but I could not find one. It might be possible to use `#[should_panic]`, but I'm not sure it would be easy to read.

SergioBenitez:
Revisiting this, given the just merged URI overhaul: I think we should implement `Deserialize` and `Serialize` for all 5 `Uri` variants. I am not sure if it we should implement it for `Uri` itself; this does not seem particularly useful.

So, here's what I would love:

  * Implement `Serialize`/`Deserialize` for all 5 URI variants in their individual files under a `serde` flag.
  * Enable the `serde` flag for `http` in `core`.
  * Add an integration test in `core` or `http` which deserializes >= 1 one of each variants from a `toml` using figment. See the tests in `config/mod.rs`. Basically:
    ```rust
    #[derive(Deserialize)]
    struct MyUris<'r> {
        absolute: Absolute<'r>,
        origin: Origin<'r>,
    }
    
    jail.create_file("Rocket.toml", r#"
        [default]
        absolute = "http://rocket.rs"
        origin = "/foo/bar?baz"
    "#)?;
    
    let uris: MyUris<'_> = Config::figment().extract()?;
    assert_eq!(...)
    ```

That should be it.

the10thWiz:
I'm moving this pull request to #1649, and closing this one.

Issue #1612, opened 2021-04-20
Provide a convenience method to map Outcome<S,E,F> to Result<S,E>


DanielJoyce:
Right now, we have success_or which throws away E, and you must provide your own error value, or success_or_else which takes a function, but the function takes no parameters. In all cases, the error value is thrown away.

This would make it easier to interop with Result style apis, the ? operator, etc.

We can of course use `match` but that's verbose.

SergioBenitez:
Sure! Happy to accept a PR, if you could send one in!

DanielJoyce:
Yeah, I have something knocking about. I'll see if I can get something together.

SergioBenitez:
Closed in https://github.com/SergioBenitez/Rocket/commit/59851a621c7826f72fcfc71d14863ec1dc03d755.

PR #1622, opened 2021-04-28
Add methods and trait implementations to Outcome

the10thWiz:
Adds an implementation of IntoOutcome for Result, along with two methods
to convert a Result into an Outcome

This addresses #1612.

### DIFF
--- a/core/lib/src/outcome.rs
+++ b/core/lib/src/outcome.rs
@@ -148,6 +148,36 @@ impl<S, E, F> IntoOutcome<S, E, F> for Option<S> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


SergioBenitez (commented):
I don't think these implementation were quite right. The `IntoOutcome` impl didn't exist because it any such implementation would have to choose how to handle error/forward, and the choice seems rather arbitrary to me.

Methods named `as_` should not panic, and while `as_result_with_forward` is fine, I think we can make the implementation a bit more generic. See 59851a621c7826f72fcfc71d14863ec1dc03d755 for what I ultimately went with.

PR #1624, opened 2021-04-29
WIP: Implement auto-documenting routes

ThouCheese:
Very large WIP, does nothing for now.

This PR aims at eventually closing #297, or at least providing some of the requested functionality. For now I have created the traits `Documented` and `Undocumented`, and I have updated the derive macro to pull in the doc comments of the route handler function. I am opening an early PR to get some feedback.

Unresolved questions (non-exhaustive :)):

1. We will probably want users to be able to use their own documentation format: be it OpenAPI, or RAML, or something custom. How do we allow users to provide their own format? Which formats do we support out of the box?
2. Rocket should probably come with one or a couple of documentation frontends built in: Redoc, Swagger, Rapi-Doc, ..... Again the question is which ones do we support out of the box? Do we provide a mechanism where users can hook in and set up their own front end or is it sufficient to just make them serve the HTML for their front end with a `Route`.
3. What info do we need to collect for each route to be able to adequately document each route? How will we handle global settings (think authentication, maybe a server-spanning description, ...)


core/codegen/src/attribute/route/parse.rs
```diff
@@ -209,9 +212,23 @@ impl Route {
...
+            .filter(|a| a.path.is_ident("doc"))
+            .filter_map(|attr| attr.parse_meta().ok())
+            .filter_map(|meta| if let NameValue(MetaNameValue { lit: Str(string), .. }) = meta {
+                Some(string.value())
+            } else {
+                None
+            })
+            .flat_map(|string| string.split('\n').map(|s| s.trim().to_owned()).collect::<Vec<_>>())
+            .fold(String::new(), |acc, elem| acc + &elem + "\n");
+        let doc_comment = (!doc_comment.is_empty()).then(|| doc_comment);
```
SergioBenitez, line 227: Devise has utilities to parse attributes. Something like this should work:
```suggestion
        let docstring = String::from_attrs("doc", &handler.attrs)?.join("\n");
```

Do we actually need the new line, though? Or does it automatically get inserted when the docstring is converted into an attribute?

core/codegen/src/attribute/route/parse.rs
```diff
@@ -29,6 +30,8 @@ pub struct Route {
     pub handler: syn::ItemFn,
     /// The parsed arguments to the user's function.
     pub arguments: Arguments,
+    /// The doc comment describing this route
+    pub doc_comment: Option<String>
```
SergioBenitez, line 34:
```suggestion
    pub docstring: String,
```

Should this be `Option`? I think an empty `String` would be fine.

core/codegen/src/attribute/route/parse.rs
```diff
@@ -1,6 +1,7 @@
 use devise::{syn, Spanned, SpanWrapped, Result, FromMeta};
 use devise::ext::{SpanDiagnosticExt, TypeExt};
 use indexmap::{IndexSet, IndexMap};
+use syn::{Lit::Str, Meta::NameValue, MetaNameValue};
```
SergioBenitez, line 4:
```suggestion
```

core/codegen/src/attribute/route/mod.rs
```diff
@@ -353,6 +360,7 @@ fn codegen_route(route: Route) -> Result<TokenStream> {
                     format: #format,
                     rank: #rank,
                     sentinels: #sentinels,
+                    doc_comment: #doc_comment,
```
SergioBenitez, line 363:
```suggestion
                    docstring: #docstring,
```

core/codegen/src/attribute/route/mod.rs
```diff
@@ -319,6 +319,13 @@ fn codegen_route(route: Route) -> Result<TokenStream> {
     let rank = Optional(route.attr.rank);
     let format = Optional(route.attr.format.as_ref());
 
+    // Get the doc comment
+    let doc_comment = route
+        .doc_comment
+        .as_ref()
+        .map(|dc| quote! { Some(#dc.to_string()) })
+        .unwrap_or_else(|| quote! { None });
+
```
SergioBenitez, line 328:
```suggestion
    let docstring = &route.docstring;

```

core/codegen/src/attribute/route/parse.rs
```diff
@@ -209,9 +212,23 @@ impl Route {
...
+                None
+            })
+            .flat_map(|string| string.split('\n').map(|s| s.trim().to_owned()).collect::<Vec<_>>())
+            .fold(String::new(), |acc, elem| acc + &elem + "\n");
+        let doc_comment = (!doc_comment.is_empty()).then(|| doc_comment);
+
         diags.head_err_or(Route {
             attr, path_params, query_params, data_guard, request_guards,
-            handler, arguments,
+            handler, arguments, doc_comment
```
SergioBenitez, line 231:
```suggestion
            handler, arguments, docstring,
```

core/lib/src/route/route.rs
```diff
@@ -190,6 +190,8 @@ pub struct Route {
     pub format: Option<MediaType>,
     /// The discovered sentinels.
     pub(crate) sentinels: Vec<Sentry>,
+    /// The doc comment associated with this route.
+    pub(crate) doc_comment: Option<String>,
```
SergioBenitez, line 194:
```suggestion
    /// The route's docstring, which may be empty.
    pub(crate) docstring: String,
```

core/lib/src/route/route.rs
```diff
@@ -253,6 +255,7 @@ impl Route {
             sentinels: Vec::new(),
             handler: Box::new(handler),
             rank, uri, method,
+            doc_comment: None,
```
SergioBenitez, line 258:
```suggestion
            doc_comment: String::new(),
```

Note that `String::new()` is guaranteed not to allocate.

core/lib/src/route/route.rs
```diff
@@ -345,6 +348,9 @@ pub struct StaticInfo {
     /// Route-derived sentinels, if any.
     /// This isn't `&'static [SentryInfo]` because `type_name()` isn't `const`.
     pub sentinels: Vec<Sentry>,
+    /// The doc comment associated with this route.
+    pub doc_comment: Option<String>,
```
SergioBenitez, line 352:
```suggestion
    /// The route's docstring, which may be empty.
    pub docstring: String,
```

core/lib/src/route/route.rs
SergioBenitez: Once we understand better what we really want to collect from a route, this `docstring: String` field will change to something semantically meaningful, derived from having parsed the docstring as needed as well as having collected information from the guards and responder.

SergioBenitez (changes requested):
Super excited for this. Some simplifications and forward-looking statements.

Thanks for submitting this early!

SergioBenitez:
> * We will probably want users to be able to use their own documentation format: be it OpenAPI, or RAML, or something custom. How do we allow users to provide their own format? Which formats do we support out of the box?

I think we have Rocket act as a data provider wherein Rocket allows _any_ consumer to access all of the available documentation information. The information should be rich enough (richer, ideally!) to allow shipping standard doc consumers in Rocket directly, such as those for `OpenAPI` and `RAML`, while also allowing external consumers for custom formats.

Effectively, Rocket's core should be unaware of any standard like `OpenAPI` or `RAML` but provide an interface rich enough to allow extracting into such standards.

> * Rocket should probably come with one or a couple of documentation frontends built in: Redoc, Swagger, Rapi-Doc, ..... Again the question is which ones do we support out of the box? Do we provide a mechanism where users can hook in and set up their own front end or is it sufficient to just make them serve the HTML for their front end with a `Route`.

As above, Rocket itself should be agnostic to doc frontends as it should be for doc standards. We might consider shipping a fairing or custom handler or something of that sort of a few front-ends, whoever.

In general, I envision the following pipeline:

```
[-------------------    -------------------------------    ------------
| Rocket Doc Data | -> | Standardized Representation | -> | Frontend |
-------------------    -------------------------------    ------------
```

Where Rocket's core participates _only_ in the first stage and everything else can be implemented externally. Standards like `OpenAPI` and `RAML` live in the second box while frontends like Swagger and Redoc live in the third. You might even have just one stage after the first, or more than two; totally up to the consumer.

> * What info do we need to collect for each route to be able to adequately document each route? How will we handle global settings (think authentication, maybe a server-spanning description, ...)

If global, auth will be a fairing, which implies fairings should be able to add docs. If local, auth will be a guard, which we already know should be able to add docs. It would be nice to be able to read the crate docstring, but that's not accessible easily. I wonder if an attribute on a docstring works:

```rust
#[rocket(doc)]
/// Some docs to make global, or something.
```

This could also be a setting/configuration in the doc exporter as apposed to something Rocket knows about. In any case, I think we punt on this until much later.

core/codegen/src/attribute/route/parse.rs
ThouCheese: I want to later detect whether the first n lines start with a pound sign, and then use those as a title attribute, so the newline may not (yet) be lost at this stage.

SergioBenitez: No, it simply gives you the raw string literal value.

I think we should preserve the docstring as-is, for now. Any muxing can be done at a later stage or when we start making semantic sense of the string. I believe the string literals will already have newlines; I'm not sure there can be more than one. I would play with this. Try routing out some single line, multi line docstrings using `///` and using a raw attribute. mu

ThouCheese:
I have created some types that can do some of the lifting when documenting a type (I'm 100% sure that they are not sufficient even for just OpenAPI, but lets start small). I think the next steps should be to modify Rocket to collect and aggregate the documentation that we have now gathered for each route and to create a macro that can derive the `Schema` trait for responders and "requesters".

SergioBenitez (changes requested):
The different files for the different impls are unnecessary; they can and should all be in one file. Eventually, we want to dedup these by using macros, too.

In terms of `Schema`, we should start with structures that are able to describe the entirety of the [serde data model](https://serde.rs/data-model.html), though several of these can be coalesced when reified. I wouldn't be surprised if structures that do so already exist somewhere. There's [`Content`](https://github.com/serde-rs/serde/blob/6366f17da719a0efa5117c780fdd5469b903929d/serde/src/private/de.rs#L195), which is a serialized version of this, and Figment's [Value](https://docs.rs/figment/0.10.5/figment/value/enum.Value.html) and this [`Value`](https://docs.rs/serde-value/0.7.0/serde_value/enum.Value.html), which is similar but excludes `Option`, which we shouldn't. So, effectively something like these but without the actual values.

I don't think `Schema` here makes too much sense. `min`/`max` don't make sense for many types, for example. I think any sort of value spec should be per-value.

In any case, the next step is to actually grab the schema for the route types much like is done now for sentinels minus the part about embeddings.

core/codegen/src/attribute/route/parse.rs
```diff
@@ -209,9 +211,11 @@ impl Route {
             })
             .collect();
 
+        let docstring = String::from_attrs("doc", &handler.attrs)?.join("\n");
```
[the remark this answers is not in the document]
ThouCheese: I believe we need it, because Rust comments are markdown, which is also supported for OpenApi and RAML. Markdown is a newline sensitive format if you will: `# title \n remainder\n` is not the same as `# title remainder`.

ThouCheese:
> The different files for the different impls are unnecessary

Gone

> we should start with structures that are able to describe the entirety of the serde data model

I have implemented HasSchema for every type in the serde data model, except for `unit_struct`, `unit_variant`,  `newtype_struct`, `newtype_variant`, `tuple_struct`, `tuple_variant`, `struct` and `struct`, because it is not possible to do a generic implementation for those.

> I think any sort of value spec should be per-value.

I agree that some of the spec must come from the value, but some of it can also come from the type. Many auto documenters provide defaults like `true` and `"string"` if no default is specified by the user. I think we should strive to do something similar.

> In any case, the next step is to actually grab the schema for the route types much like is done now for sentinels minus the part about embeddings.

So would this mean adding a field to StaticInfo with the ingoing types and the outgoing types?

SergioBenitez:
> > The different files for the different impls are unnecessary
> 
> Gone
> 
> > we should start with structures that are able to describe the entirety of the serde data model
> 
> I have implemented HasSchema for every type in the serde data model, except for `unit_struct`, `unit_variant`, `newtype_struct`, `newtype_variant`, `tuple_struct`, `tuple_variant`, `struct` and `struct`, because it is not possible to do a generic implementation for those.

What I'm suggesting is that we have a struct/enum that we can use to dynamically determine the serde data type of a given Rust type. I don't see where we would need generics for this.

> 
> > I think any sort of value spec should be per-value.
> 
> I agree that some of the spec must come from the value, but some of it can also come from the type. Many auto documenters provide defaults like `true` and `"string"` if no default is specified by the user. I think we should strive to do something similar.

What I mean is that different data types have different "specs", so we shouldn't have one "spec" kind that applies to every data type.

> > In any case, the next step is to actually grab the schema for the route types much like is done now for sentinels minus the part about embeddings.
> 
> So would this mean adding a field to StaticInfo with the ingoing types and the outgoing types?

Not quite. See how this is done with `Sentinel`s and the internal `Sentry` type. We should only add _one_ field to `Route`.

core/codegen/src/attribute/route/mod.rs
```diff
@@ -296,6 +296,38 @@ fn sentinels_expr(route: &Route) -> TokenStream {
...
+        syn::ReturnType::Default => None,
+        syn::ReturnType::Type(_, ref ty) => Some(ty.with_stripped_lifetimes())
+    };
+
+    let generic_idents: Vec<_> = route.handler.sig.generics
+        .type_params()
+        .map(|p| &p.ident)
+        .collect();
+
+    let eligible_types = route.guards()
```
SergioBenitez, line 310: We don't need any of this. We simply need the top-level types passed to the resolve macro. This only exists in sentinels due to embedded discovery.

core/codegen/src/attribute/route/mod.rs
```diff
@@ -296,6 +296,38 @@ fn sentinels_expr(route: &Route) -> TokenStream {
...
+        .map(|guard| &guard.ty)
+        .chain(ret_ty.as_ref().into_iter())
+        .flat_map(|ty| ty.unfold())
+        .filter(|ty| ty.is_concrete(&generic_idents))
+        .map(|child| (child.parent, child.ty));
+
+    let sentinel = eligible_types.map(|(parent, ty)| {
```
SergioBenitez, line 317: Perhaps this should be renamed. :)

core/codegen/src/attribute/route/mod.rs
```diff
@@ -319,6 +352,9 @@ fn codegen_route(route: Route) -> Result<TokenStream> {
     let rank = Optional(route.attr.rank);
     let format = Optional(route.attr.format.as_ref());
 
+    // Get the doc comment
+    let docstring = &route.docstring;
```
SergioBenitez, line 356: No need to separate this. Group it with the rest.

core/codegen/src/attribute/route/parse.rs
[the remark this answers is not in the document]
SergioBenitez: My question is really where rustdoc already inserts a new line or whether it infers one based on having multiple attributes. You should be able to easily check. ￼￼￼We only need the new line here if it is inferred.￼

core/lib/src/doc/has_schema.rs
```diff
@@ -0,0 +1,318 @@
...
+    Map,
+    List,
+    String,
+    Num,
+    Int,
+    Bool,
+    Set,
+}
+
+pub struct TypeSchema<T> {
```
SergioBenitez, line 12: This doesn't make much sense to me.￼ `name` seems to be a string serialization of the kind enum, so why have it at all? Or alternatively, why not have the trait simply have a const string?￼￼

I think we need to start thinking more seriously about how schema is going to be represented.￼ Otherwise, we will be bouncing back-and-forth between ideas without justification. A good approach here is the take a reasonably sized application and imagine the exact documentation we would like to render and work backwards from there. We also have the benefit of existing schema like JSON schema and so on. ￼

SergioBenitez (changes requested):
At the moment, progress here seems a bit without aim. Instead of working forward in an ambiguous direction, I'd like for us to set an ideal target. What would be the gold standard here? The answer should include:
- What the user has to do get docs.
- What the API will look like to get docs.
- How the user can progressively document their app. 

It would be ideal to work with a non-trivial example. Perhaps should a real API, like the Bitwarden API or perhaps the stripe API, imagine how endpoints might be written in Rocket, and then design a doc system that would allow rendering those kinds of docs as streamlined as possible.

core/codegen/src/attribute/route/mod.rs
ThouCheese: Do you mean add it to the `StaticInfo::schemas` field?

SergioBenitez: Sorry, I just need style wise. Basically to just remove the whitespace of and comment.

ThouCheese: Right okay

PR #1625, opened 2021-04-30
Update Template doc

rotoclone:
Based on the discussion on https://github.com/SergioBenitez/Rocket/pull/1614, I added a note to the `Template` doc clarifying that it does not contain the rendered template.

### DIFF
--- a/contrib/lib/src/templates/mod.rs
+++ b/contrib/lib/src/templates/mod.rs
@@ -146,6 +146,7 @@ use rocket::http::{ContentType, Status};
 const DEFAULT_TEMPLATE_DIR: &str = "templates";
 
 /// Responder that renders a dynamic template.
+/// Note that this only contains metadata, not the actual rendered template.
 ///
 /// # Usage
 ///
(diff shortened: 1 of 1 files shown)


contrib/lib/src/templates/mod.rs
```diff
@@ -146,6 +146,7 @@ use rocket::http::{ContentType, Status};
 const DEFAULT_TEMPLATE_DIR: &str = "templates";
 
 /// Responder that renders a dynamic template.
+/// Note that this only contains metadata, not the actual rendered template.
```
jebrosen, line 149: I think this needs another line break before it, otherwise it will be merged with the line before. This can be confirmed with `cargo doc -p rocket_contrib --open`; the short description should look the same as it is now on the very bottom of the page at <https://api.rocket.rs/v0.4/rocket_contrib/templates/index.html>.

rotoclone: Ah, yeah: https://github.com/SergioBenitez/Rocket/pull/1625/commits/73b95086f6bc1881b418936fdbd29174ffc7b584

SergioBenitez:
Awesome. Merged in bacbf073ceeba1d7b7dc83f4d20b4a1d1ac09ab3. Thank you!

PR #1636, opened 2021-05-12
Update subhead to 'Cooperative Multitasking'

stuarth:
https://rocket.rs/master/guide/state/#usage references "Cooperative Multitasking", but its heading doesn't match. Believe the right resolution is renaming the Multitasking header.

Note: Haven't verified the fix works!


SergioBenitez:
I decided to invert your change but nevertheless merged in 57f27730e7ce5f443cf298b3f3a9c6392a2df507. Thank you!

PR #1638, opened 2021-05-13
`context` macro for creating rendering contexts

Flying-Toast:
Adds a macro that can be used when rendering templates, like so:
```rust
#[get("/")]
fn index() -> Template {
    Template::render("index", context!{
        foo: "Hello world",
        bar: 83,
    })
}
```

Which is equivalent to:
```rust
#[derive(Serialize)]
struct IndexContext<'a> {
    foo: &'a str,
    bar: u64,
}

#[get("/")]
fn index() -> Template {
    Template::render("index", IndexContext {
        foo: "Hello world",
        bar: 83,
    })
}
```

### DIFF
--- a/contrib/dyn_templates/src/lib.rs
+++ b/contrib/dyn_templates/src/lib.rs
@@ -26,7 +26,7 @@
 //!
 //!      ```rust
 //!      # #[macro_use] extern crate rocket;
-//!      use rocket_dyn_templates::Template;
+//!      use rocket_dyn_templates::{Template, context};
 //!
 //!      #[launch]
 //!      fn rocket() -> _ {
@@ -35,8 +35,7 @@
 //!
 //!      #[get("/")]
 //!      fn index() -> Template {
-//!          # let context = ();
-//!          Template::render("template-name", &context)
+//!          Template::render("template-name", context! { })
 //!      }
 //!      ```
 //!
@@ -119,7 +118,8 @@
 //! Templates are rendered with the `render` method. The method takes in the
 //! name of a template and a context to render the template with. The context
 //! can be any type that implements [`Serialize`] and would serialize to an
-//! `Object` value.
+//! `Object` value. The [`context!`] macro can also be used to create inline
+//! `Serialize`-able context objects.
 //!
 //! ## Automatic Reloading
 //!
@@ -166,6 +166,9 @@ use self::context::{Context, ContextManager};
 
 use std::borrow::Cow;
 use std::path::PathBuf;
+
+#[doc(hidden)]
+pub use rocket::serde;
 
 use rocket::{Rocket, Orbit, Ignite, Sentinel};
 use rocket::request::Request;
@@ -301,10 +304,24 @@ impl Template {
     }
 
     /// Render the template named `name` with the context `context`. The
-    /// `context` can be of any type that implements `Serialize`. This is
-    /// typically a `HashMap` or a custom `struct`.
+    /// `context` is typically created using the [`context!`] macro, but it can
+    /// be of any type that implements `Serialize`, such as `HashMap` or a
+    /// custom `struct`.
     ///
-    /// # Example
+    /// # Examples
+    ///
+    /// Using the `context` macro:
+    ///
+    /// ```rust
+    /// use rocket_dyn_templates::{Template, context};
+    ///
+    /// # #[allow(unused_variables)]
+    /// let template = Template::render("index", context! {
+    ///     foo: "Hello world",
+    /// });
+    /// ```
+    ///
+    /// Using a `HashMap` as the context:
     ///
     /// ```rust
     /// use std::collections::HashMap;
@@ -434,4 +451,106 @@ impl Sentinel for Template {
 
         false
     }
+}
+
+/// A macro to easily create a template rendering context.
+///
+/// Invocations of this macro expand to a value of an anonymous type which
+/// implements [`serde::Serialize`]. Fields can be literal expressions or
+/// variables captured from a surrounding scope, as long as all fields implement
+/// `Serialize`.
+///
+/// # Examples
+///
+/// The following code:
+///
+/// ```rust
+/// # #[macro_use] extern crate rocket;
+/// # use rocket_dyn_templates::{Template, context};
+/// #[get("/<foo>")]
+/// fn render_index(foo: u64) -> Template {
+///     Template::render("index", context! {
+///         // Note that shorthand field syntax is allowed.
+///         // This is the same as `foo: foo,`
+///         foo,
+///         bar: "Hello world",
+///     })
+/// }
+/// ```
+///
+/// is equivalent to the following, but without the need to manually define an
+/// `IndexContext` struct:
+///
+/// ```rust
+/// # use rocket_dyn_templates::Template;
+/// # use rocket::serde::Serialize;
+/// # use rocket::get;
+/// #[derive(Serialize)]
+/// # #[serde(crate = "rocket::serde")]
+/// struct IndexContext<'a> {
+///     foo: u64,
+///     bar: &'a str,
+/// }
+///
+/// #[get("/<foo>")]
+/// fn render_index(foo: u64) -> Template {
+///     Template::render("index", IndexContext {
+///         foo,
+///         bar: "Hello world",
+///     })
+/// }
+/// ```
+///
+/// ## Nesting
+///
+/// Nested objects can be created by nesting calls to `context!`:
+///
+/// ```rust
+/// # use rocket_dyn_templates::context;
+/// # fn main() {
+/// let ctx = context! {
+///     planet: "Earth",
+///     info: context! {
+///         mass: 5.97e24,
+///         radius: "6371 km",
+///         moons: 1,
+///     },
+/// };
+/// # }
+/// ```
+#[macro_export]
+macro_rules! context {
+    ($($key:ident $(: $value:expr)?),*$(,)?) => {{
+        use $crate::serde::ser::{Serialize, Serializer, SerializeMap};
+        use ::std::fmt::{Debug, Formatter};
+
+        #[allow(non_camel_case_types)]
+        struct ContextMacroCtxObject<$($key: Serialize),*> {
+            $($key: $key),*
+        }
+
+        #[allow(non_camel_case_types)]
+        impl<$($key: Serialize),*> Serialize for ContextMacroCtxObject<$($key),*> {
+            fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+                where S: Serializer,
+            {
+                let mut map = serializer.serialize_map(None)?;
+                $(map.serialize_entry(stringify!($key), &self.$key)?;)*
+                map.end()
+            }
+        }
+
+        #[allow(non_camel_case_types)]
+        impl<$($key: Debug + Serialize),*> Debug for ContextMacroCtxObject<$($key),*> {
+            fn fmt(&self, f: &mut Formatter<'_>) -> ::std::fmt::Result {
+                f.debug_struct("context!")
+                    $(.field(stringify!($key), &self.$key))*
+                    .finish()
+            }
+        }
+
+        ContextMacroCtxObject {
+            $($key $(: $value)?),*
+        }
+    }};
 }
--- a/contrib/dyn_templates/tests/templates.rs
+++ b/contrib/dyn_templates/tests/templates.rs
@@ -4,7 +4,9 @@ use std::path::{Path, PathBuf};
 
 use rocket::{Rocket, Build};
 use rocket::config::Config;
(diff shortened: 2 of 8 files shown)


SergioBenitez:
@jebrosen What do you think?

jebrosen (commented):
I don't think this is something we need to re-implement within Rocket: `serde_json::json!` (or `rocket_contrib::json!`) can be used in almost exactly this way:

```rust
#[rocket::get("/")]
fn index() -> Template {
    Template::render("index", rocket_contrib::json!({
        "foo": { "name": "Hello world" },
        "bar": 83,
    }))
}
```

```jinja2
<p><strong>foo.name:</strong> {{ foo.name }}</p>
<p><strong>bar + 10</strong> = {{ bar + 10 }}</p>
```

It additionally supports nested objects and arrays, and it is the actual object model used by both template engines Rocket supports. My understanding is that it's soon going to be even easier to access that from `rocket` than it is today since `rocket_contrib::json` is likely to be moved into `rocket` core.

Although I personally prefer more strictly-typed templates such as the ones in the current examples, the `json!`-style macros are pretty convenient sometimes and I think this would be a good method to show in a few examples and/or the guide.

Flying-Toast:
Okay, sounds good

SergioBenitez:
I've been thinking about this more, and I think this is far too convenient to leave on the table. Using `json!` is both less performant and requires enabling a feature and thus importing dependencies you may not need.

@Flying-Toast If you're still okay with pushing this forward, could you rebase (likely just redo this on) master? Also, if you can manage some `macro_rules!`-foo, I think it's be _fantastic_ if the macro could support short-hand field syntax:

```rust
let foo = 1;
let bar = 2;
context! { foo, bar, baz: 10 };
```

Flying-Toast:
Sure! I just pushed a rebase and also added a pretty simple change to enable shorthand field syntax.

Nesting does work, by nesting calls to `context!` (albeit not as nice as `json!`'s built-in nesting syntax):
```rust
let foo = "abcdefg";
context! {
    foo,
    inner: context! {
        value: 22,
    },
}
```
Any thoughts?

SergioBenitez:
> @SergioBenitez Sure! I just pushed a rebase and also added a pretty simple change to enable shorthand field syntax.
> 
> 
> 
> Nesting does work, by nesting calls to `context!` (albeit not as nice as `json!`'s built-in nesting syntax):
> 
> ```rust
> 
> let foo = "abcdefg";
> 
> context! {
> 
>     foo,
> 
>     inner: context! {
> 
>         value: 22,
> 
>     },
> 
> }
> 
> ```
> 
> Any thoughts?

I think this is totally fine! Let's make sure we have an example to this effect? We should also use the shorthand syntax in an example. Finally, we need a few tests somewhere checking that the macro is doing what we think it's doing. One very complex invocation (and a few exceedingly simple ones, like an empty one, etc.) should do it.

contrib/dyn_templates/src/lib.rs
```diff
@@ -166,6 +168,9 @@ pub use self::metadata::Metadata;
 use self::fairing::TemplateFairing;
 use self::context::{Context, ContextManager};
 
+// used for `context` macro
+#[doc(hidden)]
+pub use serde as serde_internal;
```
SergioBenitez, line 173: We shouldn't need this. See my other comment.

contrib/dyn_templates/src/lib.rs
```diff
@@ -439,3 +459,70 @@ impl Sentinel for Template {
...
+///     })
+/// }
+/// ```
+#[macro_export]
+macro_rules! context {
+    ($($key:ident $(: $value:expr)?),*$(,)?) => {{
+        use $crate::serde_internal::ser::{Serialize, Serializer, SerializeMap};
+
+        #[allow(non_camel_case_types)]
+        struct AnonymousInnerContextStruct<$($key: Serialize),*> {
```
SergioBenitez, line 509: We should have some UI tests for when values aren't Serialize. I think having this bound here might result in suboptimal error messages. It should almost be unnecessary.

contrib/dyn_templates/src/lib.rs
```diff
@@ -439,3 +459,70 @@ impl Sentinel for Template {
...
+                where S: Serializer,
+            {
+                let mut map = serializer.serialize_map(None)?;
+                $(map.serialize_entry(stringify!($key), &self.$key)?;)*
+                map.end()
+            }
+        }
+
+        AnonymousInnerContextStruct {
+            $($key $(: &$value)?),*
```
SergioBenitez, line 525: Interesting, not to take ownership. I'm not sure if this is what we want. Can you argue one direction or another? The only argument I have for taking ownership is that not doing so might be surprising.￼

Perhaps more concerning is that ownership is taken when shorthand syntax is used but not when the full syntax is used.￼

SergioBenitez (changes requested):

contrib/dyn_templates/src/lib.rs
Flying-Toast: My reasoning for not taking ownership was just following in line with e.g. `println!`. Plus, taking ownership when it's not needed seems unnecessarily limiting. I agree that the differing behavior with the shorthand syntax is bad, I didn't catch that one earlier.

I'm all for borrowing (& fixing the shorthand syntax) unless you think ownership is the better path?

contrib/dyn_templates/src/lib.rs
```diff
@@ -439,3 +459,70 @@ impl Sentinel for Template {
...
+///     Template::render("index", IndexContext {
+///         foo: "Hello world",
+///         bar: param,
+///     })
+/// }
+/// ```
+#[macro_export]
+macro_rules! context {
+    ($($key:ident $(: $value:expr)?),*$(,)?) => {{
+        use $crate::serde_internal::ser::{Serialize, Serializer, SerializeMap};
```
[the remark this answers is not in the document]
Flying-Toast: I'm not seeing a reexported `serde_` anywhere. There's `::rocket::serde` though - is that one good?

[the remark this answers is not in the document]
SergioBenitez: Sorry, I confused this with another crate. I would suggest adding it as an export without a rename:

```rust
#[doc(hidden)]
pub use serde;
```

contrib/dyn_templates/src/lib.rs
```diff
@@ -321,6 +329,18 @@ impl Template {
...
     /// ```
+    ///
+    /// Using the `context` macro:
+    ///
+    /// ```rust
+    /// #[macro_use] extern crate rocket_dyn_templates;
+    /// use rocket_dyn_templates::Template;
+    ///
+    /// # #[allow(unused_variables)]
+    /// let template = Template::render("index", context!{
```
SergioBenitez, line 340:
```suggestion
    /// let template = Template::render("index", context! {
```

Can we make this style change elsewhere, too?

contrib/dyn_templates/src/lib.rs
SergioBenitez: I think ownership is the more obvious path. `println!` doesn't generate a value, so its approach in sensible there. Given we _do_ generate a value, I don't think borrowing by default makes sense. It would make it possible, for instance, to do something like:

```rust
fn f() -> impl Serialize {
    let foo = String::new();
    context! { foo }
}
```

If we want to get creative, we can introduce some new shorthand syntax (which I've quickly grown fond of):

```rust
let foo = String::new();
context! { ref foo } == context! { foo: &foo }
```

Or just leave it as is.

Flying-Toast: Good point, makes sense. I just changed the macro to take ownership of fields.
I think we should move forward with the syntax as-is for a MVP, and we can possibly add `ref` later on after we get a feel for how the macro is being used.

contrib/dyn_templates/src/lib.rs
Flying-Toast: Removing that trait bound allows code to compile if values *aren't* `Serialize`, as long as the `Serialize` impl of the returned value isn't used. i.e. this compiles fine:
```rust
enum NonSerialize { A, B, C }

let ctx = context! {
    foo: NonSerialize::A,
};
```
but not this:
```rust
enum NonSerialize { A, B, C }

let ctx = context! {
    foo: NonSerialize::A,
};
Template::render("index", ctx);
```
As of now, the error messages are actually quite good:
```
error[E0277]: the trait bound `NonSerialize: Serialize` is not satisfied
  --> src/main.rs:18:11
   |
18 |   let ctx = context! {
   |  ___________^
19 | |     foo: NonSerialize::A,
20 | | };
   | | ^
   | | |
   | |_the trait `Serialize` is not implemented for `NonSerialize`
   |   required by `ContextMacroCtxObject`
```

I'll add some ui tests. Where/how should I add them?

site/guide/5-responses.md
```diff
@@ -441,6 +441,24 @@ a template and a context to render the template with. The context can be any
 type that implements `Serialize` and serializes into an `Object` value, such as
 structs, `HashMaps`, and others.
 
+You can also use the [`context`] macro to create ad-hoc context objects, without
+the need to define a separate struct:
```
SergioBenitez, line 445:
```suggestion
You can also use [`context!`] to create ad-hoc context objects without defining a new type:
```

contrib/dyn_templates/tests/templates.rs
```diff
@@ -98,6 +100,84 @@ fn test_sentinel() {
...
+            value_b: u8,
+        }
+
+        let a = true;
+        let value_b = 123;
+        let outer_string = String::from("abc 123");
+
+        assert_same_object!(
+            context! {
+                s: outer_string.clone(),
```
SergioBenitez, line 160: Can we make this `&outer_string`, just to make sure it works?

SergioBenitez (changes requested):
This is looking really good. Can we add a few more test?

  * Ensuring the object owns its data. (likely a compile_fail test)
  * Ensuring we can store borrows.
  * I'd like to see UI tests. What does using the same field name twice look like? What does sending in a value that isn't `Serialize` look like?
  * Ensuring we can return the struct as `impl Serialize` or similar.

Flying-Toast:
Okay, I added some more tests as you requested. Anything else?

SergioBenitez:
Yes, absolutely! Anywhere we use a HashMap in examples now we should instead use this macro.

Flying-Toast:
Awesome. Everything should be all set now.

SergioBenitez:
Can you rebase on `master`? It looks like the commit bases have diverged.

Flying-Toast:
> Can you rebase on `master`? It looks like the commit bases have diverged.

Done.

Flying-Toast:
Looks like some changes in the rebase broke the build, I'll fix those now.

Flying-Toast:
@SergioBenitez `rocket_dyn_templates` no longer has `serde` and `serde_json` as dependencies, but they're needed by the crate's tests. Should I add them as dev-dependencies?

SergioBenitez:
No. Nothing in `dyn_templates` should depend on `serde_json`. `serde` should be depended on via `rocket::serde`.

contrib/dyn_templates/src/lib.rs
```diff
@@ -435,3 +452,104 @@ impl Sentinel for Template {
...
+///         bar: "Hello world",
+///     })
+/// }
+/// ```
+///
+/// is equivalent to the following, but without the need to manually define an
+/// `IndexContext` struct:
+///
+/// ```rust
+/// # #[macro_use] extern crate rocket;
```
SergioBenitez, line 485: This is unnecessary.

contrib/dyn_templates/src/lib.rs
```diff
@@ -435,3 +452,104 @@ impl Sentinel for Template {
...
+/// # use rocket_dyn_templates::Template;
+/// # use serde::Serialize;
```
SergioBenitez, line 487:
```suggestion
/// # use rocket::serde::Serialize;
```

And then you'll need the serde(crate) attribute.

SergioBenitez (changes requested):

Flying-Toast:
> No. Nothing in `dyn_templates` should depend on `serde_json`. `serde` should be depended on via `rocket::serde`.

What deserializer should I use in the tests?
```rust
macro_rules! assert_same_object {
    ($ctx:expr, $obj:expr $(,)?) => {{
        let ser_ctx = serde_json::to_string(&$ctx).unwrap();
        let deser_ctx = serde_json::from_str(&ser_ctx).unwrap();
        assert_eq!($obj, deser_ctx);
    }};
}
```

SergioBenitez:
Figment's Value.

Flying-Toast:
All set.

SergioBenitez:
This is awesome. Stellar job, @Flying-Toast. Merged in 8e3ad40beb6a14e606d22065ae537eea2378b79b.

PR #1643, opened 2021-05-19
Move rocket 0.4 to try_trait_v2

scottmcm:
I can't get the whole thing to link, because it's not finding sqlite3.lib, but otherwise I think this is right.


scottmcm:
You're too fast, @SergioBenitez -- <https://github.com/SergioBenitez/Rocket/commit/425f741b873424046e20855f3fe13a3eb01b0c5e> got in between when I forked and finishing the PR 😆

PR #1645, opened 2021-05-19
Fix the types in one of the `FromResidual` implementations

scottmcm:
Without this change, `?` on `Result` doesn't actually work as intended.

The compiler error from the new test I added, before fixing the types:
```rust
error[E0277]: the `?` operator can only be used in a function that returns `Result` or `Option` (or another type that implements `FromResidual`)
   --> core\lib\src\outcome.rs:665:25
    |
664 | /         fn demo() -> Outcome<i128, String, f32> {
665 | |             Err("hello")?;
    | |                         ^ cannot use the `?` operator in a function that returns `outcome::Outcome<i128, std::string::String, f32>`
666 | |             unreachable!()
667 | |         }
    | |_________- this function should return `Result` or `Option` to accept `?`
    |
    = help: the trait `FromResidual<std::result::Result<Infallible, &str>>` is not implemented for `outcome::Outcome<i128, std::string::String, f32>`
    = note: required by `from_residual`
```

And while I was there, I tweaked the code to remove the need for some `#[allow]`s.

### DIFF
--- a/core/lib/src/outcome.rs
+++ b/core/lib/src/outcome.rs
@@ -657,6 +657,26 @@ impl<S, E, F> fmt::Display for Outcome<S, E, F> {
 mod test {
     use super::Outcome::{self, *};
 
+    macro_rules! fake_try {
+        ($e:block) => {
+            (||{
+                std::ops::Try::from_output($e)
+            })()
+        }
+    }
+
+    #[test]
+    fn outcome_try_trait() {
+        let r: Outcome<u16, String, f64> = fake_try! {{ 3 }};
+        assert_eq!(r, Success(3));
+        let r: Outcome<u16, String, f64> = fake_try! {{ Success::<_, &'static str, f32>(3)? }};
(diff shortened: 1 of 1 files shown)


SergioBenitez:
Ah, that's interesting; I assumed the std-lib impl used `!`, but that's clearly not the case. Thanks for catching this.

SergioBenitez:
Could you add a test for `Outcome` propagation, forward and failure, as well? Ideally we'd check that the variant and values match.

scottmcm:
Hopefully the current work to stabilize `!` will make it finally stick, and then it will, but for now it doesn't.  (The plan is for `Infallible` to become a type alias for `!` at that point, so people don't have to migrate.)

Context for why, if you're curious: https://github.com/rust-lang/rust/pull/84092#discussion_r616934554

SergioBenitez:
Awesome. Merged in f2a56f422257b7239b6aec475ec749ca9b75c765.

PR #1646, opened 2021-05-20
Allow reading a TempFile without persisting it

arnodb:
This PR implements an old suggestion to make `TempFile` implement `AsyncRead`.

Technically it is an indirect implementation of `AyncRead`, and it allows conversion to `'static` lifetimes via `into_owned`.

If eventually one calls `into_owned` in the middle of reading operations (which is strongly discouraged though), then the reading continues where it was left, not at the beginning of the file.

### DIFF
--- a/core/lib/src/data/mod.rs
+++ b/core/lib/src/data/mod.rs
@@ -14,6 +14,6 @@ pub use self::from_data::{FromData, Outcome};
 pub use self::limits::Limits;
 pub use self::capped::{N, Capped};
 pub use ubyte::{ByteUnit, ToByteUnit};
-pub use temp_file::TempFile;
+pub use temp_file::{TempFile, TempFileRead};
 
 pub(crate) use self::data_stream::StreamReader;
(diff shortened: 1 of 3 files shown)


SergioBenitez:
I think we should implement an`open()` method on `TempFile` that returns an `impl AsyncRead`. The concrete type should likely be an `Either<File, &'v str>` from https://docs.rs/tokio-util/0.6.7/tokio_util/either/enum.Either.html. Then we don't need any manual implementation of `AsyncRead` nor any intermediary type.

arnodb:
In the end I don't need it. Two options:

- I reimplement in the suggested way and we manage to merge this PR if you want it
- close and forget the PR

This will be your call.

SergioBenitez:
For posterity: this was closed because we discovered, on Matrix, that the ultimate intent was to store the file upload on Azure. Using `TempFile` as an intermediary was inefficient and unnecessary, precisely the reason why `TempFile` does not and likely will not implement `AsyncRead`.

PR #1649, opened 2021-05-22
Uri serde Implementation

the10thWiz:
This is a continuation of #1621, since I started over on a separate branch. This should be nearly complete, I don't expect any major changes to happen.

A few notes:

- I've implemented Serialize and Deserialize for all 5 Uri variants, and added documentation noting the implementation. I felt it was important to include some information about `#[serde(borrow)]`, since it took me a while to figure that out for the test.
- I believe the `serde` flag for `http` in `core` is already enabled, let me know if I'm wrong about that.
- I added the integration test in `core`, since Figment isn't a depedency in `http`.

I agree that implementing `Deserialize` on `Uri` itself isn't particularly useful. I did however implement `Serialize` for `Uri`, since there are less issues there. I'm not sure if this will actually be used, but I didn't think it would cause any issues. I did add documentation to `Uri` noting the implementation, and recommending `Reference` for the generic usecase.

I also added `parse_owned` implementations to the types that didn't have an implementation. I don't think there should be any issues with my implementation, but I'm not particularly familiar with the parser itself, so let me know.

### DIFF
--- a/core/http/src/uri/absolute.rs
+++ b/core/http/src/uri/absolute.rs
@@ -115,6 +115,42 @@ impl<'a> Absolute<'a> {
         crate::parse::uri::absolute_from_str(string)
     }
 
+    /// Parses the string `string` into an `Absolute`. Parsing will never
+    /// allocate. This method should be used instead of
+    /// [`Absolute::parse()`](crate::uri::Absolute::parse()) when the source URI is
+    /// already a `String`. Returns an `Error` if `string` is not a valid absolute
+    /// URI.
+    pub fn parse_owned(string: String) -> Result<Absolute<'static>, Error<'static>> {
+        // We create a copy of a pointer to `string` to escape the borrow
+        // checker. This is so that we can "move out of the borrow" later.
+        //
+        // For this to be correct and safe, we need to ensure that:
+        //
+        //  1. No `&mut` references to `string` are created after this line.
+        //  2. `string` isn't dropped while `copy_of_str` is live.
+        //
+        // These two facts can be easily verified. An `&mut` can't be created
+        // because `string` isn't `mut`. Then, `string` is clearly not dropped
+        // since it's passed in to `source`.
+        // let copy_of_str = unsafe { &*(string.as_str() as *const str) };
+        let copy_of_str = unsafe { &*(string.as_str() as *const str) };
+        let absolute = Absolute::parse(copy_of_str)?;
+        debug_assert!(absolute.source.is_some(), "Origin source parsed w/o source");
+
+        let absolute = Absolute {
+            scheme: absolute.scheme.into_owned(),
+            authority: absolute.authority.into_owned(),
+            origin: absolute.origin.into_owned(),
+            // At this point, it's impossible for anything to be borrowing
+            // `string` except for `source`, even though Rust doesn't know it.
+            // Because we're replacing `source` here, there can't possibly be a
+            // borrow remaining, it's safe to "move out of the borrow".
+            source: Some(Cow::Owned(string)),
+        };
+
+        Ok(absolute)
+    }
+
     /// Returns the scheme part of the absolute URI.
     ///
     /// # Example
@@ -458,5 +494,46 @@ impl std::fmt::Display for Absolute<'_> {
         }
 
         Ok(())
+    }
+}
+
+#[cfg(feature = "serde")]
+mod serde {
+    use std::fmt;
+
+    use super::Absolute;
+    use _serde::{ser::{Serialize, Serializer}, de::{Deserialize, Deserializer, Error, Visitor}};
+
+    impl<'a> Serialize for Absolute<'a> {
+        fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+            serializer.serialize_str(&self.to_string())
+        }
+    }
+
+    struct AbsoluteVistor;
+
+    impl<'a> Visitor<'a> for AbsoluteVistor {
+        type Value = Absolute<'a>;
+        fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+            write!(formatter, "Expecting a valid URI string")
+        }
+
+        fn visit_str<E: Error>(self, v: &str) -> Result<Self::Value, E> {
+            Absolute::parse_owned(v.to_string()).map_err(Error::custom)
+        }
+
+        fn visit_string<E: Error>(self, v: String) -> Result<Self::Value, E> {
+            Absolute::parse_owned(v).map_err(Error::custom)
+        }
+
+        fn visit_borrowed_str<E: Error>(self, v: &'a str) -> Result<Self::Value, E> {
+            Absolute::parse(v).map_err(Error::custom)
+        }
+    }
+
+    impl<'a> Deserialize<'a> for Absolute<'a> {
+        fn deserialize<D: Deserializer<'a>>(deserializer: D) -> Result<Self, D::Error> {
+            deserializer.deserialize_str(AbsoluteVistor)
+        }
     }
 }
--- a/core/http/src/uri/absolute.rs
+++ b/core/http/src/uri/absolute.rs
@@ -64,6 +64,20 @@ use crate::uri::{Authority, Path, Query, Data, Error, as_utf8_unchecked, fmt};
 /// #   assert!(!Absolute::parse(uri).unwrap().is_normalized());
 /// # }
 /// ```
+///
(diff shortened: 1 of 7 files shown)


core/http/src/uri/absolute.rs
```diff
@@ -115,6 +129,26 @@ impl<'a> Absolute<'a> {
...
+    /// [`Absolute::parse()`](crate::uri::Absolute::parse()) when the source URI is
+    /// already a `String`. Returns an `Error` if `string` is not a valid absolute
+    /// URI.
+    pub fn parse_owned(string: String) -> Result<Absolute<'static>, Error<'static>> {
+        let absolute = Absolute::parse(&string).map_err(|e| e.into_owned())?;
+        debug_assert!(absolute.source.is_some(), "Origin source parsed w/o source");
+
+        let absolute = Absolute {
+            scheme: absolute.scheme.into_owned(),
+            authority: absolute.authority.into_owned(),
```
SergioBenitez, line 143: This will allocate. This is why the method was not implemented for this type. I'm not quite sure what to do about this.

core/http/src/uri/absolute.rs
```diff
@@ -460,3 +495,44 @@ impl std::fmt::Display for Absolute<'_> {
...
+            serializer.serialize_str(&self.to_string())
+        }
+    }
+
+    struct AbsoluteVistor;
+
+    impl<'a> Visitor<'a> for AbsoluteVistor {
+        type Value = Absolute<'a>;
+        fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+            write!(formatter, "Expecting a valid URI string")
```
SergioBenitez, line 517: This should say something like "an absolute URI"; no "expecting", lowercase start, and we don't need "valid". The other expecting strings should be fixed accordingly.

core/http/src/uri/authority.rs
```diff
@@ -109,6 +123,25 @@ impl<'a> Authority<'a> {
         crate::parse::uri::authority_from_str(string)
     }
 
+    /// Parses the string `string` into an `Authority`. Parsing will never
```
SergioBenitez, line 126: This is good. Can we make the doc string match that of `Origin` in the latest commit?

core/http/src/uri/uri.rs
```diff
@@ -33,6 +33,14 @@ use crate::uri::error::{Error, TryFromUriError};
 /// methods of the internal structure.
 ///
 /// [RFC 7230]: https://tools.ietf.org/html/rfc7230
+///
```
SergioBenitez, line 36: I think it's weird that it implements one but not the other. I can't think of any utility this offers. I would suggest either implementing both (using parse_any for deserialize) or neither.

core/lib/tests/http_uri_serde.rs
```diff
@@ -0,0 +1,49 @@
...
+            [default]
+            asterisk = "*"
+            origin = "/foo/bar?baz"
+            authority = "user:pass@rocket.rs:80"
+            absolute = "https://rocket.rs/foo/bar"
+            reference = "https://rocket.rs:8000/index.html"
+        "#)?;
+
+        let uris: UriContainer<'_> = Config::figment().extract()?;
+        assert_eq!(uris, UriContainer {
```
SergioBenitez, line 32: Don't use any of the hidden constructors here. Instead, use the uri! macro.

core/lib/tests/http_uri_serde.rs
```diff
@@ -0,0 +1,49 @@
...
+            asterisk: Asterisk,
+            origin: Origin::new("/foo/bar", Some("baz")),
```
SergioBenitez, line 34:
```suggestion
            origin: uri!("/foo/bar?baz"),
```

core/http/src/uri/absolute.rs
SergioBenitez, line 517:
```suggestion
            write!(formatter, "an absolute URI")
```

core/http/src/uri/asterisk.rs
```diff
@@ -7,3 +11,44 @@ impl std::fmt::Display for Asterisk {
...
+        }
+
+        // This method should be the only one that needs to be implemented, since the
+        // other two methods (`visit_string` & `visit_borrowed_str`) have default implementations
+        // that just call this one. We don't benefit from taking ownership or borrowing from the
+        // deserializer, so this should be perfect.
+        fn visit_str<E: Error>(self, v: &str) -> Result<Self::Value, E> {
+            if v == "*" {
+                Ok(Asterisk)
+            }else {
```
SergioBenitez, line 43:
```suggestion
            } else {
```

SergioBenitez (changes requested):
Looks mostly good. We need to test serialization; you should be able to round trip through Figment for this.

What's more, se should also test that that the `'static` versions are `DeserializeOwned` and don't require any sort of `#[serde(borrow)]` attribute.

SergioBenitez:
Note that the CI is failing

the10thWiz:
This is kind of embarrassing. The round trip test I wrote was the one failing. I'm not sure if I didn't run it, but I guess I must not have. The issue should be fixed now.

core/lib/tests/http_uri_serde.rs
```diff
@@ -44,15 +44,15 @@ fn uri_serde() {
 
 #[test]
 fn uri_serde_round_trip() {
-    let tmp = Figment::from(Serialized::default("default", UriContainer {
+    let tmp = Figment::from(Serialized::default("uris", UriContainer {
```
SergioBenitez, line 47: Use [`Serialized::defaults()`](https://docs.rs/figment/0.10.5/figment/providers/struct.Serialized.html#method.defaults),

the10thWiz: Thanks, I have switched to that. I'm also fixing the CI failure (It looks like shield was added since I started, and I have updated it to reflect the slightly different Absolute and Reference Uris)

SergioBenitez:
This is a bit difficult to review as-is. Can you either split this into two PRs, one for the changes to `Absolute` and the other for the De(serialize) impls, or squash into two commits of the same nature?

the10thWiz:
I'm in the process of rebasing this PR onto master, and removing the changes to parse_owned. I will be working out a solution to prevent parse_owned from allocating for `Absolute` and `Reference`, but for this PR I'm just going to update the documentation to say that they allocate.

core/lib/tests/http_uri_serde.rs
SergioBenitez: This should use the `uri!` macro.

SergioBenitez (changes requested):
Looks mostly good. The `uri_serde` test should use `uri!()` in the deserialize test too. We also still need to test that deserializing into `'static` variants works as expected, IE, these types are `DeserializeOwned` and don't require any kind of `serde(borrow)` attribute. I suggest adding an `OwnedUriContainer` struct with only `'static` variants (`Absolute<'static>`, etc.) and repeat the tests.

the10thWiz:
> We also still need to test that deserializing into `'static` variants works as expected, IE, these types are `DeserializeOwned` and don't require any kind of `serde(borrow)` attribute. I suggest adding an `OwnedUriContainer` struct with only `'static` variants (`Absolute<'static>`, etc.) and repeat the tests.

I went ahead and added the test, and it works, but the `serde(borrow)` attribute is still required.

SergioBenitez:
> > We also still need to test that deserializing into `'static` variants works as expected, IE, these types are `DeserializeOwned` and don't require any kind of `serde(borrow)` attribute. I suggest adding an `OwnedUriContainer` struct with only `'static` variants (`Absolute<'static>`, etc.) and repeat the tests.
> 
> 
> 
> I went ahead and added the test, and it works, but the `serde(borrow)` attribute is still required.
> 
> 
> 
> 

This is not good. This should not be the case.

the10thWiz:
Unfortunately, I think we don't have a choice.

SergioBenitez:
> Unfortunately, I think we don't have a choice.

Of course we do. The deserialize implementation is simply incorrect. Consider that the deserialize test you have now isn't borrowing from anywhere: the data is owned by Figment and isn't being passed as a reference. Consider too that Absolute<'static> borrows from nowhere and clearly shouldn't need a "borrow" attribute.

If you're not sure how to fix this, do let me know and I'll address it myself.

SergioBenitez:
Thank you so much for all of your efforts! Merged in 12335187338460a79992c9fb527b3c256f49d463 with some fixup in faaa9c90653c26e53ff10c23106297f7851910e0.

PR #1652, opened 2021-05-24
Docs: Extend second wrapping example

RotesWasser:
I had trouble immediately understanding how to send custom HTTP status codes that were not covered by the named Responders from `rocket::response::status`, so I'm suggesting including it in the second example. This could also highlight two 'bigger' Responders wrapping one another.


SergioBenitez:
You're much better off returning `(Status, T)` than `status::Custom(..)`.

SergioBenitez:
Merged in spirit in 9f5a587dc3146228045fb6e410d94f70ce45d8a7. Thank you for the suggestion!

PR #1671, opened 2021-06-08
Fix some clippy lints

MikailBag:
This pull request mostly just makes Clippy a bit happier (hopefully not hurting code readability), but it also contains two changes:
1. Several `Into` implementations are now `From` implementations. Makes UI errors a bit longer, so I'll roll back this change if needed.
2. Please, review `Accept::preferred`. I think the old implementation had a bug, and I'm not sure if the new one is correct.

### DIFF
--- a/core/http/src/header/accept.rs
+++ b/core/http/src/header/accept.rs
@@ -147,24 +147,8 @@ impl Accept {
         let mut all = self.iter();
         let mut preferred = all.next().unwrap_or(&ANY);
         for media_type in all {
-            if media_type.weight().is_none() && preferred.weight().is_some() {
-                // Media types without a `q` parameter are preferred.
+            if media_type.better_than(preferred) {
                 preferred = media_type;
-            } else if media_type.weight_or(0.0) > preferred.weight_or(1.0) {
-                // Prefer media types with a greater weight, but if one doesn't
-                // have a weight, prefer the one we already have.
-                preferred = media_type;
-            } else if media_type.specificity() > preferred.specificity() {
-                // Prefer more specific media types over less specific ones. IE:
-                // text/html over application/*.
-                preferred = media_type;
-            } else if media_type != preferred {
-                // Finally, all other things being equal, prefer a media type
-                // with more parameters over one with fewer. IE: text/html; a=b
-                // over text/html.
-                if media_type.params().count() > preferred.params().count() {
-                    preferred = media_type;
-                }
             }
         }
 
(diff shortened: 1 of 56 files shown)


core/http/src/header/accept.rs
```diff
@@ -147,24 +147,8 @@ impl Accept {
...
-                preferred = media_type;
-            } else if media_type.specificity() > preferred.specificity() {
-                // Prefer more specific media types over less specific ones. IE:
-                // text/html over application/*.
-                preferred = media_type;
-            } else if media_type == preferred {
-                // Finally, all other things being equal, prefer a media type
-                // with more parameters over one with fewer. IE: text/html; a=b
-                // over text/html.
-                if media_type.params().count() > preferred.params().count() {
```
MikailBag, line 165: This `if` can not be entered, because at this point `media_type == preferred`.

I assumed that this is intended to be a kind of lexicography comparator (check several conditions, and each one is strictly more important than the next) when I rewrote it. Please correct me if I misunderstood the intended algorithm.

SergioBenitez: The first equality check does not consider parameters, as documented. Please don't change this method at all.

SergioBenitez:
> * Several `Into` implementations are now `From` implementations. Makes UI errors a bit longer, so I'll roll back this change if needed.

Let's roll back these for now.

core/lib/src/form/error.rs
```diff
@@ -719,7 +719,9 @@ impl<'v> Error<'v> {
             | Multipart(FieldSizeExceeded { .. })
             | Multipart(StreamSizeExceeded { .. }) => Status::PayloadTooLarge,
             Unknown => Status::InternalServerError,
-            Io(_) | _ if self.entity == Entity::Form => Status::BadRequest,
+            // FIXME: is this intended behavior?
```
SergioBenitez, line 722: Yes.

core/lib/src/fairing/fairings.rs
```diff
@@ -84,7 +84,7 @@ impl Fairings {
                 .collect();
 
             // Reverse the dup indices so `remove` is stable given shifts.
-            dups.sort(); dups.dedup(); dups.reverse();
+            dups.sort_unstable(); dups.dedup(); dups.reverse();
```
SergioBenitez, line 87: We want a stable sort. It's even in the comment above. :)

core/lib/src/cookies.rs
```diff
@@ -259,15 +259,15 @@ impl<'a> CookieJar<'a> {
     /// ```
     pub fn get_pending(&self, name: &str) -> Option<Cookie<'static>> {
         let ops = self.ops.lock();
-        for op in ops.iter().rev().filter(|op| op.cookie().name() == name) {
```
SergioBenitez, line 262: We need to return as soon as possible from here so we drop the lock. This implementation makes that fact less obvious. Please revert.

SergioBenitez (commented):
I'm generally in favor of this, but after looking at this even briefly, I've spotted a few cases where semantics have changed. Please remove any changes that could even remotely change semantics. We should only make changes that retain semantics _exactly_. When in doubt, please don't change anything.

I'm happy to review the PR once it is exceedingly conservative.

core/http/src/header/accept.rs
MikailBag: I think I'll roll back this change for now and discuss it in other PR, to make review easier.

core/lib/src/fairing/fairings.rs
MikailBag: `dups` is a vector of `usize`. Two `usize` values compare equal only if they are bitwise equal, so `sort_unstable` produces the same result as `sort`. So I think that clippy is not wrong here.

SergioBenitez:
Rebased and merged (mostly) in be933ce3983b45a85840c63adfc145a0b2c78a16. Thanks for keeping this fairly up to date!

PR #1676, opened 2021-06-10
Fix wording on data::Data::peek

timando:
The original wording said max of the actual length, the amount requested and 512, which doesn't make a whole lot of sense.


jebrosen (approved):

jebrosen:
Thank you! Merged in af1aa07.

PR #1678, opened 2021-06-10
Fix the link to the JSON example in the docs

t-eckert:
In the docs, the link to the JSON example is broken. My guess is you folded it into the serialization example. I changed the link to point to the JSON file in the example.

If you would like me to instead change the copy so that it specifies that file and link the user to the root of the serialization example, I am happy to make that modification.


jebrosen:
Merged this in 7cced84 - thanks!

PR #1681, opened 2021-06-10
site: prevent an unused Result warning

paulsmith:
`launch()` returns a `Result` type that is not handled in the example `async fn main()`, which leads to a warning when compiling:

```
warning: unused `Result` that must be used
  --> src/main.rs:19:5
   |
19 | /     rocket::build()
20 | |         .mount("/", routes![index, delay])
21 | |         .launch()
22 | |         .await;
   | |_______________^
   |
   = note: `#[warn(unused_must_use)]` on by default
   = note: this `Result` may be an `Err` variant, which should be handled
```

The simplest change is just to `unwrap()` it, but the guide might want to explicitly mention the fact of `launch()` return type signature.

### DIFF
--- a/site/guide/3-overview.md
+++ b/site/guide/3-overview.md
@@ -213,7 +213,8 @@ async fn main() {
     rocket::build()
         .mount("/hello", routes![world])
         .launch()
-        .await;
+        .await
+        .unwrap();
 }
 ```
 


SergioBenitez:
We _want_ to drop the result without using it. I know we have a note somewhere explaining why. 

Still, it would be nice for this example to not emit a warning. I'll think about this a bit. Thanks for bringing this up!

SergioBenitez:
Merged something similar in dc1b729f29f2939ed56dcd38ff0c9dedf3ac9f70. Thanks again for bringing this up!

PR #1686, opened 2021-06-11
Make error type of `FromParam` for `String` `Infallible`

### LINKED ISSUE #1679: FromParam impl for String uses different error type from &str impl
I just noticed when updating a project to 0.5.0-alpha.1 and updating a `FromParam` implementation that the `String` and `&str` implementations of `FromParam` use different error types (`std::convert::Infallible` vs `&str`), which seems unintentional given that both are infallible.

JarvisCraft:
This fixes #1679.


jebrosen (approved):

SergioBenitez:
Merged in 4339c2af4861b4c95b9c2648f74b0d730ca40118.

PR #1688, opened 2021-06-12
implement Responder for Arc<T> and Box<T>

udoprog:
`Arc<str>` (or `Arc<String>`) in particular are pretty common for types which contain fixed content that is cheaply cloned. Having them supported out-of-the-box is therefore useful.

I've only added the wide pointer alternatives for `str` and `[u8]`. But it's plausible that someone might want types such as `Arc<String>` and `Arc<Vec<u8>>` as well.

### DIFF
--- a/core/lib/src/response/responder.rs
+++ b/core/lib/src/response/responder.rs
@@ -1,5 +1,6 @@
 use std::fs::File;
 use std::io::Cursor;
+use std::sync::Arc;
 
 use crate::http::{Status, ContentType, StatusClass};
 use crate::response::{self, Response};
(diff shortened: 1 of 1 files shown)


SergioBenitez:
Interesting. Where are these `Arc`s arising in your application? And, why can they be put in request-local state instead?

udoprog:
> Interesting. Where are these `Arc`s arising in your application? And, why can't they be put in request-local cache instead?

Right now they're being pulled out of a `State<Content>`-like thing, who's content is being updated elsewhere in the application.

SergioBenitez:
Can you clarify further? If you it's in managed state, then you can return the `&str` directly. That is, something like:

```rust
struct Content { string: Arc<String> }

#[get("/")]
fn foo(content: &State<Content>) -> &str {
    &*content.string
}
```

udoprog:
Right. So that field can't be updated, because `string` is just an `Arc<String>`. You need something like this:

```rust
use std::sync::RwLock;

struct Content { string: RwLock<Arc<str>> }

#[get("/")]
n foo(content: &State<Content>) -> Arc<str> {
    content.string.read().unwrap().clone()
}
```

At which point we can't return a reference because the lock guard doesn't outlive the function. And we also don't want the lock to be held for longer than necessary.

SergioBenitez:
Ok. To prevent more back-and-forth, can you be _as specific as possible_ about what your application is doing and why you're passing around `RwLock<Arc<str>>` in managed state? My concern is two-fold: 1) these are more impls to maintain without any generality to their implementation, and 2) their existence could encourage bad practices. A convincing use-case could help me overcome the latter.

udoprog:
> Ok. To prevent more back-and-forth, can you be _as specific as possible_ about what your application is doing and why you're passing around `RwLock<Arc<str>>` in managed state?

The type is an `Mutex<HashMap<Box<str>, Arc<str>>>` right now. Let's call this `Content`.

> Note: the real `Content` structure is actually CPU-sharded to reduce contention. This is still a simplification but including that would make it more confusing for you.

There is a background process which watches a directory for changes. If this directory changes it processes this directory and fills the `Content` structure with data. This happens while the rocket part of the application is running.

This is then used in a route like this:

```rust
#[get("/content/<name>")]
async fn get_debate(debates: &State<Content>, name: &str) -> Result<Arc<str>, NotFound<&'static str>> {
    let content = content.lock().await;

    if let Some(entry) = content.get(name) {
        Ok(entry.clone())
    } else {
        Err(NotFound("no such entry"))
    }
}
```

This is an extremely typical pattern for applications which perform some form of live reloading or on-the-fly updates. Usually for tools which performs some form of batch processing but has a mode where you can see the changes live.

Like `mdbook` or `zola`.

udoprog:
Of course, if there is a different way to accomplish this I'd be all for it. My current solution simply involves wrapping the `Arc<str>` in a newtype, almost identical to what I'm proposing here but defined locally in my project.

SergioBenitez:
Merged with some changes in 0ddad7a6c1dac9d7f6d23f32622b40ce37c655f7.

PR #1689, opened 2021-06-12
Fix typos and missing asserts in guide/requests

Heliozoa:
Just a couple things I noticed while going through the guide. Great work on 0.5!


SergioBenitez:
Ah, right. I was trusting that the asserts in the route would make the test fail if the example was wrong, but it'll only result in 500 error. Good catch!

jebrosen (approved):

SergioBenitez:
Merged in 2bd5cefedde7bd8fbbd5a910ce333bb80c57e6ab. Thanks!

PR #1690, opened 2021-06-12
Update irc network references to Libera.Chat

ijackson:
Hi again.  I am very excited about Rocket 0.5!  I may file more issues/MRs as I work through updating my Rocket applications.  But in the meantime:

I was reading the docs and noticed references to the old irc network Freenode.  I checked with my irc client and `#rocket` on `irc.libera.chat` seems active and has the promised Matrix bridge.  So I think it's probably just that the website hasn't been updated yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -689,7 +689,7 @@ In addition to new features and major improvements, Rocket saw the following imp
 ## Infrastructure
 
   * CI was moved from Travis to Azure Pipelines; Windows support is tested.
-  * Rocket's chat moved to [Matrix] and [Freenode].
+  * Rocket's chat moved to [Matrix] and Freenode.
 
 [`Debug`]: https://api.rocket.rs/v0.4/rocket/response/struct.Debug.html
 [`NoContent`]: https://api.rocket.rs/v0.4/rocket/response/status/struct.NoContent.html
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -689,15 +689,14 @@ In addition to new features and major improvements, Rocket saw the following imp
 ## Infrastructure
 
   * CI was moved from Travis to Azure Pipelines; Windows support is tested.
-  * Rocket's chat moved to [Matrix] and [Freenode].
```
SergioBenitez, line 692: We should never change the past. Please revert this change.

README.md
```diff
@@ -45,14 +45,14 @@ Rocket is extensively documented:
 [API Documentation]: https://api.rocket.rs/rocket/
 
 The official community support channels are [`#rocket:mozilla.org`] on Matrix
-and the bridged [`#rocket`] IRC channel on Freenode at `chat.freenode.net`. We
+and the bridged [`#rocket`] IRC channel on Freenode at `irc.libera.chat`. We
```
SergioBenitez, line 48: This still says Freenode.

ijackson: Oops, fixed, thanks.

CHANGELOG.md
ijackson: I don't agree that this is changing the past.  But it's your project, so I have reverted that change, as you request.

ijackson:
Oops, reverted the wrong commit.  Have force pushed it.  Hope it LGTY now.  Thanks.

CHANGELOG.md
SergioBenitez: This is a CHANGELOG entry for 0.4.4, where Rocket's chat moved to Freenode. Changing it to say we moved to Libera is plain wrong as a 0.4.4 entry.

SergioBenitez:
Merged in 42f84f06214ebe09d2e9f8f7b97606db7e35a77c. Thanks!

Issue #1187, opened 2019-12-17
Expose pool directly from generated database pools


cassc:
Rocket can create and manage a pooled database connection which can be accessed inside HTTP handlers. 

Can I access this database pool outside these handlers?

jebrosen:
Not easily, unfortunately. The code generated by `#[database]` creates a hidden type for the connection pool and puts it in managed state, which makes sharing hard for two reasons:

* The name of this hidden type is technically an implementation detail.
* IIRC there's no way to get an `Arc<Pool>` out, just an `&Pool` or `State<'_, Pool>` that is tied to the lifetime of the `Rocket` instance. This makes it impossible to call `launch()`, which moves the `Rocket` instance.

In case you only need a single connection before launch (e.g. for initialization), `get_one` does currently exist for that purpose.

eon8ight:
Follow-up question: so you've got `get_one`, but that requires a `Rocket` instance. What if you have a request guard that requires a database connection? How do you get the running `Rocket` instance to pass into it?

jebrosen:
If you're already in a request guard, you can use [`request.guard::<ConnectionType>()`](https://api.rocket.rs/v0.4/rocket/struct.Request.html#method.guard) to get a connection.

eon8ight:
Worked perfectly - thank you!

SergioBenitez:
@cassc Do you have a case where you need access to the actual pool as opposed to a connection from the pool (which we allow, via `get_one()`)?

cassc:
@SergioBenitez 
My application has an HTTP server and also a UDP server, both need to access the database.

KitaitiMakoto:
I have the same desire. I'm developing a web application [Plume][] using Rocket and [Riker][], an actor library. Actors are running on other threads and some actors need to connect to database. In this case, we want to make actors retrieve database connections from the pool on demand rather than hold connection for each. If we cannot share database pool between Rocket and actors, we need build two pools.

[Plume]: https://joinplu.me/
[Riker]: https://riker.rs/

PR #1696, opened 2021-06-14
Introduce `db_pools` contrib crate, for asynchronous database connection pooling

### LINKED ISSUE #1117: Tracking Issue for Asynchronous Database Pool Support

jebrosen:
Resolves #1117.

This implementation is similar to `sync_db_pools` and derives most of its documentation from that crate, but there are a few notable differences:

* Database types are declared via a `#[derive(Database)]` in combination with the `#[database(name = "")]` attribute, which makes it more straightforward to see what methods are available on the decorated type.
* Databases are declared in terms of the pool type, not the connection type. This helps avoid requiring adapters to go through a particular pool implementation such as `r2d2`, and allows us to use built-in connection pooling support that is provided by some database drivers.
* The decorated database type is placed directly in managed state by its fairing, so that `.pool()` and `.get()` are easily accessible after attaching the fairing and from within routes. This is especially useful for integrations that provide additional methods on the pool type which would otherwise be inaccessible.
* Direct access to the underlying connection pool is available (#1187), via the `.pool()` method.
* All relevant APIs are `async`, and no additional wrapper (a la `run()`) is necessary to bridge between `async` and blocking code.

TODO:

* [ ] Solidify the configuration API. In `sync_db_pools`, each implementation of `Poolable` deserializes its config from a `Rocket` instance. In this implementation, the configuration is restricted any `DeserializeOwned` type and the deserialization is done by the database fairing - this makes it easier to correctly implement `Pool` since it is no longer responsible for looking up its configuration, but it is impossible to query the worker count when choosing a pool size. I would appreciate any ideas on how to best reconcile this.
* [x] `pool_size` - Different database pools accept different combinations of min or max pool size, and currently most `Pool` implementations use the default.
* [x] Configurable `timeout` for connection retrieval.
* [ ] Tests.
* [ ] Documentation additions and corrections, especially the guide. Most of the documentation here has simply been copied from `sync_db_pools` and adjusted for the different trait names and provided integrations.

### DIFF
--- a/contrib/db_pools/codegen/Cargo.toml
+++ b/contrib/db_pools/codegen/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "rocket_db_pools_codegen"
+version = "0.1.0-dev"
+authors = ["Sergio Benitez <sb@sergio.bz>", "Jeb Rosen <jeb@jebrosen.com>"]
+description = "Procedural macros for rocket_db_pools."
+repository = "https://github.com/SergioBenitez/Rocket/contrib/db_pools"
+readme = "../README.md"
+keywords = ["rocket", "framework", "database", "pools"]
+license = "MIT OR Apache-2.0"
+edition = "2018"
+
+[lib]
+proc-macro = true
+
+[dependencies]
+proc-macro2 = "1"
+quote = "1"
+syn = "1"
+proc-macro2-diagnostics = "0.9"
--- a/contrib/db_pools/codegen/Cargo.toml
+++ b/contrib/db_pools/codegen/Cargo.toml
@@ -13,7 +13,5 @@ edition = "2018"
 proc-macro = true
 
 [dependencies]
-proc-macro2 = "1"
+devise = { git = "https://github.com/SergioBenitez/Devise.git", rev = "f2431b" }
 quote = "1"
-syn = "1"
-proc-macro2-diagnostics = "0.9"
--- a/contrib/db_pools/codegen/Cargo.toml
+++ b/contrib/db_pools/codegen/Cargo.toml
@@ -13,5 +13,5 @@ edition = "2018"
 proc-macro = true
 
 [dependencies]
-devise = { git = "https://github.com/SergioBenitez/Devise.git", rev = "f2431b" }
+devise = "0.3"
 quote = "1"
--- a/contrib/db_pools/lib/src/config.rs
+++ b/contrib/db_pools/lib/src/config.rs
@@ -1,6 +1,6 @@
+use rocket::figment::{self, Figment, providers::Serialized};
 use rocket::serde::{Deserialize, Serialize};
-
-use crate::Error;
+use rocket::{Build, Rocket};
 
 /// A base `Config` for any `Pool` type.
 ///
@@ -18,7 +18,8 @@ use crate::Error;
 /// # use rocket_db_pools::Config;
 /// Config {
 ///     url: "postgres://root:root@localhost/my_database".into(),
-///     pool_size: Some(10),
+///     pool_size: 10,
+///     timeout: 5,
 /// };
 /// ```
 ///
@@ -31,18 +32,33 @@ use crate::Error;
 pub struct Config {
     /// Connection URL specified in the Rocket configuration.
     pub url: String,
-    /// Requested pool size. Defaults to the number of Rocket workers * 4.
-    pub pool_size: Option<i64>,
-    // TODO: timeout?
+    /// Initial pool size. Defaults to the number of Rocket workers * 4.
+    pub pool_size: u32,
+    /// How long to wait, in seconds, for a new connection before timing out.
+    /// Defaults to `5`.
+    // FIXME: Use `time`.
+    pub timeout: u8,
 }
 
 impl Config {
-    /// Returns the requested pool size, or `default`
-    pub fn pool_size_or_default<E>(&self, default: i64) -> Result<usize, Error<E>> {
-        use std::convert::TryInto;
-        match self.pool_size.unwrap_or(default).try_into() {
-            Ok(p) => Ok(p),
-            Err(_) => Err(Error::config("pool_size was outside the valid range")),
+    pub fn from(db_name: &str, rocket: &Rocket<Build>) -> Result<Self, figment::Error> {
+        Self::figment(db_name, rocket).extract::<Self>()
+    }
+
+    pub fn figment(db_name: &str, rocket: &Rocket<Build>) -> Figment {
+        let db_key = format!("databases.{}", db_name);
+        let default_pool_size = rocket.figment()
+            .extract_inner::<u32>(rocket::Config::WORKERS)
+            .map(|workers| workers * 4)
+            .ok();
+
+        let figment = Figment::from(rocket.figment())
+            .focus(&db_key)
+            .join(Serialized::default("timeout", 5));
+
+        match default_pool_size {
+            Some(pool_size) => figment.join(Serialized::default("pool_size", pool_size)),
+            None => figment
         }
     }
 }
--- a/contrib/db_pools/lib/src/config.rs
+++ b/contrib/db_pools/lib/src/config.rs
@@ -1,0 +1,48 @@
+use rocket::serde::{Deserialize, Serialize};
+
+use crate::Error;
+
+/// A base `Config` for any `Pool` type.
+///
+/// For the following configuration:
+///
+/// ```toml
+/// [global.databases.my_database]
+/// url = "postgres://root:root@localhost/my_database"
+/// pool_size = 10
+/// ```
+///
+/// ...the following struct would be passed to [`Pool::initialize()`]:
+///
+/// ```rust
+/// # use rocket_db_pools::Config;
+/// Config {
+///     url: "postgres://root:root@localhost/my_database".into(),
+///     pool_size: Some(10),
+/// };
+/// ```
+///
+/// If you want to implement your own custom database adapter and need some more
+/// configuration options, you may need to define a custom `Config` struct.
+///
+/// [`Pool::initialize()`]: crate::Pool::initialize
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[serde(crate = "rocket::serde")]
+pub struct Config {
+    /// Connection URL specified in the Rocket configuration.
+    pub url: String,
+    /// Requested pool size. Defaults to the number of Rocket workers * 4.
+    pub pool_size: Option<i64>,
+    // TODO: timeout?
+}
+
+impl Config {
+    /// Returns the requested pool size, or `default`
+    pub fn pool_size_or_default<E>(&self, default: i64) -> Result<usize, Error<E>> {
+        use std::convert::TryInto;
+        match self.pool_size.unwrap_or(default).try_into() {
+            Ok(p) => Ok(p),
+            Err(_) => Err(Error::config("pool_size was outside the valid range")),
+        }
+    }
+}
--- a/contrib/db_pools/lib/src/database.rs
+++ b/contrib/db_pools/lib/src/database.rs
@@ -1,0 +1,137 @@
+use rocket::fairing::{Info, Kind};
+use rocket::futures::future::BoxFuture;
+use rocket::{Build, Ignite, Rocket, Sentinel};
+
+use crate::Pool;
+
+/// Trait implemented to define a database connection pool.
+pub trait Database: Sized + Send + Sync + 'static {
+    /// The name of this connection pool in the configuration.
+    const NAME: &'static str;
+
+    /// The underlying connection type returned by this pool.
+    /// Must implement [`Pool`].
+    type Pool: Pool;
+
+    /// Returns a fairing that attaches this connection pool to the server.
+    fn fairing() -> Fairing<Self>;
+
+    /// Direct shared access to the underlying database pool
+    fn pool(&self) -> &Self::Pool;
+
+    /// get().await returns a connection from the pool (or an error)
+    fn get(&self) -> BoxFuture<'_, Result<Connection<Self>, <Self::Pool as Pool>::GetError>> {
+        Box::pin(async move { self.pool().get().await.map(Connection)} )
+    }
+}
+
+/// A connection. The underlying connection type is determined by `D`, which
+/// must implement [`Database`].
+pub struct Connection<D: Database>(<D::Pool as Pool>::Connection);
+
+impl<D: Database> std::ops::Deref for Connection<D> {
+    type Target = <D::Pool as Pool>::Connection;
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl<D: Database> std::ops::DerefMut for Connection<D> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
+
+#[rocket::async_trait]
+impl<'r, D: Database> rocket::request::FromRequest<'r> for Connection<D> {
+    type Error = <D::Pool as Pool>::GetError;
+
+    async fn from_request(
+        req: &'r rocket::Request<'_>,
+    ) -> rocket::request::Outcome<Self, Self::Error> {
+        use rocket::http::Status;
+        use rocket::request::Outcome;
+
+        let db: &D = match req.rocket().state() {
+            Some(p) => p,
+            _ => panic!(
+                "tried to use database connection pool {}, but its fairing was not attached",
+                D::NAME,
+            ),
+        };
+
+        match db.pool().get().await {
+            Ok(conn) => Outcome::Success(Connection(conn)),
+            Err(e) => Outcome::Failure((Status::ServiceUnavailable, e)),
+        }
+    }
+}
+
+impl<D: Database> Sentinel for Connection<D> {
+    fn abort(rocket: &Rocket<Ignite>) -> bool {
+        use rocket::yansi::Paint;
+
+        if rocket.state::<D>().is_none() {
+            let dbtype = Paint::default(std::any::type_name::<D>()).bold();
+            let fairing = Paint::default(format!("{}::fairing()", dbtype)).wrap().bold();
+            error!("requesting `{}` DB connection without attaching `{}`.", dbtype, fairing);
+            info_!("Attach `{}` to use database connection pooling.", fairing);
+            return true;
+        }
+
+        false
+    }
+}
+
+/// The database fairing for pool types created with the `pool!` macro.
+pub struct Fairing<D: Database>(fn(D::Pool) -> D);
+
+impl<D: Database> Fairing<D> {
+    /// Create a new database fairing with the given constructor.  This
+    /// constructor will be called to create an instance of `D` after the pool
+    /// is initialized and before it is placed into managed state.
+    pub fn new(ctor: fn(D::Pool) -> D) -> Self {
+        Self(ctor)
+    }
+}
+
+#[rocket::async_trait]
+impl<D: Database> rocket::fairing::Fairing for Fairing<D> {
+    fn info(&self) -> Info {
+        Info {
+            name: "rocket_db_pools connection pool",
+            kind: Kind::Ignite,
+        }
+    }
+
+    async fn on_ignite(&self, rocket: Rocket<Build>) -> Result<Rocket<Build>, Rocket<Build>> {
+        let db_config = match rocket
+            .figment()
+            .find_value(&format!("databases.{}", D::NAME))
+        {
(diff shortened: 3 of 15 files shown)


SergioBenitez:
I am so stoked to see this! This is huge for 0.5. Will prioritize reviewing this ASAP. Thank you!

contrib/db_pools/lib/src/database.rs
```diff
@@ -0,0 +1,137 @@
...
+    }
+}
+
+#[rocket::async_trait]
+impl<'r, D: Database> rocket::request::FromRequest<'r> for Connection<D> {
+    type Error = <D::Pool as Pool>::GetError;
+
+    async fn from_request(
+        req: &'r rocket::Request<'_>,
+    ) -> rocket::request::Outcome<Self, Self::Error> {
```
SergioBenitez, line 51: Import these types. No need to qualify everything and make this harder to read.

contrib/db_pools/lib/src/database.rs
```diff
@@ -0,0 +1,137 @@
...
+        use rocket::http::Status;
+        use rocket::request::Outcome;
+
+        let db: &D = match req.rocket().state() {
+            Some(p) => p,
+            _ => panic!(
+                "tried to use database connection pool {}, but its fairing was not attached",
+                D::NAME,
```
SergioBenitez, line 59: We shouldn't panic. Sentinels are not guaranteed.

contrib/db_pools/lib/src/database.rs
```diff
@@ -0,0 +1,137 @@
...
+    pub fn new(ctor: fn(D::Pool) -> D) -> Self {
+        Self(ctor)
+    }
+}
+
+#[rocket::async_trait]
+impl<D: Database> rocket::fairing::Fairing for Fairing<D> {
+    fn info(&self) -> Info {
+        Info {
+            name: "rocket_db_pools connection pool",
```
SergioBenitez, line 102: This should use the actual database name.

contrib/db_pools/lib/src/database.rs
```diff
@@ -0,0 +1,137 @@
...
+            }
+        };
+
+        let config = match db_config.deserialize() {
+            Ok(c) => c,
+            Err(e) => {
+                error!("error deserializing configuration: {}", e);
+                return Err(rocket);
+            }
+        };
```
SergioBenitez, line 125: This is not how figment should be used here. The database config should be extracted to a pool-specific config structure using figment, not "deserialize". We lose all of the provenance information by doing this. And, write much more code than necessary. Use Figment!

The thing to do here is to:
  1. Create a figment focused on the database config.
  2. Join a set of defaults (size = worker count * k, timeout, etc.).
  3. Extract to the provided config type.

contrib/db_pools/lib/src/error.rs
```diff
@@ -0,0 +1,40 @@
...
+/// [`Pool::initialize`] can return an error for any of several reasons:
+///
+///   * Missing or incorrect configuration, including some syntax errors
+///   * An error connecting to the database.
+///
+/// [`Pool::initialize`]: crate::Pool::initialize
+#[derive(Debug)]
+pub enum Error<E> {
+    /// An error in the configuration
+    Config(Cow<'static, str>),
```
SergioBenitez, line 15: This should be a `figment::Error`. And this should implement `From<figment:Error>` instead of the `Error::config()` method.

contrib/db_pools/lib/src/error.rs
```diff
@@ -0,0 +1,40 @@
...
+    pub fn config(message: impl Into<Cow<'static, str>>) -> Self {
+        Self::Config(message.into())
+    }
+}
+
+impl<E: fmt::Display> fmt::Display for Error<E> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Error::Db(e) => e.fmt(f),
+            Error::Config(e) => write!(f, "database connection pool configuration error: {}", e),
```
SergioBenitez, line 35:
```suggestion
            Error::Config(e) => write!(f, "bad configuration: {}", e),
```

contrib/db_pools/lib/src/config.rs
```diff
@@ -0,0 +1,48 @@
...
+/// };
+/// ```
+///
+/// If you want to implement your own custom database adapter and need some more
+/// configuration options, you may need to define a custom `Config` struct.
+///
+/// [`Pool::initialize()`]: crate::Pool::initialize
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[serde(crate = "rocket::serde")]
+pub struct Config {
```
SergioBenitez, line 31: We shouldn't need this structure at all. Each implementation should define their own config structure.

See my other comments in config.

contrib/db_pools/lib/src/pool.rs
```diff
@@ -0,0 +1,235 @@
...
+    type Connection;
+
+    /// The configuration type this database pool expects.
+    type Config: rocket::serde::DeserializeOwned + Send;
+
+    /// The error type returned by `initialize`.
+    type InitError: std::error::Error;
+
+    /// The error type returned by `get`.
+    type GetError: std::error::Error;
```
SergioBenitez, line 57: This feels wrong to me. Why isn't there just one `Error`?

contrib/db_pools/lib/src/pool.rs
```diff
@@ -0,0 +1,235 @@
...
+    ///
+    /// This method returns an error if a connection could not be retrieved,
+    /// such as a preconfigured timeout elapsing or when the database server is
+    /// unavailable.
+    async fn get(&self) -> Result<Self::Connection, Self::GetError>;
+}
+
+#[cfg(feature = "deadpool-postgres")]
+#[async_trait]
+impl crate::Pool for deadpool_postgres::Pool {
```
SergioBenitez, line 87:
```suggestion
impl Pool for deadpool_postgres::Pool {
```

Here and everywhere else.

contrib/db_pools/lib/src/pool.rs
```diff
@@ -0,0 +1,235 @@
...
+    /// accepts a (required) `url` and an (optional) `pool_size`.
+    ///
+    /// ## Errors
+    ///
+    /// This method returns an error if the configuration is not compatible, or
+    /// if creating a pool failed due to an unavailable database server,
+    /// insufficient resources, or another database-specific error.
+    ///
+    /// [`Config`]: crate::Config
+    async fn initialize(config: Self::Config) -> Result<Self, Self::InitError>;
```
SergioBenitez, line 73: Should this return `Result<T, Error<Self::Error>>`? Or should `Self::Error` be bound by `Into<Error>`? (where `Error` is `rocket_db_pools::Error`).

examples/databases/src/sqlx.rs
```diff
@@ -42,65 +47,47 @@ async fn list(db: &State<Db>) -> Result<Json<Vec<i64>>> {
...
 
 async fn init_db(rocket: Rocket<Build>) -> fairing::Result {
-    use rocket_sync_db_pools::Config;
-
-    let config = match Config::from("sqlx", &rocket) {
-        Ok(config) => config,
-        Err(e) => {
-            error!("Failed to read SQLx config: {}", e);
-            return Err(rocket);
+    match rocket.state::<Db>() {
```
SergioBenitez, line 75: This feels like an abstraction has just leaked. I think we want something like https://api.rocket.rs/v0.5-rc/rocket/struct.State.html#method.get. Maybe `Db::pool(&rocket);` and `Db::conn(&rocket).await;`.

SergioBenitez (changes requested):
I've left some comments. The biggest issues I see are:

  * Config needs to be redone. I've laid out a structure in the comments.
  * Libraries aren't being used to their fullest extent. Use `devise` and `figment`.
  * Lots of items are fully qualified, increasing the verbosity of code unnecessarily.
  * The effective need to declare a type alias is unfortunate. Maybe an associated type helps.

contrib/db_pools/lib/src/database.rs
```diff
@@ -0,0 +1,137 @@
...
+    type Pool: Pool;
+
+    /// Returns a fairing that attaches this connection pool to the server.
+    fn fairing() -> Fairing<Self>;
+
+    /// Direct shared access to the underlying database pool
+    fn pool(&self) -> &Self::Pool;
+
+    /// get().await returns a connection from the pool (or an error)
+    fn get(&self) -> BoxFuture<'_, Result<Connection<Self>, <Self::Pool as Pool>::GetError>> {
```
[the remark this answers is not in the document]
jebrosen: A previous design of this did use `async_trait`, but you expressed a wish to avoid `async_trait`. Although maybe I misunderstood where/why you wanted to avoid it?

examples/databases/src/sqlx.rs
```diff
@@ -1,13 +1,18 @@
...
 use futures::stream::TryStreamExt;
 use futures::future::TryFutureExt;
-use sqlx::ConnectOptions;
 
-type Db = sqlx::SqlitePool;
+#[derive(rocket_db_pools::Database)]
+#[database(name = "sqlx")]
+struct Db(sqlx::SqlitePool);
+
+type Connection = rocket_db_pools::Connection<Db>;
```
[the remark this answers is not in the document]
jebrosen: The syntax with an associated type doesn't work for this, since it must be fully qualified: <https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=dc2bdcdce2130b927224fec4f20115aa>.

contrib/db_pools/lib/src/database.rs
jebrosen: The main obstacle here is that name must be an `&'static str`. But I think that can still be worked around fairly neatly by shuffling around some of the code generation.

contrib/db_pools/lib/src/pool.rs
jebrosen: `InitError` is `rocket_db_pools::Error`, which represents either bad configuration or the database-specific error. `GetError` could be the same error type, but it would always be the non-configuration variant.

examples/databases/src/sqlx.rs
jebrosen: Leaking the abstraction in this particular way (putting the derived database type directly in managed state) was intentional, because I thought it was much simpler to understand. `Db::pool` and `Db::conn` already take `&self`, so this would logically be adding `Db::pool_from_rocket(&rocket)` and `Db::conn_from_rocket(&rocket).await`. Happy to make that change, though.

contrib/db_pools/lib/src/database.rs
jebrosen: Can you describe what changes should be made to the division of work between the trait, impls, and `fairing()` to achieve this? In `sync_db_pools` today, each implementation of `Poolable` must call `Config::from(db_name, rocket)?`. Similarly, I hoped to avoid requiring each implementation of `Pool` (or associated configuration structure) to individually write out e.g. `Figment::from(rocket.figment()).focus(&db_name)` if that boilerplate could be brought into the fairing instead.

contrib/db_pools/lib/src/pool.rs
jebrosen: For some database drivers (e.g. `deadpool-postgres`), the error type returned by initialization vs by `get()` are actually different. In that case one error type is a superset of the other, but I wouldn't want to assume that about future drivers.

contrib/db_pools/lib/src/database.rs
```diff
@@ -0,0 +1,117 @@
...
+pub trait Database: Sized + Send + Sync + 'static {
+    /// The name of this connection pool in the configuration.
+    const NAME: &'static str;
+
+    /// The underlying connection type returned by this pool.
+    /// Must implement [`Pool`].
+    type Pool: Pool;
+
+    /// Returns a fairing that attaches this connection pool to the server.
+    fn fairing() -> Fairing<Self>;
```
SergioBenitez, line 20: This `Fairing` type feels pretty weird to me. It seems to me that a `Database` is itself a `Fairing`. Or something like that. There's something not quite right about the setup right now with respect to this fairing, the `Fairing` type, and the `Connection` type.

contrib/db_pools/lib/src/database.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+/// The database fairing for pool types created with the `pool!` macro.
+pub struct Fairing<D: Database>(&'static str, fn(D::Pool) -> D);
+
+impl<D: Database> Fairing<D> {
+    /// Create a new database fairing with the given constructor.  This
+    /// constructor will be called to create an instance of `D` after the pool
+    /// is initialized and before it is placed into managed state.
+    pub fn new(fairing_name: &'static str, ctor: fn(D::Pool) -> D, ) -> Self {
```
SergioBenitez, line 92: This is really weird.

contrib/db_pools/lib/src/database.rs
```diff
@@ -0,0 +1,117 @@
...
+    async fn on_ignite(&self, rocket: Rocket<Build>) -> Result<Rocket<Build>, Rocket<Build>> {
+        let pool = match <D::Pool>::initialize(D::NAME, &rocket).await {
+            Ok(p) => p,
+            Err(e) => {
+                error!("error initializing database connection pool: {}", e);
+                return Err(rocket);
+            }
+        };
+
+        Ok(rocket.manage((self.1)(pool)))
```
SergioBenitez, line 115: This will panic if one is already being managed. We should probably check and emit a nicer error.

contrib/db_pools/lib/src/pool.rs
SergioBenitez: Sure, but that doesn't mean the pool integration can't wrap those into one type.

I thin there's a lot of effort right now spent on trying to creates these reusable structures like `Config` and `Error` when, in reality, each database driver (at least those from each library) will be pretty different. I would advocate for eschewing both of these types entirely and instead creating an `Error` and `Config` for each database variant/library as needed. With `Config` specifically, this is pretty important.

contrib/db_pools/lib/src/pool.rs
```diff
@@ -0,0 +1,287 @@
...
+    /// integrations provided in `rocket_db_pools` use [`Config`], which
+    /// accepts a (required) `url` and an (optional) `pool_size`.
+    ///
+    /// ## Errors
+    ///
+    /// This method returns an error if the configuration is not compatible, or
+    /// if creating a pool failed due to an unavailable database server,
+    /// insufficient resources, or another database-specific error.
+    async fn initialize(db_name: &str, rocket: &Rocket<Build>)
+        -> Result<Self, Error<Self::InitError>>;
```
SergioBenitez, line 73: This shouldn't be formatted this way. At worst:
```suggestion
    async fn initialize(
        db_name: &str,
        rocket: &Rocket<Build>
    ) -> Result<Self, Error<Self::InitError>>;
```

But I really think this needs to be `Self::Error`.

examples/databases/src/sqlx.rs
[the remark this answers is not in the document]
SergioBenitez: Not of the trait, but of the type directly.

examples/databases/src/sqlx.rs
SergioBenitez: The problem is that this presents two different interfaces: an abstracted `Db::fairing()` and now a `Db` in managed state. If the user had instead been asked to do `.manage(Db::init()?)` or something to that effect, this would be fine. But otherwise, `.attach(Db::fairing())` says nothing about something going into managed state.

SergioBenitez (changes requested):
Something feels quite wrong here with the way the trait is defined:

  * It's _very_ weird that there are two error types in `Database`. There really needs to be one.
  * We need to specify configs per database. The existence of a common `Config` type in 0.4 was a mistake rooted in the inability to generalize configuration. With Figment, this mistake doesn't need to be made again. Each database type (or library, or wherever the distinction lies) needs to define its own configuration structure. This immediately resolves all weirdness around configuration and allows the user to configure all aspects of each database.
  * The `Fairing` struct is very odd. It effectively demands that the database type is `Fn(Self::Pool) -> Self` without having that fact be codified or enforced anywhere. Even so, _that_ feels like a strange restriction. It seems we really just want some way to go from a database pool type to a structure. Perhaps the bound we want on `Database` is `From<Self::Pool>` (which would imply generating such an implementation), or simply we want a `fn new(pool: Db::Pool) -> Self;` (or `from` or other) on `Database`. I'm not sure exactly which of these is the right approach, but the current `Fairing` situation feels like the wrong one.
  * Something needs to be done about the effective need for a type alias. Whether that's generating an inherent associated type or something similar, I don't know. But it needs to be trivial to get a connection for the given database without adding any more code.

examples/databases/src/sqlx.rs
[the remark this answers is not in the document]
jebrosen: Unfortunately, inherent associated types [are unstable](https://github.com/rust-lang/rust/issues/8995) :(

contrib/db_pools/lib/src/database.rs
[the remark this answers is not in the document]
jebrosen: Okay... in that case, should the derive generate an inherent `async` method, an `#[async_trait]` trait method, or both?

contrib/db_pools/codegen/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+readme = "../README.md"
+keywords = ["rocket", "framework", "database", "pools"]
+license = "MIT OR Apache-2.0"
+edition = "2018"
+
+[lib]
+proc-macro = true
+
+[dependencies]
+devise = { git = "https://github.com/SergioBenitez/Devise.git", rev = "f2431b" }
```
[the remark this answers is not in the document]
jebrosen: Weird! Not sure what branch I could have copy/pasted that version from.

contrib/db_pools/lib/src/pool.rs
jebrosen: > when, in reality, each database driver (at least those from each library) will be fairly distinct

So, I tried using a separate error type for each implementation and ran into every implementation of `Pool::initialize` needing to return a `figment::Error`. I feel again like I'm completely misunderstanding something, because the single `Error` type with 3 variants now seems like a lot less effort than writing and documenting the same 3 variants of a separate error enum for each database.

paulzhang5511:
How to test a query statement with `rocket_db_pools`?

SergioBenitez:
Merged in https://github.com/SergioBenitez/Rocket/commit/f7f068be11efb33844d308dec7bcfcd770e2bffa with design changes in https://github.com/SergioBenitez/Rocket/commit/5b1a04deabc75981fde566c4381b4cba0d3b7cc8.

PR #1697, opened 2021-06-14
Update handlebars to 4.0 and fix partial inclusion

sunng87:
This patch attempts to update built-in handlebars to 4.0 and fixes partial usage in example.

Actually to use a variable as parent partial name is a special use case. I suggest to simplify this and make the example focus on handlebars API and basic features. If you are ok I will update the example for that.


SergioBenitez:
> Actually to use a variable as parent partial name is a special use case.

Can you use a static string to name the parent? The only documentation and example for this that I've ever seen uses a variable name.

sunng87:
> Can you use a static string to name the parent?

Yes, that's the part I'm planning to change.

SergioBenitez:
> > Can you use a static string to name the parent?
> 
> 
> 
> Yes, that's the part I'm planning to change. 

Okay, I think that'd be fine. Nevertheless, I'm hesitant to update `handlebars` to 4. I'm concerned that the release still has some bugs to shake out, and that the number of breaking changes is much higher than that CHANGELOG alludes to.

sunng87:
4.0 fixed a few issues that incompatible with the original javascript version, for example `{{> (parent)}}`  was one of it. Some of these changes cannot be clearly reflected from CHANGELOG, because it's a change in subexpression resolution. But anyway the 4.x branch is the future. Semver can deal with the remaining bugs (I just pushed 4.0.1).

SergioBenitez:
> 4.0 fixed a few issues that incompatible with the original javascript version, for example `{{> (parent)}}`  was one of it. Some of these changes cannot be clearly reflected from CHANGELOG, because it's a change in subexpression resolution.

Why not? I don't understand why the CHANGELOG can't simply state: "This worked before but doesn't now. To do the same thing in v4, do this."

SergioBenitez:
@sunng87 Please let me know when the CHANGELOG is in a more compete state. At the moment, I think the potential footguns in v4 are too numerous to pass on to Rocket's users.

sunng87:
I have updated handlebars to 4.1 to address some whitespace and empty line issue introduced in 4.0

Changelog has been updated to reflect subexpression changes, too.

SergioBenitez:
Tests are failing. Can you rebase and check out the failures?

PR #1702, opened 2021-06-15
Remove reference to Future from Responder trait docs.

timando:


core/lib/src/response/responder.rs
```diff
@@ -22,9 +22,6 @@ use crate::request::Request;
 ///
 /// # Return Value
 ///
-/// A `Responder` returns a `Future` whose output type is an `Ok(Response)` or
```
jebrosen, line 25: We shouldn't remove the entire sentence, just the part about `Future`. This is what it said in Rocket v0.4:

"A `Responder` returns an `Ok(Response)` or an `Err(Status)`:"

jebrosen (changes requested):

jebrosen (approved):

jebrosen:
Thanks! Merged this in c3dc718.

PR #1708, opened 2021-06-16
Add missing documentation about `#[database]`-generated `run()`

### LINKED ISSUE #1704: rocket_sync_db_pools::database doesn't document generating run method
## Bug Reports

Bug reports _must_ include:

  1. The version of Rocket you're using. Ensure it's the latest, if possible.

         0.5-rc

  2. The operating system (distribution and version) where the issue occurs.

         NixOS 21.05.906.cedcf2565c6

  3. A brief description of the bug that includes:
    * The nature of the bug.
    * When the bug occurs.
    * What you expected vs. what actually happened.

     https://docs.rs/rocket_sync_db_pools/0.1.0-rc.1/rocket_sync_db_pools/attr.database.html says that macro database generates two methods, but it actually generates three, see https://docs.rs/rocket_sync_db_pools_codegen/0.1.0-rc.1/src/rocket_sync_db_pools_codegen/database.rs.html.

     This bug always occurs. I expected the documentation to say that the `run` method is being generated.

  4. How you uncovered the bug. Short, reproducible tests are especially useful.

      By trying to figure out where `run` method comes from.

  5. Ideas, if any, about what Rocket is doing incorrectly.

      Documentation should be updated to mention that it creates three methods.

JarvisCraft:
This resolves #1704.

### DIFF
--- a/contrib/sync_db_pools/codegen/src/lib.rs
+++ b/contrib/sync_db_pools/codegen/src/lib.rs
@@ -30,7 +30,7 @@ use proc_macro::TokenStream;
 /// retrieves a connection from the database pool or fails with a
 /// `Status::ServiceUnavailable` if connecting to the database times out.
 ///
-/// The macro also generates two inherent methods on the decorated type:
+/// The macro also generates three inherent methods on the decorated type:
 ///
 ///   * `fn fairing() -> impl Fairing`
 ///
--- a/contrib/sync_db_pools/codegen/src/lib.rs
+++ b/contrib/sync_db_pools/codegen/src/lib.rs
@@ -41,6 +41,11 @@ use proc_macro::TokenStream;
 ///
 ///     Retrieves a connection wrapper from the configured pool. Returns `Some`
 ///     as long as `Self::fairing()` has been attached.
+//!
+//!   * `async fn run<R: Send + Static>(&self, impl FnOnce(&mut Db) -> R) -> R`
+//!
+//!     Asynchronously runs the specified function returning its result of type `R`
+//!     using the instance of the connection where `Db` is the underlying `Poolable` type.
 ///
 /// [`FromRequest`]: rocket::request::FromRequest
 #[proc_macro_attribute]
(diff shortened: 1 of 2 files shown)


contrib/sync_db_pools/codegen/src/lib.rs
```diff
@@ -41,6 +41,11 @@ use proc_macro::TokenStream;
 ///
 ///     Retrieves a connection wrapper from the configured pool. Returns `Some`
 ///     as long as `Self::fairing()` has been attached.
+//!
+//!   * `async fn run<R: Send + Static>(&self, impl FnOnce(&mut Db) -> R) -> R`
```
jebrosen, line 45: How did you determine this method signature? It doesn't match the generated signature and is missing a few bounds.

contrib/sync_db_pools/codegen/src/lib.rs
```diff
@@ -41,6 +41,11 @@ use proc_macro::TokenStream;
 ///
 ///     Retrieves a connection wrapper from the configured pool. Returns `Some`
 ///     as long as `Self::fairing()` has been attached.
+//!
```
jebrosen, line 44: We should also change "two" to "three" up above while we're here.

jebrosen (changes requested):
Drat, I did update the usage/examples but neglected to document this properly. Thanks for filing this.

contrib/sync_db_pools/codegen/src/lib.rs
JarvisCraft: Thanks, fixed in 3e0a6df

contrib/sync_db_pools/codegen/src/lib.rs
JarvisCraft: My bad, I relied on https://github.com/SergioBenitez/Rocket/blob/5a4e66ec439411d30f16e5c045f8e4986f5883a4/contrib/sync_db_pools/codegen/src/database.rs#L104-L107 but have incorrectly attempted to simplify the signature (i.e. thought that `FnOnce(&mut X) -> R + Send + 'static`'s `Send + 'static` relate to `R` rather than `F`).

JarvisCraft: Attempted to fix this in cea7264

jebrosen:
Thanks! Merged in d9858cf, with a slight rephrasing after further consideration.

PR #1714, opened 2021-06-21
Update known_media_types.rs -  "application/sep+xml" as known media type.

gustavowd:
Add support for "application/sep+xml" as known media type. It is used in the IEEE Standard for Smart Energy Profile Application Protocol (IEEE 2030.5).


jebrosen:
https://www.iana.org/assignments/media-types/application/sep+xml indicates a "required parameter" to specify the version (currently either `-S0` or `+S0`). Can you comment on this, e.g. is either or both more commonly used in practice?

This media type might need to be split into two media types under the current design, or to be somehow handled differently on the request (with `format=`) and response sides. This media type may be more complicated than we want to support in Rocket directly, in which case it would be better handled by a Data guard than by `format=`.

gustavowd:
@jebrosen the last standard ((IEEE 2030.5-2018) only demands for the "required parameter" for the media type "application/sep-exi".

"When specifying an “application/sep-exi” media type, an extensibility level (“level”) SHALL be specified in the HTTP Accept header for schema version negotiation. "

The “application/sep+xml” do not demand such parameter. It recommends the parameter to be level -S1, though.

Here is an example:
Accept: application/sep+xml; level=-S1

"The extensibility level defines the base schema and its capability for arbitrary extension. The extensibility level is one of “-S1” or “+S1”. The S1 indicates the base schema version: IEEE Std 2030.5-2018. “-S1” indicates the node does not accept arbitrary tags that are not defined in the base schema. Due to strict memory constraints, some nodes may not be able to parse invalid parts of an EXI document encoded using the built-in grammar. To avoid such errors, a node may declare its inability to receive arbitrary extensions using the “-” (minus) prefix in the extensibility level.";

SergioBenitez:
I'm sorry I didn't get to this until now. If you'd like to move forward with this, could you provide a screenshot from IEEE 2030.5 indicating that what IANA considers a requirement isn't actually so? I'm closing this in case you don't care to move forward, but please let me know if you do.

PR #1717, opened 2021-06-21
update.cargo.

Saan37:
rust errors for global worldwide


Saan37:
[Unicode whitespace character]()

SergioBenitez:
I don't know if you're doing this intentionally or not, but please don't submit pull requests like this.

PR #1730, opened 2021-06-25
Drop `tx` channel handle before task `await` to prevent deadlocking local Client

### LINKED ISSUE #1729: Local asynchronous `Client` never finishes trying to read JSON data
Version: `0.5.0-rc.1`
OS: Windows 10

Code to reproduce:
```rust
use rocket::local::asynchronous::Client;
use rocket::serde::json::Json;

#[rocket::get("/")]
fn hello_world() -> Json<i32> {
    Json(5)
}

#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    let rocket = rocket::build().mount("/", rocket::routes![hello_world]);

    let client = Client::tracked(rocket).await?;
    assert_eq!(
        client
            .get("/")
            .dispatch()
            .await
            .into_string()
            .await
            .as_deref(),
        Some("5")
    );
    assert_eq!(
        client.get("/").dispatch().await.into_json::<i32>().await,
        Some(5)
    );

    Ok(())
}
```
results in
```
Configured for debug.
   >> address: 127.0.0.1
   >> port: 8000
   >> workers: 8
   >> ident: Rocket
   >> keep-alive: 5s
   >> limits: bytes = 8KiB, data-form = 2MiB, file = 1MiB, form = 32KiB, json = 1MiB, msgpack = 1MiB, string = 8KiB
   >> tls: disabled
   >> temp dir: C:\Users\<snip>\AppData\Local\Temp\
   >> log level: normal
   >> cli colors: true
   >> shutdown: ctrlc = true, force = true, grace = 2s, mercy = 3s
Routes:
   >> (hello_world) GET /
Fairings:
   >> Shield (liftoff, response, singleton)
Shield:
   >> X-Frame-Options: SAMEORIGIN
   >> X-Content-Type-Options: nosniff
   >> Permissions-Policy: interest-cohort=()
Rocket has launched into local orbit
GET /:
   >> Matched: (hello_world) GET /
   >> Outcome: Success
GET /:
   >> Matched: (hello_world) GET /
   >> Outcome: Success
```
and never completes. This issue doesn't occur with `.into_string()` so it seems related to the `blocking_read` never dropping the `tx` after the loop so the `rx` keeps waiting for more data?

https://github.com/SergioBenitez/Rocket/blob/c74bcfd40a47b35330db6cafb88e4f3da83e0d17/core/lib/src/local/asynchronous/response.rs#L132-L183

repnop:
Fixes #1729.

This change ensures that the `rx` isn't deadlocked waiting for more writes that will never arrive after `tx` is finished sending data in the loop.

### DIFF
--- a/core/lib/src/local/asynchronous/response.rs
+++ b/core/lib/src/local/asynchronous/response.rs
@@ -179,6 +179,8 @@ impl LocalResponse<'_> {
             }
         }
 
+        drop(tx);
+        
         reader.await.ok()
     }
 


repnop:
I'll have to do that later on today but can do :)

SergioBenitez:
No worries! Merged in c3ee34e295f30099f5fb16069bf914e6a5c3e0ab with the fix and regression tests.

repnop:
oh, thanks! :D

PR #1745, opened 2021-07-04
Fix ROCKET_CLI_COLORS syntax

silas:
The documented `ROCKET_CLI_COLORS=off` fails with the following error:

```
Error: Rocket configuration extraction from provider failed.
   >> invalid value string "off", expected true or false
   >> for key CLI_COLORS
   >> in `ROCKET_` environment variable(s)
```


jebrosen (approved):

jebrosen:
Good catch! Merged this in 31d06ee.

PR #1775, opened 2021-07-20
Callout `json` feature in guide

the10thWiz:
Let me know if there is any improvements to make here. I think there may be other places to call out Rocket feature gates, and I will look for them, let me know if you want them in this PR or a separate PR.

### DIFF
--- a/site/guide/4-requests.md
+++ b/site/guide/4-requests.md
@@ -678,7 +678,7 @@ struct Task<'r> {
 fn new(task: Json<Task<'_>>) { /* .. */ }
 ```
 
-! NOTE: Json support requires the `json` feature flag to be enabled in Rocket
+! NOTE: JSON support requires the `json` feature flag to be enabled in Rocket
 
   Rocket has intentionally chosen to place Json support, as well as a few other formats
   behind feature flags.
--- a/site/guide/4-requests.md
+++ b/site/guide/4-requests.md
@@ -678,7 +678,12 @@ struct Task<'r> {
 fn new(task: Json<Task<'_>>) { /* .. */ }
 ```
 
-See the [JSON example](@example/serialization/src/json.rs) on GitHub for a complete example.
+! NOTE: Json support requires the `json` feature flag to be enabled in Rocket
+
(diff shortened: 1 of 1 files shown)


site/guide/4-requests.md
```diff
@@ -678,7 +678,12 @@ struct Task<'r> {
 fn new(task: Json<Task<'_>>) { /* .. */ }
 ```
 
-See the [JSON example](@example/serialization/src/json.rs) on GitHub for a complete example.
+! NOTE: Json support requires the `json` feature flag to be enabled in Rocket
```
maxdeviant, line 681: Could we use proper casing for "JSON" when used in a sentence?

```suggestion
! NOTE: JSON support requires the `json` feature flag to be enabled in Rocket
```

the10thWiz: I've seen both, but I guess we should be consistent. Thanks for pointing this out.

site/guide/4-requests.md
```diff
@@ -678,7 +678,12 @@ struct Task<'r> {
 fn new(task: Json<Task<'_>>) { /* .. */ }
 ```
 
-See the [JSON example](@example/serialization/src/json.rs) on GitHub for a complete example.
+! NOTE: JSON support requires the `json` feature flag to be enabled in Rocket
+
+  Rocket has intentionally chosen to place JSON support, as well as a few other formats
+  behind feature flags.
```
jebrosen, line 684: Can we add a link to <https://api.rocket.rs/master/rocket/#features> somewhere in this sentence, which explains the available features and shows how to enable them?

jebrosen (changes requested):
This seems like the first, and most likely, place for a new user to need to enable crate features, so I'm all in favor of adding a note here.

site/guide/4-requests.md
the10thWiz: Yes, absolutely. I added a sentence about the other features, and linked to the api reference. I also added a small code example to demonstrate how to enable the feature in Cargo.toml.

SergioBenitez:
Merged in a05b3e138c558cd5cdb4ebc3a560a545513ee682.

PR #1788, opened 2021-07-26
Add support for MacAddress params

hannesschulze:
Add support for using `MacAddress` from the `mac_address` crate as a dynamic path segment. This would introduce `mac_address` as an optional dependency.

### DIFF
--- a/core/lib/Cargo.toml
+++ b/core/lib/Cargo.toml
@@ -31,6 +31,7 @@ uuid = ["uuid_", "rocket_http/uuid"]
 # Serialization dependencies.
 serde_json = { version = "1.0.26", optional = true }
 rmp-serde = { version = "0.15.0", optional = true }
+mac_address = { version = "1.1.1", optional = true }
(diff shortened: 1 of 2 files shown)


SergioBenitez:
This isn't something that should exist in `rocket`. It's simply unsustainable (and unnecessary) to try to implement traits for external crates. If you maintain the crate, consider implementing the trait there under a feature flag. Otherwise, you can write wrapper type that implements the trait.

hannesschulze:
Thank you for your swift response – that's understandable, I'll look into it :)

repnop:
Hi, `mac_address` crate owner here! I just happened across this PR, but @hannesschulze if you'd be interested in putting in a PR to add optional support for this in the crate, feel free to do so and I can review, merge, and publish that shortly after :) though to reduce churn, I think I'd ideally want `rocket` to publish `0.5` before merging it so it can be compatible with future `0.5.x` publishes of `rocket` without needing to go and update it afterwards.

PR #1800, opened 2021-07-31
Fix typo in docs `iff` -> `if`

FalseDev:


SergioBenitez:
This is correct. "If and only if"

PR #1802, opened 2021-08-01
Static files ,IfMatch/IfModifiedSince/ComputeRange/ComputeIfRange

mokeyish:
Related issues:
- https://github.com/SergioBenitez/Rocket/issues/95
- https://github.com/SergioBenitez/Rocket/pull/1376
- https://github.com/SergioBenitez/Rocket/issues/989

1.  [rust-embed](https://github.com/pyros2097/rust-embed) - [examples](https://github.com/pyros2097/rust-embed/blob/master/examples/rocket.rs)

    ```rust
    
    use std::path::PathBuf;
    use std::ffi::OsStr;
    use std::io::Cursor;
    use rocket::{
        http::{
            ContentType,
            DateTimeOffset,
            EntityTagHeaderValue
        },
        Request,
        response::{
            self, Responder,
        },
        fs::StaticFileContext
    };
    
    
    #[derive(rust_embed::RustEmbed)]
    #[folder = "dist"]
    struct Asset;
    
    pub struct EmbeddedFile {
        file: rust_embed::EmbeddedFile,
        content_type: Option<ContentType>,
    }
    
    
    impl EmbeddedFile {
        pub fn get(file_name: PathBuf) -> Option<Self> {
            let file: Option<rust_embed::EmbeddedFile>  = file_name.to_str().and_then(Asset::get);
    
            let file = file?;
    
            let content_type: Option<ContentType> = file_name
                .extension()
                .and_then(OsStr::to_str)
                .and_then(ContentType::from_extension);
            Some(Self { file, content_type })
        }
    }
    
    impl<'r> Responder<'r, 'static> for EmbeddedFile {
        fn respond_to(self, req: &'r Request<'_>) -> response::Result<'static> {
            let Self{file, content_type} = self;
            let mut static_ctx = StaticFileContext::from((
                file.data.len() as u64, content_type,
                EntityTagHeaderValue::from(&file.metadata.sha256_hash()),
                file.metadata.last_modified().map(|t| t * 1000).map(DateTimeOffset::from).unwrap_or_else(DateTimeOffset::now)
            ));
            static_ctx.comprehend_request_headers(req);
            static_ctx.proceed(req, |(builder, pos ,len)| {
                let mut data = Cursor::new(file.data);
                if pos > 0 {
                    data.set_position(pos);
                }
                builder.sized_body(len as usize, data);
            })
        }
    }
    
    ```

### DIFF
--- a/core/http/src/header/typed_header/content_range_header_value.rs
+++ b/core/http/src/header/typed_header/content_range_header_value.rs
@@ -1,20 +1,29 @@
-use super::{header_utilities, header_names, RangeItemHeaderValue, Header};
+use super::{header_utilities, header_names, Header};
 
-
+/// Represents a `Content-Range` response HTTP header.
 pub struct ContentRangeHeaderValue {
+    /// The start of the range.
     pub from: Option<u64>,
+
+    /// The end of the range.
     pub to: Option<u64>,
+
+    /// The total size of the document.
     pub length: Option<u64>,
+
+    /// The unit in which ranges are specified.
     pub unit: String,
 }
 
 impl ContentRangeHeaderValue {
+    ///  Gets a value that determines if `length` has been specified.
     pub fn has_length(&self) -> bool {
         self.length.is_some()
     }
 
+    /// Gets a value that determines if `from` and `to` have been specified.
     pub fn has_range(&self) -> bool {
-        self.from.is_some()
+        self.from.is_some() && self.to.is_some()
     }
 }
 
@@ -77,11 +86,11 @@ impl ToString for ContentRangeHeaderValue {
     }
 }
 
-impl<'h> Into<Header<'h>> for ContentRangeHeaderValue {
-    fn into(self) -> Header<'h> {
+impl From<ContentRangeHeaderValue> for Header<'_> {
+    fn from(v: ContentRangeHeaderValue) -> Self {
         Header {
             name: header_names::CONTENT_RANGE.into(),
-            value: self.to_string().into()
+            value: v.to_string().into()
         }
     }
 }
--- a/core/http/src/header/typed_header/content_range_header_value.rs
+++ b/core/http/src/header/typed_header/content_range_header_value.rs
@@ -1,6 +1,7 @@
 use super::{header_utilities, header_names, Header};
 
 /// Represents a `Content-Range` response HTTP header.
+#[derive(Clone)]
 pub struct ContentRangeHeaderValue {
     /// The start of the range.
     pub from: Option<u64>,
--- a/core/http/src/header/typed_header/content_range_header_value.rs
+++ b/core/http/src/header/typed_header/content_range_header_value.rs
@@ -51,15 +51,6 @@ impl From<(u64, u64)> for ContentRangeHeaderValue {
     }
 }
 
-impl From<&RangeItemHeaderValue> for ContentRangeHeaderValue {
-    fn from(range: &RangeItemHeaderValue) -> Self {
-        let start = range.from.unwrap();
-        let end = range.to.unwrap();
-        let len = end - start + 1;
-        (start, end, len).into()
-    }
-}
-
 impl ToString for ContentRangeHeaderValue {
     fn to_string(&self) -> String {
         let mut s = String::new();
--- a/core/http/src/header/typed_header/content_range_header_value.rs
+++ b/core/http/src/header/typed_header/content_range_header_value.rs
@@ -70,9 +70,7 @@ impl ToString for ContentRangeHeaderValue {
         if self.has_range(){
             s.push_str(self.from.unwrap().to_string().as_str());
             s.push('-');
-            if let Some(to) = self.to {
-                s.push_str(to.to_string().as_str());
-            }
+            s.push_str(self.to.unwrap().to_string().as_str());
         } else {
             s.push('*');
         }
--- a/core/http/src/header/typed_header/content_range_header_value.rs
+++ b/core/http/src/header/typed_header/content_range_header_value.rs
@@ -1,0 +1,96 @@
+use super::{header_utilities, header_names, RangeItemHeaderValue, Header};
+
+
+pub struct ContentRangeHeaderValue {
+    pub from: Option<u64>,
+    pub to: Option<u64>,
+    pub length: Option<u64>,
+    pub unit: String,
+}
+
+impl ContentRangeHeaderValue {
+    pub fn has_length(&self) -> bool {
+        self.length.is_some()
+    }
+
+    pub fn has_range(&self) -> bool {
+        self.from.is_some()
+    }
+}
+
+impl From<u64> for ContentRangeHeaderValue {
+    fn from(len: u64) -> Self {
+        Self {
+            from: None,
+            to: None,
+            length: Some(len),
+            unit: header_utilities::BYTES_UNIT.to_string()
+        }
+    }
+}
+
+impl From<(u64, u64, u64)> for ContentRangeHeaderValue {
+    fn from((from, to, len): (u64, u64, u64)) -> Self {
+        Self {
+            from: Some(from),
+            to: Some(to),
+            length: Some(len),
+            unit: header_utilities::BYTES_UNIT.to_string()
+        }
+    }
+}
+/// Scenario: "Content-Range: bytes 12-34/*"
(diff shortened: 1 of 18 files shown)


core/lib/src/fs/static_file.rs
```diff
@@ -0,0 +1,494 @@
...
+    }
+
+    fn send_range<'r>(mut self, _req: &'r Request<'_>) -> response::Result<'static> {
+        // do range
+        if let Some(ref range) = self.range {
+            let mut builder = Response::build();
+            let content_range_header = ContentRangeHeaderValue::from(range);
+            let from = content_range_header.from.unwrap();
+            let length = content_range_header.length.unwrap();
+            builder.header(content_range_header);
```
mateuszkj, line 397: Total length of the file is wrong. Below is response from server for file with size 79356416 bytes:

```
  content-range: bytes 29151232-79356415/50205184
```

According to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range after `/` there should be total file size, not remaining.

`wget` works after changing above lines to:
```rust
let mut builder = Response::build();
let from = range.from.unwrap();
let to = range.to.unwrap();
let content_range_header = ContentRangeHeaderValue::from((from, to, self.len));
let length = to - from + 1;
builder.header(content_range_header);
```

BTW. great PR, this is exactly what I need today :)

mateuszkj (commented):
https://github.com/SergioBenitez/Rocket/pull/1802/files#diff-3e244947af99be0afc53cf2b8214c42022e6869aab7cdc349068c21ebfa78889R397

core/lib/src/fs/static_file.rs
mokeyish: Yes, you are right.

core/lib/src/fs/static_file.rs
```diff
@@ -0,0 +1,495 @@
...
+{
+    Unspecified,
+    NotModified,
+    ShouldProcess,
+    PreconditionFailed,
+}
+
+#[repr(u8)]
+#[derive(EnumFlags, Copy, Clone, Eq, PartialEq)]
+enum RequestType {
```
mateuszkj, line 490: Can you get rid of `EnumFlags` ? I got undefined behavior  in release mode (in debug mode everything works fine).

But on release i got strange behavior e.g.
```
---request begin---
GET /software/myfile HTTP/1.1
User-Agent: Wget/1.20.3 (linux-gnu)
Accept: */*
Accept-Encoding: identity
Host: 127.0.0.1:9854
Connection: Keep-Alive

---request end---
Żądanie HTTP wysłano, oczekiwanie na odpowiedź... 
---response begin---
HTTP/1.1 416 Range Not Satisfiable
content-range: bytes */79355392
server: Rocket
x-frame-options: SAMEORIGIN
x-content-type-options: nosniff
permissions-policy: interest-cohort=()
content-length: 0
date: Wed, 04 Aug 2021 14:21:10 GMT
```

There should not be `Range Not Satisfiable` in this situation (It's 200 OK in debug mode).


After ding a little I found problem in lines:
```rust
if is_range_request {
    self.request_type |= RequestType::IsRange
} else {
    self.request_type -= RequestType::IsRange
}
```
When `is_range_request == false` it entered in first branch and `self.request_type` was set to `RequestType::IsRange`. After adding some `pirntln!()`` I got core dump.

I rewrote `RequestType` to 

```rust
#[derive(Copy, Debug, Clone, Eq, PartialEq, Default)]
struct RequestType {
    _is_range: bool,
}

impl RequestType {
    fn is_range(&self) -> bool {
        self._is_range
    }

    fn set_range(&mut self, v: bool) {
        self._is_range = v;
    }
}
```
and it's started to work in release and debug mode.

In fact you don't need enum `RequestType` at all. Fields other that `IsRange` are not used.

mateuszkj:
@mokeyish Could you fix also clippy warning? There is plenty of them, like: unussed_imports, let_and_return, len_zero, field_reassign_with_default, redundant_closure.

Just run `cargo clippy`

core/lib/src/fs/static_file.rs
mokeyish: Could you send your code above as PR to my  branch? I'am busy these days.

mokeyish:
> @mokeyish Could you fix also clippy warning? (only in added files) There is plenty of them, like: unussed_imports, let_and_return, len_zero, field_reassign_with_default, redundant_closure.
> 
> Just run `cargo clippy`

Crash in release mode has been fixed.  But `cargo clippy` seems a lot of work, Can you fix it and send a PR to my branch? I din't have enough time to fixe it, since I am busy these days .

core/http/src/header/typed_header/content_range_header_value.rs
```diff
@@ -0,0 +1,97 @@
...
+impl ToString for ContentRangeHeaderValue {
+    fn to_string(&self) -> String {
+        let mut s = String::new();
+        s.push_str(self.unit.as_str());
+        s.push(' ');
+
+        if self.has_range(){
+            s.push_str(self.from.unwrap().to_string().as_str());
+            s.push('-');
+            if let Some(to) = self.to {
```
rfael, line 73: You checked if `to` is some in `has_range` method

core/http/src/header/typed_header/http_rule_parser.rs
```diff
@@ -0,0 +1,252 @@
...
+    // If decimal values are not allowed, we pretend to have read the '.' character already. I.e. if a dot is
+    // found in the string, parsing will be aborted.
+    let mut have_dot = !allow_decimal;
+
+    // The RFC doesn't allow decimal values starting with dot. I.e. value ".123" is invalid. It must be in the
+    // form "0.123". Also, there are no negative values defined in the RFC. So we'll just parse non-negative
+    // values.
+    // The RFC only allows decimal dots not ',' characters as decimal separators. Therefore value "1,23" is
+    // considered invalid and must be represented as "1.23".
+    if input[current] == '.' {
```
rfael, line 76: There is no `imput` length validation

core/http/src/header/typed_header/http_rule_parser.rs
```diff
@@ -0,0 +1,252 @@
...
+    input: &[char],
+    start_index: usize,
+    open_char: char,
+    close_char: char,
+    supports_nesting: bool,
+    nested_count: &mut usize,
+    length: &mut usize
+) -> HttpParseResult {
+    *length = 0;
+    if input[start_index] != open_char {
```
rfael, line 145: `input` length should be checked too

core/http/src/header/typed_header/http_rule_parser.rs
```diff
@@ -0,0 +1,252 @@
...
+
+
+    HttpParseResult::InvalidFormat
+}
+
+/// quoted-pair = "\" CHAR
+/// CHAR = <any US-ASCII character (octets 0 - 127)>
+pub fn get_quoted_pair_length(input: &[char], start_index: usize, length: &mut usize) -> HttpParseResult {
+    *length = 0;
+    if input[start_index] != '\\' {
```
rfael, line 217: `input` length should be checked too

mokeyish:
> Are you sure about this? It seems to me it should be `if current >= input.len() || input[current] == '.' {` You should return 0 if `current` points out of `input` buffer. Same mistake is repeated in other changes in this commit.

Sorry :sweat_smile:

SergioBenitez:
I'm sorry I'm coming to this a year and a half later. Perhaps if I had reviewed it earlier, we would have made progress on this by now. Unfortunately, I cannot accept the PR as is because it reimplements functionality that already exists in Rocket. I do not understand why a new responder type `StaticFile` is being added when Rocket has (and has had) `NamedFile`. I also don't understand how these changes apply to anything existing, including `FileServer` and `NamedFile`. It feels to me that this is a dump of an external library into Rocket's core, which is generally not something we'd like to do.

Since this can live outside of Rocket and doesn't integrate into any of Rocket's existing infrastructure (and duplicates much of it), I'm closing this out. If you'd like to pursue this further, please submit a PR that begins with a design idea and perhaps a minimal implementation of it. For example, an implementation that includes two headers, one which has a value and one which doesn't, and an example enabling those headers for `FileServer` and `NamedFile`.

Thank you for the contribution.

PR #1804, opened 2021-08-03
Fix IPv6 based URI when rocket as launched

breard-r:
Currently, when Rocket is configured to listen on an IPv6 address, the URI displayed during the launch is invalid:

```
🚀 Rocket has launched from http://::1:8000
```

[RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) section 3.2.2 requires that a host identified by an IPv6 enclosed the IP within square brackets:

```
🚀 Rocket has launched from http://[::1]:8000
```

Since the RFC states "version 6 [...] or later", the test has been made using `.is_ipv4()`. This way, an eventual newer protocol would be correctly handled.

### DIFF
--- a/core/lib/src/rocket.rs
+++ b/core/lib/src/rocket.rs
@@ -1,6 +1,7 @@
 use std::fmt;
 use std::ops::{Deref, DerefMut};
 use std::convert::TryInto;
+use std::net::SocketAddr;
 
 use yansi::Paint;
 use either::Either;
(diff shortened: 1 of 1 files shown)


core/lib/src/rocket.rs
```diff
@@ -619,7 +619,11 @@ impl Rocket<Ignite> {
             rkt.fairings.handle_liftoff(&rkt).await;
 
             let proto = rkt.config.tls_enabled().then(|| "https").unwrap_or("http");
-            let addr = format!("{}://{}:{}", proto, rkt.config.address, rkt.config.port);
+            let addr = if rkt.config.address.is_ipv4() {
```
jebrosen, line 622: We could maybe use the existing `Display` impl for `SocketAddr`, which [does the right thing](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=d3ae03653bef6043a77034d6aec03aca).

jebrosen (approved):

core/lib/src/rocket.rs
breard-r: Indeed, thank you.
I've just add a commit to the PR in order to use it.

SergioBenitez:
Merged in 7a54a1293e61c09ecb934844c0c1dff2154932db. Thank you!

PR #1813, opened 2021-08-08
Remove Semi-colon from Docs

king-11:
The semicolon in the code snippet if used results in a compilation error hence should be removed in case someone copies them over for a run.


SergioBenitez:
Whether you need a semicolon or not is context dependent. The examples will not compile as changed, as evidenced by the CI failures.

king-11:
🤔 am not sure about context 😅 I was just trying to run the sample snippets and it was showing errors due to the semicolon. Can it work even with it?

Issue #1070, opened 2019-08-02
Provide a lower-level connection interface


jebrosen:
Currently Rocket's public API only provides a way to listen on TCP sockets (and Unix sockets have also been proposed) with optional TLS. Rather than adding support for every possible I/O layer in Rocket itself, it might be nice to provide an API for custom I/O types. Such an API now exists in `http/src/listener.rs`, but it is private. This issue is about making this or a similar API public, or integrating with another existing crate that provides such an API.

---

The original issue described here was implemented long ago without fanfare. Here's the original text:

> hyper's `AddrIncoming` is not sufficient for our needs because it does not support TLS. We should implement an alternative to `AddrIncoming` that abstracts over plain vs TLS and (in the future) over TCP vs Unix sockets. This type should support most or all of the connection-level options hyper does, such as timeouts. It also needs to have a `remote_addr` method.
> 
>See also #594, which implements some changes to the configuration system to handle non-TCP addresses.

jebrosen:
It would also be nice to support custom listeners. This would allow (among other things) custom TLS implementations using something other than `rustls`, unix sockets, or FastCGI to be implemented outside of Rocket.

Even if it's not exposed (yet), I think it's worth implementing one of these APIs internally, because it should make it somewhat easier to abstract over TLS and unix sockets which we definitely do want to support.

I have come up with two main ideas for a design of this.

## "Rocket-driven":
```rust
rocket::ignite().listen_on::<ListenerType>(executor)

trait Listener {
    type Connection: Connection;
    
    async fn bind(address, keepalive, read_timeout, options...) -> Result<Self, io::Error>;
    fn local_addr(&self) -> Option<SocketAddr>;
    async fn accept(&mut self) -> Result<Self::Connection, io::Error>;
}
```

In this model, Rocket will call `bind` with the configured address and port. Its main weakness that I can see is the assumption there is an "address" to bind to, which would not be the case for unix sockets (a file path) or for something like CGI (stdin). #594 makes `Address` an enum of either `Tcp` or `Unix`, which is another possibility to consider here.

Note that `listen_on` as I wrote it takes an `executor`, similar to what was implemented in #1097.

## User-driven:
```rust
rocket::ignite().listen_on(listener, executor)

trait Listener {
    type Connection: Connection;
    
    fn local_addr(&self) -> Option<SocketAddr>;
    fn set_keepalive(&mut self, keepalive);
    async fn accept(&mut self) -> Self::Connection;
}
```

[...]

PR #1820, opened 2021-08-13
Add hyper integration

MikailBag:
Solves #1070 without exposing too many APIs

### DIFF
--- a/core/lib/src/lib.rs
+++ b/core/lib/src/lib.rs
@@ -124,6 +124,7 @@ pub mod route;
 pub mod serde;
 pub mod shield;
 pub mod fs;
+pub mod service;
 
 // Reexport of HTTP everything.
 pub mod http {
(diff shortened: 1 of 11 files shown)


svenstaro:
This would be great to have. I'm currently trying to integrate Sentry but there's no single future I can call to attach the Hub to via `bind_hub()` so the integration is very lackluster right now. Also for tracing this would be really nice to have.

SergioBenitez:
My apologies for coming to this a year and a half later.

I don't think exposing `hyper` types is the approach we want to take on this one. I think the approach we want to take is one significantly less heavy-handed. Something like:

```rust
rocket::build()
    .listen(anything_implementing_listener)
```

One of the issues with doing this before was configuration. Given the monumental improvements in configuration, I think we can now say that the proper approach here would be to allow the listener to configure itself. The default TCP/TLS listener would read values as it does today, but otherwise the listener can do whatever it wants.

PR #1829, opened 2021-08-16
Fix invalid ROCKET_CLI_COLORS value in guide.

Sejsel:
This seems to have been changed from `off` to `false` in 1fb061496d0234f84495723e6c95590064778785. `0` works correctly.

When trying to use `RUST_CLI_COLORS=off` with `0.5.0-rc.1`, it fails with the following error:
```
Error: Rocket configuration extraction from provider failed.
   >> invalid value string "off", expected true or false
   >> for key CLI_COLORS
   >> in `ROCKET_` environment variable(s)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -137,6 +137,27 @@ We **strongly** advise all application authors to review this list carefully.
   * [`CookieJar`] `get()`s do not return cookies added during request handling. See
     [`CookieJar`#pending].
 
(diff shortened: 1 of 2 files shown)


jebrosen:
It looks like this change was already merged, in #1745.

Sejsel:
Ah, I didn't notice that the default branch is not the one with active development; sorry about that!

PR #1843, opened 2021-08-25
Update todo example's index.html.tera

xelivous:
The [example rework](https://github.com/SergioBenitez/Rocket/commit/50c9e88cf91cbcc6298cbd04282aebf6fba69b88#diff-8ef15e4657ed4039d24756f97d0af86db9576c69eb8d8ea7ef7e07700dc59cc7L26) changed `msg` to `flash`, but the template was never updated to match this new name.
Without this, flash messages never appear to the enduser.

### DIFF
--- a/examples/todo/src/tests.rs
+++ b/examples/todo/src/tests.rs
@@ -152,6 +152,26 @@ fn test_bad_form_submissions() {
 
         let mut cookies = res.headers().get("Set-Cookie");
         assert!(cookies.any(|value| value.contains("error")));
+        
+        // Check that the index now contains the flash message that we were expecting.
+        let body = client
+            .get("/")
+            .dispatch()
+            .await
+            .into_string()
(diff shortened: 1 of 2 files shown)


examples/todo/src/tests.rs
```diff
@@ -152,6 +152,26 @@ fn test_bad_form_submissions() {
...
+        
+        // Check that the index now contains the flash message that we were expecting.
+        let body = client
+            .get("/")
+            .dispatch()
+            .await
+            .into_string()
+            .await
+            .unwrap();
+        assert!(body.contains("Description cannot be empty."));
```
kolbma, line 164: There is no flash cookie set on request, why the flash should be shown?

SergioBenitez:
Merged in 13d8e7465571680e162fe2433bd31bbd168e2a64.

PR #1872, opened 2021-08-30
allow setting the route name

cataggar:
The route name is currently derived from the handler function name. This allows setting the route name. Here, the route name is set to `Hello`.

``` rust
#[get("/", name = "Hello")]
fn hello() -> &'static str {
    "hello"
}
```

In my use case, I use the Swagger operationId as the route name.

### DIFF
--- a/core/codegen/src/attribute/route/mod.rs
+++ b/core/codegen/src/attribute/route/mod.rs
@@ -422,6 +422,7 @@ fn incomplete_route(
         data: method_attribute.data,
         format: method_attribute.format,
         rank: method_attribute.rank,
+        name: method_attribute.name,
     };
 
     codegen_route(Route::from(attribute, function)?)
--- a/core/codegen/src/attribute/route/mod.rs
+++ b/core/codegen/src/attribute/route/mod.rs
@@ -330,6 +330,13 @@ fn codegen_route(route: Route) -> Result<TokenStream> {
     let rank = Optional(route.attr.rank);
     let format = Optional(route.attr.format.as_ref());
 
+    let route_name = match &route.attr.name {
+        Some(name) => quote! { #name },
+        None => {
+            quote! { stringify!(#handler_fn_name) }
+        }
+    };
+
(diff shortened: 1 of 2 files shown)


SergioBenitez:
Why not name the function `Hello`?

cataggar:
That was a bad example. Here is a more complete example. I some routes like:

``` rust
pub fn routes() -> Vec<rocket::Route> {
    routes![
        operations::list,
        locations::check_trial_availability,
        locations::check_quota_availability,
        private_clouds::list,
        private_clouds::list_in_subscription,
        private_clouds::get,
        private_clouds::create_or_update,
        private_clouds::update,
        private_clouds::delete,
        clusters::list,
        clusters::get,
        clusters::create_or_update,
        clusters::update,
        clusters::delete,
        private_clouds::list_admin_credentials,
        hcx_enterprise_sites::list,
        hcx_enterprise_sites::get,
        hcx_enterprise_sites::create_or_update,
        hcx_enterprise_sites::delete,
        authorizations::list,
        authorizations::get,
        authorizations::create_or_update,
        authorizations::delete
    ]
}
```

I want the route name function `operations::list` to be [Operations_List](https://github.com/Azure/azure-rest-api-specs/blob/21ffb64fdcbc0da039117af64f13c028c20d1286/specification/vmware/resource-manager/Microsoft.AVS/stable/2020-03-20/vmware.json#L103) to match its operationId instead of `list`.

SergioBenitez:
For what reason? What are you gaining, exactly?

cataggar:
`operationId` is what we use to name each route. I'm able to test all method + uri combinations route currently to a operationId.

kolbma:
Why you don't use `handler` and a custom mapping during the test?  
Runtime code then doesn't have to handle stuff you need for a test.

SergioBenitez:
I think this needs more motivation. I recommend opening up a "suggestion" issue with a concrete motivating use-case or two.

PR #1873, opened 2021-08-31
make Router public

cataggar:
Please make `Router` public. Related to #1872, I want to be to test that a `Router` routes correctly to a specific named route.


SergioBenitez:
We can't just mark things public. This is a recipe for unsupportable libraries. Please open an issue making a request that follows the template.

PR #1877, opened 2021-08-31
Fix warnings emitted by markdown linter.

silvioprog:


README.md
```diff
@@ -32,17 +32,17 @@ route won't get called, resulting in a 404 error.
...
-[Overview]: https://rocket.rs/overview/
-[Guide]: https://rocket.rs/guide/
-[API Documentation]: https://api.rocket.rs/rocket/
+- [Overview]: A brief look at what makes Rocket special.
+- [Quickstart]: How to get started as quickly as possible.
+- [Getting Started]: How to start your first Rocket project.
+- [Guide]: A detailed guide and reference to Rocket.
+- [API Documentation]: The "rustdocs".
+
+[quickstart]: https://rocket.rs/guide/quickstart
```
jebrosen, line 41: Do the links still work even if the case is mismatched? This looks like either a net neutral or net negative change, depending on the parser.

README.md
```diff
@@ -73,17 +73,17 @@ The `core` directory contains the three core libraries: `lib`, `codegen`, and
 `http` published as `rocket`, `rocket_codegen` and `rocket_http`, respectively.
 The latter two are implementations details and are reexported from `rocket`.
 
-### Testing
+## Testing
```
jebrosen, line 76: This changes the organization of the README. It's confusing that there are both cosmetic and functional changes in the same commit.

jebrosen (changes requested):
*What* linter in particular suggested these changes? There would be no purpose for us to merge the stylistic changes unless we could check or enforce the same style again in the future, and I'm concerned about the semantic changes.

silvioprog:
> _What_ linter in particular suggested these changes? There would be no purpose for us to merge the stylistic changes unless we could check or enforce the same style again in the future, and I'm concerned about the semantic changes.

This one: https://github.com/DavidAnson/markdownlint/blob/v0.23.1/doc/Rules.md

README.md
silvioprog: Yes, they still working fine. It can be checked at the forked repo: https://github.com/silvioprog/Rocket

README.md
silvioprog: Oops, sorry, it was a typo. I'm going to fix it...

silvioprog: Done.

SergioBenitez:
I see no reason to do this. What's more, because the markdown renderer we use for the website is unspecified, there's no way to test or even begin to ensure that such changes don't break rendering.

PR #1887, opened 2021-09-08
Bump `mongodb` dependency to `2.0.0`

isabelatkinson:
Bumps the `mongodb` dependency to the newly-released `2.0.0` version. This notably includes support for tokio 1.0; a full changelog can be viewed [here](https://github.com/mongodb/mongo-rust-driver/releases/tag/v2.0.0).

### DIFF
--- a/contrib/db_pools/lib/Cargo.toml
+++ b/contrib/db_pools/lib/Cargo.toml
@@ -53,6 +53,8 @@ optional = true
 
 [dependencies.mongodb]
 version = "2.0.0"
+default-features = false
+features = ["tokio-runtime"]
 optional = true
 
 [dependencies.sqlx]
--- a/contrib/db_pools/lib/Cargo.toml
+++ b/contrib/db_pools/lib/Cargo.toml
@@ -52,9 +52,7 @@ features = ["rt_tokio_1"]
 optional = true
 
 [dependencies.mongodb]
-version = "1"
-default-features = false
-features = ["tokio-runtime"]
+version = "2.0.0"
 optional = true
 
 [dependencies.sqlx]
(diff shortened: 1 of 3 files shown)


contrib/db_pools/lib/Cargo.toml
```diff
@@ -52,9 +52,7 @@ features = ["rt_tokio_1"]
 optional = true
 
 [dependencies.mongodb]
-version = "1"
-default-features = false
-features = ["tokio-runtime"]
+version = "2.0.0"
```
isabelatkinson, line 55: The only default feature for `mongodb` is `tokio-runtime`.

contrib/db_pools/lib/src/pool.rs
```diff
@@ -254,7 +254,6 @@ mod mongodb {
             opts.min_pool_size = config.min_connections;
             opts.max_pool_size = Some(config.max_connections as u32);
             opts.max_idle_time = config.idle_timeout.map(Duration::from_secs);
-            opts.wait_queue_timeout = Some(Duration::from_secs(config.connect_timeout));
```
isabelatkinson, line 257: This option was removed in the `2.0.0` release; see [RUST-757](https://jira.mongodb.org/browse/RUST-757) for more information.

contrib/db_pools/lib/Cargo.toml
jebrosen: I believe this was actually intentional (for all database dependencies)! This way, `rocket_db_pools` avoids accidentally enabling any other default features that are added in future versions, which not all projects may need/want.

contrib/db_pools/lib/src/pool.rs
jebrosen: The suggested client-side change appears to be to wrap the affected operations with a timeout manually. Which operations honored the `wait_queue_timeout`, and are any of those already covered by `connect_timeout`?

jebrosen (changes requested):
Thanks!

We should definitely make sure to replace that timeout, if it's necessary.

kolbma:
Bumped mongodb to 2.1 in https://github.com/rocket-org/Rocket/pull/3

SergioBenitez:
Looks like I inadvertently took care of this already. Apologies for the duplicate effort!

PR #1896, opened 2021-09-12
unused deps?

gilescope:
Not needed?


jebrosen:
To my memory most of these are indeed used, particularly in tests. Have you tested these changes, e.g. with `./scripts/test.sh` locally as described [in the README](https://github.com/SergioBenitez/Rocket#testing)?

kolbma:
All is used in (doc)tests/examples and rand is just a fixation for the version, because rand is used by many other dependencies.

PR #1912, opened 2021-09-21
Change logging level of launch messages to info.

### LINKED ISSUE #1828: Log Warnings at startup
**Description**

When Rocket starts (launches) (in 0.5) it lists a lot of log messages that are marked as `warn` instead of `info` as it was in 0.4.
Below you can see the messages as they are now (0.5.0-rc.1)
```
WARN :rocket::launch - 🔧 Configured for debug.
WARN :rocket::launch_ - address: 127.0.0.1
WARN :rocket::launch_ - port: 8000
WARN :rocket::launch_ - workers: 8
...
WARN :rocket::launch - 🛰  Routes:
...
```
(note that the format is custom because of my log implementation)
The `WARN` is the `record.level()` so all warning added by `log::warn!(...)`.
In Rocket 0.4.x these messages where printed using `log::info!(...)`.
This was desirable as now my warning logs would be full or warnings even though nothing was wrong.

**To Reproduce**

Any example will do.
In order to show the log level you can use the code below (applied this to `examples/hello`)
Add to toml: `log = "0.4.14"`
Code:
```rust
use log::{Record, Level, Metadata};

struct SimpleLogger;

static LOGGER: SimpleLogger = SimpleLogger;

#[launch]
fn rocket() -> _ {
    log::set_logger(&LOGGER).expect("Could not setup logger");
    log::set_max_level(log::LevelFilter::Info);
    rocket::build()
        .mount("/", routes![hello])
        .mount("/hello", routes![world, mir])
        .mount("/wave", routes![wave])
}

impl log::Log for SimpleLogger {
    fn enabled(&self, metadata: &Metadata) -> bool {
        metadata.level() <= Level::Info
    }

    fn log(&self, record: &Record) {
        if self.enabled(record.metadata()) {
            println!("{} - {}", record.level(), record.args());
        }
    }

    fn flush(&self) {}
}
```

**Expected Behavior**

Log messages that are expected should be marked as `info` not as `warning`.

**Environment:**

 - OS Distribution and Kernel: Linux
 - Rocket Version: 0.5.0-rc.1

davidv1992:
Fixes #1828. Since the launch messages are expected, it is unnatural to log them as warnings, in particular as they will not require action in almost all cases.

### DIFF
--- a/core/lib/src/log.rs
+++ b/core/lib/src/log.rs
@@ -32,8 +32,8 @@ define_log_macro!(warn, warn_);
 define_log_macro!(info, info_);
 define_log_macro!(debug, debug_);
 define_log_macro!(trace, trace_);
-define_log_macro!(launch_info: warn, "rocket::launch", $);
-define_log_macro!(launch_info_: warn, "rocket::launch_", $);
+define_log_macro!(launch_info: info, "rocket::launch", $);
+define_log_macro!(launch_info_: info, "rocket::launch_", $);
 
 #[derive(Debug)]
 struct RocketLogger;


SergioBenitez:
Merged in 83e7fe1157000704240a393e2a610e1008b86f86.

PR #1934, opened 2021-10-05
Allowed testing of client certificates

blbuford:
Hello,

I'm fairly new to Rust/Rocket. I saw an opportunity to add a feature that would make unit testing my API a lot easier and I wanted to see what y'all thought. mTLS is a fairly new feature in Rocket, and as such, it does not seem that there is a way to unit test providing client certs to "route" functions. I looked at the tests in examples/tls and verified as much. I added an identity method to Client such that you can provide a String path to a client cert that would be provided to the route. I updated the tests in the TLS example to show how it can it can used.


SergioBenitez:
Thank you! Merged with significant changes in aa6ad7030ad3b374295ae4a11465a4c14e0dbf0f. I'm sorry it took so long.

PR #1939, opened 2021-10-09
adds missing derive attributes for serde in doc

ValJed:
Hello,

I'm currently learning Rust and Rocket which is a really good tool.
I just made some fixes in the documentation by adding some missing derive attributes for serde.
I had to dive into the code examples to understand why it wasn't working for me.

I'm a newbie so maybe I'm wrong somewhere, let me know.
I may also not be targeting the right branch.

Thanks for your work.

### DIFF
--- a/core/lib/src/local/request.rs
+++ b/core/lib/src/local/request.rs
@@ -231,6 +231,7 @@ macro_rules! pub_request_impl {
     /// use rocket::http::ContentType;
     ///
     /// #[derive(Serialize)]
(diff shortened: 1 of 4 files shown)


kolbma:
This should be only needed if you use somewhere a different serde version and not `rocket::serde`.   
Have you put an own serde dependency in the Cargo.toml of your project? Can you confirm this?!

SergioBenitez:
It looks like this was handled in 761ffb009ea9d35c32d3c8eecd948ec4434cd0a3. Thanks!

PR #1968, opened 2021-10-28
Fix `Result` namespacing in `context!` dyn templates macro

Misterio77:
Hey!

I was getting an error while trying to use `context!` in any files where i have a `Result` alias scoped.

This quick patch should explicitly make the `Result` at line 532 be the std one.

PS: Also, thank you so much for this project! I've been keeping up with development for quite a while. It's my framework of choice for both work and personal projects, it's been a joy to see just how much polished Rocket gets with each and every update!


SergioBenitez:
Excellent! Merged in 66d18bf66517e2765494d082629e9b9748ff8ad6.

Issue #1884, opened 2021-09-07
Get access to database connection pool


threema-danilo:
**Is your feature request motivated by a concrete problem? Please describe.**

I handle requests that trigger the processing of long-running tasks in the background. These tasks are executed in the context of a thread pool (let's call it the worker thread).

Because the worker thread needs access to the database, and potentially to multiple connections (due to further parallelization of tasks), the API currently expects that a reference to the database connection pool is passed in.

When using the `rocket_sync_db_pools` crate, I can create a database and use that through the request guard. However, from what I can tell by reading the source code, that request guard only gets access to a single connection and does not expose the connection pool itself. And the connection (at least in the case of `diesel::PgConnection`) is not cloneable.

**Why this feature can't or shouldn't live outside of Rocket**

As long as the connection pool is not exposed in any way, there cannot be any solution outside of `rocket_sync_db_pools` (unless a completely separate solution is written).

**Ideal Solution**

Ideally the `#[database(...)]`-annotated struct that wraps the connection would get an automatically implemented method `fn get_pool(&self) -> P` which returns a reference to the database connection pool.

**Alternatives Considered**

I could create a second, completely independent database connection and pass that into the handlers using `&RocketState<MyPool>`. However, that seems hacky.

Zopieux:
Ah, fun timing, there is some context and half-solutions in a related (but closed) bug that I just commented on: https://github.com/SergioBenitez/Rocket/issues/1640#issuecomment-917320655

Since July 18th and the implementation of async `rocket_db_pools` (5b1a04deabc75981fde566c4381b4cba0d3b7cc8), it looks like this became possible, so I've asked Sergio about the expected usage pattern. I feel like this should be documented in a "background processing" section of some sort, given the multiple FRs we've seen.

threema-danilo:
The async pools will only work with async-capable backends though (e.g. no diesel).

If the general approach of exposing a `get_pool(&self) -> Option<Pool>` in the `sync_db_pools` crate is deemed to be OK (I'm not sure if exposing the pool can have any side effects), then I'll happily try to contribute an implementation!

SergioBenitez:
> If the general approach of exposing a `get_pool(&self) -> Option<Pool>` in the `sync_db_pools` crate is deemed to be OK (I'm not sure if exposing the pool can have any side effects), then I'll happily try to contribute an implementation!

[...]

PR #1972, opened 2021-11-02
sync_db_pools: Expose connection pool

threema-danilo:
This is a proposed fix for #1884.

It adds a connection pool reference to every connection handle and exposes that through a generated method.

Is this the right approach? If not, how else should it be implemented?

### DIFF
--- a/contrib/sync_db_pools/codegen/src/database.rs
+++ b/contrib/sync_db_pools/codegen/src/database.rs
@@ -108,6 +108,11 @@ pub fn database_attr(attr: TokenStream, input: TokenStream) -> Result<TokenStrea
             {
                 self.0.run(__f).await
             }
+
+            /// Retrieves a copy of the inner connection pool.
+            pub fn get_pool(&self) -> #root::r2d2::Pool<<#conn_type as #root::Poolable>::Manager> {
+                self.0.get_pool()
+            }
         }
 
         #[#rocket::async_trait]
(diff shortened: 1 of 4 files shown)


kolbma:
We definitely don't want to `clone` the pool here.  
What is the usecase to access the pool?

threema-danilo:
Cloning the pool only increments a reference count and is thus cheap, right?

The use case is an application where a request launches multiple background worker threads that need database access. Every thread needs its own connection, and a connection cannot be cloned. However, the pool can be cloned and it's cheap to do so. By passing a clone of the pool (i.e. a reference to the pool) to every worker thread, the thread an fetch its own connection.

kolbma:
Ok, understand, will look at it in detail. Thanks.

SergioBenitez:
We want the signature of the generated inherent message on the database type to look like `fetch` from `db_pools`:

```rust
    fn fetch<P: Phase>(rocket: &Rocket<P>) -> Option<&Self> {
```

We don't want to be exposing the raw `r2d2::Pool` type.

threema-danilo:
@SergioBenitez I can't quite follow what that means in practice. What is `&Self` in this case?

AndrolGenhald:
I thought the point was to expose the pool so that it could be used without a reference to `Rocket`? Wouldn't making `rocket_sync_db_pools::ConnectionPool::get` public accomplish that? Then the `ConnectionPool` could be retrieved from the `Rocket` after ignition, but it wouldn't need a reference to the `Rocket` to retrieve a connection, so it could easily be used outside of rocket at that point.

SergioBenitez:
Given the state of `rocket_sync_db_pools`, this is much trickier than one would expect. I went ahead and implemented this in https://github.com/SergioBenitez/Rocket/commit/04819d8cfdae92e69162f051c5571234b14c5387. Thank you for this PR, however.

PR #1973, opened 2021-11-03
Fixed spelling mistake

an-toni-ok:


p-fruck:
Hey there! I guess this usage of **iff** is intended, see [here](https://en.wikipedia.org/wiki/If_and_only_if) for more details :)
Maybe this article could be linked in the docs since many users might not be familiar with this term?

SergioBenitez:
See https://github.com/SergioBenitez/Rocket/pull/2096#issuecomment-1049306058.

PR #1984, opened 2021-11-13
Add missing 'return' to luhn validation example

wezm:
I copied this this code to write my own custom validation function and then noticed that the `return` was missing. This PR fixes that.


wezm:
Oh I just noticed the `?`.

PR #1993, opened 2021-11-17
[DRAFT] Add type safety to fairings availability

petar-dambovaliev:
I was updating rocket, while having an issue where i wasn't aware that fairings can only be accessed after the `Ignite` state.
This change would be tell a user immediately what the problem is, as a compilation error.
This is only for means of showing what my intentions are and is not the final work i will commit.
If this is approved, i will test it and add a compilation error test.

### DIFF
--- a/contrib/sync_db_pools/codegen/src/database.rs
+++ b/contrib/sync_db_pools/codegen/src/database.rs
@@ -92,7 +92,7 @@ pub fn database_attr(attr: TokenStream, input: TokenStream) -> Result<TokenStrea
             /// instance. Returns `Some` as long as `Self::fairing()` has been
             /// attached.
             pub async fn get_one<P>(__rocket: &#rocket::Rocket<P>) -> Option<Self>
-                where P: #rocket::Phase,
+                where P: #rocket::fairing::HasFairings,
             {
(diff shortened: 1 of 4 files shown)


petar-dambovaliev:
As far as i can see, Pull requests are looked at for months.
It seems this project is not actively maintained.
I am closing this.

SergioBenitez:
This change is logically incorrect, in either case, in the sanse that it makes it impossible to access managed state outside of the orbit and ignite phases. Fairings are not the only way to add state to an instance of Rocket, and that state needs to be accessible at all times. I'm certain this would fail the CI.

petar-dambovaliev:
> This change is logically incorrect, in either case, in the sense that it makes it impossible to access managed state outside of the orbit and ignite phases. Fairings are not the only way to add state to an instance of Rocket, and that state needs to be accessible at all times. I'm certain this would fail the CI.

I don't know what the solution is.
However, what i do know is that i get a runtime error that something is missing, when i just put it in rocket 1 line before.
It is obviously a problem.
Whether you want to fix that and how to go about it is not for me to say, as i am not super familiar with the framework and it's internals.

PR #2006, opened 2021-12-03
Add Request::try_local_cache method

HaronK:
In some cases we need to retrieve value from the cache only if it exists. This method returns Option<&T> without value creation.


kolbma:
Your example won't work...

SergioBenitez:
This has been requested before, and every time it was due to poor design, so I'm not convinced this is needed. Can you open an issue with why you think you need this method?

PR #2013, opened 2021-12-10
Implement listening on Unix domain sockets

clubby789:
Implement the ability to bind to a Unix domain socket

TODO:
 - [ ] TLS over Unix socket﻿

### DIFF
--- a/core/http/src/lib.rs
+++ b/core/http/src/lib.rs
@@ -28,6 +28,8 @@ mod status;
 mod raw_str;
 mod parse;
 mod listener;
+#[cfg(unix)]
+mod unix;
 
 /// Case-preserving, ASCII case-insensitive string types.
 ///
--- a/core/http/src/lib.rs
+++ b/core/http/src/lib.rs
@@ -42,6 +42,8 @@ pub mod uncased {
 #[doc(hidden)]
 #[path = "."]
 pub mod private {
+    #[cfg(unix)]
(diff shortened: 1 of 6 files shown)


clubby789:
Making TLS work would probably require refactoring the `Listener` and `Connection` traits to not expect IP-based listeners

kolbma:
What is the use case at your place?

lUNuXl:
Can't speak for the author, but in my case running a rust-powered website on a server with php-only website support is the main use case :p

Also communication over unix sockets is faster

SergioBenitez:
Thank you for this! And my apologies for coming to this only now.

The way I want to approach this is via #1070. Specifically, I want to:

  * Remove all connection configuration from the top-level `Config`.
  * Modify or create a composable `Listener` trait that knows how to:
    - Extract its configuration from the configured provider.
    - Set itself up and provide something for Rocket to read data from.
    - Emit an "address" in its "native" format. 
  * Implement `Listener` for what Rocket does internally today with TCP/TLS.
  * Implement `Listener` for Unix Domain Sockets.
  * Implement `Listener` for some default listener that composes that previous two.
  * Add a builder method to `Rocket` to set a different `Listener`.
  * Set the default listener to be the composed listener from the previous point.

Given that this PR doesn't take this approach (of course), I'm closing it out for now. But I'd be really excited for someone to tackle the above.

PR #2032, opened 2021-12-22
Implement #1560

### LINKED ISSUE #1560: It would be nice if Forward could include a Error to return if Forwarding fail to find another route
Imagine you have two routes,  

``` rust
#[get("/one")]
pub async fn get_protected_endpoint(user: Authenticated){}
#[get("/one", rank = 2)]
pub async fn get_public_endpoint(){}

#[get("/two")]
pub async fn get_protected_endpoint(user: Authenticated){}
```

A guard takes care of enforcing Authenticated.

if the guard returns Outcome::Error wrapping NotAuthorized, then the second route for /one will never get called 

If the guard returns Outcome::Forward, then for /two the final result will be NotFound instead of UnAuthorized.

If Forward could return a Error, then the routing could be "Try and find another matching route. If nothing works, then return this error"

So Forward(NotAuthorized) would represent this case

To work around this, you need a dummy catchall route to return 'unauthorized', or mess around with a special fairing.

bewee:
This aims to close #1560

### DIFF
--- a/core/codegen/src/attribute/route/mod.rs
+++ b/core/codegen/src/attribute/route/mod.rs
@@ -37,7 +37,7 @@ fn query_decls(route: &Route) -> Option<TokenStream> {
     }
 
     define_spanned_export!(Span::call_site() =>
-        __req, __data, _log, _form, Outcome, _Ok, _Err, _Some, _None
+        __req, __data, _log, _form, Outcome, _Ok, _Err, _Some, _None, Status
     );
 
     // Record all of the static parameters for later filtering.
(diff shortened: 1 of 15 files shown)


bewee:
Hi, I just stumbled across this old PR and saw that it gathered some conflicts. Is there still any interest in this feature? Then I could try to resolve the conflicts

bewee:
Hello??? @SergioBenitez ? Anybody here? Just any life sign, please?

LMH01:
Hello @bewee 

I would really appreciate if you would resolve the conflicts and if this pull request could be merged as I stumbled across a problem where I would require this feature.

SergioBenitez:
Merged in https://github.com/SergioBenitez/Rocket/commit/055ad107df9753765c2cfe02bccb99942750d57d! Thank you!

Issue #2049, opened 2022-01-05
HSTS Preload Mising includeSubdomains


Soham3-1415:
**Description**

According to the HSTS Preload requirements at https://hstspreload.org/:
> The includeSubDomains directive must be specified.

and in order to maintain a domain on the preload list:
> You must make sure your site continues to satisfy the submission requirements at all times.

The `rocket::shield::Hsts` has a `Preload` variant that is supposed to add an HSTS header with the preload directive. However, this variant does not have the `includeSubdomains` directive. This contradicts the requirements specified for the preload list.

As an additional note, the `max-age` directive must be at least `31536000` seconds, and there is no mechanism to enforce or warn the user about this. However, I do not consider this particular behavior to be a bug as it is still possible for the header to be used correctly.

**To Reproduce**
Cargo.toml:
```toml
[package]
name = "oops"
version = "0.1.0"
edition = "2021"

[dependencies]
rocket = "0.5.0-rc.1"
time = "0.2"
```

main.rs:
```rust
#[macro_use]
extern crate rocket;

use rocket::shield::{Shield, Hsts};

#[launch]
fn oops() -> _ {
  rocket::build().attach(Shield::new().enable(Hsts::Preload(time::Duration::new(63072000, 0))))
}
```

Steps:
1. Run with: `cargo run`
2. Run `curl -I localhost:8000`

Response:
```
HTTP/1.1 404 Not Found
content-type: text/html; charset=utf-8
server: Rocket
strict-transport-security: max-age=63072000; preload
content-length: 383
date: Wed, 05 Jan 2022 22:54:59 GMT
```
Notice that strict-transport-security has the `preload` directive but is missing the `includeSubdomains` directive.
(date is subject to change and is not relevant)

**Expected Behavior**
The strict-transport-security header ought to display as (source: https://hstspreload.org/):
```
Strict-Transport-Security: max-age=63072000; includeSubDomains; preload
```
(header capitalization is irrelevant)

**Environment:**

 - OS Distribution and Kernel: core-11.1.0ubuntu3-noarch:security-11.1.0ubuntu3-noarch
 - Rocket Version: 0.5.0-rc.1

**Additional Context**
The issue seems to be at line `320` in `/core/lib/src/shield/policy.rs`:
```rust
Hsts::Preload(age) => format!("max-age={}; preload", age.whole_seconds()),
```

somehowchris:
[...]

PR #2050, opened 2022-01-05
Fixed broken HSTS preload header

Soham3-1415:
Implements a fix for  #2049

### DIFF
--- a/core/lib/src/shield/policy.rs
+++ b/core/lib/src/shield/policy.rs
@@ -317,7 +317,7 @@ impl From<&Hsts> for Header<'static> {
             Hsts::IncludeSubDomains(age) => {
                 format!("max-age={}; includeSubDomains", age.whole_seconds())
             }
-            Hsts::Preload(age) => format!("max-age={}; preload", age.whole_seconds()),
+            Hsts::Preload(age) => format!("max-age={}; includeSubDomains; preload", age.whole_seconds()),
         };
 
         Header::new(Hsts::NAME, policy_string)
--- a/core/lib/src/shield/policy.rs
+++ b/core/lib/src/shield/policy.rs
@@ -286,12 +286,24 @@ pub enum Hsts {
 
     /// Google maintains an [HSTS preload service] that can be used to prevent
     /// the browser from ever connecting to your site over an insecure
-    /// connection. Read more at [MDN]. Don't enable this before you have
-    /// registered your site.
+    /// connection. Read more at [MDN]. This will need to be enabled before
+    /// registering your site.
+    ///
+    /// A flag is available to include (iff true) the `includeSubDomains`
+    /// directive. The directive is necessary to join and stay on Google's
+    /// preload list. Furthermore, a `max-age` (specified by the Duration
+    /// parameter) of at least 31536000 seconds (1 year) is necessary to
+    /// be added and maintain presence on Google's preload list. This `preload`
+    /// directive need only be present on the base domain. The `preload`
+    /// directive is not an official spec. Be careful when using a lengthy
+    /// `max-age`, `includeSubDomains`, and/or `preload`. HSTS mistakes are
+    /// are difficult to fix. Starting with a small `max-age` and checking for
+    /// errors in deployment is recommended.
+    /// 
     ///
     /// [HSTS preload service]: https://hstspreload.org/
     /// [MDN]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security#Preloading_Strict_Transport_Security
-    Preload(Duration),
+    Preload(Duration, bool),
 }
 
 /// Defaults to `Hsts::Enable(Duration::days(365))`.
@@ -317,7 +329,8 @@ impl From<&Hsts> for Header<'static> {
             Hsts::IncludeSubDomains(age) => {
                 format!("max-age={}; includeSubDomains", age.whole_seconds())
             }
-            Hsts::Preload(age) => format!("max-age={}; includeSubDomains; preload", age.whole_seconds()),
+            Hsts::Preload(age, false) => format!("max-age={}; preload", age.whole_seconds()),
+            Hsts::Preload(age, true) => format!("max-age={}; includeSubDomains; preload", age.whole_seconds()),
         };
 
         Header::new(Hsts::NAME, policy_string)


core/lib/src/shield/policy.rs
```diff
@@ -286,12 +286,24 @@ pub enum Hsts {
 
     /// Google maintains an [HSTS preload service] that can be used to prevent
     /// the browser from ever connecting to your site over an insecure
-    /// connection. Read more at [MDN]. Don't enable this before you have
-    /// registered your site.
+    /// connection. Read more at [MDN]. This will need to be enabled before
+    /// registering your site.
+    ///
+    /// A flag is available to include (iff true) the `includeSubDomains`
```
kolbma, line 292: There is a typo

core/lib/src/shield/policy.rs
```diff
@@ -286,12 +286,24 @@ pub enum Hsts {
...
+    /// directive is not an official spec. Be careful when using a lengthy
+    /// `max-age`, `includeSubDomains`, and/or `preload`. HSTS mistakes are
+    /// are difficult to fix. Starting with a small `max-age` and checking for
+    /// errors in deployment is recommended.
+    /// 
     ///
     /// [HSTS preload service]: https://hstspreload.org/
     /// [MDN]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security#Preloading_Strict_Transport_Security
-    Preload(Duration),
+    Preload(Duration, bool),
```
kolbma, line 306: I would fix this without this not descriptive bool.  
At the moment there is no functionality, because preloading works in real life only with includedSubDomains.

core/lib/src/shield/policy.rs
Soham3-1415: That iff was meant to be short for if and only if.

core/lib/src/shield/policy.rs
Soham3-1415: I agree. However, I was advised in #2049 to include the flag.

somehowchris: How about internal networks?

kolbma: Internal networks and HSTS preload, how should this work? This is a "google" service to secure the web.

somehowchris: Wait, are we talking about using HSTS & Preload or the preload list?

somehowchris: hit me with an RTFM if im completely lost

Soham3-1415: As far as I know Google is the only provider and maintainer of the list that major browsers use. Since preloading is not an RFC, I think in theory it is possible for any browser to use any list they want, but they could also use any other directive as well. I think the preload directive is associated with Google's list. In order to minimize the size of the list included with the browser, Google mandates that automatic submissions to the list use includeSubDomains. Google themselves do not follow that requirement, and you can see several sub domains of google.com on common lists. However, I am not aware of any accessible way for others to use preload and not include all subdomains.

That being said, since it's not a spec, Google could change the requirements of the list at anytime. Or Mozilla could, for example, start their own public submission list, and they could use an entirely different directive. However, the current behavior (and it has been so for a while), as far as I am aware, is that browsers will include the Chromium HSTS list.

We can take a look at Firefox's behavior: https://wiki.mozilla.org/Privacy/Features/HSTS_Preload_List
> Google maintains a list in their Chrome source ... We can use this list, since Google has invested in maintaining and sharing it.

> Unfortunately, we can't simply take every entry in Chrome's list with mode:"force-https".

> Our preload list must contain only sites that we can say are HSTS with a high degree of confidence. So, starting with Chrome's list (because we have to start somewhere), we will only include sites that:
> 1. Actually send the HSTS header
> 2. Have a max-age larger than or equal to 18 weeks (10886400)
> 
> Additionally, whether or not to include subdomains will be based on the header the site actually sends, not the contents of Chrome's list.

I see no mention of whether Firefox actually verifies the existence of the preload directive, but we can see that Firefox does take into account the includeSubdomains directive.

Soham3-1415: Here's a base64 encoded version of the preload list:
https://chromium.googlesource.com/chromium/src/+/refs/heads/main/net/http/transport_security_state_static.json?format=TEXT

I think Firefox and other browsers will be pulling this in, but how they deal with it clearly differs. If you look through the file, the details are very messy.

It is possible to get manually added to the preload list: https://github.com/chromium/hstspreload.org/wiki/Preload-List-Processes#requirements-for-manual-hsts-entries

Originally, I was against including a flag for what seemed to be a required feature to use preloading effectively. I thought that adding an additional flag would just create confusion and serve no practical purpose. However, after some research, it is clear that HSTS preloading is an evolving non standard directive that is used in many different ways. While I maintain that the _vast_ majority of users will need includeSubDomains in addition to preload, there seem to be possible use cases where that is not the case. In order to support those cases, I think a flag is necessary.

kolbma: No. First the wiki of mozilla is much outdated. The page is a draft in the age of version 16 when they started to think about implementing it.
It totally makes no sense because you will not land with your domain on any list without the argument set. It is against the idea and job of preload, which is anything at my place will be delivered with https and the client knows about it before requesting anything next to the list to circumwent any redirect roundtrips. Only tlds get on the list. In logic it makes no sense to send a preload header without includesubdomains because there will be no preload functionality. This is only possible with the domain on list. The list can be the only authority because an intruder or mitm could change the headers of the server and so browsers may not trust this header or the browser breaks security. This would be a bug.

Soham3-1415: Good point about the wiki of Mozilla being outdated.

> It totally makes no sense because you will not land with your domain on any list without the argument set.

https://github.com/chromium/hstspreload.org/wiki/Preload-List-Processes#requirements-for-manual-hsts-entries
There is a section on manual entries. Google, at their discretion can add sites to the list without the includeSubDomains directive, and if you look at the relatively recent list, there are sites on there that do not have includeSubDomains.

Take united.com, for example.
```bash
curl -I https://united.com
```
```
HTTP/1.1 302 Moved Temporarily
Server: AkamaiGHost
Content-Length: 0
Location: https://www.united.com/
Expires: Sun, 23 Jan 2022 17:26:39 GMT
Cache-Control: max-age=0, no-cache, no-store
Pragma: no-cache
Date: Sun, 23 Jan 2022 17:26:39 GMT
Connection: keep-alive
Strict-Transport-Security: max-age=15768000; preload
```
If you check with the preload list, they are on there without any issues.


> Only tlds get on the list.
eTLDs can apply to get on the list, but most submissions are for eTLDs+1. There are also some subdomains on the list as well. 

wordpress.com and www.wordpress.com are on the list. Both have preload. Neither have includeSubDomains.


I agree that includeSubDomains should be strongly encouraged. However, I think documentation is the best place to do that because it is completely possible to preload without includeSubDomains.

Issue #2031, opened 2021-12-21
rocket-worker-thread panics whith PoisonError after recieving Os Error { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" } once


FrankenApps:
**Description**

I have a simple rocket application running on a server and it receives around 150 requests per day. After a week or two the service stops working until I restart it.

This is the output from `stderr`:
```
thread 'rocket-worker-thread' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.6.1/src/runtime/blocking/pool.rs:261:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'rocket-worker-thread' panicked at 'called `Result::unwrap()` on an `Err` value: PoisonError { .. }', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.6.1/src/runtime/blocking/pool.rs:287:86
thread 'rocket-worker-thread' panicked at 'called `Result::unwrap()` on an `Err` value: PoisonError { .. }', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.6.1/src/runtime/blocking/pool.rs:287:86
...
thread 'rocket-worker-thread' panicked at 'called `Result::unwrap()` on an `Err` value: PoisonError { .. }', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.6.1/src/runtime/blocking/pool.rs:287:86
thread 'rocket-worker-thread' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.6.1/src/runtime/blocking/pool.rs:261:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'rocket-worker-thread' panicked at 'called `Result::unwrap()` on an `Err` value: PoisonError { .. }', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.6.1/src/runtime/blocking/pool.rs:287:86
[...]

PR #2064, opened 2022-01-12
Add blocking_worker config value

the10thWiz:
Add a configuration value to set the maximum number of blocking threads. It has the same default value as Tokio (512).

This might help address issues like #2031, by allowing the value to be set lower than the OS limit for the number threads spawned. I don't think adjusting the default is necessary (or a good idea), since the Tokio default seems to be pretty good for most use cases.

### DIFF
--- a/core/lib/src/config/config.rs
+++ b/core/lib/src/config/config.rs
@@ -73,6 +73,8 @@ pub struct Config {
     /// _**Note:** Rocket only reads this value from sources in the [default
     /// provider](Config::figment())._
     pub workers: usize,
+    /// Limit on threads to start for synchronous tasks. **(default: `512`)**
+    pub sync_threads: usize,
     /// How, if at all, to identify the server via the `Server` header.
     /// **(default: `"Rocket"`)**
     pub ident: Ident,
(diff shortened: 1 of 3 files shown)


kolbma (commented):
This needs to be added to https://rocket.rs/v0.5-rc/guide/configuration/#overview

https://github.com/SergioBenitez/Rocket/blob/8cae077ba1d54b92cdef3e171a730b819d5eeb8e/site/guide/9-configuration.md?plain=1#L24

As well further below some sentences to the worker section.

SergioBenitez:
I've rebased the PR on the latest `master`. I've also changed the name of the config value to `sync_threads`. Unfortunately, I am still unhappy with this name, which is why I haven't merged the PR.

I'd like a name that:
  * isn't too long
  * somehow talks about blocking/synchronous _threads_
  * ideally clarifies that this is an upper limit

I'm open to suggestions.

SergioBenitez:
Settled on `max_blocking`. Merged in 47946cc55c0cb8071e79a738b1e34d0696e29102. Thank you again!

Issue #2019, opened 2021-12-15
SIGTERM is ignored when process has no TTY


darakian:
**Description**

Ran into an issue killing a rocket process and tracked it down to a missing tty. SIGKILL still kills the process, but a graceful shutdown would be nice even when missing a tty. SIGQUIT works as expected.

**To Reproduce**

Launch a rocket web server as a background process then log out the active user.

**Expected Behavior**

The root user (or other management users) should be able to kill the process using the kill command as `kill <pid>` rather than needed to specify SIGQUIT.

**Environment:**

 - OS Distribution and Kernel: CentOS 8 stream `4.18.0-269.el8.x86_64`
 - Rocket Version: `0.5.0-rc.1` via crates.io

**Additional Context**

```
[root@node ~]# ps -U web
    PID TTY          TIME CMD
 918027 pts/1    00:00:00 rocket_process
[root@node ~]# kill -s SIGTERM 918027
[root@node ~]# ps -U web
    PID TTY          TIME CMD
[root@node ~]# ps -U web
    PID TTY          TIME CMD
 918066 ?        00:00:00 rocket_process
[root@node ~]# kill -s SIGTERM 918066
[root@node ~]# ps -U web
    PID TTY          TIME CMD
 918066 ?        00:00:00 rocket_process
```

Launch config
```
Configured for release.
   >> address: 0.0.0.0
   >> port: 80
   >> workers: 1
   >> ident: Rocket
   >> keep-alive: 5s
   >> limits: bytes = 8KiB, data-form = 2MiB, file = 1MiB, form = 32KiB, forms = 32KiB, json = 1MiB, msgpack = 1MiB, string = 8KiB
   >> tls: disabled
   >> temp dir: /tmp
   >> log level: critical
   >> cli colors: true
   >> shutdown: ctrlc = true, force = true, signals = [SIGTERM], grace = 2s, mercy = 3s
```

kolbma:
I can confirm this.  
It works if you redirect the process output...

`executable >out.log 2>&1`

or

`executable 2>&1 | tee out.log`

Will see if this can be fixed in Rocket or it is a problem in tokio::signal.

kolbma:
@darakian Does Rocket successful service requests in this detached state?

It panics in trying to `warn!("Received {}. Requesting shutdown.", sig);` because print to stdout is not possible anylonger.

kolbma:
[...]

PR #2070, opened 2022-01-17
fix(log): implement print macros without panic

kolbma:
I don't have an explanation why the rust std print macros panic, but this fixes https://github.com/SergioBenitez/Rocket/issues/2019

Tested with and without redirecting. Log output looks like usual.


SergioBenitez:
Merged in https://github.com/SergioBenitez/Rocket/commit/cc98f98aa75be5561d2db3a0d13f3db0d2968963. Thank you!

Issue #2063, opened 2022-01-10
CookieJar::get_private_pending doesn't return pending private cookies


jamesbirtles:
**Description**

As far as I can tell, `CookieJar::get_private_pending` doesn't do what it claims to be able to do, or at least what it does do isn't very helpful.

[The documentation](https://api.rocket.rs/v0.5-rc/rocket/http/struct.CookieJar.html#method.get_private_pending) says
> Retrives the original or pending Cookie inside this collection with the name name and authenticates and decrypts the cookie’s value. If the cookie cannot be found, or the cookie fails to authenticate or decrypt, None is returned.

However calling `get_private_pending` after calling `add_private` with the same cookie name, does not return the cookie just added. 

**To Reproduce**

```rust
#[get("/")]
fn handler(jar: &CookieJar<'_>) {
    jar.add_private(Cookie::new("name", "value"));
    assert!(jar.get_private_pending("name").is_some());
}
```

**Expected Behavior**

I would expect that `get_pending_cookie` would return the pending cookie if there is one, or the original cookie decrypted, or None,

**Environment:**

 - OS Distribution and Kernel: Manjaro Quonos 21.2.1 (Kernel 5.14.21)
 - Rocket Version: 0.5.0-rc.1

**Additional Context**

From looking at the code, I believe this bug is because  `get_private_pending` attempts to decrypt the cookie it found via `get_pending` regardless of if it was a newly added cookie, or one from the request. This is a problem because add_private doesn't actually appear to encrypt the cookie, instead it flags it as private and I assume it is encrypted later on. Due to this, the attempt to decrypt the pending cookie fails and so it appears as if there was no cookie at all.

kolbma:
Have you set a `secret_key`? https://docs.rs/rocket/0.5.0-rc.1/rocket/http/struct.CookieJar.html#encryption-key

jamesbirtles:
> Have you set a `secret_key`? https://docs.rs/rocket/0.5.0-rc.1/rocket/http/struct.CookieJar.html#encryption-key

Of course, not that it would matter though as firstly rocket would generate one if not specified, and secondly `get_private_pending` is only relevant within the same request so the fact the secret key would regenerate on restart wouldn't matter here

PR #2071, opened 2022-01-17
Fix for Issue 2063 CookieJar::get_private_pending()

kolbma:
Fixes https://github.com/SergioBenitez/Rocket/issues/2063

In cookies::CookieJar::get_private_pending() is no decrypt of pending cookie possible, because pending cookies are just operated before server sends response and so cookies are not yet encrypted. Just return the cookie as it is.

Also updates dependency of crate cookie from rc to release 0.16.0 and add feature secure.

### DIFF
--- a/core/http/Cargo.toml
+++ b/core/http/Cargo.toml
@@ -39,7 +39,7 @@ pear = "0.2.3"
 pin-project-lite = "0.2"
 memchr = "2"
 stable-pattern = "0.1"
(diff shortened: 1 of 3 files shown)


SergioBenitez:
This fix elucidates the lack of necessity for `get_private_pending()` in general, so I removed the method in favor of `get_pending()`. Merged in spirit in 1ba41b3990c8202ba50de60ff0e7157a9ee2887a. Thank you!

PR #2072, opened 2022-01-17
build!: update MSRV 1.51.0

kolbma:
A few up-to-date dependencies requires at least rust version 1.51.0.  
As this is also quite old, it should be no problem to move this on.


kolbma:
Found some newer restrictions in trying to get the scipts/test.sh to work.

PR #2077, opened 2022-01-18
build: cleanup and fixes for successful scripts/test.sh run

kolbma:
scripts/test.sh runs now with rust 1.53.0 and also todays nightly


SergioBenitez:
I had an old commit that addressed this. My apologies for not pushing it sooner. Thank you!

PR #2096, opened 2022-01-30
FromData trait, Data Guards section: add context link for "iff"

admwrd:
Add context for what "iff" means.

For those unfamiliar with mathematics terminology, it may be beneficial to clarify what _iff_ means, as some may not know that this is a term, and see it as a spelling mistake (and thus miss its meaning).

For clarity, it may be best to expand from _iff_ to _if and only if_ or to point to the definition at a specific resource (such as Wikipedia).

I've added the second option to this commit, but am fine making other changes if something better is suggested.

This comes after me seeing the term and thinking it was a spelling error (thinking "if" was italic for emphasis rather than technical usage).

_Ignore from this line down. It's saved for posterity from when this was seen as a spelling error._

---
Spelling issue can be seen on the docs page for [FromData (§ Data Guards)](https://api.rocket.rs/v0.5-rc/rocket/data/trait.FromData.html#data-guards).

Original
>  In other words, a type is a data guard _iff_ it implements `FromData`.

Commit change
> In other words, a type is a data guard _if_ it implements `FromData`.

It seems to have come from commit 63a14525d86595a8033715e3bdcd2bf2581eecb1 ([ref. to line](https://github.com/SergioBenitez/Rocket/commit/63a14525d86595a8033715e3bdcd2bf2581eecb1#diff-a50bf8c7c0ff968d7b05542b7ed9ecde3e9a509b6de581c47dffcf217b8f39daR38)) on on Mar 4, 2021 with tag [`v0.5.0-rc.1`](https://github.com/SergioBenitez/Rocket/releases/tag/v0.5.0-rc.1)

### DIFF
--- a/core/lib/src/data/from_data.rs
+++ b/core/lib/src/data/from_data.rs
@@ -36,7 +36,7 @@ impl<'r, S, E> IntoOutcome<S, (Status, E), Data<'r>> for Result<S, E> {
 ///
 /// A data guard is a guard that operates on a request's body data. Data guards
 /// validate and parse request body data via implementations of `FromData`. In
-/// other words, a type is a data guard _iff_ it implements `FromData`.
+/// other words, a type is a data guard _[iff](https://en.wikipedia.org/wiki/If_and_only_if)_ it implements `FromData`.
 ///
 /// Data guards are the target of the `data` route attribute parameter:
 ///


blackghost1987:
This is not a typo, this is intended, see: https://github.com/SergioBenitez/Rocket/pull/1800

The italic formatting is trying to show that it's not a word, but a special shorthand. Maybe a bold style or some extra explanation in parenthesis would be better. Or maybe an "alt" like pop-up explanation message? There were like 3 pull requests on this already so changing it might worth considering...

admwrd:
Ah, I see. Thanks for pointing this out. 

So there are a few options I see that can be taken.

1. It could be left as-is, but potentially trip up others unfamiliar with mathematical/logic terminology.
2. It could be expanded to _if and only if_ 
3. It could be left as _iff_ but perhaps link to the Wikipedia or some other explanation of the term. Something like:
  * > In other words, a type is a data guard _[iff](https://en.wikipedia.org/wiki/If_and_only_if)_ it implements `FromData`.

I've updated my commit to the third option, but happily leave the decision to the repo owners' discretion. 

I've also updated the title of this MR to match the current intent.

SergioBenitez:
We use `iff` at least 8 times in the codebase. Explaining the term each time seems unfortunate. It's a rather common term wherever logic is used slightly more formally, so I'm inclined to suggest we leave this as is and hope that folks either know the term or clue into the italics. If this continues to be a problem, I would opt for simply spelling out "if and only if," though I imagine the term itself won't have much meaning (beyond just if) to folks that aren't familiar with "iff", e.g, "iff" is known iff "if and only if" is known)

Issue #2082, opened 2022-01-22
Code review of tls/mtls feature


kolbma:
I'd like to point you on this suggestion https://github.com/rocket-org/Rocket/issues/4  
Maybe you can support and step in.

SergioBenitez:
I don't think this needs its own issue. Furthermore, the vast majority of the functionality, especially the security sensitive parts, are implemented outside of Rocket. As such, I'm closing this out.

PR #2111, opened 2022-02-23
Make HTTP/2 optional (closes #2030)

### LINKED ISSUE #2030: Make HTTP/2 optional
HTTP/2 support is actually useless in some real-world cases, primarily when a web application is behind a proxy server like nginx (which uses HTTP/1.0 for proxying). Moreover, `h2` is one of the biggest crates in the resulting hello-world binary (~220-240 KiB according to "cargo bloat --release"). I would be happy if I could significantly reduce the size of the binary by disabling the feature that I never plan to use.

If I understand the code correctly, Rocket doesn't use `h2` directly but just has few configuration lines ([1](https://github.com/SergioBenitez/Rocket/blob/7ffe3a73601aabdfe3190a809e660538acd5bdce/core/lib/src/server.rs#L462), [2](https://github.com/SergioBenitez/Rocket/blob/7ffe3a73601aabdfe3190a809e660538acd5bdce/core/http/src/tls/listener.rs#L73)), so I hope it's easy to implement.

See also related PR in warp: https://github.com/seanmonstar/warp/pull/836

UPD: also, axum has http2 feature disabed by default https://github.com/tokio-rs/axum/blob/axum-v0.4.3/axum/Cargo.toml#L16

edgarogh:
See SergioBenitez/Rocket#2030

This very simple PR adds a new "`http2`" feature to `core/http`, that itself enables or disables `hyper`'s similarly-named feature. An  "`http2`" feature is also added to the main crate, `core/lib` and is set as default here.

All statements that depend on HTTP/2 are put behind `#[cfg]` attributes.

`rocket` compiles both with and without this new feature.

### DIFF
--- a/core/http/Cargo.toml
+++ b/core/http/Cargo.toml
@@ -18,6 +18,7 @@ edition = "2018"
 default = []
 tls = ["rustls", "tokio-rustls", "rustls-pemfile"]
 mtls = ["tls", "x509-parser"]
+http2 = ["hyper/http2"]
 private-cookies = ["cookie/private", "cookie/key-expansion"]
 serde = ["uncased/with-serde-alloc", "serde_"]
 uuid = ["uuid_"]
@@ -50,7 +51,7 @@ optional = true
 [dependencies.hyper]
 version = "0.14.9"
 default-features = false
-features = ["http1", "http2", "runtime", "server", "stream"]
+features = ["http1", "runtime", "server", "stream"]
 
 [dependencies.serde_]
 package = "serde"
--- a/core/http/src/tls/listener.rs
+++ b/core/http/src/tls/listener.rs
@@ -64,7 +64,12 @@ impl TlsListener {
             .map_err(|e| io::Error::new(io::ErrorKind::Other, format!("bad TLS config: {}", e)))?;
 
         tls_config.ignore_client_order = c.prefer_server_order;
-        tls_config.alpn_protocols = vec![b"h2".to_vec(), b"http/1.1".to_vec()];
+
+        tls_config.alpn_protocols = vec![b"http/1.1".to_vec()];
+        if cfg!(feature = "http2") {
+            tls_config.alpn_protocols.insert(0, b"h2".to_vec());
+        }
+
         tls_config.session_storage = ServerSessionMemoryCache::new(1024);
         tls_config.ticketer = rustls::Ticketer::new()
             .map_err(|e| io::Error::new(io::ErrorKind::Other, format!("bad TLS ticketer: {}", e)))?;
--- a/core/lib/Cargo.toml
+++ b/core/lib/Cargo.toml
@@ -19,9 +19,10 @@ edition = "2018"
 all-features = true
 
 [features]
-default = []
+default = ["http2"]
 tls = ["rocket_http/tls"]
 mtls = ["rocket_http/mtls", "tls"]
+http2 = ["rocket_http/http2"]
 secrets = ["rocket_http/private-cookies"]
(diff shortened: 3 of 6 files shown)


SergioBenitez:
The flag needs to be added to the set of flags to toggle when testing in test.sh.

SergioBenitez:
Additionally, the docs need to be updated to reflect this feature's existence.

edgarogh:
I don't have a good computer right now and tests take forever to run. I don't know if they all pass yet (it will mostly depend on whether or not there is a test specific to HTTP/2, else I don't see what could fail). Don't bother enabling my CI rights yet, I'll be patient.

By the way, should I try to implement a test that checks if the server does **not** correctly respond to HTTP/2 requests, or that's not useful ? It's the first time I contribute to such a big codebase, I don't know what the standards are.

edgarogh:
Here's the (only) thing I added that can be considered "docs" (1d1e0381cc3aa81b69c1db5047f223f458695fa4):

https://github.com/SergioBenitez/Rocket/blob/1d1e0381cc3aa81b69c1db5047f223f458695fa4/core/lib/src/lib.rs#L59-L84

Obligatory « English is not my first language ». Feel free to suggest a rephrasing if a sentence doesn't sound natural.

This feature looks slightly too specific to add to the guide too (in my humble opinion), but if you have an idea of a page where it would fit, I can do the job.

edgarogh:
The tests seem to run to completion :heavy_check_mark:

SergioBenitez:
Looks like this needs to be rebased on master to be able to run the CI.

SergioBenitez:
> I don't have a good computer right now and tests take forever to run. I don't know if they all pass yet (it will mostly depend on whether or not there is a test specific to HTTP/2, else I don't see what could fail). Don't bother enabling my CI rights yet, I'll be patient.

Yeah, it's unfortunate how long the test suite takes to run. I hope to have some dedicated infrastructure for Rocket contributors at some point to ease this.

> By the way, should I try to implement a test that checks if the server does **not** correctly respond to HTTP/2 requests when the feature is disabled, or that's not useful ? It's the first time I contribute to such a big codebase, I don't know what the standards are.

That would be nice, but we don't have any infrastructure to write/run these kinds of tests at the moment. See #2082 as well.

edgarogh:
First time I rebase commits, I'm confident I did it correctly but we never know :grin:

The non-rebased (previous) version is tagged just in case.

SergioBenitez:
See CI failures.

edgarogh:
Yup, I probably fixed the problem, I'm at least `cargo check`ing the code locally this time because apparently I can't even write 4 lines valid Rust code :smile:

edgarogh:
```diff
***************
*** 102,104 ****
  error[E0728]: `await` is only allowed inside `async` functions and blocks
!   --> tests/ui-fail-stable/async-entry.rs:73:17
     |
--- 102,104 ----
  error[E0728]: `await` is only allowed inside `async` functions and blocks
!   --> tests/ui-fail-stable/async-entry.rs:73:41
     |
***************
*** 107,109 ****
  73 |         let _ = rocket::build().launch().await;
!    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ only allowed inside `async` functions and blocks
  
--- 107,109 ----
  73 |         let _ = rocket::build().launch().await;
!    |                                         ^^^^^^ only allowed inside `async` functions and blocks
```

SergioBenitez:
Thank you! Merged in https://github.com/SergioBenitez/Rocket/commit/ff7cf68461ea8b3f52d6d816d7684ba4723d3042.

PR #2117, opened 2022-03-04
Allow listening on Unix sockets

### LINKED ISSUE #545: Support for listening on Unix sockets
I would like to utilize my existing infrastructure on [atthost.pl](https://atthost.pl). They are running nginx proxy and allowing me to bind it to any service exposing Unix socket (so I can run servers written in Nodejs, RoR, Golang etc.). The proxy handles SSL/TLS and communicates with servers through socket file. I cannot f.e. open a random port and use `socat` as proxy between TCP and Unix sockets (due to missing `netstat` command).

Rust `std` has UnixListener, but `hyper` 0.10.x does not. Should anyone find time to implement this feature, here is some relevant code:

- guide for hyper 0.11.x: https://github.com/hyperium/hyper/issues/1284
- `hyperlocal` crate for hyper 0.10.x: https://github.com/softprops/hyperlocal

mattfbacon:
This will close #545.


mattfbacon:
Why does the `on_launch_fairing_can_inspect_port` test expect the port to not equal zero? After making this change, it is equal to zero (thus the test fails). I suppose I could edit the test but I want to know why it exists in the first place.

mattfbacon:
TODO:

- UDP listening (or remove)
- `on_launch_fairing_can_inspect_port` test (or edit test to pass)

@SergioBenitez please advise on these matters.

SergioBenitez:
I recommend looking through the initial pull request adding UDS support to Rocket. You should ensure that you're addressing all of the issues we encountered there, and more importantly, note the suggestions for how to resolve them. The most pressing issue here is what to do about the `port` config value as it makes no sense when the address type is a domain socket. Other issues include creating/deleting/locking the socket file, failing gracefully when the file exists, and proper error messages or support for Windows users.

mattfbacon:
## Issues you mentioned

### `port` config value

I handled the port config value in 603bf37379e2f0864c888a24778a6b6fce45f93e and 87a4201cf926ded1369f173a615d08c3c690d5ad. Let me know what you think about that approach.

### Socket file lifetime

You said in the issue:

> I think the most natural thing for Rocket to do is to fail on start-up if a socket already exists, create a socket on start-up if it doesn't, and delete the socket on shut-down.

Which is great because it means that no lockfiles are required: the socket simply existing means that it's in use, and Rocket will fail to launch. This exactly matches my implementation so far though I have not checked the error message for address-in-use to see if it needs to be improved.

### Proper error messages

I have not focused much on this but will do so as my next step once all the essential functionality has been implemented.

### Windows users

Hmm.. **do we want to gate the Unix variant of the BindableAddr enum with `#[cfg(unix)]`?** Seems like the best approach to me.

## Other issues I found in #545 

### TLS

Already handled in 2f3b6aba385f2959c3878bc9304e8c42d119d95e with some fixes in subsequent commits.

### UDP listening

**I still need feedback on this. Should we add it?** UDP is connectionless, so (AFAICT) it would require some significant changes to internal Rocket code to handle the lack of `UdpListener`-`UdpStream` pair similar to the ones that exist for the TCP and Unix-socket protocols. It could be done; I just want to know if you are interested in it, or if I should just drop it. YAGNI?

## Summary of Necessary Input

- In terms of Windows "support", just gate the `Unix` variant of `BindableAddr` behind `#[cfg(unix)]` and call it a day?
- Do we need UDP listening at all? Or more specifically do we need it in this PR?
- What is the deal with that test? What is it trying to test, and what do you suggest I do with it considering it's currently the only failing test in the suite?

mattfbacon:
BTW, could you approve running workflows for this PR?

IniterWorker:
It seems to me a lot of CI errors are related to cfg(unix) not present and `on_ignite_fairing_can_inspect_port`.

Do you have any plan to push it forward?

mattfbacon:
Yes! I just got busy recently. I will have some time later this week to work on this.

mattfbacon:
Workflows mostly succeeded when I ran locally. Was not able to test Windows so please approve the workflows.

mattfbacon:
Here's what's left:

- UDP listening
- Support for Windows domain sockets via `uds_windows` crate

SergioBenitez:
> ### `port` config value
> I handled the port config value in [603bf37](https://github.com/SergioBenitez/Rocket/commit/603bf37379e2f0864c888a24778a6b6fce45f93e) and [87a4201](https://github.com/SergioBenitez/Rocket/commit/87a4201cf926ded1369f173a615d08c3c690d5ad). Let me know what you think about that approach.

This looks pretty good, though I'm concerned with serde's inability to handle `flatten` correctly, specially for `enum`s, which according to serde's docs aren't actually supported:

> The flatten attribute inlines keys from a field into the parent struct. flatten may be used any number of times within the same struct. It is supported only within structs that have named fields, and the field to which it is applied must be a struct or map type.

I also want to make it a priority to make every existing config _just work_ as it did before. That is, without needing to change the `ADDRESS` or `PORT`.

> Which is great because it means that no lockfiles are required: the socket simply existing means that it's in use, and Rocket will fail to launch. This exactly matches my implementation so far though I have not checked the error message for address-in-use to see if it needs to be improved.

Doing the deletion on `Drop` seems incredibly error prone. In general, `Drop` should not be used for cleanup that _needs_ to happen. In this case, it doesn't really _need_ to happen, but if it doesn't happen, the application will fail to restart. This seems incredibly problematic when coupled with supervisors that will restart an application that fails. We could easily cause an application to go into a restart loop.

In general, I think I've changed my position on this issue so that if a file exists with the desired socket name, Rocket tries to delete it unless there's a lock file around. The OS will ensure the lock file is gone if the application crashes, which coupled with retrying to lock a few times, will allow the Rocket application to restart reliably if it happens to fail.

> ### Windows users
> Hmm.. **do we want to gate the Unix variant of the BindableAddr enum with `#[cfg(unix)]`?** Seems like the best approach to me.

Whatever we do, the error message needs to be incredibly clear. A few `cfg`s perhaps coupled with `compile_error!` might do the trick.

> ### UDP listening
> **I still need feedback on this. Should we add it?** UDP is connectionless, so (AFAICT) it would require some significant changes to internal Rocket code to handle the lack of `UdpListener`-`UdpStream` pair similar to the ones that exist for the TCP and Unix-socket protocols. It could be done; I just want to know if you are interested in it, or if I should just drop it. YAGNI?

I think we can leave this for later. 

> * What is the deal with that test? What is it trying to test, and what do you suggest I do with it considering it's currently the only failing test in the suite?

This is trying to make sure that an OS-chosen port can be read by the application via the config once the application starts. Specifically, it tests the follow scenario:

  1. The user sets the port to `0`, which asks the OS to choose an available port. `config.port = 0`.
  2. Rocket try to bind to `IP:0`. `config.port` is still `0`.
  3. Bind succeeds. The OS has chosen an available port. `config.port` should now be the OS chosen port.
  4. The user reads `config.port` which is the chosen port.

SergioBenitez:
I've pushed a commit that fixes the UI tests for the latest rustc stable release.

mattfbacon:
OK @SergioBenitez I believe this is ready to merge so I have changed the status to "Ready for Review".

mattfbacon:
(Pending final review of course)

mattfbacon:
What did we decide about the management of the socket in terms of creation, deletion, etc?

SergioBenitez:
> In general, I think I've changed my position on this issue so that if a file exists with the desired socket name, Rocket tries to delete it unless there's a lock file around. The OS will ensure the lock file is gone if the application crashes, which coupled with retrying to lock a few times, will allow the Rocket application to restart reliably if it happens to fail.
> 
> >

> In general, I think I've changed my position on this issue so that if a file exists with the desired socket name, Rocket tries to delete it unless there's a lock file around. The OS will ensure the lock file is gone if the application crashes, which coupled with retrying to lock a few times, will allow the Rocket application to restart reliably if it happens to fail.

mattfbacon:
I'll try to fix the conflicts some time this week.

SergioBenitez (changes requested):
I started a review but quickly had to stop as the changeset, in its current form, contains far too many changes that have little to do with the proposal at hand. Please remove all style related changes to existing code, including those to Rust source code and Markdown files. None will be accepted as part of this PR, and their presence makes reviewing this PR difficult. Furthermore, please remove any changes that are not related to adding support for UDS.

SergioBenitez:
> When you say "Rocket tries to delete it unless there's a lock file around" are you referring to flock? I'm not sure that works on Windows, and also it only works for other Rocket processes.

Yes. The point is that we ensure two Rocket servers aren't trying to compete for the same socket file. This would result in all kinds of issues if allowed. Of course, if you require exclusive creation of the socket file to succeed before continuing, then this isn't necessary. As it stands, the implementation takes neither approach. 

My concern from earlier remains:

> This [behavior of failing to restart because of an existing socket file] seems incredibly problematic when coupled with supervisors that will restart an application that fails. We could easily cause an application to go into a restart loop.

I cannot foresee accepting this PR while also allowing this failure scenario to remain possible. While it _is_ an edge case, it is also a possibility. Rocket takes great measures to ensure it doesn't crash, but there's nothing we can do about Hyper crashing, for instance, or folks setting panic to abort, and h2 is known to have plenty of bugs that lead to panics. Were to happen, we facilitate a "permanent", one-shot DoS attack.

One possible approach is to have a mechanism by which we know both whether a socket file was created by Rocket and whether it's still in use. This way, we can delete socket files that are no longer in use but were known to be created by Rocket. We would fail to start if we can neither create the file nor delete an existing file given the conditions above; this means the file corresponds to a something created for another purpose, and thus we can't reasonably ensure we have exclusive access, or one in use by Rocket.

One possible concrete approach, which I've already mentioned, is to exclusively create a lock file (socket.lock.rocket).  If the lock file is present, the socket was created by Rocket. If it is locked, the socket is in use by Rocket. If it is not, then it is not in use by Rocket, and the existing socket can be re-used after locking.

mattfbacon:
The majority of the changes present in this PR are now due to its branch's old base commit. I think maybe the best path forward is to reapply the core UDS changes in a new PR that is based on the current head. So I'll probably close this one and make a new one.

SergioBenitez:
> The majority of the changes present in this PR are now due to its branch's old base commit. I think maybe the best path forward is to reapply the core UDS changes in a new PR that is based on the current head. So I'll probably close this one and make a new one.

As far as I can tell, the majority of changes in the diff are style changes that are likely applied by a tool automatically. Nevertheless, you'll need to rebase on master. You don't need to create a new PR. Just force push to the PR branch and this PR will update.

SergioBenitez:
> > As it stands, the implementation takes neither approach.
> 
> 
> 
> You can't create a socket that already exists, so exclusivity is maintained.

Apologies, that's correct. I wrote that comment with the second comment in mind. 

> 
> 
> > One possible concrete approach, which I've already mentioned, is to exclusively create a lock file (socket.lock.rocket). If the lock file is present, the socket was created by Rocket. If it is locked, the socket is in use by Rocket. If it is not, then it is not in use by Rocket, and the existing socket can be re-used after locking.
> 
> 
> 
> I'm trying to understand the scenario you're imagining. For the lock file to be relevant, two different applications, one of them not being Rocket, would have to be contending for the same socket path. That doesn't seem like something that we should have to handle, and it would probably break most other applications anyway.

The scenario is: Rocket runs, binds to a UDS, crashes, restarts; we'd like it to run again, binding to the same socket. Unless we do something to get rid of the file on crash, we won't be able to restart. One possibility, on unix at least, is to unlink before we bind. If the process crashes or exits, the file is gone.

mattfbacon:
> You don't need to create a new PR. Just force push to the PR branch and this PR will update.

Will do.

BTW, @SergioBenitez, how important is backwards compatibility? Even with the efforts I made to make transitioning to the new protocol addresses fairly easy, it's still a breaking change, so I guess this will have to be 0.6? Or are breaking changes OK in beta? Not sure how you're orchestrating this release model.

IniterWorker:
> The scenario is: Rocket runs, binds to a UDS, crashes, restarts; we'd like it to run again, binding to the same socket. Unless we do something to get rid of the file on crash, we won't be able to restart. Another possibility, on unix at least, is to unlink before we bind. If the process crashes or exits, the file is gone.

It makes sense to unlink before binding. Some well-known applications remove the UNIX domain socket at startup. It seems fairly more reliable than removing from a crash. It will prevent an infinite restart loop from the external daemon in the wrost scenario.

mattfbacon:
Looks like due to the nature of the upstream changes I effectively have to re-implement most or all of the UDS functionality, and I don't have the time or energy for that right now. I'm going to close this and maybe someone else can pick up on the general idea.

PR #2125, opened 2022-03-19
Migrate to Edition 2021

xfix:
Since Rocket 0.5 has not yet been released, this may be a good opportunity to increase MSRV so that it won't have to be done after its release.

### DIFF
--- a/benchmarks/Cargo.toml
+++ b/benchmarks/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "rocket-benchmarks"
 version = "0.0.0"
-edition = "2018"
+edition = "2021"
 publish = false
 
 [workspace]
--- a/contrib/db_pools/codegen/Cargo.toml
+++ b/contrib/db_pools/codegen/Cargo.toml
@@ -7,7 +7,7 @@ repository = "https://github.com/SergioBenitez/Rocket/contrib/db_pools"
 readme = "../README.md"
 keywords = ["rocket", "framework", "database", "pools"]
 license = "MIT OR Apache-2.0"
-edition = "2018"
+edition = "2021"
 
 [lib]
 proc-macro = true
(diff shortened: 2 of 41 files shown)


somehowchris:
Nice, maybe this would be a nice time to use the `rust-version` attribute inside `Cargo.toml`. Yes in order to use the 2021 edition we already need `1.56.0` but I would see it as a benefit in order to actually declare the msrv inside the `Cargo.toml` explicitly

https://github.com/SergioBenitez/Rocket/pull/2076#discussion_r819901188

SergioBenitez:
Let's do it, but indeed as @somehowchris suggests, lets set `rust-version` appropriately. Can you rebase on `master` and make this change?

Also, did we really not need to make any changes for `2021`? Or are there no improvements as a result? We should, at least, be able to remove `TryInto`, `TryFrom`, and `FromIterator` imports and use `.into_iter()` directly. Did you run `cargo fix --edition`?

xfix:
> Also, did we really not need to make any changes for `2021`? Or are there no improvements as a result? We should, at least, be able to remove `TryInto`, `TryFrom`, and `FromIterator` imports and use `.into_iter()` directly. Did you run `cargo fix --edition`?

One change was necessary - one of documentation tests used `panic!(())` which doesn't work in Edition 2021 as `panic!` macro cannot panic with an arbitrary object anymore. I changed that to use `panic!()` as I think double parentheses were a typo.

That said, I made some manual tweaks in separate commits - removing those imports and using `.into_iter()` directly. `cargo fix --edition-idioms` doesn't apply those changes.

SergioBenitez:
Merged in https://github.com/SergioBenitez/Rocket/commit/7bbe0457a5ecc9d4b3ca0a2bc6120ffbe1e7febd and https://github.com/SergioBenitez/Rocket/commit/4d258739f5f9943185dcc51009313ca75fa285be.

PR #2126, opened 2022-03-20
Replace deprecated 'std::array::IntoIter::new()'

JOE1994:
Non-Functional-change.
Helps reduce noise from "deprecated" warnings :)


SergioBenitez:
Let's tackle this as part of #2125.

PR #2139, opened 2022-03-28
Rename type alias Result to DieselResult

graves501:
Thanks for providing these examples! To me it was a bit confusing that `Result` was aliased. So to make the type alias a bit more distinct and easier to understand - especially for Rust newbies - I suggest to rename it to something like `DieselResult` or something similar. :smile:


SergioBenitez:
It's actually rather common (and thus idiomatic) to create domain-specific `Result` aliases in Rust, believe it or not! See, for instance, https://doc.rust-lang.org/stable/std/io/type.Result.html, https://doc.rust-lang.org/stable/std/fmt/type.Result.html, and https://doc.rust-lang.org/stable/std/thread/type.Result.html

PR #2141, opened 2022-03-30
better error message when unsupported TLS key header

### LINKED ISSUE #2128: Better error message needed for invalid TLS key header
https://github.com/SergioBenitez/Rocket/blob/8825f936efcf05d1b8382fcdb2d528ff7c3ff933/core/http/src/tls/util.rs#L25

I had to dig through the rocket source and rustls issue tracker to find that `"invalid key header"` means something closer to "Unsupported key header, available formats are: RSA, PKCS8".

awoimbee:
Fixes #2128.


SergioBenitez:
Merged in https://github.com/SergioBenitez/Rocket/commit/4df97f0e250df720d42a45b9511b179841a1d6bf. Thanks!

PR #2145, opened 2022-04-15
Update 'sqlite' contrib dependencies

tobymurray:
This bumps the dependencies needed for using SQLite + sync DB pools to their latest versions.

The same chore was tackled almost a year ago here: https://github.com/SergioBenitez/Rocket/commit/f4ecc3e3bdb01fa2f79b0f89acc8a80d0a11a677.

### DIFF
--- a/contrib/sync_db_pools/lib/Cargo.toml
+++ b/contrib/sync_db_pools/lib/Cargo.toml
@@ -27,8 +27,8 @@ diesel = { version = "1.0", default-features = false, optional = true }
 postgres = { version = "0.19", optional = true }
 r2d2_postgres = { version = "0.18", optional = true }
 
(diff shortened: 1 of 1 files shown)


SergioBenitez:
This hadn't been done because the `-sys` dependency generally causes an irreconcilable version conflict. Let's see if we have working versions now.

SergioBenitez:
See the CI failures. I don't think we can update the sqlite dependencies without usptream properly configuring their `-sys` dependencies.

tobymurray:
> I don't think we can update the sqlite dependencies without usptream properly configuring their -sys dependencies.

Yeah, seems like this is a bit of a contentious dependency. I guess first stop is Diesel

Issue #1878, opened 2021-08-31
testable routing


cataggar:
I want to be be able to test that routes resolve correctly, given the HTTP method and URI.

``` rust
#[cfg(test)]
mod test {
    use super::*;
    use rocket::Router;

    #[test]
    fn test_route_name() -> Result<(), Error> {
        let mut router = Router::new();
        for route in api_routes() {
            router.add_route(route);
        }

        assert_eq!("Hello".to_owned(), route_name(&router, "GET", "/")?);
        // a lot more asserts
        Ok(())
    }
}

// Get the route name from the HTTP method + URI combination.
fn route_name(router: &Router, method: &str, uri: &str) -> Result<String, Error> {
    let client = rocket::local::blocking::Client::debug_with(Vec::new()).map_err(Error::CreateLocalClient)?;
    let request_method = Method::from_str(method).map_err(|_| Error::InvalidHttpMethod(method.to_owned()))?;
    let request_uri = rocket::http::uri::Origin::parse(uri).map_err(|_| Error::ParseUri(uri.to_owned()))?;
    let request = client.req(request_method, request_uri);
    let mut routes = router.route(&request);
    let route = routes.next().ok_or_else(|| Error::RouteNotFound {
        method: method.to_owned(),
        uri: uri.to_owned(),
    })?;
    let name = route.name.as_ref().ok_or_else(|| Error::UnnamedRoute {
        method: method.to_owned(),
        uri: uri.to_owned(),
    })?;
    Ok(name.to_owned().to_string())
}

#[non_exhaustive]
#[derive(Debug, thiserror::Error)]
pub enum Error {
    #[error("unable to create rocket client: {0}")]
    CreateLocalClient(rocket::Error),
    #[error("unable to parse uri: {0}")]
    ParseUri(String),
    #[error("invalid http method: {0}")]
    InvalidHttpMethod(String),
    #[error("route not found for {method} {uri}")]
    RouteNotFound { method: String, uri: String },
    #[error("unnamed route for {method} {uri}")]
    UnnamedRoute { method: String, uri: String },
}
```

**Existing Functionality**

`Router`

**Suggested Changes**

The above code works if `Router` is made public #1873 and the route name is settable #1872.

**Alternatives Considered**

**Additional Context**

SergioBenitez:
> I want to be be able to test that routes resolve correctly, given the HTTP method and URI.

[...]

PR #2167, opened 2022-05-09
Route tracing

the10thWiz:
Adds support for testing which route generated a response. This is an implementation of the ideas I put forward in #1878, with a few minor changes. First, I did not create a macro, since my implementation works just fine with a generic type parameter. Writing a macro is pretty trivial, if it's desired. I also added tests to the hello world example to demonstrate what the actual tests look like.

There are a few improvements that could be made: 
- The methods added to `LocalResponse` (`routed_by`, `caught_by` and `was_caught`) could probably have better names.
- The Request object could (in testing mode) track which routes were attempted, and some basic information about the result. This should be pretty trivial to add, although it may require more that just the route's `TypeId` to actually use this to produce meaningful information.

The only regression I see is that `Catcher` now has a private member, so other crates cannot directly construct a `Catcher`, but I think this is acceptable, since it should probably be constructed with the constructor as before.


core/lib/src/route/route.rs
```diff
@@ -17,7 +24,8 @@ use crate::sentinel::Sentry;
 ///
 /// ```rust
 /// # #[macro_use] extern crate rocket;
-/// # use std::path::PathBuf;
+/// #
+/// use std::path::PathBuf;
```
peddermaster2, line 28: Is this change nessesary?

core/lib/src/catcher/catcher.rs
```diff
@@ -8,6 +9,12 @@ use crate::catcher::{Handler, BoxFuture};
 
 use yansi::Paint;
 
+/// A generic trait for catcher types. This should be automatically implemented on the structs
+/// generated by the codegen for each catcher, and manually implemented on custom cater types.
```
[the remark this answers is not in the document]
the10thWiz: Thanks!

the10thWiz:
It looks like the issue with Windows Debug is a memory or hard drive error.

SergioBenitez:
Indeed, we're running out of disk space. Let's disregard it for now.

the10thWiz:
I think I've reached an API I'm pretty happy with. Since the type tracking is done internally, this could be extended to support listing the routes attempted in order, but I don't think that's necessary for this PR. I'm also happy enough with the syntax presented here, I don't feel the need to introduce a new assert macro. @SergioBenitez, this PR is ready for a review, whenever you get a chance.

core/lib/src/catcher/catcher.rs
```diff
@@ -127,6 +134,9 @@ pub struct Catcher {
     ///
     /// This is -(number of nonempty segments in base).
     pub(crate) rank: isize,
+
+    /// A unique route type to identify this route
+    pub(crate) catcher_type: Option<TypeId>,
```
cataggar, line 139: Very cool! I had to read about [TypeId](https://doc.rust-lang.org/std/any/struct.TypeId.html).

cataggar (approved):
Very cool! This may fix #1878.

SergioBenitez:
I've rebased on the latest master and push to your fork so that I can review.

SergioBenitez (commented):
The way I'm understanding this PR is as making fundamentally two changes:

  1. Associate routes and catchers with an ID. Here, the ID is of type `TypeId` which is indirect obtained via new traits `CatcherType` and `RouteType`, both of which require `Any`. The new traits exist so that we can write conditionals of the form `response.was_routed_by::<super::world>()`, where `super::world` is a function annotated with `#[route]`.

  2. Before a route handler or catcher is called, set state in `Request` indicating the route or catcher. The state can then be queried to check which route or catcher actually handled the request.

**On 1:** I wish we could simple generate a unique ID whenever `Route::new()` and `Catcher::new()` are called and use that instead of introducing a trait that requires manual implementation for custom routes/catchers, and where failing to implement the trait doesn't result in a compile-time error but inhibits functionality. I would really like to be able to avoid these drawbacks while still being able to write something like `response.was_routed_by::<super::world>()`.

One idea is to convert `Route` and `Catcher` into traits. Not only would this resolve the above, but it would also obviate the need for several namespace hacks we have today and pave a way to interact with code generated routes, something sorely missing today. I'm not sure how this approach changes everything else, in particular what custom routes will look like, but it's worth considering.

**On 2:** I'm concerned about the myriad `set()` calls sprinkled about. Is there a way to centralize these calls so we can't forget them?

core/lib/src/local/response.rs
```diff
@@ -180,6 +180,134 @@ macro_rules! pub_response_impl {
...
+    /// [`FileServer`](crate::fs::FileServer) implementes `RouteType`, so a route that should
+    /// return a static file can be checked against it. Libraries which provide custom Routes should
+    /// implement `RouteType`, see [`RouteType`](crate::route::RouteType) for more information.
+    ///
+    /// # Note
+    ///
+    /// This method is marked as `cfg(test)`, and is therefore only available in unit and
+    /// integration tests. This is because the list of routes attempted is only collected in these
+    /// testing environments, to minimize performance impacts during normal operation.
+    #[cfg(test)]
```
SergioBenitez, line 262: This is unfortunately not the way `cfg(test)` works. `cfg(test)` only gets enabled on the top-level crate being tested, not on its dependencies. So `rocket` compiled as a dependency of an application, even when compiling that application's tests, will never see this `cfg` enabled.

PR #2176, opened 2022-05-11
Avoid causing clippy warnings with default fields

xfix:
Fixes an issue that causes `#[field(default = ...)]` to cause Clippy warnings (`clippy::useless_conversion` and `clippy::needless_late_init`). Works around rust-lang/rust-clippy#7845.


SergioBenitez:
Our general policy is to fix any issues presented by clippy _without_ working around bugs in clippy. As such, I've kept the relevant changes in 6f223518e43961a46f228641f1fc08319e076e51. Thank you!

PR #2181, opened 2022-05-13
Update default Permission-Policy header to disable browsing-topics

arjentz:
Since Google abandoned FLoC, and introduced Topics, hereby suggesting these changes. The feature itself is still experimental: https://bugs.chromium.org/p/chromium/issues/detail?id=1294456

### DIFF
--- a/core/lib/src/shield/policy.rs
+++ b/core/lib/src/shield/policy.rs
@@ -486,7 +486,7 @@ impl From<&Prefetch> for Header<'static> {
 /// # Default
 ///
 /// The default returned via [`Permission::default()`] blocks access to the
-/// `interest-cohort` feature, otherwise known as FLoC, which disables using the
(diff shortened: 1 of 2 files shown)


xfix:
https://github.com/patcg-individual-drafts/topics says that `Permissions-Policy: interest-cohort=()` will also work.

SergioBenitez:
Based on https://github.com/SergioBenitez/Rocket/pull/2181#issuecomment-1129741758, it looks like we don't need to change anything, and _not_ making this change means we do what we intend even on ~older browsers.

arjentz:
I don't think there is any disadvantage to setting both, using:
```
Permissions-Policy: browsing-topics=(), interest-cohort=()
```
They are also both still experimental, and not enabled by default (yet), I believe.

PR #2183, opened 2022-05-13
Salvage bad request

the10thWiz:


SergioBenitez:
I don't think we should have that logic inside the `from_hyp` method. Instead, I think we should build that request where we were already building the fake request, just do so with as much of the information as we can.

the10thWiz:
The main issue with that approach is that the `ConnectionMeta` struct is not copy, which means it would need to be cloned for every request. I'll look into this again, and see what I can do.

the10thWiz:
@SergioBenitez I've refactored this PR to use a separate method, and it avoids cloning the `ConnectionMeta` struct as well.

SergioBenitez:
Merged the general idea in 5cb70ec58ca9c38f7b54d6a7a2e84b3a674e76ba. I had to play with it a bit to see what made sense, code-wise. I finally settled on a "pattern" of sorts I use in other places: keep track of errors in a `Vec`, check if the `Vec` is empty at the end and if it isn't return the error. This is similar to your approach in the second commit but to make sense here, I created a new `BadRequest` error type to _always_ return a `Request` even when there are errors.

PR #2194, opened 2022-05-19
Convert Form Error impl to Const Generic

the10thWiz:
Converts an older style array impl to use const generics, allowing any
length, not just the listed sizes.

I'm not sure if the old implementation was intentional (limiting the array to a length of 9), but I would assume it wasn't. I believe the minimum complier version for Rocket is high enough that const generics are guaranteed to be supported.

Please let me know if there are any issues with this.


SergioBenitez:
This is great, thanks! Merged in f21da79f44c773e2761fe7c9aae3c008c999e62d.

PR #2195, opened 2022-05-19
Update form Error types to improve lifetimes

the10thWiz:
Adds an override of the `extend` method that enables derived form
validators to return `form::Result<'static, ()>`. It appears the central
issue is the `Cow<'v, [Cow<'v, str>]>` potentially held by the
`ErrorKind` struct, which Rust can't automatically shorten the lifetime
for. In theory, it should be safe to perform the same transform via an
unsafe transmute, but I would rather avoid unsafe code, and rely on the
compiler to recogize that it's effectively a no-op.

### DIFF
--- a/core/lib/src/form/error.rs
+++ b/core/lib/src/form/error.rs
@@ -210,6 +210,40 @@ pub enum ErrorKind<'v> {
     Io(io::Error),
 }
 
+/// Map the choices Cow from any lifetime to a shorter one
+fn cow_map<'r, 'v>(c: Cow<'r, [Cow<'r, str>]>) -> Cow<'v, [Cow<'v, str>]> where 'r: 'v {
+    match c {
+        Cow::Owned(v) => Cow::Owned(v),
+        Cow::Borrowed(v) => Cow::Borrowed(v),
+    }
+}
+
+impl<'v> ErrorKind<'v> {
+    /// Explicitly shorten the ErrorKind's lifetime
+    fn with_lifetime<'r>(s: ErrorKind<'r>) -> Self where 'r: 'v {
+        match s {
+            ErrorKind::InvalidLength { min, max } => Self::InvalidLength { min, max },
+            ErrorKind::InvalidChoice { choices } => Self::InvalidChoice {
+                choices: cow_map(choices)
+            },
+            ErrorKind::OutOfRange { start, end } => Self::OutOfRange { start, end },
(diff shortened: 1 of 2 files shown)


SergioBenitez:
It would be nice to get this in. Sadly, it seems that some change in the `FromForm` codegen makes this no longer work. I would more than welcome someone taking a look to make this work.

For posterity, I rewrote the change to `extend` to look like this:

```rust
    // "Override" the `extend` method so we can use a longer lifetime `'r`.
    #[doc(hidden)]
    pub fn extend<'r: 'v>(&mut self, iter: Errors<'r>) {
        self.0.extend(iter.0.into_iter().map(|e| Error {
            name: e.name.map_or(None, |name: NameBuf<'r>| Some(name)),
            value: e.value,
            entity: e.entity,
            kind: match e.kind {
                ErrorKind::InvalidLength { min, max } => ErrorKind::InvalidLength { min, max },
                ErrorKind::InvalidChoice { choices } => ErrorKind::InvalidChoice {
                    choices: match choices {
                        Cow::Owned(v) => Cow::Owned(v),
                        Cow::Borrowed(v) => Cow::Borrowed(v),
                    }
                },
                ErrorKind::OutOfRange { start, end } => ErrorKind::OutOfRange { start, end },
                ErrorKind::Validation(e) => ErrorKind::Validation(e),
                ErrorKind::Duplicate => ErrorKind::Duplicate,
                ErrorKind::Missing => ErrorKind::Missing,
                ErrorKind::Unexpected => ErrorKind::Unexpected,
                ErrorKind::Unknown => ErrorKind::Unknown,
                ErrorKind::Custom(s, e) => ErrorKind::Custom(s, e),
                ErrorKind::Multipart(e) => ErrorKind::Multipart(e),
                ErrorKind::Utf8(e) => ErrorKind::Utf8(e),
                ErrorKind::Int(e) => ErrorKind::Int(e),
                ErrorKind::Bool(e) => ErrorKind::Bool(e),
                ErrorKind::Float(e) => ErrorKind::Float(e),
                ErrorKind::Addr(e) => ErrorKind::Addr(e),
                ErrorKind::Io(e) => ErrorKind::Io(e),
            }
        }))
    }
```

I'd prefer to just keep everything local to this method, as above.

SergioBenitez:
Closing due to https://github.com/SergioBenitez/Rocket/pull/2195#issuecomment-1503879931. My apologies.

Issue #2211, opened 2022-05-30
basic clippy warnings get triggered by macros


toxeus:
**Description**
1. `rocket::post()` proc macro triggers `clippy::unused-unit`
1. `rocket::routes!` triggers `clippy::no-effect-underscore-binding`

**To Reproduce**

Enable corresponding lints and run clippy.

**Expected Behavior**

No clippy warnings for basic clippy warnings.

**Environment:**

 - OS Distribution and Kernel: Linux
 - Rocket Version: 0.5rc2

SergioBenitez:
We don't consider these bugs, but if we can improve codegen to prevent these lints from triggering, then great! Please feel free to send a PR.

I would suggest running `cargo expand` to determine which part of the generated code is causing the lint to fire and then looking at/grepping the codegen crate to find/modify the generation site.

toxeus:
> We don't consider these bugs

Agreed, it's technically not a bug. But it's terrible macro hygiene. Your generated code shouldn't trigger lints in my crates.

SergioBenitez:
> > We don't consider these bugs
> 
> 
> 
> Agreed, it's technically not a bug. But it's terrible macro hygiene. Your generated code shouldn't trigger lints in my crates.

This doesn't have anything to do with macro hygiene. What's more, this isn't rocket triggering warnings in your code, this is an external tool, clippy, that you're running yourself, triggering a warning, often erroneously. Finally, I'm confused by the seeming condescension (my apologies if I'm reading your message incorrectly). I have no opposition to improving codegen to resolve issues of this sort; my entire prior comment except for the first sentence, should make this clear. My "we don't consider these bugs" comment was simply a matter of ontology.

toxeus:
My apologies if this came across condescending. That's not the intention. I don't agree with your statement but I also think we both have made our points already.

PR #2212, opened 2022-05-30
codegen: fix clippy::no-effect-underscore-bindings

toxeus:
This lint checks for binding to underscore prefixed variables without side-effects.

Further usage of such variables is not checked, which can lead to false positives if they are used later in the code.

To fix the false positive we use Rust's `r#` variable prefix to avoid collisions with Rust keywords.

Fixes first part of #2211

### DIFF
--- a/core/codegen/src/bang/mod.rs
+++ b/core/codegen/src/bang/mod.rs
@@ -20,9 +20,9 @@ fn struct_maker_vec(
     // Parse a comma-separated list of paths.
     let paths = <Punctuated<Path, Token![,]>>::parse_terminated.parse(input)?;
     let exprs = paths.iter().map(|path| {
-        let expr = map(quote_spanned!(path.span() => ___struct));
+        let expr = map(quote_spanned!(path.span() => r#struct));
         quote_spanned!(path.span() => {
-            let ___struct = #path {};
+            let r#struct = #path {};
             let ___item: #ty = #expr;
             ___item
         })


toxeus:
The tests fail for me also without this commit :man_shrugging:

SergioBenitez:
The reason we use `___foo` instead of `foo` or `r#foo` is because of the lack of hygiene for proc-macros. The concern is that if we use `foo` in our codegen, an errant `foo` in the user's code will cause a naming conflict and wreak all sorts of havoc. In the original PR for the lint (https://github.com/rust-lang/rust-clippy/pull/499), this use-case was considered and as far as I can tell, two exceptions to the lint were made: 1) idents with two underscores aren't considered `__anything`, and 2) idents with differing hygiene than their context. 2) doesn't apply given we're in a proc-macro, but 1) should, since `___response` starts with `__`. So, why is this triggering?

I think this is an issue for `clippy` to resolve. Using `r#struct` opens us up to the same issues as using a bare `foo`, even if it is less likely.

toxeus:
> The reason we use `___foo` instead of `foo` or `r#foo` is because of the lack of hygiene for proc-macros. 

Thanks for providing some context here. Given that those variables clash with rust keywords (`vec`) and macros imported by the std preamble (`vec`) it makes the impression that the underscores have been chosen to avoid those conflicts instead of working around macro hygiene.  

> The concern is that if we use `foo` in our codegen, an errant `foo` in the user's code will cause a naming conflict and wreak all sorts of havoc.

I don't see how that can be a concern given that all variables live in their own scope:
https://github.com/SergioBenitez/Rocket/blob/7275df9fdfab84c493c0fe4cc3f16e66018f2f14/core/codegen/src/bang/mod.rs#L31-L34

PR #2225, opened 2022-06-07
Added implementation of #[field(flatten)].

azewierzejew:
Limited to on flattened struct.

### DIFF
--- a/core/codegen/src/derive/form_field.rs
+++ b/core/codegen/src/derive/form_field.rs
@@ -19,6 +19,7 @@ pub struct FieldAttr {
     pub validate: Option<SpanWrapped<syn::Expr>>,
     pub default: Option<syn::Expr>,
     pub default_with: Option<syn::Expr>,
+    pub flatten: Option<bool>,
 }
 
 impl FieldAttr {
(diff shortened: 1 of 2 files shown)


azewierzejew:
I'm proposing here an implementation (without error handling) of the flatten attribute in `FromForm` derive macro.
Before finishing it I would like to discuss the semantics of this trait.

With current `FromForm` trait interface, there is no way which flattened sub-struct uses given field.
One solution for that would be to pass the field to all flattened structs.
But that doesn't really work either because `DataField` can't be copied.
Also that would produce different results than serde for names duplicated by flattening, e.g:
```
#[derive(FromForm)]
struct X {
    name: String,
}
#[derive(FromForm)]
struct Y {
    #[field(flatten)]
    inner1: X,
    #[field(flatten)]
    inner2: X,
}
```
In above case `serde` initializes `inner1.name` with provided value and `inner2.name` with default value (or gives an Error).

Therefore I believe restricting flattening to only one field is the only way without serious API changes.

Azumi4han:
any updates here? 😭😭😭😭

PR #2235, opened 2022-06-19
Fix typo and imporve echo command in guide/pastebin

Indosaram:
1. Fix a typo.
2. A string in `echo` command does not need to be wrapped by double-quotes.


Indosaram:
Issue #2236

SergioBenitez:
That's true! But `echo "hello world"` will do something different than `echo "hello.   world"`. I'd rather use the quotes as a result. Thank you!

PR #2237, opened 2022-06-20
also evaluate sentinels of data guards

pennae:
sentinels of FromData arguments to should also be evaluated.


pennae:
looks like this was merged manually in the interim, closing :)

PR #2243, opened 2022-06-22
Add ULID support

DXist:
See ULID specification at https://github.com/ulid/spec

### DIFF
--- a/core/http/Cargo.toml
+++ b/core/http/Cargo.toml
@@ -23,6 +23,7 @@ http2 = ["hyper/http2"]
 private-cookies = ["cookie/private", "cookie/key-expansion"]
 serde = ["uncased/with-serde-alloc", "serde_"]
 uuid = ["uuid_"]
+ulid = ["ulid_"]
 
 [dependencies]
 smallvec = "1.0"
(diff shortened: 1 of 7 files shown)


SergioBenitez:
The features that we implement in `rocket` directly are those that are likely to see wide use, for which using a wrapper type because of coherence issues is undesired, or both. There are quite a few replacements for UUID in the wild. Why should Rocket prefer ULID over anything else? Can this be implemented in the the ULID library directly, which would then expose a `rocket` feature? Are the ergonomics of implementing this in a separate library entirely too unfortunate to bear?

DXist:
I agree that ULID spec has niche usage. It doesn't reguire centralized coordination during ID generation and is a good option when there is a preference to use monotonically increasing identifiers in distributed environments - like sharded databases or push ids for mobile clients.

So the choice is more tied to storage layer than a web framework like Rocket.

I'll try to implement it in the ULID library itself as a `rocket` feature. I chose `rusty_ulid` because it supports lenient parsing of ULIDs according to Crockford base32 specification - it normalizes similarly looking characters like 0/O or I/1/l to 0 and 1.

SergioBenitez:
> I'll try to implement it in the ULID library itself as a `rocket` feature. I chose `rusty_ulid` because it supports lenient parsing of ULIDs according to Crockford base32 specification - it normalizes similarly looking characters like 0/O or I/1/l to 0 and 1.

Got it. That sounds good. If that doesn't pan out, I would suggest publishing a `rocket-ulid` library as another alternative. If that sees considerable usage, we can integrate the support into `rocket` itself.

PR #2269, opened 2022-07-15
Fix limits from Config not working

### LINKED ISSUE #2268: Limits in Rocket.toml not working
**Description**

I was trying to config some limits in the Rocket.toml like:
```
[global.limits]
json = 104857600
bytes = 104857600
"PhactoryAPI.DispatchBlocks" = 104857600
```
Found it doesn't work.

It is because that rocket deserializes the Limits from a Figment with `figment::util::vec_tuple_map`:

https://github.com/SergioBenitez/Rocket/blob/f64029f079882c1fe8e47d65d4a57ec260729bea/core/lib/src/data/limits.rs#L125

And later trying to binary search the enties in the limits would fail because the deserialized limits are unsorted .

kvinwang:
Fixes #2268

### DIFF
--- a/core/lib/src/data/limits.rs
+++ b/core/lib/src/data/limits.rs
@@ -122,8 +122,8 @@ use crate::http::uncased::Uncased;
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(transparent)]
 pub struct Limits {
-    #[serde(with = "figment::util::vec_tuple_map")]
-    limits: Vec<(Uncased<'static>, ByteUnit)>
+    #[serde(with = "serde_limits")]
+    limits: Vec<(Uncased<'static>, ByteUnit)>,
 }
 
 impl Default for Limits {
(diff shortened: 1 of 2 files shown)


SergioBenitez:
Good catch, good find!

Can we instead just directly implement `Deserialize` for `Limits` so we don't run into this problem again? That is, remove the derive `Deserialize` and add an `impl Deserialize for Limits` which internally does the logic you have here, using `vec_tuple_map`. Let's also add a regression test to ensure this works as expected.

kvinwang:
> Can we instead just directly implement `Deserialize` for `Limits` so we don't run into this problem again? That is, remove the derive `Deserialize` and add an `impl Deserialize for Limits` which internally does the logic you have here, using `vec_tuple_map`. 

I think the current approach is better. Because [`impl Deserialize for Limits`](https://serde.rs/deserialize-struct.html) would be too verbose and break the symmetry between serilization and deserialization (serializing the struct via derive but deserialzing it via custom impl). The `#[serde(with = "...")]` way is much clearer.

> Let's also add a regression test to ensure this works as expected.

A regression test added.

SergioBenitez:
Agreed! Merged with a slight change in f7e703bc5e5d83af34669e8c44c3e928a30b5410. Thank you!

Issue #2270, opened 2022-07-15
Adding support for modifying fairings


the10thWiz:
I believe Rocket should provide support for modifying fairings during the ignite phase. This would enable, for example, an API for fairings to register template functions automatically. I'm working on a rocket-auth crate that provides convenient and secure authentication, and I would like the ability to register a template function that provides a pre-built login template, or the Sign in with Google button. This can't simply be a template macro, since at least the Google button requires some things like the API key in the HTML.

**Why this feature can't or shouldn't live outside of Rocket**

There are two parts to the feature described above, and only the first needs to live in Rocket's core. To provide a simple and flexible system, Rocket should likely only provide a mutable reference to previously registered fairings. The ability to add functions to the template engine should be implemented in `rocket_dyn_templates`. In this way, Fairings can define in what ways they can be modified, preventing unexpected interactions.

**Ideal Solution**

I think the ideal solution should look something like this. Since every other method on `Rocket<Build>` takes `self`, I think it's appropriate for modify_fairing to as well. This should also be quite reasonable to call from within an ignite fairing, since the return type works for it.

```rust
impl Rocket<Build> {
    pub fn modify_fairing<F: Fairing, E: Error>(self, f: Fn(&mut F) -> Result<(), E>) -> Result<Self, Self>;
}
```

There are a few issues to consider. First, Fairings are held as `Box<dyn Fairing>`, however this is actually pretty trivial since Fairings must implement Any. Second, it's possible there are multiple Fairings of a given type. In that case, I think it's best to simply run the function on each instance. If any instance fails (or there are none), modify_fairing should return Err.

**Alternatives Considered**

At the moment, something like this is kind of possible (at least for template functions), but it requires some somewhat ugly code in the main function. Ideally, the somewhat ugly code should be in the library providing the template, if it cannot be eliminated entirely.

SergioBenitez:
This was resolved by https://github.com/rwf2/Rocket/pull/2598 in https://github.com/rwf2/Rocket/commit/d3323391abb9923f07082006c49aca7a2e20b757.

PR #2271, opened 2022-07-15
Modify fairings

the10thWiz:
Add a method to enable modification of fairings during the build & on_ignite phases. Implemented for #2270.


core/lib/src/fairing/fairings.rs
```diff
@@ -42,6 +42,10 @@ impl Fairings {
         Fairings::default()
     }
 
+    pub(crate) fn iter_mut<'a>(&'a mut self) -> impl Iterator<Item = &mut Box<dyn Fairing>> + 'a {
+        self.all_fairings.iter_mut()
+    }
```
SergioBenitez, line 47:
```suggestion
    pub(crate) fn iter_mut(&mut self) -> impl Iterator<Item = &mut dyn Fairing> + '_ {
        self.all_fairings.iter_mut().map(|boxed| &mut **boxed)
    }
```

core/lib/src/rocket.rs
```diff
@@ -468,6 +469,61 @@ impl Rocket<Build> {
...
+    ///                 f.register_function("some_function", some_function)
+    ///             })
+    ///         })))
+    /// }
+    /// ```
+    ///
+    /// If multiple instances of the same type of Fairing have been attached, the modify function
+    /// will be run on each instance.
+    #[must_use]
+    pub fn modify_fairings<F: Fairing, E: fmt::Debug>(mut self, modifier: impl Fn(&mut F) -> Result<(), E>) -> Result<Self, Self> {
```
SergioBenitez, line 497: Could this be the following?

```rust
    pub fn modify_fairings<F: Fairing>(mut self, modifier: impl FnMut(&mut F)) -> Self;
```

This is much simpler to reason about and implement, though perhaps I'm missing something that we gain with the other signature.

core/lib/src/rocket.rs
```diff
@@ -468,6 +469,61 @@ impl Rocket<Build> {
...
+            Failure,
+        }
+        let mut state = State::None;
+        for f in self.fairings.iter_mut() {
+            let f = f.as_mut();
+            if f.type_id() == TypeId::of::<F>() {
+                // SAFTEY: `f` is a valid pointer since it was obtained from a Box, and it points
+                // at an `F` since it has the same TypeId. Converting from dyn type to a concrete
+                // type is handled automatically by the compilier
+                let f = unsafe { &mut *((f as *mut dyn Fairing) as *mut F) };
```
SergioBenitez, line 511: This is really scary. Minimally, we should enable `trait_upcasting` on nightly and then add something like the following:

```rust
#[cfg(nightly)]
fn downcast<T>(fairing: &mut dyn Fairing) -> Option<&mut T> {
    (fairing as &mut (dyn Any + Send + Sync).downcast_mut()
}

#[cfg(not(nightly))]
fn downcast<T>(fairing: &mut dyn Fairing) -> Option<&mut T> {
   if fairing.type_id() == TypeId::of::<T> {
       return unsafe { Some(...) }
   }

  None
}
```

At least then when the CI runs on nightly ,we ask the compiler to verify that this is sensible.

SergioBenitez (changes requested):
Okay, I'm in favor of this, barring the requested changes.

SergioBenitez:
Closing due to inactivity (apologies as I know it's largely on me), though I'm entirely open to this PR with the requested changed above.

PR #2288, opened 2022-07-24
FAQ: fix typo in production readiness question

fxsn:
My assumption is that the intention is zero-downtime, not zero-downtown, certificate rotation.


SergioBenitez:
Thanks! Merged in ef5b7f89a5a830a6f0190fca32a69bdd4a06c400.

Issue #2281, opened 2022-07-20
Rocket doesn't understand TLS keys with comments


Alibirb:
**Description**

Rocket rejected my PKCS 8/PEM formatted RSA TLS key file, claiming "bad TLS private key: invalid key header; supported formats are: RSA, PKCS8". After removing the leading and trailing comment lines (beginning with "#"), Rocket accepted it, but I don't think we'd be allowed to remove the comments as they're used to mark the confidentiality of the files.

According to [RFC 7468](https://datatracker.ietf.org/doc/html/rfc7468#section-2), "Data before the encapsulation boundaries are permitted, and parsers MUST NOT malfunction when processing such data."

**To Reproduce**

Add a new line with `# test comment` at the beginning of `examples/tls/private/rsa_sha256_key.pem`, and then run the tls example.

**Expected Behavior**

Should recognize the file as containing an RSA private key.

**Environment:**

 - OS Distribution and Kernel: CentOS container
 - Rocket Version: 0.5.0-rc.2

**Additional Context**

I'm not sure if this issue is in Rocket itself, or an underlying TLS library. The stacktrace just pointed to `error.rs` and a bunch of stuff in `std`, so I'm not sure where Rocket actually reads the key file.

Rocket seems to have no issue with the leading comments in the PEM-formated certificate file; just the key file.

julianbuettner:
Hi there!  
Thanks for reporting. The stack trace was not very helpful but I found the Übeltäter (person making trouble, german).  

[core/http/src/tls/util:L19](https://github.com/SergioBenitez/Rocket/blob/master/core/http/src/tls/util.rs#L19)  
It literally just reads the first line and excpect it to be the `--- BEGIN ---` line.  
I am considering creating a pull request as it turned out to be a rather simple issue.

PR #2315, opened 2022-08-09
Fix 2281 - Ignore comment lines in TLS private keys

julianbuettner:
Resolves issue [2281](https://github.com/SergioBenitez/Rocket/issues/2281).  

Lines before the first pre-encapsulation boundary are ignored as specified in [RFC7468](https://datatracker.ietf.org/doc/html/rfc7468#section-2).  
Lines afterwards are ignores by the then chosen parsing function.  

Behaves the same otherwise.


SergioBenitez:
Rewritten but merged in daa157f872343552e7990a94ae0a71f4fc71aab9. Thank you!

Issue #2337, opened 2022-09-10
RUSTSEC-2020-0016 net2 is oudated; transitive dep of notify<-mio<-rocket_dyn_templates


jblachly:
See #1440 

Existing objections in that thread have been addressed (namely, notify 5.0.0 was released as stable 11 days ago, and has been updated to use mio > 0.6, thus net2 is no longer a transitive dependency)

https://crates.io/crates/notify/5.0.0/dependencies
https://crates.io/crates/mio/0.8.4/dependencies

**SUGGESTED FIX**
Update rocket_dyn_templates to use a notify version in the 5.0 series.

SergioBenitez:
We should absolutely do this! Happy to accept a PR updating `notify`.

PR #2343, opened 2022-09-17
Update the notify dependency to 5.0.0.

jacobsimpson:
Addresses https://github.com/SergioBenitez/Rocket/issues/2337.

According to the notify upgrading advice, the `raw_watcher` has been removed in favor of the raw behavior being the default. And `RawEvent` has been updated to `Event`.

Where the example in 4.0.16 used a [raw watcher](https://github.com/notify-rs/notify/blob/v4.0.16/examples/monitor_raw.rs#L13), the example in 5.0.0 is updated to just use a [recommended watcher](https://github.com/notify-rs/notify/blob/notify-5.0.0/examples/monitor_raw.rs#L19).

### DIFF
--- a/contrib/dyn_templates/Cargo.toml
+++ b/contrib/dyn_templates/Cargo.toml
@@ -18,7 +18,7 @@ handlebars = ["handlebars_"]
 
 [dependencies]
 glob = "0.3"
-notify = "4.0.6"
+notify = "5.0.0"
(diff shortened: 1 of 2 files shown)


SergioBenitez:
Wonderful! Excellent work. Thank you for this! Merged in 6c3d35e7e5bceaf11c329623d48f81c276407030.

PR #2360, opened 2022-10-06
Implement basic request id

the10thWiz:
This provides a resolution to #2352.

# Motivation

For easier debugging and tracing with Rocket, @thomasmost wanted to associate a unique id with every request to associate logs with a specific request. The primary goal isn't just to provide an id, but make it possible to access without explicitly passing it along with the request. In Rocket 0.4, this would have been possible with thread-local values, but for Rocket 0.5, `tokio::task_local` is required. Tokio's task local API requires the ability to scope the task local value, something `Fairing`s cannot do. As a solution, this PR integrates the necessary functionality into Rocket itself.

Integrating this into Rocket means Rocket can take advantage of the id, and include it in the logs generated by default, making tracking much easier.

# Design

The proposed API is a `RequestId` struct, which uniquely identifies a request. This is generated when Rocket receives the request, and is placed into task local storage for the handler. `RequestId` is opaque, and implements `Display` & `Debug`. There is an additional method on `Request` to retrieve the request's id, along with a free function `rocket::request::current_request` that pulls the value from task local storage. It can also be used as a request guard.

The logging macros provided by Rocket automatically retrieve the request id from local storage and prefix the log message with it.

Internally, an id is a `u64`, and ids are assigned sequentially. Right now, the id only implements formatting and copy. Implementing `Eq` and `Hash` would be trivial, but I don't believe this is necessary for this feature. This can trivially be expanded in the future if it turns out to be useful. The current implementation also means `Ord` could be used to indicate whether a request came before or after another, but I don't think this is a good idea.

### DIFF
--- a/core/lib/src/local/asynchronous/client.rs
+++ b/core/lib/src/local/asynchronous/client.rs
@@ -2,6 +2,7 @@ use std::fmt;
 
 use parking_lot::RwLock;
 
+use crate::request::RequestId;
 use crate::{Rocket, Phase, Orbit, Ignite, Error};
 use crate::local::asynchronous::{LocalRequest, LocalResponse};
 use crate::http::{Method, uri::Origin, private::cookie};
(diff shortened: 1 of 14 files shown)


SergioBenitez:
So, I completely agree with the motivation for this, but I feel that the implementation here is far too specific to one thing to be worthwhile. Specifically, the functionality proposed here would be enabled by an implementation of #21. I'm going to close this in favor of #21, which I will prioritize implementing.

Issue #2148, opened 2022-04-19
Simpler alternative to TempFile in multipart form file upload (Like using Vec<u8> directly)


Alvenix:
For file upload, I have been using TempFile. However, I don't feel that it is very ergonomic. You have to store the file first to read it and later delete it.

Here is some pseudo code which showcase how sometimes I had to use it.
```
#[post("/upload", data = "<file>")]
async fn process_file(mut file: TempFile<'_>) -> std::io::Result<()> {
    let uuid = Uuid::new_v4();
    let random_file_name = format!("/tmp/{uuid}");
    file.persist_to(&random_file_name).await?;
  
    let content = fs::read(&random_file_name)?;

    fs::delete(&random_file_name)?;

    Ok(())
}
```
I did not find existing alternative in Rocket.

SergioBenitez:
Well, what are you trying to do with the data? A `TempFile` is to be used when you're trying to save the data to disk, or when the amount of data might exceed what you can store in memory. Given that you read the entire contents of the file into memory, your code is either wrong, you're accepting the crash due to memory exhaustion, or you're properly limiting file size.

If you're doing the latter, then indeed this is inefficient. Unfortunately Rocket doesn't and can't have `Vec<u8>` mean "a byte buffer" because `Vec<u8>` already means "a sequence of 8-bit unsigned integers," as far as forms go. We could add a `Bytes` or `Buffer` type or similar with the intended meaning, however.

If you're doing some sort of cheap computation with the data, then you probably don't want to buffer it all beyond, say, 4k streaming chunks, and store only the computed value. This is relatively easy to implement in Rocket via the `FromFormField` trait and gives you complete control over data limiting, buffering, and streaming.

Alvenix:
> Well, what are you trying to do with the data? A `TempFile` is to be used when you're trying to save the data to disk, or when the amount of data might exceed what you can store in memory. Given that you read the entire contents of the file into memory, your code is either wrong, you're accepting the crash due to memory exhaustion, or you're properly limiting file size.
> 
> If you're doing the latter, then indeed this is inefficient. Unfortunately Rocket doesn't and can't have `Vec<u8>` mean "a byte buffer" because `Vec<u8>` already means "a sequence of 8-bit unsigned integers," as far as forms go. We could add a `Bytes` or `Buffer` type or similar with the intended meaning, however.
> 
[...]

PR #2361, opened 2022-10-08
Support of &[u8] for form fields

badoken:
Will allow to define forms with &[u8] fields, like
```rust
#[derive(FromForm)]
struct MyForm<'r> {
    foo: &'r [u8],
}
```

Should address the issue https://github.com/SergioBenitez/Rocket/issues/2148


SergioBenitez:
Merged! Thank you!

PR #2362, opened 2022-10-08
Add Bytes DataGuard

chinoto:
This will hopefully be sufficient for most of those concerned with issue #2148.
I might have gone the route of #2361 (implementing for `&[u8]`), but some people may want to get the filename and content-type as well, though I think that PR is still nice to have if all you need is the data.

### DIFF
--- a/core/lib/src/data/bytes.rs
+++ b/core/lib/src/data/bytes.rs
@@ -1,0 +1,88 @@
+use std::borrow::Cow;
+use std::io;
+
+use crate::data::{Capped, FromData, Limits, Outcome, N};
+use crate::form::{DataField, Errors, FromFormField, ValueField};
+use crate::http::{ContentType, Status};
+use crate::{fs::FileName, outcome::IntoOutcome, Data, Request};
+
+#[derive(Debug)]
+pub struct Bytes<'v> {
+    pub file_name: Option<&'v FileName>,
+    pub content_type: Option<ContentType>,
+    pub content: Cow<'v, [u8]>,
+}
+
+impl<'v> Bytes<'v> {
+    async fn from<'a>(
+        req: &Request<'_>,
+        data: Data<'_>,
+        file_name: Option<&'a FileName>,
+        content_type: Option<ContentType>,
+    ) -> io::Result<Capped<Bytes<'a>>> {
+        let limit = { content_type.as_ref() }
+            .and_then(|ct| ct.extension()?.limits().find(&["file", ext.as_str()]))
+            .or_else(|| req.limits().get("file"))
+            .unwrap_or(Limits::FILE);
+
+        let Capped { value, n } = data.open(limit).into_bytes().await?;
+
+        Ok(Capped::new(
+            Bytes {
+                content_type,
+                file_name,
+                content: value.into(),
+            },
+            n,
+        ))
+    }
+}
+
+#[crate::async_trait]
+impl<'v> FromFormField<'v> for Capped<Bytes<'v>> {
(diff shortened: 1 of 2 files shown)


core/lib/src/data/bytes.rs
```diff
@@ -0,0 +1,88 @@
...
+
+impl<'v> Bytes<'v> {
+    async fn from<'a>(
+        req: &Request<'_>,
+        data: Data<'_>,
+        file_name: Option<&'a FileName>,
+        content_type: Option<ContentType>,
+    ) -> io::Result<Capped<Bytes<'a>>> {
+        let limit = { content_type.as_ref() }
+            .and_then(|ct| ct.extension()?.limits().find(&["file", ext.as_str()]))
```
odiseo0, line 24: I think there's a mistake here, shouldn't be
```rs
            .and_then(|ct| ct.extension())
            .and_then(|ext| req.limits().find(&["file", ext.as_str()]))
```
?

chinoto: Sorry for the delay, been busy working.
There is no need for a second `.and_then(...)` because the `?` after `.extension()` extracts the value from `Option::Some(T)` or returns `Option::None` depending on which variant it happened to be, allowing for more concise code.
Apparently it was [stabilized back in 2016](https://github.com/rust-lang/rust/pull/36995), feels newer. There's a [section in the book about it](https://doc.rust-lang.org/book/ch09-02-recoverable-errors-with-result.html#a-shortcut-for-propagating-errors-the--operator).

core/lib/src/data/bytes.rs
```diff
@@ -0,0 +1,88 @@
...
+use crate::data::{Capped, FromData, Limits, Outcome, N};
+use crate::form::{DataField, Errors, FromFormField, ValueField};
+use crate::http::{ContentType, Status};
+use crate::{fs::FileName, outcome::IntoOutcome, Data, Request};
+
+#[derive(Debug)]
+pub struct Bytes<'v> {
+    pub file_name: Option<&'v FileName>,
+    pub content_type: Option<ContentType>,
+    pub content: Cow<'v, [u8]>,
```
GorgWorgington, line 13: Just wondering, is there a reason for using `Cow<[u8]>` here instead of `Vec<u8>`? As far as I can tell converting a `Cow<[T]>` back into a `Vec<T>` would cause a reallocation of the entire file even if it's owned and I would rather have the `Vec` for my use-case.

chinoto: I'm finally done with having to work so much, so hopefully I'll be more responsive now.
On line 52, I use it for the borrowed variant, so there is a purpose in using it and there should not be an unnecessary (re)allocation because:
> [Calling into_owned on a Cow::Owned returns the owned data. The data is moved out of the Cow without being cloned.](https://doc.rust-lang.org/std/borrow/enum.Cow.html#method.into_owned)

For extra clarification, the type of `content` doesn't mention `Vec` because it is inferred from the [ToOwned implementation for slices](https://doc.rust-lang.org/src/alloc/slice.rs.html#782) that specifies its `Owned` variant is `Vec`.

SergioBenitez (changes requested):
This implementation should just be delegating to the existing `&[u8]` and `Vec<u8>` `FromData` implementations. See 569bc09a1dea1068cffaa6751cba9b3ad9392eed for how this was done for the `&[u8]` `FromForm` implementation.

SergioBenitez:
Closing due to inactivity (apologies as I know it's largely on me), though I'm entirely open to this PR with the requested changed above.

PR #2365, opened 2022-10-13
Showcasing the ability to wrap guards in Option

Arteneko:
The ability to wrap guards in Option is useful for routes that may behave differently whether the guard is set or not, but where making two separate handlers isn't justified (for example, having this info passed to the template rendering, without any impact on the handler logic). I didn't see it documented, but I tested it as working as expected on my daemon, so I'm adding a bit of documentation on the topic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,11 @@
-# Version 0.5.0-rc.2 (May 9, 2022)
+# Version 0.5.0-rc.2 (May 09, 2022)
 
 ## Major Features and Improvements
 
   * Introduced [`rocket_db_pools`] for asynchronous database pooling.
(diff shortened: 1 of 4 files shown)


SergioBenitez:
Yeah! Thank you for this. This has totally been missing from the guide. I took the opportunity to expand on your contribution and merged it in 9584edcf5987714cb156920652507c34befccaa3.

I'm noticing that GitHub is not attributing the contribution to you. I took your name and email from the patch in this PR. Is that email address associated with your GitHub handle? I'd love for your contributions to be properly recognized!

SergioBenitez:
Ah, I see, you merged a different PR into yours, and that's what GitHub created the first patch from. Sigh. My apologies, totally misattributed this PR. In the future, please don't merge other PRs into yours - it's best to keep things separate.

Issue #2391, opened 2022-11-10
Remove sqlx features due to a bug in rustls/webpki failing to resolve IP addresses 


aadityadhruv:
**Description**

 Currently `sqlx` in `db_pools` uses `runtime-tokio-rustls` as its feature. This has a [bug](https://github.com/launchbadge/sqlx/issues/1095) which results in a `InvalidDNSNameError` for raw IP addresses as a part of the url. This is being [fixed](https://github.com/rustls/webpki/pull/5), however, it would be nice to let the user choose their TLS back-end. 

**To Reproduce**

Setting the database url to a raw IP address as follows: 
```
mysql://foo:bar@<IP>:3306/baz
```

**Expected Behavior**

The IP shouldn't need any DNS resolution. 

**Environment:**

 - OS Distribution and Kernel: [Fedora 36, Kernel 6.0.5]
 - Rocket Version: [0.5.0-rc.2]

**Additional Context**

A simple fix is to remove the feature from sqlx, and let the user choose the backend.

aadityadhruv:
Here is a simple fix to resolve this issue https://github.com/SergioBenitez/Rocket/pull/2392

tn-dev-fl:
i got this error **error occurred while attempting to establish a TLS connection: InvalidDNSNameError .** it worked fine on my main machine . i tried to remove the features sqlx runtime i get another error 
compile_error!( "one of the features ['runtime-actix-native-tls', 'runtime-async-std-native-tls', \
      'runtime-tokio-native-tls', 'runtime-actix-rustls', 'runtime-async-std-rustls', \
      'runtime-tokio-rustls'] must be enabled"
 );
   ^

aadityadhruv:
You need to specify a runtime, you can use the `tokio-native-tls` one since `rustls` cannot resolve direct IP addresses yet.

SergioBenitez:
I'm considering this an upstream isusue: https://github.com/SergioBenitez/Rocket/pull/2392#issuecomment-1487728679.

aadityadhruv:
Yeah, I read that comment, I think that may be a suitable idea.

SergioBenitez:
Looks like the new rustls release fixes this! Shall we update dependencies and see if that's so?

SergioBenitez:
[...]

PR #2392, opened 2022-11-10
Fixes Issue #2391 - Remove sqlx features from db_pools

aadityadhruv:
By removing sqlx features from the crate itself, the choice is put upon the user, which lets them choose between `rustls` and `native-tls`.

### DIFF
--- a/contrib/db_pools/lib/Cargo.toml
+++ b/contrib/db_pools/lib/Cargo.toml
@@ -61,7 +61,7 @@ optional = true
 [dependencies.sqlx]
 version = "0.6"
 default-features = false
-features = ["runtime-tokio-rustls"]
+features = ["runtime-async-std-native-tls"]
 optional = true
 
 [dev-dependencies.rocket]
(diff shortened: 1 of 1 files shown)


SergioBenitez:
Remove the feature means that `sqlx` doesn't compile. _Not_ removing the feature means we have this issue. Adding a separate feature to enable `native_tls` instead means we can't compile the crate with `--all-features`. I don't see a solution here that is good. The problems, in my opinion, are in `sqlx` (for how they implemented features, which they've fixed in 0.7) and `rustls/webpki`, for the initial source of the problem with resolving IP addresses.

I am unfortunately inclined to suggest that we wait for a fix upstream. Unless there's a solution I'm missing that:
  1. Doesn't require everyone to depend on `sqlx` for use `db_pools` with `sqlx`.
  2. Allowing use to build `contrib` with `--all-features`.
  3. Allows selecting either `rustls` or `native-tls`

aadityadhruv:
That sounds good to me!

PR #2393, opened 2022-11-11
add two form validation test cases that should pass but don't

ChristofferEnne:
As subject says


SergioBenitez:
This was merged earlier. Thank you!

PR #2400, opened 2022-11-16
Fix typo in doc

kiritoxkiriko:
sream -> stream


SergioBenitez:

PR #2411, opened 2022-12-01
Add Diesel support

pooroligarch:
This PR adds support for the `postgres`, `mysql` and `sqlite` backends of Diesel ORM using the `r2d2` connection pool. There's no documentation yet but it works just fine in a real project without panics or perf problems. 

Questions:
* Is there a cleaner way to group together common dependencies without the `_diesel` feature?
* `r2d2::Pool` has no "close" or "destroy" method. Is the empty function `async fn close(&self) {}` the right way to implement this?


SergioBenitez:
Diesel is already supported by `rocket_sync_db_pools`. `rocket_db_pools` is for `async` database drivers, and `diesel` is _not_ `async`, so this implementation would cause all sorts of issues.

SergioBenitez:
Closing for the above reasons.

PR #2440, opened 2023-01-01
Use path instead of database connection string

### LINKED ISSUE #2439: `url` is not passed to rusqlite anymore
The SQLite database settings are passed to rusqlite via 
```toml
[default.databases.rusqlite]
url = "file:rusqlite?mode=memory&cache=shared"
```
in `Rocket.toml` according to the examples.

However, it looks like the `url` is not used by rusqlite anymore since https://github.com/rusqlite/rusqlite/pull/1248. Specifically, according to the tests in that PR, `Connection::open("")` now has the same effect as `Connection::open("file:dummy.db?mode=memory&cache=shared)`. In other words, the `url` is ignored:

```rust
    #[test]
    fn test_path() -> Result<()> {
        let tmp = tempfile::tempdir().unwrap();
        let db = Connection::open("")?;
        assert_eq!(Some(""), db.path());
        let db = Connection::open_in_memory()?;
        assert_eq!(Some(""), db.path());
        let db = Connection::open("file:dummy.db?mode=memory&cache=shared")?;
        assert_eq!(Some(""), db.path());
        let path = tmp.path().join("file.db");
        let db = Connection::open(path)?;
        assert!(db.path().map(|p| p.ends_with("file.db")).unwrap_or(false));
        Ok(())
    }
```

rikhuijzer:
Closes #2439.

### DIFF
--- a/examples/databases/Rocket.toml
+++ b/examples/databases/Rocket.toml
@@ -1,5 +1,5 @@
 [default.databases.rusqlite]
-url = "file:rusqlite?mode=memory&cache=shared"
+url = "rusqlite"
 
 [default.databases.sqlx]
 url = "db/sqlx/db.sqlite"


rikhuijzer:
This doesn't work and gives errors when running in parallel.

SergioBenitez:
Can you point to documentation (beyond the unit test) that shows that what we had before and what you have now are identical?

rikhuijzer:
> Can you point to documentation (beyond the unit test) that shows that what we had before and what you have now are identical?

I haven't been able to find documentation for this.

By running the example, I can verify that `file:rusqlite?mode=memory&cache=shared` does not create a file `rusqlite` in the current directory:

```shell
$ git clone https://github.com/SergioBenitez/Rocket.git

$ cd Rocket/examples/databases

$ cargo test

$ ls 
Cargo.toml  db  README.md  Rocket.toml   sqlx-data.json  src
```
and after changing `url` in `Rocket.toml` to `rusqlite`, it becomes:
```shell
$ cargo test

$ ls
Cargo.toml  db  README.md  Rocket.toml  rusqlite  sqlx-data.json  src
```
But I also noticed that `rusqlite` is not in the `.gitignore`, so it was never the intention to store the file in the local dir?

In any way, I think that specifying the path is better. I'm running Rocket + rusqlite in production for weeks now without those flags (`?mode=memory&cache=shared`). The concurrent tests and concurrent load in production all work flawlessly.

SergioBenitez:
That's right - want to use to an in-memory database for this example. See my comment in your issue for an explanation of what's going on in that rusqlite PR.

PR #2460, opened 2023-01-30
Fix Result namespacing in FromFormField derive

jhinch:
Previously, if a module used or defined a type alias for Result, FromFormField derives would fail to compile as it would use the type alias instead of the fully qualified type

### DIFF
--- a/core/codegen/src/derive/from_form_field.rs
+++ b/core/codegen/src/derive/from_form_field.rs
@@ -47,7 +47,7 @@ pub fn derive_from_form_field(input: proc_macro::TokenStream) -> TokenStream {
             .with_output(|_, output| quote! {
                 fn from_value(
                     __f: #_form::ValueField<'__v>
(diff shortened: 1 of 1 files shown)


jhinch:
The builds are failing but I get the same failures on master

SergioBenitez:
Indeed. Without even looking at the failures, I can almost guarantee that these are UI compiletest failures due to new output in new compilers. This has been a rather unfortunate issue for quite some time.

I'll look into removing these UI tests from the list of CI that must pass.

SergioBenitez:
Okay, I've gone ahead and tweaked the testing infrastructure to avoid this issue now and in the future. If you rebase, we can run a real series of tests.

jhinch:
PR has been rebased

SergioBenitez:
Awesome. Merged in 910994f982aad7fccc6f18944aeb2f84939e851f.

PR #2467, opened 2023-02-15
chore: replace atty with is-terminal

Unpublished:
Inspired by https://github.com/clap-rs/clap/pull/4249


SergioBenitez:
Merged into the RC branch. Thank you so much for this!

PR #2474, opened 2023-02-21
`impl Fairing for Arc<dyn Fairing>`

dnbln:


SergioBenitez:
Merged into the RC branch. Thank you!

PR #2499, opened 2023-03-23
Add support for the anyhow Error as a Responder

Munksgaard:
This PR adds feature-gated support for the Error struct from anyhow as a Responder. I'm not even sure if you're interested in it, but I like using anyhow, and I think a lot of others do to. Feel free to close this PR if you don't want it in Rocket.

Fixes https://github.com/dtolnay/anyhow/issues/105

### DIFF
--- a/core/lib/Cargo.toml
+++ b/core/lib/Cargo.toml
@@ -30,6 +30,9 @@ msgpack = ["rmp-serde", "tokio/io-util"]
 uuid = ["uuid_", "rocket_http/uuid"]
 
 [dependencies]
+# error dependencies
+anyhow = { version = "1.0", optional = true }
+
 # Serialization dependencies.
 serde_json = { version = "1.0.26", optional = true }
 rmp-serde = { version = "1", optional = true }
--- a/core/lib/src/anyhow.rs
+++ b/core/lib/src/anyhow.rs
@@ -1,0 +1,17 @@
+//! Support for using [`Error`](crate::anyhow::Error) from [`mod@anyhow`]
+//! as a [`Responder`].
+
+#[doc(inline)]
+pub use anyhow::*;
+
+use crate::http::Status;
+use crate::request::Request;
+use crate::response::{self, Responder};
+
+/// Prints a warning with the error and forwards to the `500` error catcher.
+impl<'r> Responder<'r, 'static> for anyhow::Error {
+    fn respond_to(self, _: &'r Request<'_>) -> response::Result<'static> {
+        warn_!("Error: {:#}", yansi::Paint::default(self));
+        Err(Status::InternalServerError)
+    }
+}
--- a/core/lib/src/lib.rs
+++ b/core/lib/src/lib.rs
@@ -68,6 +68,7 @@
 //! | `json`    | Support for [JSON (de)serialization].                   |
 //! | `msgpack` | Support for [MessagePack (de)serialization].            |
 //! | `uuid`    | Support for [UUID value parsing and (de)serialization]. |
+//! | `anyhow`  | Support for the [anyhow] Error type as a Responder.     |
 //!
 //! Disabled features can be selectively enabled in `Cargo.toml`:
 //!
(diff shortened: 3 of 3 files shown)


Munksgaard:
I'm unsure how best to print the errors from anyhow. Following the example from the rest of the rocket logging, I'm now using `{:#}`, which prints the entire backtrace on a single line. Example:

```
GET /foo text/html:
   >> Matched: (index) GET /foo/
   >> Error: get base data: get accounts: request: error decoding response body: missing field `collection` at line 1 column 393: missing field `collection` at line 1 column 393
   >> Outcome: Failure
   >> No 500 catcher registered. Using Rocket default.
   >> Response succeeded.
```

Alternatively, we can use `{:?}` to print the trace over multiple lines, like this:

```
GET /foo text/html:
   >> Matched: (index) GET /foo/
   >> Error: get base data

Caused by:
    0: deserializing json
    1: error decoding response body: missing field `agreementNumber` at line 1 column 375
    2: missing field `agreementNumber` at line 1 column 375
   >> Outcome: Failure
   >> No 500 catcher registered. Using Rocket default.
   >> Response succeeded.
```

I think the former is preferable for e.g. grepping?

More information: https://docs.rs/anyhow/latest/anyhow/struct.Error.html#display-representations

SergioBenitez:
I really appreciate the motivation behind this PR. The alternative is a wrapper type around `anyhow::Error` which is 1) annoying to implement, and 2) occasionally annoying to use. That being said, I think accepting this PR would constitute establishing precedence for adding support to Rocket for external types via feature flags, which isn't something I think we should do. This is largely a maintenance issue in the sense that we now need to track the external library: if it bumps a major version, we need to bump (potentially major) versions, or else users of Rocket are restricted to the older release. This is too big of a commitment, which is why we don't generally do this in Rocket. Even database support is split into an external library via `rocket_db_pools` and `rocket_sync_db_pools`, for the very same reason.

I've long though about an officially sanctioned but community driven list of crates that are advertised on Rocket's website. I think a crate that externally implements a wrapper type around `anyhow::Error`, and makes it as convenient to use as possible, would be a perfect fit. That list need not exist, of course, to write such a crate, and I encourage you to do that as well.

I hope this reasoning makes sense! Thank you for the PR!

Munksgaard:
Thank you for your detailed explanation.

I completely understand the desire to be unencumbered by external dependencies to the highest degree possible. I'd argue that `anyhow` is a relatively simple crate that doesn't have many external dependencies, but it still has some.

For anyone else coming this way, it _is_ possible to use my branch to patch Rocket with `anyhow` support with the following addition to Cargo.toml, but I **do not** recommend doing so:

```
[patch.crates-io]
rocket = { git = "https://github.com/Munksgaard/Rocket", branch = "anyhow" }
```

I will not be maintaining that branch, and it may go away without any notice.

With regards to a separate crate, I think you may be right, that is the best way forward. I see that [`rocket_anyhow`](https://crates.io/crates/rocket_anyhow/0.1.1) ([repo](https://github.com/thanadolps/rocket_anyhow)) exists, but it is out of date, and I haven't taken a serious look at figuring out how hard it would be to update it. Perhaps I will do so at some point.

Thanks again for considering my PR, and for your work on Rocket in general.

Munksgaard:
Just for completeness, another alternative would be to add something like the following to Rocket:

```rust
impl<'r, T: std::error::Error> Responder<'r, 'static> for T {
    fn respond_to(self, _: &'r Request<'_>) -> response::Result<'static> {
        warn_!("Error: {:?}", yansi::Paint::default(self));
        Err(Status::InternalServerError)
    }
}
```

Unfortunately, that implementation conflicts with other implementations for Responder, like `&str` and `String` and many others.

Munksgaard:
After some initial experimentation, it seems like you'd really need to reimplement all the functionality of anyhow in a new crate in order to get the full benefit. Otherwise, handy helpers such as the [`anyhow!` macro](https://docs.rs/anyhow/latest/anyhow/macro.anyhow.html) will not work.

Issue #2425, opened 2022-12-13
Make private cookies cookie marked as secure by default


arjentz:
**Existing Functionality**

Rocket is a framework using "secure defaults". However, cookies are not marked as secure by default.

The functionality to mark a cookie as secure is there, but when libraries set cookies for you it is nearly impossible to mark it as secure. For example https://github.com/jebrosen/rocket_oauth2/blob/0.5.0-rc.1/src/lib.rs#L693 is not marked as secure and cannot be done in any trivial way.

**Suggested Changes**

Mark private cookies as secure by default. I.e. add 3 lines of code to: https://github.com/SergioBenitez/Rocket/blob/v0.5.0-rc.2/core/lib/src/cookies.rs#L497

**Alternatives Considered**

This could probably be done using some proxy like nginx, but it would be very hacky. When not using a proxy it is hard to do when using existing libraries like Rocket OAUTH2.

**Additional Context**

Options to consider:
- only do so when a secure connection is used
- only make it the default for production builds

arjentz:
So apparently, browsers do not allow to set secure cookies over an insecure connection (makes sense), so it would indeed make the code setting the cookies useless for insecure connections. I am not sure if it is possible to detect if a connection runs over TLS in a reliable manner (e.g. if a reverse proxy is used), but if there is I'd like to hear about it.

So the only viable option I see now is to use a configuration option that toggles compatibility with insecure connections, but I'm not sure what your vision is on these kind of options.

arjentz:
So, I've looked into some other frameworks, and Django has the following options regarding TLS: https://docs.djangoproject.com/en/4.1/topics/security/#ssl-https

In particular, the following setting looks like a proper solution on how to implement this in a workable way: https://docs.djangoproject.com/en/4.1/ref/settings/#std-setting-SECURE_PROXY_SSL_HEADER

arjentz:
Apparently, using this header is also the way Kubernetes ingress and mainstream load balancers do this: https://cloud.google.com/load-balancing/docs/https#target-proxies

SergioBenitez:
I'd be in favor of enabling this conservatively. That is, as long as we're as certain as we can be that the connection is being served over a secure tunnel. It seems this means that either:
  1) Rocket was launched with TLS enabled and is running.
  2) A (default to `None`) configurable header is set that indicates the connection is being proxied over a secure tunnel.

arjentz:
[...]

PR #2519, opened 2023-04-12
Forwarded proto

arjentz:
PR for #2425. Some notes:
* `Protocol` is based largely on `ProtocolName` from https://github.com/hyperium/hyper-old-types/blob/0426d7ac64c955af6327279c595168c39e7b1415/src/header/common/upgrade.rs#L84
* Renamed `ip_header::deserialize` to `http_header::deserialize` to be able to re-use it
* Used https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto as a reference guide
* Everything else is mostly based on the way `x-real-ip` header is handled: https://github.com/SergioBenitez/Rocket/commit/9377af5978c2505911769bb4f873b2a800e8e35a
* `CookieJar` has an extra field `proxy_secure` that is set internally when it makes sense to do so
* `Request` has a `proxy_proto` and `proxy_secure` method used internally
* `FromRequest` is implemented for `Protocol`
* Configuration of `proto_header` defaults to `None`/`false`


SergioBenitez:
Thanks for the work on this. It's top on my list for early next week. I appreciate the patience.

core/lib/src/config/config.rs
```diff
@@ -89,8 +89,19 @@ pub struct Config {
     /// that the value must syntactically be a valid HTTP header name.
     ///
     /// **(default: `"X-Real-IP"`)**
-    #[serde(deserialize_with = "crate::config::ip_header::deserialize")]
+    #[serde(deserialize_with = "crate::config::http_header::deserialize")]
     pub ip_header: Option<Uncased<'static>>,
+    /// The name of a header, whose value is typically set by an intermediary
+    /// server or proxy, which contains the protocol used by the connecting
+    /// client. Used by [`Request::proxy_proto()`] and
+    /// [`Request::proxy_secure()`].
```
SergioBenitez, line 97: We need some more guidance here. It's _really_ important that this header is trusted. Otherwise, we might assume we're in a secure context when we're not. For this PR (cookies being marked as secure), we just create an application that doesn't. But that might change in the future.

We could also say that this should probably be `X-Forwarded-Proto` and point to docs for it.

core/lib/src/cookies.rs
```diff
@@ -163,6 +164,7 @@ impl<'a> Clone for CookieJar<'a> {
             jar: self.jar.clone(),
             ops: Mutex::new(self.ops.lock().clone()),
             config: self.config,
+            proxy_secure: self.proxy_secure,
```
SergioBenitez, line 167: Why do we need this?

core/lib/src/request/from_request.rs
```diff
@@ -169,6 +169,12 @@ impl<S, E> IntoOutcome<S, (Status, E), ()> for Result<S, E> {
 ///     via [`Request::client_ip()`]. If the client's IP address is not known,
 ///     the request is forwarded.
 ///
+///   * **Protocol**
+///
+///     Extracts the protocol of the incoming request as a [`Protocol`] via
+///     [`Request::proxy_proto()`] (HTTP or HTTPS). If the used protocol is not
+///     known, the request is forwarded.
+///
```
SergioBenitez, line 177: This seems problematic. I would expect such a guard to succeed 99.9% of the time, given the name of the type. But this will only work when the header is set. Is there a benefit to exposing this? If we really want to do this, we should change the name to something less general, like `ForwardedProtocol`, or something.

SergioBenitez (changes requested):

SergioBenitez:
My initial inclination is that this is doing far too much. I would have expected the implementation to:

  * Add a new config parameter that configures a header to check.
  * Add a method somewhere that allows checking whether we're in a secure context that effectively returns `tls_enabled() || have_secure_header()` .
  * Call that new method from the relevant `default()` methods in `Cookie`.

I'd like to change this implementation such that we get there. In particular, I really dislike `set_proxy_secure`: it adds states to a `Cookie` where that was none before. Instead the cookie just needs to know whether it should set the `Secure` flag or not, and we should provide a way to do that.

I'm not convinced by the name `Protocol`. I think we should use something more specific to the context. Perhaps `ForwardedProto`, or `ProxyProtocol`, or something like that.

core/lib/src/request/from_request.rs
arjentz: While working on this I thought I would need this, but I realised that in the end we didn't but it might be useful to someone else. As it's trivial to implement, I will just remove it.

core/lib/src/cookies.rs
arjentz: I wholeheartedly agree that this isn't the nicest solution, but when working on it I couldn't come up with a better one. The problem I wanted to solve is that only the `Request` knows whether we are in a secure context (as forwarded by the proxy) but the `CookieJar` itself has no way to determine this.

I would love to remove this, and if you could provide some guidance in a better way to determine whether we are in a secure context (when setting the defaults) I will make changes to do so. Or feel free to do them yourself

core/lib/src/config/config.rs
arjentz: I updated to docs to add some more guidance and a link to the docs, hope its better now.

core/lib/src/request/from_request.rs
arjentz: Decided to keep it, as the tests I added are using it, but I renamed it to `ForwardedProtocol`.

arjentz:
I renamed everything to `ForwardedProtocol` and `forwarded_secure` (except the part in `CookieJar` as I assumed we will remove it). As said, I could use some guidance on implementing a better way to determine whether we are in a secure context when setting the defaults for a new cookie.

core/http/src/header/forwarded_protocol.rs
```diff
@@ -0,0 +1,38 @@
...
+/// A protocol used to identify a specific protocol forwarded by an HTTP proxy.
+// Names are case-insensitive
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum ForwardedProtocol {
+    /// `http` value, Hypertext Transfer Protocol
+    Http,
+    /// `https` value, Hypertext Transfer Protocol Secure
+    Https,
+    /// Any other protocol name not known to us
+    Unregistered(String),
```
SergioBenitez, line 15: Let's call this `Unknown`. Also, should `String` be `Uncased`?

core/http/src/header/forwarded_protocol.rs
```diff
@@ -0,0 +1,38 @@
+use std::{
+    fmt::{self, Display},
+    str::FromStr,
+};
```
SergioBenitez, line 4:
```suggestion
use std::fmt;
```

Then use `fmt::Display` and `std::str::FromStr` later.

core/lib/src/request/request.rs
```diff
@@ -386,6 +386,67 @@ impl<'r> Request<'r> {
...
+            .and_then(|proto| {
+                proto.parse()
+                    .ok()
+                    .map(|p| {
+                        if let ForwardedProtocol::Unregistered(_) = p {
+                            warn_!("'{proto_header}' header is malformed: {proto}");
+                        }
+                        p
+                    })
+            })
```
SergioBenitez, line 423:
```suggestion
            .and_then(|proto| proto.parse().ok())
```

I don't think we need this warning, and we certainly shouldn't emit it every time we call this.

SergioBenitez (changes requested):
After some consideration, I feel like we should really call it `ProxyProtocol` or `ProxyProto`, as it is quite literally the protocol that the proxy is using to communicate. Similarly, the config parameter should be called `proxy_proto_header` or similar, and the rest of the code should be adjusted accordingly.

With respect to how the `CookieJar` determines whether it's in a secure context, I really feel like we need a better approach. Right now, you have to call `set_proxy_secure` everywhere that we create a `Request`. But this is really error prone. It's clear that `CookieJar` needs to have everything it needs when it's created. This way, we can't forget to set something later.

The ideal would be to have `CookieJar::new()` ask for an `&Request`. Unfortunately this would mean that `Request` would need to hold a borrow to itself (`Request` contains `RequestState` which contains `CookieJar`), which is no fun. And it also wouldn't really work since we don't know the headers the request will have until later, in the case of `LocalRequest`. I think the next best thing to do is to:

  1. Remove the `set_()` method(s).
  2. Change the code so we have a new `request.context_is_likely_secure()` method.
  3. Have a single constructor for `CookieJar` that takes both the `&Config` (which we need for the secret key) and a new `secure_context: bool` which will be set to `config.tls_enabled() || request.context_is_likely_secure()`. When we don't have the final request, we can't really do anything. But we should change `Request::from_hyp()` so that it actually calls `CookieJar::new()` instead of adding to the existing cookie jar. At least then we're not mutating something. The `CookieJar` created in `RequestState` will just have the context `config.tls_enabled()`, which is correct but suboptimal.

I'll give this a second look later today to see if there's something more robust we can do.

arjentz:
Implement your suggestions, except for the `from_hyp` rewrite, as I didn't have time for that yet. Feel free to work on it if you find the time, but otherwise I'll take a look soon hopefully

SergioBenitez:
These latest changes look great! Pending consistently using the `CookieJar` constructor and fixing the testing (you can run `./scripts/test.sh` locally to get the same output), we should be good to go. I'll let you push this across the finish line.

arjentz:
Unexpectedly had some time today. Is this more or less what you had in mind?

I also optimised the `ProxyProtocol` implementation a tiny bit

core/lib/src/config/config.rs
```diff
@@ -89,8 +89,23 @@ pub struct Config {
...
+    /// to determine whether a request is handled in a secure context.
+    ///
+    /// [`X-Forwarded-Proto`]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto
+    ///
+    /// To disable using any header for this purpose, set this value to `false`.
+    /// Deserialization semantics are identical to those of [`ip_header`].
+    ///
+    /// **(default: `None`)**
+    #[serde(deserialize_with = "crate::config::http_header::deserialize")]
+    pub proxy_proto_header: Option<Uncased<'static>>,
```
SergioBenitez, line 108: We also need to add this to the "stringy" set of constants.

core/lib/src/config/config.rs
```diff
@@ -395,6 +411,11 @@ impl Config {
             None => launch_meta_!("IP header: {}", bold("disabled"))
         }
 
+        match self.proxy_proto_header {
+            Some(ref name) => launch_meta_!("Protocol header: {}", bold(name)),
+            None => launch_meta_!("Protocol header: {}", bold("disabled"))
```
SergioBenitez, line 416:
```suggestion
            Some(ref name) => launch_meta_!("Proxy-Proto header: {}", bold(name)),
            None => launch_meta_!("Proxy-Proto header: {}", bold("disabled"))
```

core/lib/src/request/from_request.rs
```diff
@@ -169,6 +169,12 @@ impl<S, E> IntoOutcome<S, (Status, E), ()> for Result<S, E> {
 ///     via [`Request::client_ip()`]. If the client's IP address is not known,
 ///     the request is forwarded.
 ///
+///   * **Protocol**
+///
+///     Extracts the protocol of the incoming request as a [`Protocol`] via
+///     [`Request::forwarded_proto()`] (HTTP or HTTPS). If the used protocol is
+///     not known, the request is forwarded.
+///
```
SergioBenitez, line 177: This should reflect the new "proxy protocol" naming/wording.

core/lib/src/request/request.rs
```diff
@@ -386,6 +386,60 @@ impl<'r> Request<'r> {
...
+    /// # let client = rocket::local::blocking::Client::debug(rocket::custom(figment)).unwrap();
+    /// # let request = client.get("/");
+    /// let request = request.header(Header::new("x-forwarded-proto", "https"));
+    /// assert_eq!(request.proxy_proto(), Some(ProxyProto::Https));
+    /// ```
+    pub fn proxy_proto(&self) -> Option<ProxyProto<'_>> {
+        let proxy_proto_header = self.rocket().config.proxy_proto_header.as_ref()?.as_str();
+        self.headers()
+            .get_one(proxy_proto_header)
+            .and_then(|proto| Some(proto.into()))
```
SergioBenitez, line 414:
```suggestion
        self.rocket().config
            .proxy_proto_header
            .as_ref()
            .and_then(|header| self.headers().get_one(header))
            .map(ProxyProto::from)
```

SergioBenitez (changes requested):
Awesome. These suggestions are largely for me. I'll make the changes and get this committed. Thank you!

site/guide/9-configuration.md
```diff
@@ -17,23 +17,24 @@ is configured with. This means that no matter which configuration provider
...
+| key                  | kind              | description                                     | debug/release default   |
+|----------------------|-------------------|-------------------------------------------------|-------------------------|
+| `address`            | `IpAddr`          | IP address to serve on                          | `127.0.0.1`             |
+| `port`               | `u16`             | Port to serve on.                               | `8000`                  |
+| `workers`*           | `usize`           | Number of threads to use for executing futures. | cpu core count          |
+| `max_blocking`*      | `usize`           | Limit on threads to start for blocking tasks.   | `512`                   |
+| `ident`              | `string`, `false` | If and how to identify via the `Server` header. | `"Rocket"`              |
+| `ip_header`          | `string`, `false` | IP header to inspect to get [client's real IP]. | `"X-Real-IP"`           |
+| `proxy_proto_header` | `string`, `false` | Header to inspect for identifying the protocol. | `"X-Forwarded-For"`     |
```
arjentz, line 28:
```suggestion
| `proxy_proto_header` | `string`, `false` | Header to inspect for identifying the protocol. | `None`                  |
```

arjentz:
Thanks! I see now that I forgot to change some comments to the new wording. Do you want me to do any of those changes? I did one more suggestion for a mistake of mine in the docs

SergioBenitez:
Thanks for noticing that! Caught it in a commit I just pushed that addresses all of the issues I've brought up.

The original ideas and implementation were buggy as well as they didn't react correctly if the header was set after the request was received, e.g., by a request fairing. This commit fixes this by going through the existing header cache busting infra.

This work brought to light a different issue, however: any cookies add to the jar _before_ such a header is set won't be marked "secure". This was the case in previous iterations as well, of course, it's just that this implementation made that issue obvious.

It's unclear to me whether this is a big concern or not. Certainly, what's we're doing now is better than what we did before. However, we might also be providing a false sense of security that has edge cases, which might be worse. The potential paths forward I see are:
  1. Scrap the entire idea.
  2. Make `cookie.state.secure` user modifiable, maybe along with 1.
  3. Set `Secure` right before we generate the final response. Headers are no longer modifiable at this point, so the edge case is removed. This introduces some unfortunate complexity to the code as doing this in a performant manner requires a custom `Header` type or else we have to clone cookies.
  4. Document this edge case really well.

I'm inclined to go with 4. It's really unlikely this edge case will be hit, and even more unlikely that it will cause any sort of security issue. But I can be convinced to go with 3.

SergioBenitez:
Oh, I should say, the following things still need to be done:

  - [ ] Document the defaulting behavior in `Cookie::{add, add_private}()`.
  - [ ] Fix the `set_{private}defaults` methods' docs.
  - [ ] Probably document the configuration variable specifically in the configuration guide.

arjentz:
Thanks for the cleanup, it’d gotten a bit of mess after the back and forth renaming. I really like what you did with the `CookieState`, that is much nicer.

So, to get to the best decision, I’d like to reiterate the idea behind the PR. In order to be able to add one more secure default to cookies set by Rocket running behind a proxy, the `x-forwarded-for` header is inspected. This header is intended to be set by the proxy as a way for Rocket to identify the protocol used for a connection. If the proxy terminates the TLS connection, there’d be no other way for Rocket to know about it. This all happens before the request reaches Rocket. Therefore I think it is not necessary to take into account any request fairings modifying the headers, as the proxy proto header is not intended to set in there. However, I do think that cookies added by fairings should make use of the new default, if possible, as it would be desirable and expected that these have the same defaults as cookies set in request handlers.

Hope this helps in reaching a conclusion.

arjentz:
Thanks! I will add that, hopefully somewhere this week, otherwise next week. Do you agree that moving the logic back to `from_hyp` is the way to go? I can create a little POC to demonstrate the reasoning behind why it belongs there.

SergioBenitez:
> Do you agree that moving the logic back to `from_hyp` is the way to go?

Not as of now. Can you explain why you think this is the way to go? As far as I see it, doing this has the following immediate drawbacks:
  * It means a fairing can never set or remove the header and have it impact the `Secure` setting for `Cookie`. While you're right that the header likely comes from the original request, we cannot simply rule out the possibility of a sane reason for a fairing to set or remove it. For example, one might do this because multiple proxies are being used, and some combination of headers set by those proxies should be used as the real header value. Or a fairing might want to remove the header value entirely because it doesn't want to trust the header value in some dynamic cases. Both of these approaches become impossible.
  * There must now be independent logic between "real" requests (where `from_hyp` is called) and local requests (i.e, via `Client`) so that the secure context state in `CookieJar` is set appropriately. This makes the implementation harder to reason about and harder to get right.

The only drawback to the current approach is an additional string comparison when a header is set.

> any cookies set in fairings should default to secure (if the header is set to `https`) as well.

This the case with the either approach.

core/lib/src/request/request.rs
```diff
@@ -934,6 +1012,10 @@ impl<'r> Request<'r> {
             if self.accept().is_none() || replace {
                 self.state.accept = Storage::new();
             }
+        } else if self.rocket().config.proxy_proto_header.as_ref().map_or(false, |h| h == name) {
+            if !self.cookies().state.secure || replace {
+                self.cookies_mut().state.secure = self.context_is_likely_secure();
```
arjentz, line 1017: I think this will not work, as the header is not added te the header map yet. Therefore, `context_is_likely_secure` will call `proxy_proto` which will always return `None`, am I right?

arjentz:
@SergioBenitez did you see my comment [here](https://github.com/SergioBenitez/Rocket/pull/2519#discussion_r1249340266)? I used the following code to test:
```

use rocket::http::{Cookie, CookieJar};

#[get("/")]
fn hello(jar: &CookieJar<'_>) -> &'static str {
    jar.add_private(Cookie::new("a", "a"));
    "Hello, world!"
}

#[launch]
fn rocket() -> _ {
    let config = rocket::Config{proxy_proto_header: Some("x-forwarded-proto".into()), ..rocket::Config::debug_default()};
    rocket::build().mount("/", routes![hello]).configure(config)
}
```

But it doesn't seem to work now (`rev = "e8d326c9d9584f8295989ae6d771fec632bdbcef"`):
```
$ curl -sI -H "x-forwarded-proto: https" 127.0.0.1:8000 | grep "set-cookie"
set-cookie: a=6segmjzwEo1YzIHEo306Rl6X8l1exTLc%2FUkEFCs%3D; HttpOnly; SameSite=Strict; Path=/; Expires=Fri, 25 Aug 2023 19:10:48 GMT
```

Even though it did before (`rev = "52bd6fd2c1a12e41f6999b0e92e3118be8a948c6"`):
```
$ curl -sI -H "x-forwarded-proto: https" 127.0.0.1:8000 | grep "set-cookie"
set-cookie: a=kxg8sjz8m0SpSGoXqssn1p9yKCtDNO4Ctd0vV9s%3D; HttpOnly; SameSite=Strict; Secure; Path=/; Expires=Fri, 25 Aug 2023 19:10:59 GMT
```

core/lib/src/config/config.rs
```diff
@@ -90,8 +90,23 @@ pub struct Config {
...
+    /// The name of a header, whose value is typically set by an intermediary
+    /// server or proxy, which contains the protocol (HTTP or HTTPS) used by the
+    /// connecting client. This should probably be [`X-Forwarded-Proto`], as
+    /// that is the de facto standard. Used by [`Request::forwarded_proto()`]
+    /// to determine the forwarded protocol and [`Request::forwarded_secure()`]
+    /// to determine whether a request is handled in a secure context.
+    ///
+    /// [`X-Forwarded-Proto`]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto
+    ///
+    /// To disable using any header for this purpose, set this value to `false`.
```
SergioBenitez, line 104:
```suggestion
    /// To disable using any header for this purpose, set this value to `false` or `None`.
```

core/lib/src/request/from_request.rs
```diff
@@ -483,6 +489,18 @@ impl<'r> FromRequest<'r> for IpAddr {
...
 }
 
+#[crate::async_trait]
+impl<'r> FromRequest<'r> for ProxyProto<'r> {
+    type Error = std::convert::Infallible;
+
+    async fn from_request(request: &'r Request<'_>) -> Outcome<Self, Self::Error> {
+        match request.proxy_proto() {
+            Some(proto) => Success(proto),
+            None => Forward(Status::NotFound),
```
SergioBenitez, line 499:
```suggestion
            None => Forward(Status::InternalServerError),
```

core/lib/src/request/request.rs
```diff
@@ -386,6 +386,84 @@ impl<'r> Request<'r> {
...
+    /// # let req = c.get("/");
+    /// let req = req.header(Header::new("x-forwarded-proto", "xproto"));
+    /// assert_eq!(req.proxy_proto(), Some(ProxyProto::Unknown("xproto".into())));
+    /// ```
+    pub fn proxy_proto(&self) -> Option<ProxyProto<'_>> {
+        self.rocket().config
+            .proxy_proto_header
+            .as_ref()
+            .and_then(|header| self.headers().get_one(header.as_str()))
+            .map(ProxyProto::from)
```
SergioBenitez, line 429:
```suggestion
        self.rocket()
            .config
            .proxy_proto_header
            .as_ref()
            .and_then(|header| self.headers().get_one(header.as_str()))
            .map(ProxyProto::from)
```

core/lib/src/request/request.rs
SergioBenitez: This method is called any time a header is added or replaced in the header map. If it starts in the map, then `state.secure` == `true`. If it's added later, than this method will correct the value. That's the idea, at least.

SergioBenitez:
> @SergioBenitez did you see my comment [here](https://github.com/SergioBenitez/Rocket/pull/2519#discussion_r1249340266)? I used the following code to test:
> 
> ```
> #[macro_use] extern crate rocket;
> 
> use rocket::http::{Cookie, CookieJar};
> 
> #[get("/")]
> fn hello(jar: &CookieJar<'_>) -> &'static str {
>     jar.add_private(Cookie::new("a", "a"));
>     "Hello, world!"
> }
> 
> #[launch]
> fn rocket() -> _ {
>     let config = rocket::Config{proxy_proto_header: Some("x-forwarded-proto".into()), ..rocket::Config::debug_default()};
>     rocket::build().mount("/", routes![hello]).configure(config)
> }
> ```
> 
> But it doesn't seem to work now (`rev = "e8d326c9d9584f8295989ae6d771fec632bdbcef"`):
> 
> ```
> $ curl -sI -H "x-forwarded-proto: https" 127.0.0.1:8000 | grep "set-cookie"
> set-cookie: a=6segmjzwEo1YzIHEo306Rl6X8l1exTLc%2FUkEFCs%3D; HttpOnly; SameSite=Strict; Path=/; Expires=Fri, 25 Aug 2023 19:10:48 GMT
> ```
> 
> Even though it did before (`rev = "52bd6fd2c1a12e41f6999b0e92e3118be8a948c6"`):
> 
> ```
> $ curl -sI -H "x-forwarded-proto: https" 127.0.0.1:8000 | grep "set-cookie"
> set-cookie: a=kxg8sjz8m0SpSGoXqssn1p9yKCtDNO4Ctd0vV9s%3D; HttpOnly; SameSite=Strict; Secure; Path=/; Expires=Fri, 25 Aug 2023 19:10:59 GMT
> ```

Let me rebase on `master` to retest.

SergioBenitez:
Ah, right, yes, you're absolutely right. In the `bust_header_cache`, we check for the header, but we call that method _before_ we actually insert the header. So, indeed, this won't work as you suggest. Let me fix this.

SergioBenitez:
> This work brought to light a different issue, however: any cookies added to the jar _before_ such a header is set won't be marked "secure". This was the case in previous iterations as well, of course, it's just that this implementation made the issue obvious.

Note that this can only really occur in local testing, where the jar is retrieved manually before setting a header. Otherwise, we only return a `Request` when we've finished parsing header in an incoming request.

SergioBenitez:
Okay, I've fixed the issue while also simplifying the impl. Please take a look and let me know what you think. Also check out the unresolved review items as well as my commentary for what's still left to be done.

core/lib/src/cookies.rs
arjentz: Superseded by https://github.com/rwf2/Rocket/pull/2519/commits/678b7c1f8dfe59d43a6295c0221d102ba720d520#diff-c8751afca2b03eeea5b65a1e8cc20f0cf291dcce2a129b96d3d405247c1d7222R178

core/lib/src/request/from_request.rs
arjentz: Let me get this right; if one adds `ProxyProto` as a request guard to a particular route, and no such header is set (and no other matching route exists), this change would make it return an internal server error?

core/lib/src/config/config.rs
arjentz: Included in commit

core/lib/src/request/request.rs
arjentz: This has been removed by now. It was copied from `ip_header` earlier.

core/lib/src/config/config.rs
arjentz: Noticed this one wasn't included yet. Did it just now

core/lib/src/request/request.rs
arjentz: Fixed in 678b7c1f8dfe59d43a6295c0221d102ba720d520 (by you)

core/lib/src/request/request.rs
arjentz: Already fixed in 678b7c1f8dfe59d43a6295c0221d102ba720d520

core/lib/src/request/from_request.rs
arjentz: This discussion still stands, unless keeping it is now deemed acceptable. Also see https://github.com/rwf2/Rocket/pull/2519#discussion_r1428635476

arjentz:
Thanks for taking the time. I really like how you simplified it, especially the `CookieState` in the `CookieJar`. I checked all open discussions and resolved what was fixed, superseded or outdated. I've also added a few more updates to comments, including the first two points in https://github.com/rwf2/Rocket/pull/2519#issuecomment-1536949567. The third one is still to be done, which I want to make sure to do right. I'll try to look at it this week, if you find the time, could you have a look at the other changes and two discussions I left open in the meantime?

core/lib/src/request/from_request.rs
SergioBenitez: That's right. The idea is that if a route requests a `ProxyProto` directly, then that implies that they expect it. If it's not present, then that's a logic error, which we signal via the `500`. If instead the user is unsure if the value is present, they can instead use an `Option<ProxyProto>`, for example, which is infallible.

core/lib/src/config/config.rs
```diff
@@ -409,6 +425,11 @@ impl Config {
             None => launch_meta_!("IP header: {}", "disabled".paint(VAL))
         }
 
+        match self.proxy_proto_header {
+            Some(ref name) => launch_meta_!("Proxy-Proto header: {}", name.paint(VAL)),
```
SergioBenitez, line 429: Slight simplification:

```suggestion
        match self.proxy_proto_header.as_ref() {
            Some(name) => launch_meta_!("Proxy-Proto header: {}", name.paint(VAL)),
```

core/lib/src/cookies.rs
```diff
@@ -287,7 +279,10 @@ impl<'a> CookieJar<'a> {
     ///    * `path`: `"/"`
     ///    * `SameSite`: `Strict`
     ///
-    /// Furthermore, if TLS is enabled, the `Secure` cookie flag is set.
+    /// Furthermore, if TLS is enabled or handled by a proxy (as determined by
+    /// `Request::context_is_likely_secure`), the `Secure` cookie flag is set.
```
SergioBenitez, line 283:
```suggestion
    /// [`Request::context_is_likely_secure()`]), the `Secure` cookie flag is set.
```

core/lib/src/cookies.rs
```diff
@@ -327,9 +322,10 @@ impl<'a> CookieJar<'a> {
     ///    * `HttpOnly`: `true`
     ///    * `Expires`: 1 week from now
     ///
-    /// Furthermore, if TLS is enabled, the `Secure` cookie flag is set. These
-    /// defaults ensure maximum usability and security. For additional security,
-    /// you may wish to set the `secure` flag.
+    /// Furthermore, if TLS is enabled or handled by a proxy (as determined by
+    /// `Request::context_is_likely_secure`), the `Secure` cookie flag is set.
```
SergioBenitez, line 326:
```suggestion
    /// [`Request::context_is_likely_secure()`]), the `Secure` cookie flag is set.
```

core/lib/src/cookies.rs
```diff
@@ -515,8 +511,9 @@ impl<'a> CookieJar<'a> {
     ///    * `path`: `"/"`
     ///    * `SameSite`: `Strict`
     ///
-    /// Furthermore, if TLS is enabled, the `Secure` cookie flag is set.
-    fn set_defaults(config: &Config, cookie: &mut Cookie<'static>) {
+    /// Furthermore, if TLS is enabled or handled by a proxy (as determined by
+    /// `Request::context_is_likely_secure`), the `Secure` cookie flag is set.
```
SergioBenitez, line 515:
```suggestion
    /// [`Request::context_is_likely_secure()`]), the `Secure` cookie flag is set.
```

core/lib/src/cookies.rs
```diff
@@ -554,17 +551,12 @@ impl<'a> CookieJar<'a> {
     ///    * `HttpOnly`: `true`
     ///    * `Expires`: 1 week from now
     ///
-    /// Furthermore, if TLS is enabled, the `Secure` cookie flag is set.
+    /// Furthermore, if TLS is enabled or handled by a proxy (as determined by
+    /// `Request::context_is_likely_secure`), the `Secure` cookie flag is set.
```
SergioBenitez, line 555:
```suggestion
    /// [`Request::context_is_likely_secure()`]), the `Secure` cookie flag is set.
```

core/lib/src/request/request.rs
```diff
@@ -386,6 +385,85 @@ impl<'r> Request<'r> {
...
+
+    /// Returns whether we are *likely* in a secure context.
+    ///
+    /// A request is in a "secure context" if it was initially sent over a
+    /// secure (TLS, via HTTPS) connection. If TLS is configured and enabled,
+    /// then the request is guaranteed to be in a secure context. Otherwise, if
+    /// [`Request::proxy_proto()`] evaluates to `Https`, then we are _likely_ to
+    /// be in a secure context. We say _likely_ because it is entirely possible
+    /// for the header to indicate that the connection is being proxied via
+    /// HTTPS while reality differs. As such, this values should not be trusted
```
SergioBenitez, line 440:
```suggestion
    /// HTTPS while reality differs. As such, this value should not be trusted
```

site/guide/9-configuration.md
```diff
@@ -17,26 +17,29 @@ is configured with. This means that no matter which configuration provider
...
-parameters are only read from the [default provider](#default-provider).</small>
+| key                  | kind              | description                                     | debug/release default   |
+|----------------------|-------------------|-------------------------------------------------|-------------------------|
+| `address`            | `IpAddr`          | IP address to serve on                          | `127.0.0.1`             |
+| `port`               | `u16`             | Port to serve on.                               | `8000`                  |
+| `workers`*           | `usize`           | Number of threads to use for executing futures. | cpu core count          |
+| `max_blocking`*      | `usize`           | Limit on threads to start for blocking tasks.   | `512`                   |
+| `ident`              | `string`, `false` | If and how to identify via the `Server` header. | `"Rocket"`              |
+| `ip_header`          | `string`, `false` | IP header to inspect to get [client's real IP]. | `"X-Real-IP"`           |
+| `proxy_proto_header` | `string`, `false` | Header identifying client to proxy protocol.    | `None`                  |
```
SergioBenitez, line 28: Maybe this should link `ProxyProto` the way `LogLevel` does, for example?

SergioBenitez (changes requested):
Some minor changes here, nothing interested. I think we're at the finish line!

For the documentation on when `context_is_likely_secure()`, we need to say that the `Secure` flag is set _as long as_ it's not explicitly set to `false`. It's already mentioned that this is a "default", but I think being explicit here is worthwhile.

site/guide/9-configuration.md
arjentz: I linked `proxy_proto` the way `ip_header` does. I figured `ProxyProto` itself is not particularly useful here, as it is not used as the setting itself.

arjentz:
Rebased the branch and added the configuration entry. Happy to iterate on that a bit more if you think it needs that.

core/http/src/header/proxy_proto.rs
```diff
@@ -0,0 +1,51 @@
+use std::fmt;
+
+use uncased::{UncasedStr, AsUncased};
+
+/// A protocol used to identify a specific protocol forwarded by an HTTP proxy.
+/// Value are case-insensitive.
```
SergioBenitez, line 6: I think we need to add some links to where this is used (`Config::proxy_proto_header`, the cookie defaults) and where in Rocket someone might interact with it (as a request guard, `Request::proxy_proto_header()`, context_is_likely_secure()).

core/lib/src/config/config.rs
```diff
@@ -85,13 +85,28 @@ pub struct Config {
...
+    /// except that the value must syntactically be a valid HTTP header name.
     ///
     /// **(default: `"X-Real-IP"`)**
-    #[serde(deserialize_with = "crate::config::ip_header::deserialize")]
+    #[serde(deserialize_with = "crate::config::http_header::deserialize")]
     pub ip_header: Option<Uncased<'static>>,
+    /// The name of a header, whose value is typically set by an intermediary
+    /// server or proxy, which contains the protocol (HTTP or HTTPS) used by the
+    /// connecting client. This should probably be [`X-Forwarded-Proto`], as
+    /// that is the de facto standard. Used by [`Request::forwarded_proto()`]
```
SergioBenitez, line 98: The wording "this should probably be" begets the question: why isn't this the default, then?

I think we need to be much, much clearer here about what this value control, when it should be set, and if set, what values might be typical.

core/lib/tests/config-proxy-proto-header.rs
```diff
@@ -0,0 +1,118 @@
...
+#[get("/")]
+fn inspect_proto(proto: Option<ProxyProto>) -> String {
+    proto
+        .map(|proto| match proto {
+            ProxyProto::Http => "http".to_owned(),
+            ProxyProto::Https => "https".to_owned(),
+            ProxyProto::Unknown(s) => s.to_string(),
+        })
+        .unwrap_or("<none>".to_owned())
+}
```
SergioBenitez, line 15:
```suggestion
#[get("/")]
fn inspect_proto(proto: ProxyProto) -> String {
    proto.to_string()
}
```

And then adjust the tests so they look for a `500` when there is no such header.

core/lib/tests/config-proxy-proto-header.rs
```diff
@@ -0,0 +1,118 @@
...
+        assert_eq!(response.into_string().unwrap(), "http");
+
+        let response = client.get("/")
+            .header(Header::new("X-Url-Scheme", "https"))
+            .dispatch();
+
+        assert_eq!(response.into_string().unwrap(), "https");
+
+        let response = client.get("/").dispatch();
+        assert_eq!(response.into_string().unwrap(), "<none>");
```
SergioBenitez, line 51: e.g.:

```suggestion
        assert_eq!(response.status(), Status::InternalServerError);
```

site/guide/9-configuration.md
```diff
@@ -148,7 +152,8 @@ workers = 16
 max_blocking = 512
 keep_alive = 5
 ident = "Rocket"
-ip_header = "X-Real-IP" # set to `false` to disable
+ip_header = "X-Real-IP" # set to `false` or `None` to disable
+proxy_proto_header = `false` # set to `false` or `None` to disable
```
SergioBenitez, line 156: Is saying `None` correct here? This implies you can write:

```toml
ip_header = None
proxy_proto_header = None
```

And it'll disable the header. But that isn't the case.

```suggestion
ip_header = "X-Real-IP" # set to `false` to disable
proxy_proto_header = `false` # set to `false` to disable
```

site/guide/9-configuration.md
```diff
@@ -358,6 +363,25 @@ mutual TLS.
...
 
+### Proxied TLS
+
+If Rocket is running behind a reverse proxy that terminates TLS, it is useful to
+know whether the original connection was made securely. Therefore, Rocket offers
+the option to configure a `proxy_proto_header` that is used to determine if the
+request is handled in a secure context. The outcome is available via
+[`Request::context_is_likely_secure()`] and used to set cookies' secure flag by
+default. To enable this behaviour, configure the header as set by your reverse
+proxy. For example:
```
SergioBenitez, line 374: Again, we should be more specific: why is it useful? I'd suggest inverting this paragraph to something like:

```suggestion
The `proxy_proto_header` configuration parameter allows Rocket applications to
determine when and if a client's initial connection was likely made in a
secure context by examining the header with the configured name. The
header's value is parsed into a [`ProxyProto`] value and retrievable via
[`Request::proxy_proto()`].

That value is in-turn inspected to determine if the initial protocol was secure
(i.e, over TLS) and the result made available via
[`Request::context_is_likely_secure()`]. The value returned by this method
influences cookie defaults. In particular, if the method returns `true` (the
request context is likely secure) secure, the cookie `Secure` cookie flag is set
by default.
```

SergioBenitez (changes requested):
I think we're set functionality wise. The docs need a bit of work. I'll fix them up, push a commit, and get this merged in a bit.

PR #2525, opened 2023-04-17
Update Hyper from 0.14.9 -> 0.14.26 for security reasons

x10an14:
This upgrade the transitive dependency of H2 from <0.3.17 (0.3.9) -> 0.3.17.
This removes CVE-2023-26964 / GHSA-f8vr-r385-rh5r, see: https://github.com/advisories/GHSA-f8vr-r385-rh5r.

### DIFF
--- a/core/http/Cargo.toml
+++ b/core/http/Cargo.toml
@@ -51,7 +51,7 @@ version = "0.13"
 optional = true
 
 [dependencies.hyper]
-version = "0.14.9"
+version = "0.14.26"
 default-features = false
 features = ["http1", "runtime", "server", "stream"]
 


x10an14:
Tested to pass all test for the following commits:
1. 9b0564ed27f90686b333337d9f6ed76484a84b27 (git default branch - master?)
2. 91f6288ea4aeb3d5a502b2f18b2b9677a85463ea (v0.5-rc branch - where I'd like this to end up!)

SergioBenitez:
This doesn't meaningfully change anything. A cargo version specified without an equality marker is equivalent to `>=`, so all this does is raise the minimum version. Old projects will still pin to the old version, and new projects will already default to the new version.

Still, there's no reason _not_ to make this change.

x10an14:
> Perhaps you're aware of this, but a cargo version specified without an equality marker is equivalent to >=, so this commit only raises the minimum version. Old projects will still pin to the old version, concerned projects can simply cargo update to get the latest version, and new projects will already default to the new version.

That may be true, I don't know myself and trust what you're saying about it =)

What's an interesting datapoint though is that I `cargo add`ed `rocket@v0.5.0-rc3` _last week_.
The `0.14.25` version of Hyper is reported by https://crates.io/crates/hyper to have been published ~2023.03.25.

Yet I got the mentioned dependabot error, when - if I understood your logic correctly - I should've gotten the latest version of Hyper.

Maybe it's Hyper that's not leveraging default `>=` setting you're explaining?

PR #2537, opened 2023-05-11
Language updates for the hello example

SerheyDolgushev:


SergioBenitez:
Thanks for the PR!

We generally don't make changes to the codebase that don't modify the existing functionality nor exhibit new functionality. Russian was chosen for the example purely because it uses a UTF-8 encoded script outside of the ASCII range for the standard character set, allowing us to exhibit Rocket's ability to handle any Unicode character in UTF-8. I'm happy to choose any other language that has the same properties as long as the change is motivated by technical reasons. Given this, I'm closing the PR, but please feel free to submit another PR that follows these general guidelines.

Thanks again!

Issue #2363, opened 2022-10-12
Hot reloading tls certificates


cobalthex:
**Is your feature request motivated by a concrete problem? Please describe.**

I am not sure if this feature already exists, but most SSL certs have an expiration and it would be nice to have the web server able to reload the certs (either automatically from the file specified, or manually by my own code)

**Why this feature can't or shouldn't live outside of Rocket**

Rocket implements TLS

**Ideal Solution**

Rocket provides a way to supply tls after launch, or supports hot reloading

cobalthex:
Not sure if this is a second ticket worthy but it would be nice to perhaps provide support for this via a callback which can allow for returning different values based on the incoming domain

SergioBenitez:
> Not sure if this is a second ticket worthy but it would be nice to perhaps provide support for this via a callback which can allow for returning different values based on the incoming domain

Agreed. I think the ideal interface here is one where both Rocket can provide some default functionality (like updating the TLS certs if they change on disk) but applications can plug in their own dynamic TLS functionality at will as well. I don't have a concrete proposal for an api just yet, but I'm more than in favor of this change. Perhaps this can be part of #1070?

cobalthex:
Sure, ideally as long as it's not all managed or all user driven (ie it would be nice to say here's my tls function without having to manage every aspect of the connection manually)

SergioBenitez:
> Sure, ideally as long as it's not all managed or all user driven (ie it would be nice to say here's my tls function without having to manage every aspect of the connection manually)

I don't think that interface would be a problem. The custom listener could simply forward the majority of the implementation to an existing listener.

GentBinaku:
Can I work on this?

SergioBenitez:
Here's a guide on how to work on this issue. If you decide to work on this, please let me know. I'll assign the issue.

## Getting Familiar

The first step, as always, is to get familiar with the relevant code. 

* Rocket's [`http` crate](https://github.com/SergioBenitez/Rocket/tree/9b0564ed27f90686b333337d9f6ed76484a84b27/core/http) contains all of the relevant HTTP server handling code, including handling TLS connections and certificates. The latter functionality is contained in the [`tls` module](https://github.com/SergioBenitez/Rocket/tree/9b0564ed27f90686b333337d9f6ed76484a84b27/core/http/src/tls).

[...]

PR #2542, opened 2023-05-18
[WIP] Hot reloading tls certificates #2363

GentBinaku:


GentBinaku:
Hi @SergioBenitez  this is a draft PR so you can see the progress and maybe make suggestions here and there

core/http/src/tls/listener.rs
```diff
@@ -52,16 +60,15 @@ pub struct TlsStream {
     state: TlsState,
     certs: Certificates,
 }
-
```
SergioBenitez, line 55: Why are these empty lines going away? We always want an empty line between unrelated items.

SergioBenitez (changes requested):
The use of `AtomicPtr` here is really not necessary and it's likely incorrect, though I haven't taken a close look. I may have pointed you to it before, but I meant it as an example. It is possible to use `AtomicPtr` to get an atomic swap safely, but while you develop this, it's much safer to just use something like `RwLock` either from `std` or from `tokio` in the interim.

The spawned task, if it's going to be doing this, needs to also watch out for [shutdown](https://api.rocket.rs/v0.5-rc/rocket/struct.Shutdown.html) or else it will stall shutdown.

PR #2545, opened 2023-05-19
Remove `Send` constraint from async runtime futures

nathanfranke:
This may or may not break everything

Allows me to do this:

```rs
#[rocket::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    rocket().await?.launch().await?;
    Ok(())
}
```


SergioBenitez:
Given the above and lack of response, I'm closing this PR. I'm happy to revisit this should something suggest that this change is the right approach.

PR #2546, opened 2023-05-19
Update tungstenite crate

BlackDex:
- Update tungstenite to the latest version v0.19

Would be cool if this could be merged to v0.5 :)


SergioBenitez:
Merged into master and v0.5!

PR #2555, opened 2023-06-04
Upgrade diesel to v2.1.0

ParisLiakos:
https://github.com/diesel-rs/diesel/releases/tag/v2.1.0


SergioBenitez:
This PR doesn't meaningfully change anything. Just run cargo update and you'll get the latest version. See other PRs that have updated versions similarly without effect for the rationale.

PR #2566, opened 2023-06-18
Allow responding with named MessagePack data.

### LINKED ISSUE #2107: Allow named msgpack serialization
Since @SergioBenitez is returning, imma pin that one here https://github.com/rocket-org/Rocket/issues/18 @Cloudef

> **Is your feature request motivated by a concrete problem? Please describe.**
> 
> Currently the `serde::msgpack::MsgPack` included with rocket only does "compact serialization", meaning it doesn't serialize key names, only values in arrays. To be JSON compatible, and to avoid bugs it would be nice to have "named serialization".
> 
> **Why this feature can't or shouldn't live outside of Rocket**
> 
> The `serde::msgpack` is already in rocket.
> 
> **Ideal Solution**
> 
> It seems the problem is here: https://github.com/Rocket-org/Rocket/blob/cc0621626bf5e917ea1d4c1e29a327e9a7871921/core/lib/src/serde/msgpack.rs#L191 This should be changed to `rmp_serde::to_vec_named`, however since I'm still fairly new to rust, I don't know how this would be exposed so that the developer can choose either from the compact or named serialization form.

Artemis21:
Closes #2107. See also #1882.

This implementation does not change the default, but notes in the docs that users should pick an explicit option to allow for the default changing in the future.

Also, const generics are used to avoid having to duplicate implementations which would be the same for both named and compact representations. Arguably this does not provide the best user ergonomics, so let me know if you'd prefer a different implementation.

### DIFF
--- a/core/lib/src/serde/msgpack.rs
+++ b/core/lib/src/serde/msgpack.rs
@@ -43,22 +43,12 @@ pub use rmp_serde::decode::Error;
 ///
 /// ## Sending MessagePack
 ///
-/// To respond with serialized MessagePack data, return a `MsgPack<T>` type,
-/// where `T` implements [`Serialize`] from [`serde`]. The content type of the
-/// response is set to `application/msgpack` automatically.
+/// To respond with serialized MessagePack data, return either [`Named<T>`] or
+/// [`Compact<T>`] from your handler. `T` must implement [`serde::Serialize`].
 ///
-/// ```rust
-/// # #[macro_use] extern crate rocket;
-/// # type User = usize;
-/// use rocket::serde::msgpack::MsgPack;
-///
-/// #[get("/users/<id>")]
-/// fn user(id: usize) -> MsgPack<User> {
-///     let user_from_id = User::from(id);
-///     /* ... */
-///     MsgPack(user_from_id)
-/// }
-/// ```
+/// Currently, returning `MsgPack<T>` is equivalent to returning `Compact<T>`,
+/// but you should prefer to use an explicit option as this default may change
+/// in the future.
 ///
 /// ## Receiving MessagePack
 ///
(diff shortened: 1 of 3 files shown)


core/lib/src/serde/msgpack.rs
```diff
@@ -123,9 +113,61 @@ pub use rmp_serde::decode::Error;
...
+///    /* ... */
+///    msgpack::MsgPack(user_from_id)
+/// }
+/// ```
+///
+/// Prefer using [`MsgPack<T>`] for request guards, as the named/compact
+/// distinction is not relevant for request data - the correct option is
+/// implemented automatically. Using [`Named<T>`] as a request guard will
+/// NOT prevent compact requests from being accepted.
+pub type Named<T> = MsgPack<T, false>;
```
the10thWiz, line 168: We should consider only exporting the `Compact` type alias, and renaming it to be something like `CompactMsgPack`. Since the default would now be named, we don't need a type alias that doesn't change from the default.

core/lib/src/serde/msgpack.rs
```diff
@@ -123,9 +113,61 @@ pub use rmp_serde::decode::Error;
 /// msgpack = 5242880
 /// ```
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct MsgPack<T>(pub T);
+pub struct MsgPack<T, const COMPACT: bool = true>(pub T);
```
the10thWiz, line 116: I think we should change the default here. This is likely to be the more common case, and provides better interoperability with potential client code.

the10thWiz (changes requested):

the10thWiz:
Merged in db598be3!

PR #2586, opened 2023-07-30
Use `Span::mixed_site` to avoid let unit value warnings

### LINKED ISSUE #2568: False positive `clippy::let_unit_value` in empty route
[`()` implements `Responder`](https://docs.rs/rocket/latest/rocket/response/trait.Responder.html#impl-Responder%3C%27r%3E-for-()), which allows you to write, for example:
```rust
#[get("/")]
fn foo() {}
```
However, under the hood it is expand into: 
```rust
let ___responder = init();
::rocket::route::Outcome::from(__req, ___responder)
```
Clippy doesn't like that
```rust
warning: this let-binding has unit value
  --> src\main.rs:38:8
   |
38 |     fn foo() {}
   |        ^^^^ help: omit the `let` binding: `foo;`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#let_unit_value
   = note: `#[warn(clippy::let_unit_value)]` on by default
```

In my view, adding `#[allow(clippy::let_unit_value)]` to [here](https://github.com/SergioBenitez/Rocket/blob/9a9cd76c0121f46765ff0df9ef81e36563a2a31f/core/codegen/src/attribute/route/mod.rs#L245-L248) would fix this

jieyouxu:
Using `Span::mixed_site` changes the syntax context of the generated responder outcome expr to identify as macro-expanded, helping the `clippy::let_unit_value` lint and a potential `unit_bindings` rustc lint to avoid linting on generated code.

Closes #2568.
Avoids a potential rustc `unit_bindings` lint (if it is merged at https://github.com/rust-lang/rust/pull/112380).

### DIFF
--- a/core/codegen/src/attribute/route/mod.rs
+++ b/core/codegen/src/attribute/route/mod.rs
@@ -242,7 +242,7 @@ fn responder_outcome_expr(route: &Route) -> TokenStream {
         .map(|a| quote_spanned!(a.span() => .await));
 
     define_spanned_export!(ret_span => __req, _route);
-    quote_spanned! { ret_span =>
+    quote_spanned! { Span::mixed_site().located_at(ret_span) =>
         let ___responder = #user_handler_fn_name(#(#parameter_names),*) #_await;
         #_route::Outcome::from(#__req, ___responder)
     }


SergioBenitez:
This is actually correct? Doesn't this change hygiene?

jieyouxu:
> Is this correct? Doesn't this change hygiene?

It does indeed change hygiene AFAIUI, in that it becomes mixed hygienic. That is,

> Procedural macro producing tokens with `Span::mixed_site()` spans should be equivalent to a `macro_rules` macro producing the same tokens.

My understanding (might be wrong here) is that originally the proc_macro was unhygienic w.r.t. local variables (which might be why `___responder` has the three underscores?) and defaults to `Span::call_site` hygiene. Using `Span::mixed_site().located_at(ret_span)` would change the macro from being unhygienic w.r.t. local variables to being mixed hygienic w.r.t. local variables but retain the line/column information of `ret_span`.

SergioBenitez:
I've merged this. Let's see how it goes!

PR #2598, opened 2023-08-14
Modify fairings

the10thWiz:
Supersedes: #2271
Resolves: #2271

Adds a modify_fairings method to make changes to fairings before launch.

There were several comments on the original PR, which have been resolved. Some simplification of the API has been performed.

### DIFF
--- a/core/lib/src/fairing/fairings.rs
+++ b/core/lib/src/fairing/fairings.rs
@@ -1,5 +1,3 @@
-use std::collections::HashSet;
-
 use crate::{Rocket, Request, Response, Data, Build, Orbit};
 use crate::fairing::{Fairing, Info, Kind};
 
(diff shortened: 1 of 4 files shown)


SergioBenitez:
I went a bit of a different direction with this one that I think results in a nicer API that can be used in more situations. It necessitated a bit of tricky code in the fairings set, but thankfully that complexity doesn't leak to the user, and there's no `unsafe` anywhere. Let me know what you think.

the10thWiz (commented):
I like the changes. My original PR was intended to work with a specific use-case, but I think the final API works just fine.

The only thing I see is that you can technically call `fairings_mut` on a `Rocket<Orbit>`, but I don't think there's any point during the Rocket lifecycle this is actually possible outside of Rocket. I don't think this is an issue, since there isn't much you could do at that point anyway.

(Btw, I'd post an approval, but I can't approve my own PR).

PR #2613, opened 2023-09-19
Upgrade tokio-tungstenite to 0.20

fenhl:


SergioBenitez:
What's changed? Can you share why this version bump is just okay to do?

fenhl:
In tungstenite 0.20, new methods for writing data have been added that no longer implicitly flush the sinks, and the previous methods with implicit flushing have been deprecated. Since there are no deprecation warnings, I believe rocket-ws is not affected by the breaking change.

SergioBenitez (approved):
LGTM

SergioBenitez:
Merged into master! Thank you.

PR #2616, opened 2023-09-20
Fix #1224 by searching routes after failed match

### LINKED ISSUE #1224: Rocket should show 405 instead of 404 for method not allowed

jespersm:
This PR fixes #1224 by introducing more detailed HTTP status reporting.

If the request is matched, this patch does nothing. That way, routes which set their own HTTP status code are not affected even if they return 404.

If *no* route matches the request, this PR checks for "near matches", first looking to see if any routes failed only the format check, and afterwards checks for matching paths, but on a different HTTP method.

For format-check failures for POST and PUT, the HTTP status code will be 415 Unsupported Media Type (since this means that the Content-Type of the request didn't match what the route expected). For GET, etc., a failed format check will be 406 Not Acceptable, since the route couldn't produce what the client requested.

If the method and path didn't match at all, but the path *is* present as a route for one of the other HTTP methods, 405 Method Not Allowed is returned.

Otherwise, the 404 Not Found is left as is.

Tests were corrected as required.

### DIFF
--- a/core/lib/src/router/router.rs
+++ b/core/lib/src/router/router.rs
@@ -6,6 +6,8 @@ use crate::http::{Method, Status};
 use crate::{Route, Catcher};
 use crate::router::Collide;
 
+use super::matcher::{paths_match, queries_match};
+
 #[derive(Debug, Default)]
 pub(crate) struct Router {
     routes: HashMap<Method, Vec<Route>>,
@@ -62,7 +64,7 @@ impl Router {
         self.routes.get(&req.method())
             .into_iter()
             .flatten()
-            .any(|route| super::matcher::paths_match(route, req) && super::matcher::queries_match(route, req))
+            .any(|route| paths_match(route, req) && queries_match(route, req))
     }
 
     const ALL_METHODS: &[Method] = &[
@@ -79,7 +81,7 @@ impl Router {
             .filter(|method| *method != &req.method())
             .filter_map(|method| self.routes.get(method))
             .flatten()
-            .any(|route| super::matcher::paths_match(route, req))
+            .any(|route| paths_match(route, req))
     }
 
     // For many catchers, using aho-corasick or similar should be much faster.
--- a/core/lib/src/router/router.rs
+++ b/core/lib/src/router/router.rs
@@ -67,7 +67,7 @@ impl Router {
             .any(|route| paths_match(route, req) && queries_match(route, req))
     }
 
-    const ALL_METHODS: &[Method] = &[
+    const ALL_METHODS: &'static [Method] = &[
         Method::Get, Method::Put, Method::Post, Method::Delete, Method::Options,
         Method::Head, Method::Trace, Method::Connect, Method::Patch,
     ];
--- a/core/lib/src/router/router.rs
+++ b/core/lib/src/router/router.rs
@@ -55,6 +55,33 @@ impl Router {
             .flat_map(move |routes| routes.iter().filter(move |r| r.matches(req)))
     }
 
+    pub(crate) fn matches_except_formats<'r, 'a: 'r>(
+        &'a self,
+        req: &'r Request<'r>
+    ) -> bool {
+        self.routes.get(&req.method())
+            .into_iter()
+            .flatten()
+            .any(|route| super::matcher::paths_match(route, req) && super::matcher::queries_match(route, req))
+    }
+
+    const ALL_METHODS: &[Method] = &[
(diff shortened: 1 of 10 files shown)


jespersm:
(Also, compared to earlier PRs, this approach only adds runtime overhead when routes fail to match.)

core/lib/src/router/router.rs
```diff
@@ -55,6 +57,33 @@ impl Router {
...
+    pub(crate) fn matches_except_method<'r, 'a: 'r>(
+        &'a self,
+        req: &'r Request<'r>
+    ) -> bool {
+        Self::ALL_METHODS
+            .iter()
+            .filter(|method| *method != &req.method())
+            .filter_map(|method| self.routes.get(method))
+            .flatten()
+            .any(|route| paths_match(route, req))
```
palant, line 84: It should be possible to simplify this using the `Router::routes` method:

```rust
self.routes()
    .any(|route| route.method != req.method() && paths_match(route, req))
```

core/lib/src/server.rs
```diff
@@ -342,6 +342,20 @@ impl Rocket<Orbit> {
...
         error_!("No matching routes for {}.", request);
+        if request.route().is_none() {
+            // We failed to find a route which matches on path, query AND formats.
+            if self.router.matches_except_formats(request) {
+                // Tailor the error code to the interpretation of the request in question.
+                if request.method().supports_payload() {
+                    status = Status::UnsupportedMediaType;
+                } else {
+                    status = Status::NotAcceptable;
+                }
```
palant, line 353: I don’t think that returning `NotAcceptable` is correct here. This status code is about content negotiation via the request’s `Accept` header – e.g. the client wants application/json but the server can only deliver text/html. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406

IMHO this branch should always return `UnsupportedMediaType` regardless of whether payload is supported.

palant (commented):
Note that I am not the maintainer of this repository, so I’m only nit picking here and you are free to ignore my comments.

core/lib/src/router/router.rs
jespersm: Thanks, I'll check how that works out.

core/lib/src/server.rs
jespersm: I don't quite follow your reasoning here. Rocket's route "format" specifier switches on exactly whether or not the method expects payload or not ( https://rocket.rs/v0.5/guide/requests/#format )
The logic here merely mirrors that same logic, and returns 406 if the route can only produce payload which the client doesn't want.
It would be wrong to return 415 for a GET route which can't return the desired content.

palant: My bad, I didn’t realize that this restriction had different meanings depending on the method. Your implementation is correct.

reneleonhardt:
@jespersm Thank you for your contribution! Is there still something missing except rebasing?

jespersm:
> @jespersm Thank you for your contribution! Is there still something missing except rebasing?

Ouch, I'm not sure, and I'm AFK for a few days, so I can't really check or rebase at the moment.

reneleonhardt:
No stress, thank you for your work when you have time 😅

jespersm:
@reneleonhardt : I have now ported my PR to the latest main, and it seems to pass the CI builds and tests: https://github.com/jespersm/Rocket/actions/runs/10726665635/job/29747224230
I added some extra checks to make the hints more precise.

reneleonhardt:
Cool, I hope your improvements get merged soon! 🚀

PR #2629, opened 2023-10-23
Upgrade 'tokio-tungstenite' to 0.20.

BlackDex:
Cherry-Pick tungstenite  update to v0.5 branch.
This ensures applications which use the contrib/ws to use a patched version which has a known CVE https://github.com/snapview/tokio-tungstenite/blob/master/CHANGELOG.md#0201

### DIFF
--- a/contrib/ws/Cargo.toml
+++ b/contrib/ws/Cargo.toml
@@ -17,7 +17,7 @@ default = ["tungstenite"]
 tungstenite = ["tokio-tungstenite"]
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


SergioBenitez:
As noted on Matrix, the vulnerability in tungstenite doesn't affect `rocket_ws` as we don't make use of its HTTP handshaking functionality.

SergioBenitez:
I just finished cherry picking every possible commit from `master` to `0.5-rc`, which included this commit. I preferred to do it this way to keep the commits in relative order.

BlackDex:
Thx!

PR #2630, opened 2023-10-26
Avoid clippy lint in `uri!` macro

isaacthefallenapple:
Avoids the `clippy::redundant_locals` lint in `uri!` invocations (see https://stackoverflow.com/q/77366480/11423104 for reference).

This happens because the macro expands to `let parameter_name = argument_name;` which triggers the lint if `parameter_name == argument_name`. The fix is a simple `allow` attribute.


SergioBenitez:
Our policy is to not work-around clippy warnings or errors in codegen by using `allow` but instead to ensure that clippy ignores our codegen code. The proper way to resolve this is to make sure hygiene at the site is labeled "mixed". I've done that in the commit that closed this PR.

Thank you for bringing this to our attention!

PR #2648, opened 2023-11-21
Modify todo example to use postgres and diesel_async

ThouCheese:


SergioBenitez:
What's the status of this? We were looking into having a postgres DB available when the CI runs to merge this. @ELD Were you working on that?

It would be nice if this example could run without a postgres DB as well, however. Perhaps we can have a dynamic switch for the database based on a profile and implement it for both backends in the same example?

ELD:
Alright, I have Postgres working on Linux and macOS now. Just down to Windows: https://github.com/ELD/Rocket/actions/runs/7353097692

ELD:
Finished off the CI work for Windows and Linux along with some `vcpkg` changes to pull in the latest versions of `libpq` for tests (and caching via `actions/cache`). We can vendor the `vcpkg` packages, but the `actions/cache` solution seemingly works well. It'll take longer, though, when certain build deps increase their version.

We just need @ThouCheese to merge the upstream PR to then get a final review in this PR :)

PR #2654, opened 2023-11-27
Make CLI colors an enum

### LINKED ISSUE #2649: Force enable color output
**Is your feature request motivated by a concrete problem? Please describe.**

When running Rocket in local development under an orchestrator like Foreman (together with other apps), the color output gets disabled because subprocesses spawned by Foreman are not real shells (no TTY). The colored output is still desired (it looks nice and helps readability), so we would like to enable it and bypass the TTY check.

**Why this feature can't or shouldn't live outside of Rocket**

Rocket controls its color output. Faking a TTY in Foreman and friends is not obvious, e.g. requires using a fake shell like `tty` in [Python](https://stackoverflow.com/questions/56835373/python-subprocess-realtime-print-with-colors-and-save-stdout). 

**Ideal Solution**

Make `cli_colors` an `Option<bool>`. If specified, `true` will always enable color output, `false` will always disable it, and `None` will default to `true` and the current TTY check.

**Alternatives Considered**

Patching Foreman/Honcho to support color output, but this quickly became a game of cat and mouse with TTY. Can't patch every single subprocess spawner out there, and this issue is not exclusive to Foreman.

**Additional Context**

Happy to contribute a PR, although I'm a bit lost in the Deserializer implementation in Figment for `Option<bool>`. Currently Figment tests aren't compiling in `master`, so a bit of a lift to get started.

levkk:
1. Make `cli_colors` an enum that accepts "auto", "always" and "never" just like cargo `--color`. Acts the same way. Backwards compatible: 0 = never, 1 = auto, false = never, true = auto. Fix #2649 
2. Ran `cargo fmt`. The diff is huge, not sure what happened...did we not run `cargo fmt` in a long time or does it accept some settings I should tweak on my side to make formatting consistent with the project? `$ cargo fmt --version
rustfmt 1.6.0-stable (d5c2e9c 2023-09-13)`


SergioBenitez:
Please don't run cargo fmt. We don't use it because it doesn't format code the way we need it to.

SergioBenitez:
Manual formatting is what we recommend.

levkk:
Thanks, appreciate your thoughts and your work on Rocket. I'll write up a PR implementing this without the formatting.

Issue #2608, opened 2023-09-09
FromForm derive macro uses removed private_in_public lint


ItsDrike:
**Description**

Using the `FromForm` derive macro causes a warning for `renamed_and_removed_lints`, explaining that the `private_in_public` lint is no longer available to appear on newer rust versions.

**To Reproduce**

Create a structure that derives `FromForm`, and compile with rust version of 1.72 (not sure exactly in which version this lint got removed, so some older versions might also produce the warning)

**Expected Behavior**

The `#[allow(private_in_public)]` should be removed, or replaced by the new lints. I'm not sure exactly what this lint is for, and what was should it be replaced with (if anything), but it would be nice to be able to avoid this warning on every struct that derives `FromForm`.

This wouldn't be as big of an issue if people could do `#[allow(renamed_and_removed_lints)]`. But putting this before the derive macro doesn't actually cause the lint to be allowed, probably since the derive macro actually puts the line with that allow elsewhere. This makes it very annoying to use `FromForm` as the warning is unavoidable.

**Environment:**

 - OS Distribution and Kernel: Arch Linux (6.4.12-arch1-1)
 - Rocket Version: 0.5.0-rc.3

SergioBenitez:
Very annoying indeed. Pushing a fix soon.

CSharperMantle:
[Rust 1.74.0](https://github.com/rust-lang/rust/releases/tag/1.74.0) has stably deprecated `private_in_public` lint with a group of two rust-lang/rfcs#2145 compatible new lints: `private_interfaces` and `private_bounds`; see rust-lang/rust#113126 for details. Thus I think it is safe for us to move to these new lints when we are compiling on `rustc > 1.74.0`.

A PR is underway.

PR #2689, opened 2024-01-04
Make '#[allow(private_bounds)]' compiler version-aware

CSharperMantle:
The current fix for #2608 as implemented in e7ef93be495793ba570f13e43772ddff895b644c addressed the problem by introducing both the new `private_bounds` lint rule and the deprecated `private_in_public` one. To smooth out the cross-version incompatibilities, @SergioBenitez added two more lint rules, namely `renamed_and_removed_lints` and `unknown_lints`, to suppress the warnings on both old and new `rustc`s.

https://github.com/rwf2/Rocket/blob/b3abc760aee8556f5b455f13aaa4a1c7ae70bd85/core/codegen/src/derive/from_form.rs#L123-L131

However, we can take one step further by utilizing `version_check` dependency. This PR proposes that the said crate checks the availability of `private_bounds` at codegen time, allowing for fewer lint rules generated in the AST.

The fix is implemented by introducing an auxiliary variable `lints`, which expands to different attributes of lint rules when compiled on different versions of `rustc`. When `rustc` is newer than `1.74.0`, it is set to the token stream of `#[allow(private_bounds)]`, utilizing the new feature; otherwise it is equal to the old sequence of `#[allow(...)]` rules. The need for `#[allow(unknown_lints)]` is eliminated.

**Pros.** Less lint rules generated while retaining the effect of the fix in e7ef93be495793ba570f13e43772ddff895b644c.

**Cons.** None known to the author.

Feel free to make comments and suggest edits to this PR! 👋

### DIFF
--- a/core/codegen/src/derive/from_form.rs
+++ b/core/codegen/src/derive/from_form.rs
@@ -120,12 +120,9 @@ pub fn derive_from_form(input: proc_macro::TokenStream) -> TokenStream {
                 let (ctxt_ty, gen) = context_type(input)?;
                 let (impl_gen, _, where_clause)  = gen.split_for_impl();
                 let output = mapper::input_default(mapper, input)?;
-                Ok(quote_spanned! { mixed(input.span()) =>
+                Ok(quote_spanned! { mixed(input.span())=>
                     /// Rocket generated FormForm context.
                     #[doc(hidden)]
-                    #[allow(unknown_lints)]
-                    #[allow(renamed_and_removed_lints)]
-                    #[allow(private_in_public)]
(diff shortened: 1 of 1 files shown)


SergioBenitez:
> When rustc is newer than 1.74.0...

As of https://github.com/rwf2/Rocket/commit/38dbab8dd36b008108aba0a7af5ecf949ae238b9, the MSRV is now 1.75. Let's do the minimal thing that works on 1.75+ since we no longer need to support older versions.

CSharperMantle:
@SergioBenitez,

Changes are made and filed.

PR #2701, opened 2024-01-13
Draft: (sync_db_pools) Postgres TLS

olback:
Another Postgres TLS PR, supersedes #2018.

This approach allows both plain text and TLS/SSL connections, with the previous PR you had to determine at compile-time if you we're connecting with TLS/SSL.

Related PR: #2605


olback:
I have tested this PR against a Digital Ocean managed PostgreSQL 15 instance, no issues as far as i can tell.

PR #2713, opened 2024-01-25
Minor guide fixes

atezet:


site/guide/9-configuration.md
```diff
@@ -25,7 +25,7 @@ values:
 | `max_blocking`*      | `usize`           | Limit on threads to start for blocking tasks.   | `512`                   |
 | `ident`              | `string`, `false` | If and how to identify via the `Server` header. | `"Rocket"`              |
 | `ip_header`          | `string`, `false` | IP header to inspect to get [client's real IP]. | `"X-Real-IP"`           |
-| `proxy_proto_header` | `string`, `false` | Header identifying [client to proxy protocol].  | `None`                  |
```
SergioBenitez, line 28: The default is actually `None`, it's just that you can't set that via TOML.

In general, with the config docs, there's a difficulty in expressing the concepts in such a way that they're not too broad (i.e, too conceptual to the point of being inapplicable) but also not too specific (i.e, they don't obfuscate the concepts entirely and narrow applicability to a single instance). I'd really love to redo them entirely, though I'm not sure what the best approach _would_ be.

arjentz: I see. I honestly don't think the docs are particularly bad right now (the contrary), I was (apparently) just confused by this specific setting.

atezet: Noticed this was still open so rebased the branch, removed the mistaken and kept the tiny fix that was left :)

PR #2716, opened 2024-01-30
Allow for optionally serving pre-compressed files.

hcldan:
There wasn't a good way to propagate the new option into `NamedFile` so I made a new constructor and added a flag to the struct. In order to make the flag understandable, I converted to a struct with members instead of a simple tuple.

I am not sure how to test the `Content-Type` returned...  It seems that the original tests didn't check those headers and there's not really a clean way to pass that in the assert functions without refactoring a bunch of stuff.


hcldan:
One approach I tried first was to implement a wrapper for FileServer and implement a Handler that would defer to FileServer and adjust headers on the way out.

I could not figure out how to properly clone the Request in order to get a mutable value I could `set_uri()` on so I could get it to try to read the alternate file and check to see if it was found. And while something like that would have worked, if I found a way to modify the request, it's still rather kluge-y as the request *wasn't* for the gz file, it was for the *non gz* file and should be using the gz file to serve the request for the non-gz file with appropriate headers.

It makes it complicated to tamper with the request like that and have a proper response. I think it's better to make the resource-check by looking at the file system and choosing what file to send back according to the actual request.

hcldan:
closes #2718

core/lib/tests/file_server.rs
```diff
@@ -39,29 +46,44 @@ static INDEXED_DIRECTORIES: &[&str] = &[
...
-        let mut expected_contents = String::new();
-        file.read_to_string(&mut expected_contents).expect("read file");
-        assert_eq!(response.into_string(), Some(expected_contents));
+        let mut expected_contents = vec![];
+        file.read_to_end(&mut expected_contents).expect("read file");
+
+        let mut actual = vec![];
+        response.read_to_end(&mut actual).expect("read response");
+
+        let ce: Vec<&str> = response.headers().get("Content-Encoding").collect();
```
ApprenticeofEnder, line 72: Would it make sense to check the Content-Type header similar to how you have here, and have a pre-defined header for GZipped files, and then other stuff otherwise?

core/lib/src/fs/server.rs
```diff
@@ -237,8 +237,25 @@ impl Handler for FileServer {
...
+                let gzip_accepted = req.headers().get("Accept-Encoding")
+                    .find(|value| value.contains("gzip"))
+                    .is_some();
+                let check_compressed = gzip_accepted && options.contains(Options::CheckCompressed);
+                if check_compressed {
+                    if let Some(file) = p.file_name() {
+                        let mut compressed = file.to_os_string();
+                        compressed.push(".gz");
+                        let compressed_file = p.with_file_name(compressed);
+                        if compressed_file.exists() {
```
ApprenticeofEnder, line 250: Could we use magic bytes here instead of just relying on file extensions? Could save a lot of headache if the "compressed" files are encoded wrong.

ApprenticeofEnder:
Final question: Could we also perhaps have some way of providing file hashes with the server? That might be a separate PR, though it's a good thing to think about, especially with compressed files like this.

core/lib/tests/file_server.rs
hcldan: Non zipped files would not have a [content-encoding header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding) and I was only implementing gzip, so expanded values would need to be checked in the future if adding support for other compression schemes.

core/lib/src/fs/server.rs
hcldan: True, though that's more disk access than just seeing if the file exists.
And I've personally seen a long history of this technique (same file name with .gz extension) for pre-gzipped content.

I figured it would be fine since we are using the extension for the mime types of the file already.

hcldan:
Hashing the files would be nice for other reasons, like you say, but it would also introduce performance overhead. I think that's best left to another PR to implement features specific to hashing needs.

I also did not implement anything related to cache control headers, as there's a fairly long conversation in other issues and I've not seen much movement on it.

core/lib/src/fs/server.rs
ApprenticeofEnder: That's fair, I'm very used to file extensions potentially not lining up with their content (working in security), but if it's the standard and hasn't caused problems so far, no harm no foul!

hcldan:
@SergioBenitez Is this something that the Rocket team would like to merge? How do I proceed here?

ApprenticeofEnder:
> @ApprenticeofEnder Is this something that the Rocket team would like to merge? How do I proceed here?

Sorry, unfortunately I'm not on the Rocket team! Just thought I'd start some conversation surrounding your PR.

SergioBenitez:
> How do I proceed here?

Just as you are!

> @SergioBenitez Is this something that the Rocket team would like to merge? 

Potentially. It's an interesting idea. As I see it, here are the pros and cons:

- The feature asumes that foo.gz is indeed a gzipped version of foo. This may not be the case. It seems trivially simple for foo.gz to be an outdated version of foo, for example.
- This adds somewhat complicated logic to the file server, which means the file server is doing things that are more interesting than just serving files.
- This is easily accomplished outside of Rocket with a handler that handles file serving. 

The pros are:

- The logic seems simple enough and, as long as it's opt-in, shouldn't be error-prone.
- This is a convenient way to resolve a requested feature. 

In general I think I'm in favor of the idea with a different implementation. There are a few key things I would like to see changes.

1. I don't think we need to make any changes to `NamedFile`, nor do I support making those changes. If we want a different responder, we define a different one. The impl can be easily derived, for instance, for a structure that includes the file with the appropriate fields or attributes for headers. 
2. As mentioned before, it seems far too simple to serve the incorrect file. I believe we need to add some sort of consistency checking to make this viable. At minimum, we should require that the gz file has a mod time greater than the original file. Ideally however, we would be able to check that the gz file does indeed correspond to the regular file. One approach, as an example, would be to require the gz file name to include a hash of the original file. That is, foo.HASH.gz, where HASH is the hash of foo. We would then compute the hash of the original file and compare the hashes, serving the gz file only if they match. Ideally we'd only need to do this once per file, caching the results of the check for the future. 
3. It feels that there is a more generalizable mechanism at play, one that can not only be used to implement this feature, but other existing features, such as index file serving. What I'm suggesting is effectively mod_rewrite, albeit a very constrained version. If we allow `FileServer` to be constructed with dynamic "rewrite" functions in which the user (or us internally) is able to tell us which file, if any, should be served for the path identified by the request, then this can be implemented outside of Rocket entirely. This effectively makes FileServer pluggable. We would seek an API by which we can implement most if not all options currently in FileServer. Ideally, the API would also make composition and precedence  possible and obvious. For instance, should foo/index.html.gz be served for foo/ if both the Index and Compressed options are enabled? Ideally we'd have an API that would allow the user to create a FileServer that can answer the question in either direction.

If you choose to proceed, I would first like to see a sketch of an API for 3 that would handle most of lot all of the available options. The "sketch" should show that each option for FileServer could be implemented, including the one in question, and how they  compose. Ideally the implementation for each option would be rather straightforward and consist only of the bare-minimal logic necessary. For example, the implementation of the Index option should be something as close to as simple as:

```rust
path.dir().map(|dir| dir.join("index.html"))
```

The DotFiles option as simple as:

```rust
(allow_dotfiles || !path.starts_with('.')).then(path)
```

Then if each do these rewrites is nameable in some way, the default FileServer becomes a specific composition of these rewrites:

```rust
FileServer::new()
    .rewrite(Jail("/foo")) # rewrite /req/path to /foo/req/path.
    .rewrite(DotFiles(true)) # allow DotFiles
    .rewrite(IndexFile) # rewrite /foo or /foo/ to /foo/index.html
```

And new ones can be added:


```rust
FileServer::new()
    .rewrite(Jail("/foo"))
    .rewrite(DotFiles(true))
    .rewrite(IndexFile)
    .rewrite(PreCompressed)
```

Note how the previous ambiguity is removed regarding ordering is removed and both options (check for compressed before or after index.html rewriting) are easily achievable. 

I would be in favor of such a clean-up PR.

hcldan:
@SergioBenitez I removed changes to NamedFile which should address (1)

See my response above for (2) and (3)

core/lib/src/fs/maybe_zipped_file.rs
```diff
@@ -0,0 +1,110 @@
...
+///     MaybeZippedFile::open(path).await.ok()
+/// }
+/// ```
+///
+/// Always prefer to use [`FileServer`] which has more functionality and a
+/// pithier API.
+///
+/// [`FileServer`]: crate::fs::FileServer
+#[derive(Debug)]
+pub(crate) struct MaybeZippedFile {
```
palant, line 40: It may be a good idea to change `MaybeZippedFile` into something more generic like `MaybeCompressedFile`. Reason being: for example tower-http which is used by Axum supports gzip, br and deflate content encodings (see https://github.com/tower-rs/tower-http/pull/156). While deflate should be irrelevant, adding support for Brotli compression might be desirable in future.

core/lib/src/fs/maybe_zipped_file.rs
```diff
@@ -0,0 +1,110 @@
...
+            Some(e) if e == "gz" => (Some(e.to_owned()), false, NamedFile::open(path).await?),
+            // construct path to the .gz file
+            Some(e) => {
+                let bare = o_path.with_extension("");
+
+                let ct_ext = Some(e.to_owned());
+                let mut zip_ext = e.to_owned();
+                zip_ext.push(".gz");
+
+                let zipped = bare.with_extension(&zip_ext);
```
palant, line 68: I think there is a bug here. If the requested file is something like `index.something.html`, then `bare` will become `index.something`. Calling `with_extension("html.gz)` on this will remove the file extension once again and result in `index.html.gz`.

I think that you can just call `o_path.with_extension(&zip_ext)` here, `bare` is unnecessary.

core/lib/src/fs/maybe_zipped_file.rs
```diff
@@ -0,0 +1,110 @@
...
+                }
+            }
+            // construct path to the .gz file
+            None => {
+                let zipped = o_path.with_extension("gz");
+                match zipped.exists() {
+                    true  => (None, true, NamedFile::open(zipped).await?),
+                    false => (None, false, NamedFile::open(path).await?),
+                }
+            }
```
palant, line 81: It looks like the `Some(e)` and `None` branches could be consolidated. Something along the lines of:

```rust
ct_ext => {
  let zip_ext = ct_ext.map(|e| {
    let mut e = e.to_owned();
    e.push(".gz");
    e
  }).unwrap_or(OsString::from(String::new("gz"));
  let zipped = o_path.with_extension(zip_ext);
  …
}
```

core/lib/src/fs/server.rs
```diff
@@ -218,28 +218,35 @@ impl Handler for FileServer {
...
                 }
 
-                let index = NamedFile::open(p.join("index.html")).await;
-                index.respond_to(req).or_forward((data, Status::NotFound))
-            },
-            Some(p) => {
-                let file = NamedFile::open(p).await;
-                file.respond_to(req).or_forward((data, Status::NotFound))
+                let gzip_accepted = req.headers().get("Accept-Encoding")
+                    .find(|value| value.contains("gzip"))
```
palant, line 242: I’m not sure whether a substring match here is a good idea. See https://github.com/tower-rs/tower-http/pull/156/files#diff-83cd763343b90a262fdc4334e5289daea2e9cfb9b1cf76785ea7c42b66d3d5a0R85 for proper parsing code.

palant (commented):
I’m not the maintainer of this repository, so I’m merely nit-picking which you are free to ignore.

core/lib/src/fs/maybe_zipped_file.rs
hcldan: Oh! Good catch. I'll give that a whirl.

core/lib/src/fs/maybe_zipped_file.rs
palant: I’m perfectly fine with Gzip only. Brotli support is something for the future.

core/lib/src/fs/server.rs
hcldan: It depends on what you mean by "good idea" :)

From reading the spec for the header, this should cover what is needed for this change (gzip only). The header doesn't let you say "please not gzip", so if they mention it, then they support it (even if it is not preferred).

It is true that this is not ideal for future multiple compression types.
Especially if we wish to honor the browser's preference order.

osmano807: Per RFC 7231, section 5.3.4. Accept-Encoding

> 2.  If the representation has no content-coding, then it is
       acceptable by default unless specifically excluded by the
       Accept-Encoding field stating either "identity;q=0" or "*;q=0"
       without a more specific entry for "identity".

>   3.  If the representation's content-coding is one of the
       content-codings listed in the Accept-Encoding field, then it is
       acceptable unless it is accompanied by a qvalue of 0.  (As
       defined in [Section 5.3.1](https://datatracker.ietf.org/doc/html/rfc7231#section-5.3.1), a qvalue of 0 means "not acceptable".)
       
 So if it specifies `Accept-Encoding: gzip;q=0` it's *denying* gzip

hcldan: Right you are, I glossed right over that.  I'll see if I can implement something that makes more sense than "contains".

palant: Have a look here: https://github.com/rwf2/Rocket/blob/a866134212a13bfaa61a1cd1e68880461af1ab42/core/http/src/parse/media_type.rs. That’s a very similar logic, and this directory should also be the appropriate place to add this parsing code.

hcldan: This was considerably more work than I had hoped.

hcldan: @palant I would gladly take any additional unit tests, my time I can spend on this project is running short.

palant: @hcldan Unfortunately, it’s the same situation here. Unless I see some PRs being accepted in a meaningful form, I have no more time left to contribute to this project.

PR #2732, opened 2024-02-20
Fix Request::client_ip for TLS connections

palant:
`Request::client_ip` method currently always returns `None` on TLS connections if no “real ip” header is configured/present, a regression compared to Rocket 0.5. That’s because the fallback assumes `Request::remote` to return a TCP endpoint and won’t handle TLS endpoints correctly.

I considered changing `Endpoint::tcp` (and `Endpoint::unix`) method to automatically recurse for TLS endpoints instead but I doubt this is how the code logic there is supposed to be.

### DIFF
--- a/core/lib/src/request/request.rs
+++ b/core/lib/src/request/request.rs
@@ -3,7 +3,7 @@ use std::ops::RangeFrom;
 use std::sync::{Arc, atomic::Ordering};
 use std::borrow::Cow;
 use std::future::Future;
-use std::net::IpAddr;
+use std::net::{IpAddr, SocketAddr};
 
 use yansi::Paint;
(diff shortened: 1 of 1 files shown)


SergioBenitez:
Note that this is already fixed more generally on the h3 branch. I'll work on getting that branch merged into mainline later today.

palant:
I see. That’s the change I meant here:

> I considered changing Endpoint::tcp (and Endpoint::unix) method to automatically recurse for TLS endpoints instead but I doubt this is how the code logic there is supposed to be.

Guess it was supposed to be this way after all. :smile:

SergioBenitez:
This is now on master.

Issue #2731, opened 2024-02-19
Allow routes that match any method


palant:
### What's missing?

There are some scenarios which currently require cloning a route multiple times to make sure each and every method is accepted. For me these scenarios are at the very least:

* HTTP -> HTTPS redirector
* Proxying requests to another server (not redirecting because it isn’t accessible from the web)
* Internal redirects (modifying a request and dispatching it, making sure it’s processed elsewhere in the app)

### Ideal Solution

From the look of it, the following changes should suffice:

* Change type of `Route::method` into `MethodSet`, an alias for `enumset<Method>`. This should allow most of the existing code to work without changes.
* Add `#[any(…)]` attribute macro – essentially identical to `#[get(…)]` but setting `Route::method` to `MethodSet::all()`.
* Modify `#[route(…)]` attribute macro to accept multiple methods: `#[route(GET | POST, …)]`.
* Modify `#[route(…)]` attribute macro to accept the special keyword `ANY` as method parameter meaning `MethodSet::all()`. An alternative approach would making the method optional here and default to `MethodSet::all()` but personally I’d rather have things explicit.
* Modify `Route::matches()` implementation, this change is trivial. From what I can tell, Rocket doesn’t optimize route matching, so the only changes required otherwise should be some logging code.

I should be able to create a PR if this approach is approved.

### Why can't this be implemented outside of Rocket?

Creating attribute macros to mimic those built into Rocket would require considerable complexity (duplication of logic). Also, it would still require creating an excessive number of routes in this scenario.

### Are there workarounds usable today?

This limitation can certainly be worked around, see redirector in the TLS example in this repository. The work-around there is heavy on boilerplate however instead of using the usual descriptive approach of Rocket.

My work-around does slightly better:

```rust
#[rocket::get("/<_..>")]
async fn redirect(
    config: &State<Config>,
    host: &Host<'_>,
    uri: &Origin<'_>,
) -> Result<Redirect, Status> {
    …
}

let redirects = [Get, Put, Post, Delete, Options, Head, Trace, Connect, Patch]
    .into_iter()
    .map(|m| {
        let mut route: Route = rocket::routes![redirect].remove(0);
        route.method = m;
        route
    })
    .collect::<Vec<_>>();
```

However, this still creates nine routes instead of one. If nothing else, this has an impact on performance.

[...]

PR #2734, opened 2024-02-22
Add support for routes matching any method

palant:
This implements the change proposed in #2731.

Note that technically this allows invalid requests with `Method::Any` to be created manually. Such requests merely won’t be created automatically by either `Request::from_hyp` or `Rocket::preprocess`. I don’t consider that an issue but your mileage might vary.

### DIFF
--- a/core/http/src/method.rs
+++ b/core/http/src/method.rs
@@ -47,7 +47,9 @@ pub enum Method {
     /// The `CONNECT` variant.
     Connect,
     /// The `PATCH` variant.
-    Patch
+    Patch,
+    /// A wildcard variant meant to accept any method
+    Any = u8::MAX,
 }
 
 impl Method {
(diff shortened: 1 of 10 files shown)


core/http/src/method.rs
```diff
@@ -47,7 +47,9 @@ pub enum Method {
     /// The `CONNECT` variant.
     Connect,
     /// The `PATCH` variant.
-    Patch
+    Patch,
+    /// A wildcard variant meant to accept any method
+    Any = u8::MAX,
```
SergioBenitez, line 52: This breaks the `Method` enum in that `Any` is not actually a method. We need to find another way to represent "any". My recommendation is to use an `Option<Method>` for the route's method, where `Some(Method)` indicates it matches a (single) method, and `None` indicates it matches all methods.

SergioBenitez (changes requested):
Okay, I'm in favor of the idea in general, but I think we need to take a different approach to the implementation. Particularly, modifying `Method` to include a variant that is not, in fact, a method, is problematic for quite a variety of reasons. This PR shouldn't touch `Method` directly. The changes should be restricted to routing.

I think the most reasonable approach is to modify the `Route::method` field to be an `Option<Method>` where `None` represents "matches any method".  This is equivalent to how `format` is handled, for example, and has semantic meaning. Everything else should follow from there.

SergioBenitez:
Closing due to inactivity.

PR #2751, opened 2024-03-18
Update websocket.rs

gaoyia:
return accept_key


SergioBenitez:
Why do you need the `Sec-WebSocket-Key`? I'm not opposed to exposing it, as there is no security implication, but the key should be a likely protocol-level concern, so it's curious that you would need the value.

gaoyia:
I personally prefer using it as a connection ID and manually maintaining the uniqueness of Sec-WebSocket-Key.

I also want to know how to get the "Request" in WebSocket. 

thanks a lot.

SergioBenitez:
Merged.

PR #2753, opened 2024-03-21
Fix `Content-Length` header

bash:


SergioBenitez:
So much for consts helping us avoid typos.

Can you add a test to make sure this never happens again?

SergioBenitez:
Adding the test and merging this now.

SergioBenitez:
Thanks for catching and fixing this. Merged in 35a1cf12b6b2f2cf4c699153f4f7938ca289bd7d.

bash:
Wow, thanks for merging this so promptly! (I should really stop depending on master 😆)

I originally wanted to write a test, but ran into the second problem that you fixed in your commit, namely that the header is not visible to local clients.

PR #2764, opened 2024-03-31
Add licenses to crates

### LINKED ISSUE #2760: Missing license files in rocket_codegen-0.5.0.crate
### What kind of documentation problem are you reporting?

Other

### Where is the issue found?

https://crates.io/crates/rocket_codegen/0.5.0

### What's wrong?

[LICENSE-MIT](https://github.com/rwf2/Rocket/blob/master/LICENSE-MIT) and [LICENSE-APACHE](https://github.com/rwf2/Rocket/blob/master/LICENSE-APACHE) is missing in rocket_codegen-0.5.0.crate downloaded from https://crates.io/crates/rocket_codegen/0.5.0 Please add it to all your packages. I would like to package it as rpm for Fedora (which identifies this as a [mistake](https://docs.fedoraproject.org/en-US/packaging-guidelines/LicensingGuidelines/#_license_text) and encourages upstream to include it) and it would make it easier. Thanks.

### System Checks

- [X] I confirmed that the issue still exists on `master` on GitHub.
- [X] I was unable to find a previous report of this problem.

wojnilowicz:
fixes #2760


SergioBenitez:
This should be done for every crate, not just the core crates.

wojnilowicz:
> This should be done for every publishable crate, not just the core crates.

Are they all now or is something missing still? Please let me know.

SergioBenitez:
> > This should be done for every publishable crate, not just the core crates.
> 
> Are they all now or is something missing still? Please let me know.

I think we're good. Waiting for a fix to https://github.com/rust-lang/rust/issues/123573#issuecomment-2042359434, which is currently breaking the build on nightly, to push this.

PR #2784, opened 2024-04-28
Minijinja2: Improve error reporting

michaelmelanson:
This change improves the error reporting for Minijinja2 templates.

Currently, when errors occur in a Minijinja2 template, it usually gives an error message that only has a top-level explanation of the problem. When the issue comes from an included file or a layout, this error message does not include enough details to diagnose the problem when the problem.

With this change, it now prints details about the entire error chain returned, which acts as a stack trace to find the cause of the error.

It also adds the error message on 'template not found' errors.


SergioBenitez:
Merged in https://github.com/rwf2/Rocket/commit/347d9da663f1184b26b956a25c48e0dadb87cdcb. Thank you!

PR #2789, opened 2024-05-08
Implement rewrite API

the10thWiz:
The goal of this PR is to implement a Rewrite API, following the basic outline presented in #2716.

The goal is to replace the existing `Options` on `FileServer` to instead use a rewrite based API. This will make `FileServer` pluggable, making it possible to implement things like cached compression implementable outside of Rocket.

The basic usage example is as follows:
```rust
FileServer::new("static")
   .rewrite(DotFiles)
   .rewrite(NormalizeDirs)
   .rewrite(Index("index.html"))
```

There are a couple of major limitations to the rewrite trait, specifically the file to be served must exist on the local file system, and the trait is synchronous, so it should not do any IO when rewriting paths. However, I would argue that if you want to move beyond these, you should just implement your own `Handler`, and fully replace `FileServer`.

### DIFF
--- a/core/http/src/uri/segments.rs
+++ b/core/http/src/uri/segments.rs
@@ -218,6 +218,8 @@ impl<'a> Segments<'a, Path> {
         for segment in self.clone() {
             if segment == ".." {
                 buf.pop();
+            } else if segment == "." {
+                continue;
             } else if !allow_dotfiles && segment.starts_with('.') {
                 return Err(PathError::BadStart('.'))
             } else if segment.starts_with('*') {
--- a/core/http/src/uri/segments.rs
+++ b/core/http/src/uri/segments.rs
@@ -178,8 +178,9 @@ impl<'a> Segments<'a, Path> {
     }
 
     /// Creates a `PathBuf` from `self`. The returned `PathBuf` is
-    /// percent-decoded. If a segment is equal to `..`, the previous segment (if
-    /// any) is skipped.
+    /// percent-decoded and guaranteed to be relative. If a segment is equal to
+    /// `.`, it is skipped. If a segment is equal to `..`, the previous segment
+    /// (if any) is skipped.
     ///
     /// For security purposes, if a segment meets any of the following
     /// conditions, an `Err` is returned indicating the condition met:
@@ -193,7 +194,7 @@ impl<'a> Segments<'a, Path> {
     /// Additionally, if `allow_dotfiles` is `false`, an `Err` is returned if
     /// the following condition is met:
     ///
-    ///   * Decoded segment starts with any of: `.` (except `..`)
+    ///   * Decoded segment starts with any of: `.` (except `..` and `.`)
     ///
     /// As a result of these conditions, a `PathBuf` derived via `FromSegments`
     /// is safe to interpolate within, or use as a suffix of, a path without
@@ -216,10 +217,10 @@ impl<'a> Segments<'a, Path> {
     pub fn to_path_buf(&self, allow_dotfiles: bool) -> Result<PathBuf, PathError> {
         let mut buf = PathBuf::new();
         for segment in self.clone() {
-            if segment == ".." {
-                buf.pop();
-            } else if segment == "." {
+            if segment == "." {
                 continue;
+            } else if segment == ".." {
+                buf.pop();
             } else if !allow_dotfiles && segment.starts_with('.') {
                 return Err(PathError::BadStart('.'))
             } else if segment.starts_with('*') {
--- a/core/http/src/uri/segments.rs
+++ b/core/http/src/uri/segments.rs
@@ -245,39 +245,6 @@ impl<'a> Segments<'a, Path> {
 
         Ok(buf)
     }
-
-    /// Similar to `to_path_buf`, but always allows dotfiles, and reports whether
-    /// the path contains dotfiles.
-    pub fn to_path_buf_dotfiles(&self) -> Result<(PathBuf, bool), PathError> {
-        let mut buf = PathBuf::new();
-        let mut is_dotfile = false;
-        for segment in self.clone() {
-            if segment == ".." {
-                buf.pop();
-            } else if segment.starts_with('.') {
-                buf.push(segment);
-                is_dotfile = true;
-            } else if segment.starts_with('*') {
-                return Err(PathError::BadStart('*'))
-            } else if segment.ends_with(':') {
-                return Err(PathError::BadEnd(':'))
-            } else if segment.ends_with('>') {
-                return Err(PathError::BadEnd('>'))
-            } else if segment.ends_with('<') {
-                return Err(PathError::BadEnd('<'))
-            } else if segment.contains('/') {
-                return Err(PathError::BadChar('/'))
-            } else if cfg!(windows) && segment.contains('\\') {
-                return Err(PathError::BadChar('\\'))
-            } else if cfg!(windows) && segment.contains(':') {
-                return Err(PathError::BadChar(':'))
-            } else {
-                buf.push(segment)
-            }
-        }
-
-        Ok((buf, is_dotfile))
-    }
 }
 
 impl<'a> Segments<'a, Query> {
--- a/core/http/src/uri/segments.rs
+++ b/core/http/src/uri/segments.rs
@@ -245,6 +245,39 @@ impl<'a> Segments<'a, Path> {
 
         Ok(buf)
     }
+
+    /// Similar to `to_path_buf`, but always allows dotfiles, and reports whether
+    /// the path contains dotfiles.
+    pub fn to_path_buf_dotfiles(&self) -> Result<(PathBuf, bool), PathError> {
+        let mut buf = PathBuf::new();
+        let mut is_dotfile = false;
+        for segment in self.clone() {
+            if segment == ".." {
+                buf.pop();
+            } else if segment.starts_with('.') {
+                buf.push(segment);
+                is_dotfile = true;
+            } else if segment.starts_with('*') {
+                return Err(PathError::BadStart('*'))
+            } else if segment.ends_with(':') {
+                return Err(PathError::BadEnd(':'))
+            } else if segment.ends_with('>') {
+                return Err(PathError::BadEnd('>'))
+            } else if segment.ends_with('<') {
+                return Err(PathError::BadEnd('<'))
+            } else if segment.contains('/') {
+                return Err(PathError::BadChar('/'))
+            } else if cfg!(windows) && segment.contains('\\') {
+                return Err(PathError::BadChar('\\'))
+            } else if cfg!(windows) && segment.contains(':') {
+                return Err(PathError::BadChar(':'))
+            } else {
+                buf.push(segment)
+            }
+        }
+
+        Ok((buf, is_dotfile))
+    }
 }
 
 impl<'a> Segments<'a, Query> {
--- a/core/lib/src/fs/mod.rs
+++ b/core/lib/src/fs/mod.rs
@@ -34,7 +34,7 @@ crate::export! {
     ///
     /// #[launch]
     /// fn rocket() -> _ {
-    ///     rocket::build().mount("/", FileServer::from(relative!("static")))
+    ///     rocket::build().mount("/", FileServer::new(relative!("static")))
     /// }
     /// ```
     ///
--- a/core/lib/src/fs/mod.rs
+++ b/core/lib/src/fs/mod.rs
@@ -11,7 +11,6 @@ pub use server::*;
 pub use named_file::*;
 pub use temp_file::*;
 pub use file_name::*;
-pub use rewrite::{Rewrite, Rewriter};
 
 crate::export! {
     /// Generates a crate-relative version of a path.
--- a/core/lib/src/fs/mod.rs
+++ b/core/lib/src/fs/mod.rs
@@ -5,6 +5,8 @@ mod named_file;
 mod temp_file;
 mod file_name;
 
+pub mod rewrite;
+
 pub use server::*;
 pub use named_file::*;
 pub use temp_file::*;
--- a/core/lib/src/fs/mod.rs
+++ b/core/lib/src/fs/mod.rs
@@ -11,6 +11,7 @@ pub use server::*;
 pub use named_file::*;
 pub use temp_file::*;
 pub use file_name::*;
+pub use rewrite::{Rewrite, Rewriter};
 
 crate::export! {
     /// Generates a crate-relative version of a path.
--- a/core/lib/src/fs/mod.rs
+++ b/core/lib/src/fs/mod.rs
@@ -11,4 +11,53 @@ pub use server::*;
 pub use named_file::*;
 pub use temp_file::*;
 pub use file_name::*;
-pub use server::relative;
+
+crate::export! {
+    /// Generates a crate-relative version of a path.
+    ///
+    /// This macro is primarily intended for use with [`FileServer`] to serve
+    /// files from a path relative to the crate root.
+    ///
+    /// The macro accepts one parameter, `$path`, an absolute or (preferably)
+    /// relative path. It returns a path as an `&'static str` prefixed with the
+    /// path to the crate root. Use `Path::new(relative!($path))` to retrieve an
+    /// `&'static Path`.
+    ///
+    /// # Example
+    ///
+    /// Serve files from the crate-relative `static/` directory:
+    ///
+    /// ```rust
+    /// # #[macro_use] extern crate rocket;
+    /// use rocket::fs::{FileServer, relative};
+    ///
+    /// #[launch]
+    /// fn rocket() -> _ {
+    ///     rocket::build().mount("/", FileServer::from(relative!("static")))
+    /// }
+    /// ```
+    ///
+    /// Path equivalences:
+    ///
+    /// ```rust
+    /// use std::path::Path;
+    ///
+    /// use rocket::fs::relative;
+    ///
+    /// let manual = Path::new(env!("CARGO_MANIFEST_DIR")).join("static");
+    /// let automatic_1 = Path::new(relative!("static"));
+    /// let automatic_2 = Path::new(relative!("/static"));
+    /// assert_eq!(manual, automatic_1);
+    /// assert_eq!(automatic_1, automatic_2);
+    /// ```
+    ///
+    macro_rules! relative {
+        ($path:expr) => {
+            if cfg!(windows) {
+                concat!(env!("CARGO_MANIFEST_DIR"), "\\", $path)
+            } else {
+                concat!(env!("CARGO_MANIFEST_DIR"), "/", $path)
+            }
+        };
+    }
+}
--- a/core/lib/src/fs/rewrite.rs
+++ b/core/lib/src/fs/rewrite.rs
@@ -1,0 +1,329 @@
+use std::borrow::Cow;
+use std::path::Path;
+
+use crate::{Request, Response};
+use crate::http::{HeaderMap, ContentType};
+use crate::http::ext::IntoOwned;
+use crate::response::{self, Redirect, Responder};
+
+/// Trait used to implement [`FileServer`] customization.
+///
+/// Conceptually, a [`FileServer`] is a sequence of `Rewriter`s, which transform
+/// a path from a request to a final response. [`FileServer`] add a set of default
+/// `Rewriter`s, which filter out dotfiles, apply a root path, normalize directories,
+/// and use `index.html`.
+///
+/// After running the chain of `Rewriter`s,
+/// [`FileServer`] uses the final [`Option<Rewrite>`](Rewrite)
+/// to respond to the request. If the response is `None`, a path that doesn't
+/// exist or a directory path, [`FileServer`] will respond with a
+/// [`Status::NotFound`](crate::http::Status::NotFound). Otherwise the [`FileServer`]
+/// will respond with a redirect or the contents of the file specified.
+///
+/// [`FileServer`] provides several helper methods to add `Rewriter`s:
+/// - [`FileServer::rewrite()`]
+/// - [`FileServer::filter()`]
+/// - [`FileServer::map()`]
+pub trait Rewriter: Send + Sync + 'static {
+    /// Alter the [`Rewrite`] as needed.
+    fn rewrite<'r>(&self, file: Option<Rewrite<'r>>, req: &'r Request<'_>) -> Option<Rewrite<'r>>;
+}
+
+/// A Response from a [`FileServer`]
+#[derive(Debug, Clone)]
+#[non_exhaustive]
+pub enum Rewrite<'r> {
+    /// Return the contents of the specified file.
+    File(File<'r>),
+    /// Returns a Redirect.
+    Redirect(Redirect),
+}
+
+/// A File response from a [`FileServer`]
+#[derive(Debug, Clone)]
+pub struct File<'r> {
+    /// The path to the file that [`FileServer`] will respond with.
+    pub path: Cow<'r, Path>,
+    /// A list of headers to be added to the generated response.
+    pub headers: HeaderMap<'r>,
+}
+
+impl<'r> Rewrite<'r> {
+    pub fn file(&self) -> Option<&File<'r>> {
+        match self {
+            Rewrite::File(f) => Some(f),
+            _ => None,
+        }
+    }
+
+    pub fn redirect(&self) -> Option<&Redirect> {
+        match self {
+            Rewrite::Redirect(r) => Some(r),
+            _ => None,
+        }
+    }
+}
+
+impl<'r> From<File<'r>> for Rewrite<'r> {
+    fn from(value: File<'r>) -> Self {
+        Self::File(value)
+    }
+}
+
+impl<'r> From<Redirect> for Rewrite<'r> {
+    fn from(value: Redirect) -> Self {
+        Self::Redirect(value)
+    }
+}
+
+impl<'r> File<'r> {
+    pub fn new(path: impl Into<Cow<'r, Path>>) -> Self {
+        Self { path: path.into(), headers: HeaderMap::new() }
+    }
+
+    pub(crate) async fn open(self) -> std::io::Result<NamedFile<'r>> {
+        let file = tokio::fs::File::open(&self.path).await?;
+        let metadata = file.metadata().await?;
+        if metadata.is_dir() {
+            return Err(std::io::Error::new(std::io::ErrorKind::Other, "is a directory"));
+        }
+
+        Ok(NamedFile {
+            file,
+            len: metadata.len(),
+            path: self.path,
+            headers: self.headers,
+        })
+    }
+
+    /// Replace the path of this `File`.
+    pub fn map_path<F, P>(self, f: F) -> Self
+        where F: FnOnce(Cow<'r, Path>) -> P,
+              P: Into<Cow<'r, Path>>,
+    {
+        Self {
+            path: f(self.path).into(),
+            headers: self.headers,
+        }
+    }
+}
+
+impl<F: Send + Sync + 'static> Rewriter for F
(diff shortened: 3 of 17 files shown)


the10thWiz:
I've implemented a basic gzip compression rewriter, which provides an API similar to what was discussed in #2716.

https://github.com/the10thWiz/rocket-caching-layer

This particular implementation generates the compressed version of the file once it's been requested once, and responds with the uncompressed version in the mean time. It's not too hard to imagine different implementations, with different trade-offs. I built this as a proof-of-concept, to show that the new API actually worked for the intended use-case.

SergioBenitez:
I'm about to review this. Note that I tried to implement this API myself some time ago and ran into a bunch of issues. Here's my attempt, in case it helps with anything: https://paste.rs/wEcVu.rs

Hopefully you've manage to overcome the limitations I found!

the10thWiz:
@SergioBenitez 
> Hopefully you've manage to overcome the limitations I found!

I'm not sure what the exact limitations you came across, but here are a few things I've done differently:

- Rewrites can only produce the content of a file on disk. This simplifies the API somewhat, since a rewrite can only take one of a few specific actions. This also allows further modification in pretty much every case, so every rewrite is always applied in every case.
- I've implemented `DotFiles` differently. Rather than adding a rewrite that either allows or disallows them, I have set up the `FileServer` to output a `Hidden` response, that the `DotFiles` rewrite converts to a normal `File` response.
- Rather than using `Cow<Path>`, I'm using `PathBuf`. I'm pretty sure we always need to allocate the path, so we wouldn't gain much from `Cow<Path>`, but I'm open to switching if we can potentially remove some allocations.

There are a few points we need to decide on before this implementation can be finalized:
- Is calling `.is_dir()` (internally `std::fs::metadata`) acceptable in an async environment? There is an async tokio version, because this is an IO call, but the current implementation uses `.is_dir()` internally already. My current version also uses `.is_dir()`, and I'm not sure if this is a big deal or not.
- How do we want to handle `Missing` and `IndexFile`. These are an interesting case, because their primary function is a check on startup. We could extend the `Rewrite` API to include a startup check (and default to passing), but `Missing` would actually want us to ignore an error. For now, I've included `IndexFile` (which does nothing), and excluded `Missing` because I don't have a way to implement it.
- How do we want to roll this out? My current implementation is technically backwards compatible, and outputs a deprecation warning. If we intend to introduce this as part of 0.6, we could make this a breaking change, but I'd to know your thoughts before I go in on either option.
- Should we include a compression rewrite? I've created one (see the link in a previous comment) outside of Rocket, but compression is such a common requirement we should consider including it either in core, or as a contrib crate.

SergioBenitez:
The entire patch for my rewrite API implementation is here: https://paste.rs/AbR41.patch

I'll be referring to it in my review.

SergioBenitez (commented):
The primary difference I see between this API and the one I implemented, at a more fundamental level, is the degree to which it seeks to make `FileServer` completely generic. In my implementation, `FileServer` becomes purely a sequence of rewrites:

```diff
+pub struct FileServer {
+    rewriters: Vec<Arc<dyn Rewriter>>,
+}
```

Whereas here, we still have the notion of a root and some options. I think to consider this a successful API, we need something that looks more like the former: we want to be able to take a path from the client and progressively apply rewrites until we have the response we're looking for, with those rewrites being controlled completely in a generic manner.

To that effect, I _do_ think we need to make `FileServer` be:

```rust
pub struct FileServer {
    rewriters: Vec<Arc<dyn Rewriter>>,
}
```

And then be able to recover the existing functionality as a series of rewrites that we provide out of the box.

-----

The second core difference I note is in the treatment of responses, in particular with respect to headers. In my changeset, I introduce changes to `NamedFile` in which I add a `HeaderMap` to the `struct`:

```diff
+pub struct NamedFile {
+    file: File,
+    path: PathBuf,
+    headers: HeaderMap<'static>,
+}
```

(In reality I also added `Metadata`, but I think that was a mistake.)

I _think_ this makes sense as it creates parity between `NamedFile` and `FileServer`. My intent was to allow the rewriter to return a `Path` and optionally a `HeaderMap` and then use those to construct a `NamedFile` that gets used to generate the response. This proved to not be enough as we also need to support rewriting into redirects, and thus returning a `Response` was added.

In your implementation, you have an `enum` that effectively replaces the need for `Response` in my changeset.

I think the right approach is somewhere in-between.

----

At the end of my experiment, and after taking a look at the code here, I'm left with the impression that the rewrite API I suggested is fundamentally trying to be monadic in `Option<FileResponse>` of some sort, and that we should really approach that API a little more directly. That is, to boil things down to a sequence of calls of the form `Option<FileResponse>` `->` `Option<FileResponse>`, where `FileResponse` looks something like `FileServerResponse`:

```rust
enum FileResponse<'a> {
    File(File<'a>),
    Redirect(Redirect<'a>),
}

struct File<'a> {
    path: Cow<'a, Path>,
    headers: HeaderMap<'a>,
}
```

Then, `FileServer` is really just an `Option<FileResponse<'a>>` that starts as a `Some(FileResponse::File(File::from(req.uri())))` (or something morally equivalent) and proceeds through a series of `and_then()`s, `map()s`, and `filter()`s. I think this presentation would also make the ordering argument a bit clearer as we're already used to the ordering consequences between these operations.

If this path is followed, then the `Rewrite` trait becomes:

```rust
pub trait Rewrite: Send + Sync + Any {
    fn rewrite<'r>(&self, response: Option<FileResponse<'r>>) -> Option<FileResponse<'r>>;
}
``` 

It can be implemented for all function of the same form (`for<'r> F: Fn(Option<FileResponse<'r>>) -> Option<FileResponse<'r>>`) as well as the `bind` form (`FileResponse<'r> -> Option<FileResponse<'r>>`) and the `map` form (`FileResponse<'r> -> FileResponse<'r>`). We can also specialize implementations for common forms such as `File<'a> -> File<'a>`, `File<'a> -> Option<File<'a>>`, and `File<'a> -> FileResponse<'r>`.

This would mean we can implement the basic usage as something like:

```rust
FileServer::new()
    .filter_file(|file| !file.path.components().any(|c| c.starts_with('.')))
    .map_file(|file| if file.path.is_dir() && !file.path.ends_with('/') {
        Redirect(Redirect::to(file.path + '/'))
    } else {
        File(file)
    })
   .map_file(|file| if file.path.is_dir() && file.path.join("index.html").exists() {
        File(file.path.join("index.html"))
    } else {
        file
    })
```

Of course, we'd want to provide these as reusable components, but the point is that the implementation becomes significantly more trivial.

core/http/src/uri/segments.rs
```diff
@@ -245,6 +245,39 @@ impl<'a> Segments<'a, Path> {
 
         Ok(buf)
     }
+
+    /// Similar to `to_path_buf`, but always allows dotfiles, and reports whether
+    /// the path contains dotfiles.
+    pub fn to_path_buf_dotfiles(&self) -> Result<(PathBuf, bool), PathError> {
```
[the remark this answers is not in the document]
the10thWiz: I agree. Something like this was needed to support `DotFiles` the way I wanted to implement it though.

the10thWiz:
I think the APIs we have proposed (the two rewrite traits) are to some extent isomorphic. My trait has the request and file server root as additional parameters, but I'm pretty sure we could eliminate that without any issues. (I've already refactored away the need for `root`, and `NormalizedDirs` is the only one that still requires `req`, and this could be removed pretty easily). From there, `FileServerResponse::NotFound` is sort of equivalent to `None`.

The reason I opted to go with an explicit `NotFound` variant was to enable filtering dot files by default, i.e. `FileServer::new("static")` implies that dot files should *not* be served. To implement this, the `None` case needs to carry the path, enabling the `DotFiles` rewrite to convert it back to a `Some` case. This is also the reason I added `to_path_buf_dotfiles`.

I opted to retain `Options` for two reasons: first, to maintain backwards compatibility (and avoid the need to defer this to 0.6), but it sounds like you want this to be a breaking change. The second reason is that we can't mimic the behavior of `Options::Missing` and `Options::IndexFile` cleanly with a rewrite based API, since their primary function is a check during startup. If we are willing to lose this functionality, we can fully eliminate options from the `FileServer` struct.

With that in mind, `FileServer` becomes:
```rust
pub struct FileServer {
    root: PathBuf,
    rewrites: Vec<Arc<dyn Rewriter>>,
    rank: isize,
}
```

I think including `root` and `rank` is worthwhile. `rank` should be included, because it can't really be replicated by rewrites. I think we should include `root`, because we should restrict `FileServer` to a single root. This means `handle` becomes (roughly) `File(self.root.join(req.uri().path())`, a series of rewrites, and converting the final output to a `Response`.

Overall, I agree with most of the API changes you propose. The only changes I'm not willing to lose is hiding dot files by default, and some kind of startup check for whether the root directory exists.

SergioBenitez:
> The reason I opted to go with an explicit `NotFound` variant was to enable filtering dot files by default, i.e. `FileServer::new("static")` implies that dot files should _not_ be served. To implement this, the `None` case needs to carry the path, enabling the `DotFiles` rewrite to convert it back to a `Some` case. This is also the reason I added `to_path_buf_dotfiles`.

We don't need to force filtering dot files by default: it would suffice to make the default constructor include a rewrite that filters dotfiles. This way we avoid negative logic.

> I opted to retain `Options` for two reasons: first, to maintain backwards compatibility (and avoid the need to defer this to 0.6), but it sounds like you want this to be a breaking change.

0.6 is right around the corner. And in general, we shouldn't compromise on the API to avoid a breaking change unless there's a huge benefit to be had, which I don't feel is the case here. 

> The second reason is that we can't mimic the behavior of `Options::Missing` and `Options::IndexFile` cleanly with a rewrite based API, since their primary function is a check during startup. If we are willing to lose this functionality, we can fully eliminate options from the `FileServer` struct.

I think `Options::Missing` should become part of the functionality of specific rewrites. If a rewrite works from some directory, the it can/should be responsible for erroring as needed. This can be part of the constructor for that rewrite.

`Options::IndexFile` _can_ and _should_ be a rewrite, and it _is_ in my implementation. In fact, it's one of the simplest of rewrites:

```rust
|_| File(self.0)
```

> I think including `root` and `rank` is worthwhile. `rank` should be included, because it can't really be replicated by rewrites. 

Absolutely.

> I think we should include `root`, because we should restrict `FileServer` to a single root.

I think this is the case for `FileServer` today, but nothing should prevent a rewrite that looks in multiple roots, for example. Furthermore, we likely want to eventually extend `FileServer` to serve files included in the binary directly, and this would prevent that from being possible. In general, I don't think we gain anything from having `root` be endemic to `FileServer` but stand to lose quite a bit of flexibility. 

> Overall, I agree with most of the API changes you propose. The only changes I'm not willing to lose is hiding dot files by default, and some kind of startup check for whether the root directory exists.

For the more, we should include it as part of the default constructor, not as part of `FileServer` directly. For the latter, that should be a check done by the rewrite, not by `FileServer`.

the10thWiz:
@SergioBenitez I think I've got this to a pretty good place now. I'm pretty sure there aren't any subtle regressions left, but I'm not 100% sure. The test suite does pass now, so that's a good sign.

Looking forward, do we want to add a pre-computed compression rewrite to core or contrib? I'm happy to open another PR for an implementation. If not, I'll likely create and publish a separate crate.

core/lib/src/fs/server.rs
```diff
@@ -60,351 +72,492 @@ use crate::fs::NamedFile;
...
+/// # }
+/// ```
+pub fn filter_dotfiles(file: &File<'_, '_>) -> bool {
+    !file.path.iter().any(|s| s.as_encoded_bytes().starts_with(b"."))
+}
+
+/// Normalize directory accesses to always include a trailing slash.
+///
+/// Must be used after `dir_root`, since it needs the full path to check whether it is
+/// a directory.
```
SergioBenitez, line 342: This shouldn't be the case. This should just check the file being rewritten to, and if it's a directory, add a slash. We want to make these rewrites as independent as possible.

core/lib/src/fs/server.rs
```diff
@@ -60,351 +72,492 @@ use crate::fs::NamedFile;
...
+    }
+}
+
+/// A File response from a [`FileServer`]
+#[derive(Debug)]
+pub struct File<'p, 'h> {
+    /// The path to the file that [`FileServer`] will respond with.
+    pub path: Cow<'p, Path>,
+    /// The original uri from the [`Request`]
+    pub full_uri: &'p Origin<'p>,
```
SergioBenitez, line 161: What is the utility of having `full_uri` here? I think we should be able to remove this, and places where this is being can likely be improved.

core/lib/src/fs/server.rs
```diff
@@ -60,351 +72,492 @@ use crate::fs::NamedFile;
...
+            headers: self.headers,
+        }
+    }
+
+    /// Convert this `File` into a Redirect, transforming the URI.
+    pub fn into_redirect(self, f: impl FnOnce(Origin<'static>) -> Origin<'static>)
+        -> FileResponse<'p, 'h>
+    {
+        FileResponse::Redirect(Redirect::permanent(f(self.full_uri.clone().into_owned())))
+    }
```
SergioBenitez, line 196: This feels wrong to me. Remove it falls in line with my comments on `full_uri` and the `normalize_dirs` rewriter.

core/lib/src/fs/server.rs
```diff
@@ -60,351 +72,492 @@ use crate::fs::NamedFile;
...
-    ///
-    /// Note that this is different than [`Options::default()`](#impl-Default),
-    /// which enables options.
-    pub const None: Options = Options(0);
+    let path = path.as_ref();
+    if !path.is_dir() {
+        let path = path.display();
+        error!("FileServer path '{}' is not a directory.", path.primary());
+        warn!("Aborting early to prevent inevitable handler error.");
+        panic!("invalid directory: refusing to continue");
```
SergioBenitez, line 259: This will need to be rewritten to be structured `trace` messages.

core/lib/src/fs/server.rs
```diff
@@ -60,351 +72,492 @@ use crate::fs::NamedFile;
...
+/// # Panics
 ///
-/// `Options` structures can be `or`d together to select two or more options.
-/// For instance, to request that both dot files and index pages be returned,
-/// use `Options::DotFiles | Options::Index`.
-#[derive(Debug, Clone, Copy)]
-pub struct Options(u8);
+/// Panics if `path` is not directory.
+pub fn dir_root(path: impl AsRef<Path>)
+    -> impl for<'p, 'h> Fn(File<'p, 'h>) -> FileResponse<'p, 'h> + Send + Sync + 'static
```
SergioBenitez, line 250: This will be hard to understand. We _should_ be able to return `impl Rewriter`, I believe, which would result in a much simpler type signature.

core/lib/src/fs/server.rs
```diff
@@ -60,351 +72,492 @@ use crate::fs::NamedFile;
...
+/// FileServer::empty()
+///     .map_file(file_root("static/index.html"))
+/// # }
+/// ```
+///
+/// # Panics
+///
+/// Panics if `path` does not exist.
+pub fn file_root(path: impl AsRef<Path>)
+    -> impl for<'p, 'h> Fn(File<'p, 'h>) -> FileResponse<'p, 'h> + Send + Sync + 'static
```
SergioBenitez, line 283: Same here for `impl Rewriter`.

core/lib/src/fs/server.rs
```diff
@@ -60,351 +72,492 @@ use crate::fs::NamedFile;
...
+///
+/// ```rust,no_run
+/// # use rocket::fs::{FileServer, missing_root};
+/// # fn make_server() -> FileServer {
+/// FileServer::empty()
+///     .map_file(missing_root("/tmp/rocket"))
+/// # }
+/// ```
+pub fn missing_root(path: impl AsRef<Path>)
+    -> impl for<'p, 'h> Fn(File<'p, 'h>) -> FileResponse<'p, 'h> + Send + Sync + 'static
```
SergioBenitez, line 313: And here `impl Rewriter`.

core/lib/src/fs/server.rs
```diff
@@ -60,351 +72,492 @@ use crate::fs::NamedFile;
...
-/// `Options` structures can be `or`d together to select two or more options.
-/// For instance, to request that both dot files and index pages be returned,
-/// use `Options::DotFiles | Options::Index`.
-#[derive(Debug, Clone, Copy)]
-pub struct Options(u8);
+/// Panics if `path` is not directory.
+pub fn dir_root(path: impl AsRef<Path>)
+    -> impl for<'p, 'h> Fn(File<'p, 'h>) -> FileResponse<'p, 'h> + Send + Sync + 'static
+{
+    use yansi::Paint as _;
```
SergioBenitez, line 252: We should never use `yansi::Paint` outside of the `trace::subscriber` module.

core/lib/src/fs/server.rs
```diff
@@ -60,351 +72,492 @@ use crate::fs::NamedFile;
...
-///   * [`Options::NormalizeDirs`] - Redirect directories without a trailing
-///     slash to ones with a trailing slash.
+/// # Panics
 ///
-/// `Options` structures can be `or`d together to select two or more options.
-/// For instance, to request that both dot files and index pages be returned,
-/// use `Options::DotFiles | Options::Index`.
-#[derive(Debug, Clone, Copy)]
-pub struct Options(u8);
+/// Panics if `path` is not directory.
```
SergioBenitez, line 248: We should refer to the non-panicking variant here.

core/lib/src/fs/server.rs
```diff
@@ -60,351 +72,492 @@ use crate::fs::NamedFile;
...
+/// # use rocket::fs::{FileServer, file_root};
+/// # fn make_server() -> FileServer {
+/// FileServer::empty()
+///     .map_file(file_root("static/index.html"))
+/// # }
+/// ```
+///
+/// # Panics
+///
+/// Panics if `path` does not exist.
```
SergioBenitez, line 281: We should refer to the non-panicking variant here.

core/lib/src/fs/server.rs
```diff
@@ -60,351 +72,492 @@ use crate::fs::NamedFile;
...
+/// # Example
+///
+/// ```rust,no_run
+/// # use rocket::fs::{FileServer, missing_root};
+/// # fn make_server() -> FileServer {
+/// FileServer::empty()
+///     .map_file(missing_root("/tmp/rocket"))
+/// # }
+/// ```
+pub fn missing_root(path: impl AsRef<Path>)
```
SergioBenitez, line 312: I think we want to call this something different, say `file_root_permissive`, or something like that.

core/lib/src/fs/server.rs
```diff
@@ -60,351 +72,492 @@ use crate::fs::NamedFile;
...
+/// ```rust,no_run
+/// # use rocket::fs::{FileServer, filter_dotfiles, dir_root};
+/// # fn make_server() -> FileServer {
+/// FileServer::empty()
+///     .filter_file(filter_dotfiles)
+///     .map_file(dir_root("static"))
+/// # }
+/// ```
+pub fn filter_dotfiles(file: &File<'_, '_>) -> bool {
+    !file.path.iter().any(|s| s.as_encoded_bytes().starts_with(b"."))
```
SergioBenitez, line 336: This will filter out things like `foo/../bar` and `foo/./bar` which are totally okay and safe. We need to allow this paths. We should have tests to ensure that we do allow those paths. I fixed the latter case in my [patch](https://paste.rs/AbR41.patch).

core/lib/src/fs/server.rs
```diff
@@ -60,351 +72,492 @@ use crate::fs::NamedFile;
...
+/// ```rust,no_run
+/// # use rocket::fs::{FileServer, index, dir_root};
+/// # fn make_server() -> FileServer {
+/// FileServer::empty()
+///     .map_file(dir_root("static"))
+///     .map_file(index("index.html"))
+/// # }
+/// ```
+pub fn index(index: &'static str)
+    -> impl for<'p, 'h> Fn(File<'p, 'h>) -> FileResponse<'p, 'h> + Send + Sync
```
SergioBenitez, line 381: `impl Rewriter`

core/lib/src/fs/server.rs
```diff
@@ -60,351 +72,492 @@ use crate::fs::NamedFile;
...
+        // information out of `dyn Rewriter`
+        route.name = Some("FileServer".into());
+        vec![route]
     }
 }
 
-impl std::ops::BitOr for Options {
-    type Output = Self;
+#[crate::async_trait]
+impl Handler for FileServer {
```
SergioBenitez, line 532: Can you model this on the implementation in my patch? I find that version easier to read.

core/lib/src/fs/server.rs
the10thWiz: This actually won't filter out `..`, since they are already stripped out by `FileServer` when it calls `to_path_buf` on the URI. I think we should modify `to_path_buf` to filter out `.` segments in a similar way (by just ignoring them).

core/lib/src/fs/server.rs
the10thWiz: While we could, the return type would not be useable in map_file or filter_file. The only way to implement a return type that support both and_rewrite(impl Rewriter) and map_file(impl Fn(file) -> Response) would be to add traits for MapFile and FilterFile, so a custom struct could implement both. I thought this would complicate the API more, and choose to leave it with a more complex return type for now. I'm not sure what the right way to go here is, but I'm leaning towards adding MapFile and FilterFile traits.

core/lib/src/fs/server.rs
the10thWiz: I've changed the documentation to say `Should normally be used after `dir_root` (or another rewrite that adds a root), since it needs the full path to check whether a path points to a directory.` The thing is, this is a fundamental limitation of `normalize_dirs`, since it can't look forward at what rewrites later in the chain will do.

core/lib/src/fs/server.rs
the10thWiz: I've updated it (and removed yansi in the process). If there is more information we can provide, I'd be more than happy to update these messages.

core/lib/src/fs/server.rs
the10thWiz: After doing some testing, I've updated this to return `impl FileMap`, which is more or less a type alias of the full `impl Fn...`. I initially wanted to make them proper traits, but it would prevent using lambdas in a convenient way, since the Rust type inferencing can't automatically determine the HRTBs.

core/lib/src/fs/server.rs
the10thWiz: I've simplified the implementation a bit, and extracted some of the code to a separate function, but I don't think I can do much better. Many of the differences (the `.exists` check, and the new `strip_trailing_slash`) stem from more thorough testing.

the10thWiz:
I'm updated this PR (using the changes you've made), to pass CI. Overall, I like most of the changes. I'm not sure how I feel about `Rewrite` implementing `Rewriter`, and the change to using temporary redirects.

To that end, I've changed a few things:
- Added `FileServer::directory` (One doc-test uses it).
- Updated tests to reflect a change in `TrailingDirs`'s behavior - it now generates temporary redirects, unlike the old file server which generated permanent redirects.
- Re-exported `Rewrite` and `Rewriter` from `rocket::fs`.
- Added `From<File>` and `From<Redirect>` impls for `Option<Rewrite>`.

core/lib/src/fs/rewrite.rs
```diff
@@ -215,12 +221,24 @@ impl<'r> From<File<'r>> for Rewrite<'r> {
     }
 }
 
+impl<'r> From<File<'r>> for Option<Rewrite<'r>> {
+    fn from(value: File<'r>) -> Self {
+        Some(Rewrite::File(value))
+    }
+}
```
SergioBenitez, line 228: What's the utility of this `impl`? I don't see it used anywhere, and I'm not sure we really want to hide the functionality under an opaque `.into()`.

core/lib/src/fs/server.rs
```diff
@@ -94,6 +94,20 @@ impl FileServer {
...
+    /// Adds a set of default rewrites:
+    /// - [`filter_dotfiles`]: Hides all dotfiles.
+    /// - [`prefix(path)`](prefix): Applies the root path.
+    /// - [`normalize_dirs`]: Normalizes directories to have a trailing slash.
+    pub fn directory<P: AsRef<Path>>(path: P) -> Self {
+        Self::empty()
+            .filter(|f, _| f.is_visible())
+            .rewrite(Prefix::checked(path))
+            .rewrite(TrailingDirs)
+    }
```
SergioBenitez, line 109: So I had implemented this exact method but I removed it because I doubted its utility, given that this would make `new()` exactly `Self::directory(path).rewrite(DirIndex::unconditional("index.html"))`. If we're going to keep this, we should document the two methods well with why/when you would use one or the other.

core/lib/src/fs/mod.rs
```diff
@@ -11,6 +11,7 @@ pub use server::*;
 pub use named_file::*;
 pub use temp_file::*;
 pub use file_name::*;
+pub use rewrite::{Rewrite, Rewriter};
```
SergioBenitez, line 14: I don't think we should export the same type from two different spaces. My view is this: if one is going to use `Rewrite` or `Rewriter`, then they'll likely also use one of the existing rewriters. As such, it makes sense to either expose all `rewrite` in `fs` or _none_ of `rewrite` in `fs`. For clarity, I would prefer the latter.

core/lib/src/fs/rewrite.rs
```diff
@@ -72,8 +72,14 @@ impl<'r> File<'r> {
...
-    /// Returns `true` if the file is a dotfile. A dotfile is a file whose name
-    /// starts with a period (`.`) and is considered hidden.
+    /// Returns `true` if the file is a dotfile. A dotfile is a file whose
+    /// name or any directory in it's path start with a period (`.`) and is
+    /// considered hidden.
+    ///
+    /// # Windows Note
+    ///
+    /// This does *not* check the file metadata on any platform, so hidden files
+    /// on Windows will not be detected.
```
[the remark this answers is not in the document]
the10thWiz: This is bit more complex than the dotfiles check. First, it would only work after `Prefix` (much like `is_dir` or `exists`), and the current dotfiles check is typically done before `Prefix`.

Second, to match the behavior of dotfiles, we would likely want to also check all parent folders, which would likely represent a non-trivial amount of work. Right now, each `Rewrite` requires at most one syscall, and I think that's a reasonable limit.

Overall, I'm not sure if it's worthwhile. We could provide an alternate method that does check fs metadata, but I'm not sure how much value it would provide. We would also likely need to create a more complex test for this functionality, since I'm not sure git preserves Windows file properties.

core/lib/src/fs/rewrite.rs
the10thWiz: The other impl (`From<Redirect>` for `Option<Rewrite>`) is actually used in at least one doc-test.

This makes it possible for a rewrite to return `File { .. }.into()`, rather than `Some(File { .. }.into())`. I think it makes the API more succinct, without losing any information. I think it might be worth looking at the other doc-tests, and using this impl if appropriate.

core/lib/src/fs/server.rs
the10thWiz: I assumed something like that might have happened. It was still used in one of the doc-tests (of `DirIndex` specifically). I think it's worth including, but I'm thinking it might be worth changing the name in addition to writing better docs.

core/lib/src/fs/rewrite.rs
the10thWiz: When adding documentation, now both are used (in the doc-tests for `FileServer::filter` and `FileServer::map` respectively).

SergioBenitez:
Merged! https://github.com/rwf2/Rocket/commit/65e3b87d6bee4f56da62565fba638b950426953c

PR #2799, opened 2024-05-28
Add MiniJinja for templating example

va-an:
Hi!
I noticed that support for MiniJinja has been added in the new version of `rocket_dyn_templates`, and I have added a corresponding example.

### DIFF
--- a/examples/templating/src/tests.rs
+++ b/examples/templating/src/tests.rs
@@ -40,17 +40,20 @@ fn test_404(base: &str) {
     let client = Client::tracked(rocket()).unwrap();
     for bad_path in &["/hello", "/foo/bar", "/404"] {
         let path = format!("/{}{}", base, bad_path);
-        let escaped_path = RawStr::new(&path).html_escape();
+
+        let escaped_path_upper = RawStr::new(&path).html_escape();
+        let escaped_path_lower = escaped_path_upper.to_lowercase();
 
         let response = client.get(&path).dispatch();
         assert_eq!(response.status(), Status::NotFound);
         let response = response.into_string().unwrap();
 
-        assert!(response.contains(base));
-        assert! {
-            response.contains(&format!("{} does not exist", path))
-                || response.contains(&format!("{} does not exist", escaped_path))
-        };
+        assert!(response.contains(base));        
+        assert!(
+            response.contains(&format!("{} does not exist", path)) ||
+            response.contains(&format!("{} does not exist", escaped_path_upper)) ||
+            response.contains(&format!("{} does not exist", escaped_path_lower))
+        );
     }
 }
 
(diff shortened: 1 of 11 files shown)


examples/templating/src/tests.rs
```diff
@@ -40,17 +40,20 @@ fn test_404(base: &str) {
     let client = Client::tracked(rocket()).unwrap();
     for bad_path in &["/hello", "/foo/bar", "/404"] {
         let path = format!("/{}{}", base, bad_path);
-        let escaped_path = RawStr::new(&path).html_escape();
+
+        let escaped_path_upper = RawStr::new(&path).html_escape();
+        let escaped_path_lower = escaped_path_upper.to_lowercase();
```
va-an, line 45: I've ran this modified test with command `cargo test -- --nocapture`:

```rust
fn test_404(base: &str) {
    // Check that the error catcher works.
    let client = Client::tracked(rocket()).unwrap();
    for bad_path in &["/hello", "/foo/bar", "/404"] {
        let path = format!("/{}{}", base, bad_path);
        let escaped_path = RawStr::new(&path).html_escape();

        let response = client.get(&path).dispatch();
        assert_eq!(response.status(), Status::NotFound);
        let response = response.into_string().unwrap();
        
        println!("path: '{}'", path);
        println!("escaped_path: '{}'", escaped_path);
        println!("response body: {:?}", response);
    }
}
```

And the case is different for the HTML escaped characters in escaped_path and response.

For `tera`:
```
path: '/tera/foo/bar'
escaped_path: '&#x2F;tera&#x2F;foo&#x2F;bar'
response body: "... The page at &#x2F;tera&#x2F;foo&#x2F;bar does not exist! ..."
```

For `minijinja`:
```
path: '/minijinja/foo/bar'
escaped_path: '&#x2F;minijinja&#x2F;foo&#x2F;bar'
response body: "... The page at &#x2f;minijinja&#x2f;foo&#x2f;bar does not exist! ..."
```

Is this good enough or is there a better solution?

SergioBenitez:
Awesome! Please see the build failures. Will review once the CI is green.

va-an:
> Awesome! Please see the build failures. Will review once the CI is green.

Found and fixed issues with line length and whitespaces. Now `./scripts/test.sh --examples -q` works successfully on my machine.

examples/templating/src/tests.rs
va-an: Fixed, please have a look!

va-an:
Added mention of MiniJinja to the guide. Force-pushed to delete files added by mistake.

SergioBenitez:
@the10thWiz Can you finalize this PR and then squash + rebase + reword and merge?

Issue #477, opened 2017-11-12
Expose private cookies functionality.


dpc:
## Feature Requests

Expose new API: `encrypt_private(&str) -> String` and `decrypt_private(&str) -> Result<String>` or something like it, that would use the same scheme as private cookies do.

### Why you believe this feature is necessary.

It's not necessary, but it would be useful to be able to use the same encryption that private cookies use on arbitrary data/strings. This way encrypted values can be embedded eg. in rendered forms to pass authenticated, unforgeable data.

### A convincing use-case for this feature.

I'd like to embed an URL and some control data in one of my forms, to control some flow after form submission, and not have to worry about all the possible attack vectors. I can't put it in cookies, since it's page, and not session, related.

### Why this feature can't or shouldn't exist outside of Rocket.

Obviously, it can be done manually / through another library, but since Rocket already takes care of managing the encryption key, it would be much easier to just reuse it.

RalfJung:
I would like to use such a feature to embed a link into a confirmation email I send to the user, so that when the user clicks that link, I can verify that the payload has not been tampered with. Right now, I have to ask admins to generate and configure *two* secret keys (one for Rocket and one specifically for my app for the email links), which seems rather unnecessary.

morbatex:
I would also like this feature added, as it would allow me to decrypt private cookies I recieved in a LocalResponse while testing, to check, if it contains the right value.

SergioBenitez:
@morbatex You can already do that today by asking the `Cookies` jar to decrypt the cookie for you.

jebrosen:
I don't think that's currently true of `LocalResponse` -- https://github.com/SergioBenitez/Rocket/issues/948#issuecomment-472614363

SergioBenitez:
> I don't think that's currently true of `LocalResponse` -- [#948 (comment)](https://github.com/SergioBenitez/Rocket/issues/948#issuecomment-472614363)

As of https://github.com/SergioBenitez/Rocket/commit/52320020bc26387b96a0126a0d49821ad3125e7d, reality has reached my previous statement.

Kimundi:
I've also run into this on 0.5. I want to generate an URL with an encrypted query parameter for which the existing cookie secret would be the perfect fit. I tried to work around it, but so far I did not find a solution that does not involved my own copy of a secret key and encryption routine:
- Rocket does not have a API as proposed here, so I can not encrypt data directly
[...]

PR #2803, opened 2024-06-03
Encrypt data with secret key

vadim-anfv:
Implements encryption and decryption functions for arbitrary textual data, as discussed in #477.

- Added functions `encrypt` and `decrypt` to `SecretKey`.
- AAD is not used since we don't have metadata like cookie names.
- Functions for encryption/decryption are based on implementations from [cookie-rs](https://github.com/rwf2/cookie-rs/blob/master/src/secure/private.rs#L42-L88).
- Added example with suggested usage.

### DIFF
--- a/core/lib/Cargo.toml
+++ b/core/lib/Cargo.toml
@@ -74,6 +74,8 @@ tokio-stream = { version = "0.1.6", features = ["signal", "time"] }
 cookie = { version = "0.18", features = ["percent-encode"] }
 futures = { version = "0.3.30", default-features = false, features = ["std"] }
 state = "0.6"
+aes-gcm = "0.10.3"
+base64 = "0.22.1"
 
 # tracing
 tracing = { version = "0.1.40", default-features = false, features = ["std", "attributes"] }
--- a/core/lib/Cargo.toml
+++ b/core/lib/Cargo.toml
@@ -77,6 +77,8 @@ state = "0.6"
 chacha20poly1305 = "0.10.1"
 hkdf = "0.12.4"
 sha2 = "0.10.8"
+base64 = "0.22.1"
+hex = "0.4.3"
 
 # tracing
 tracing = { version = "0.1.40", default-features = false, features = ["std", "attributes"] }
--- a/core/lib/Cargo.toml
+++ b/core/lib/Cargo.toml
@@ -74,7 +74,7 @@ tokio-stream = { version = "0.1.6", features = ["signal", "time"] }
 cookie = { version = "0.18", features = ["percent-encode"] }
 futures = { version = "0.3.30", default-features = false, features = ["std"] }
 state = "0.6"
-aes-gcm = "0.10.3"
+chacha20poly1305 = "0.10.1"
 
 # tracing
 tracing = { version = "0.1.40", default-features = false, features = ["std", "attributes"] }
--- a/core/lib/Cargo.toml
+++ b/core/lib/Cargo.toml
@@ -75,7 +75,6 @@ cookie = { version = "0.18", features = ["percent-encode"] }
 futures = { version = "0.3.30", default-features = false, features = ["std"] }
 state = "0.6"
 aes-gcm = "0.10.3"
-base64 = "0.22.1"
 
 # tracing
 tracing = { version = "0.1.40", default-features = false, features = ["std", "attributes"] }
--- a/core/lib/Cargo.toml
+++ b/core/lib/Cargo.toml
@@ -75,6 +75,8 @@ cookie = { version = "0.18", features = ["percent-encode"] }
 futures = { version = "0.3.30", default-features = false, features = ["std"] }
 state = "0.6"
 chacha20poly1305 = "0.10.1"
+hkdf = "0.12.4"
+sha2 = "0.10.8"
 
 # tracing
 tracing = { version = "0.1.40", default-features = false, features = ["std", "attributes"] }
--- a/core/lib/Cargo.toml
+++ b/core/lib/Cargo.toml
@@ -26,7 +26,7 @@ workspace = true
 default = ["http2", "tokio-macros", "trace"]
 http2 = ["hyper/http2", "hyper-util/http2"]
 http3-preview = ["s2n-quic", "s2n-quic-h3", "tls"]
-secrets = ["cookie/private", "cookie/key-expansion"]
+secrets = ["cookie/private", "cookie/key-expansion", "chacha20poly1305", "hkdf", "sha2", "base64", "hex"]
 json = ["serde_json"]
 msgpack = ["rmp-serde"]
 uuid = ["uuid_", "rocket_http/uuid"]
@@ -43,6 +43,13 @@ uuid_ = { package = "uuid", version = "1", optional = true, features = ["serde"]
 
 # Optional MTLS dependencies
 x509-parser = { version = "0.16", optional = true }
+
+# Optional dependencies for "secrets" feature
+chacha20poly1305 = { version = "0.10.1", optional = true }
+hkdf = { version = "0.12.4", optional = true }
+sha2 = { version = "0.10.8", optional = true }
+base64 = { version = "0.22.1", optional = true }
+hex = { version = "0.4.3", optional = true }
 
 # Hyper dependencies
 http = "1"
@@ -74,11 +81,6 @@ tokio-stream = { version = "0.1.6", features = ["signal", "time"] }
 cookie = { version = "0.18", features = ["percent-encode"] }
 futures = { version = "0.3.30", default-features = false, features = ["std"] }
 state = "0.6"
-chacha20poly1305 = "0.10.1"
-hkdf = "0.12.4"
-sha2 = "0.10.8"
-base64 = "0.22.1"
-hex = "0.4.3"
 
 # tracing
 tracing = { version = "0.1.40", default-features = false, features = ["std", "attributes"] }
--- a/core/lib/src/config/secret_key.rs
+++ b/core/lib/src/config/secret_key.rs
@@ -11,6 +11,15 @@ use crate::request::{Outcome, Request, FromRequest};
 const NONCE_LEN: usize = 12;
 const KEY_LEN: usize = 32;
 
+#[derive(Debug)]
+pub enum Error {
+    KeyLengthError,
+    NonceFillError,
+    EncryptionError,
+    DecryptionError,
+    EncryptedDataLengthError,
+}
+
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 enum Kind {
     Zero,
@@ -194,16 +203,19 @@ impl SecretKey {
     /// use rocket::config::SecretKey;
     ///
     /// let plaintext = "I like turtles".as_bytes();
-    /// let secret_key = SecretKey::generate().expect("error generate key");
+    /// let secret_key = SecretKey::generate().unwrap();
     ///
-    /// let encrypted = secret_key.encrypt(&plaintext).expect("can't encrypt");
-    /// let decrypted = secret_key.decrypt(&encrypted).expect("can't decrypt");
+    /// let encrypted = secret_key.encrypt(&plaintext).unwrap();
+    /// let decrypted = secret_key.decrypt(&encrypted).unwrap();
     ///
     /// assert_eq!(decrypted, plaintext);
     /// ```
-    pub fn encrypt<T: AsRef<[u8]>>(&self, value: T) -> Result<Vec<u8>, &'static str> {
+    pub fn encrypt<T: AsRef<[u8]>>(&self, value: T) -> Result<Vec<u8>, Error> {
         // Convert the encryption key to a fixed-length array
-        let key: [u8; KEY_LEN] = self.key.encryption().try_into().map_err(|_| "enc key len error")?;
+        let key: [u8; KEY_LEN] = self.key
+            .encryption()
+            .try_into()
+            .map_err(|_| Error::KeyLengthError)?;
 
         // Create a new AES-256-GCM instance with the provided key
         let aead = Aes256Gcm::new(GenericArray::from_slice(&key));
@@ -211,11 +223,11 @@ impl SecretKey {
         // Generate a random nonce
         let mut nonce = [0u8; NONCE_LEN];
         let mut rng = rand::thread_rng();
-        rng.try_fill_bytes(&mut nonce).map_err(|_| "couldn't random fill nonce")?;
+        rng.try_fill_bytes(&mut nonce).map_err(|_| Error::NonceFillError)?;
         let nonce = Nonce::from_slice(&nonce);
 
         // Encrypt the plaintext using the nonce
-        let ciphertext = aead.encrypt(nonce, value.as_ref()).map_err(|_| "encryption error")?;
+        let ciphertext = aead.encrypt(nonce, value.as_ref()).map_err(|_| Error::EncryptionError)?;
 
         // Prepare a vector to hold the nonce and ciphertext
         let mut encrypted_data = Vec::with_capacity(NONCE_LEN + ciphertext.len());
@@ -228,12 +240,12 @@ impl SecretKey {
     /// Decrypts the given base64-encoded encrypted data.
     /// Extracts the nonce from the data and uses it for decryption.
     /// Returns the decrypted plaintext string.
-    pub fn decrypt<T: AsRef<[u8]>>(&self, encrypted: T) -> Result<Vec<u8>, &'static str> {
+    pub fn decrypt<T: AsRef<[u8]>>(&self, encrypted: T) -> Result<Vec<u8>, Error> {
         let encrypted = encrypted.as_ref();
 
         // Check if the length of decoded data is at least the length of the nonce
         if encrypted.len() <= NONCE_LEN {
-            return Err("length of encrypted data is <= NONCE_LEN");
+            return Err(Error::EncryptedDataLengthError);
         }
 
         // Split the decoded data into nonce and ciphertext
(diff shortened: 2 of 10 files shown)


core/lib/src/config/secret_key.rs
```diff
@@ -178,6 +187,75 @@ impl SecretKey {
...
+    /// ```rust
+    /// let plaintext = "I like turtles";
+    /// let secret_key = SecretKey::generate().expect("error generate key");
+    ///
+    /// let encrypted = secret_key.encrypt(&plaintext).expect("can't encrypt");
+    /// let decrypted = secret_key.decrypt(&encrypted).expect("can't decrypt");
+    ///
+    /// assert_eq!(decrypted, plaintext);
+    /// ```
+    pub fn encrypt(&self, plaintext: &str) -> Result<String, &'static str> {
```
the10thWiz, line 205: Should we restrict this to only encrypting strings? It wouldn't be hard to offer this for any `impl AsRef<[u8]>`.

Obviously, we would also need to implement a decrypt that decrypts to bytes, since it might not be a UTF-8 string.

Also, `Result<_, &'static str>` isn't very good practice. We should instead create an error enum, and return that.

core/lib/src/config/secret_key.rs
```diff
@@ -178,6 +187,75 @@ impl SecretKey {
...
+        // Encrypt the plaintext using the nonce
+        let ciphertext = aead.encrypt(nonce, plaintext.as_ref()).map_err(|_| "encryption error")?;
+
+        // Prepare a vector to hold the nonce and ciphertext
+        let mut encrypted_data = Vec::with_capacity(NONCE_LEN + ciphertext.len());
+        encrypted_data.extend_from_slice(nonce);
+        encrypted_data.extend_from_slice(&ciphertext);
+
+        // Return the base64-encoded result
+        Ok(URL_SAFE.encode(encrypted_data))
```
the10thWiz, line 227: Should we enforce Base64 encoding? I think a more general solution might be to simply return `Vec<u8>`, and let the user encode it if they want to.

core/lib/src/config/secret_key.rs
```diff
@@ -178,6 +187,75 @@ impl SecretKey {
...
+    }
+
+    /// Decrypts the given base64-encoded encrypted data.
+    /// Extracts the nonce from the data and uses it for decryption.
+    /// Returns the decrypted plaintext string.
+    pub fn decrypt(&self, encrypted: &str) -> Result<String, &'static str> {
```
the10thWiz, line 233: See above

examples/private-data/Rocket.toml
```diff
@@ -0,0 +1,2 @@
+[default]
+secret_key = "itlYmFR2vYKrOmFhupMIn/hyB6lYCCTXz4yaQX89XVg="
```
the10thWiz, line 2: We should avoid committing any secret key to GitHub, even for examples. Rocket should auto-generate a random key on every launch, so we should be okay with not specifying one here.

examples/private-data/src/main.rs
```diff
@@ -0,0 +1,36 @@
...
+}
+
+#[get("/decrypt/<msg>")]
+    fn decrypt_endpoint(msg: &str, config: &State<Config>) -> String {
+        let secret_key = config.secret_key.clone();
+        let decrypted = secret_key.decrypt(msg).unwrap();
+
+        info!("received message for decrypt: '{}'", msg);
+        info!("decrypted msg: '{}'", decrypted);
+
```
the10thWiz, line 27: Not sure what's going on with the formatting here.

the10thWiz (commented):
@SergioBenitez can weigh in, but I think this is pretty close to being ready for merge. There are a few specific points in the API I noted.

core/lib/src/config/secret_key.rs
va-an: Good point!
In this case we don't have to add `base64` lib to Rocket and this option is more flexible. 
I will change code and example.

examples/private-data/src/main.rs
va-an: For some reason, the formatting of the code stopped working for me; neither cargo fmt nor autoformat when saving a file in VSCode is functioning. 
I'll fix it.

the10thWiz: I believe we have intentionally disabled it, since it doesn't work properly for some of the more complex types and where clauses.

examples/private-data/Rocket.toml
va-an: Understood, then I'll just delete Rocket.toml for this example, and Rocket will generate a new secret key with every run. 
I also deleted the secret key from the [cookies example](https://github.com/rwf2/Rocket/blob/c303add07fe15520654aa388e6f387979704b1d5/examples/cookies/Rocket.toml#L2).

core/lib/src/config/secret_key.rs
```diff
@@ -178,6 +193,80 @@ impl SecretKey {
     {
         ser.serialize_bytes(&[0; 32][..])
     }
+
+    /// Encrypts the given plaintext.
+    /// Generates a random nonce for each encryption to ensure uniqueness.
+    /// Returns the base64-encoded string of the concatenated nonce and ciphertext.
```
the10thWiz, line 199: This is no longer true.

the10thWiz (changes requested):
I'm almost ready to approve these changes. @SergioBenitez should take a look, I think the implementation is solid.

core/lib/src/config/secret_key.rs
```diff
@@ -178,6 +193,80 @@ impl SecretKey {
...
+    /// let decrypted = secret_key.decrypt(&encrypted).unwrap();
+    ///
+    /// assert_eq!(decrypted, plaintext);
+    /// ```
+    pub fn encrypt<T: AsRef<[u8]>>(&self, value: T) -> Result<Vec<u8>, Error> {
+        // Convert the encryption key to a fixed-length array
+        let key: [u8; KEY_LEN] = self.key
+            .encryption()
+            .try_into()
+            .map_err(|_| Error::KeyLengthError)?;
```
SergioBenitez, line 218: This stores the key on the stack, which shouldn't be necessary. Furthermore, it reuses the `encryption()` key for a different operation, which we shouldn't do. Instead we should use a KDF (say HKDF).

core/lib/src/config/secret_key.rs
```diff
@@ -178,6 +193,80 @@ impl SecretKey {
...
+
+        // Create a new AES-256-GCM instance with the provided key
+        let aead = Aes256Gcm::new(GenericArray::from_slice(&key));
```
SergioBenitez, line 221: I think we likely want to use a more a nonce-resistant algorithm here, either AES-GCM-SIV or XChaCha20Poly1305.

core/lib/src/config/secret_key.rs
```diff
@@ -178,6 +193,80 @@ impl SecretKey {
...
+
+        // Generate a random nonce
+        let mut nonce = [0u8; NONCE_LEN];
+        let mut rng = rand::thread_rng();
+        rng.try_fill_bytes(&mut nonce).map_err(|_| Error::NonceFillError)?;
+        let nonce = Nonce::from_slice(&nonce);
```
SergioBenitez, line 227: Note that you can use the [`generate_nonce()`](https://docs.rs/aes-gcm/latest/aes_gcm/trait.AeadCore.html#method.generate_nonce) method to generate a properly sized nonce.

core/lib/src/config/secret_key.rs
```diff
@@ -178,6 +193,80 @@ impl SecretKey {
...
+    ///
+    /// let plaintext = "I like turtles".as_bytes();
+    /// let secret_key = SecretKey::generate().unwrap();
+    ///
+    /// let encrypted = secret_key.encrypt(&plaintext).unwrap();
+    /// let decrypted = secret_key.decrypt(&encrypted).unwrap();
+    ///
+    /// assert_eq!(decrypted, plaintext);
+    /// ```
+    pub fn encrypt<T: AsRef<[u8]>>(&self, value: T) -> Result<Vec<u8>, Error> {
```
SergioBenitez, line 213: A return type of `Vec<u8>` is not particularly useful. It would be nice to return something that encapsulates the `Vec<u8>` with convenient methods, like [`blake3::Hash`](https://docs.rs/blake3/latest/blake3/struct.Hash.html). For example, something lIke:

```rust
struct Cipher(Vec<u8>);

impl Cipher {
    fn from_bytes(&[u8]) -> Result<Self>;
    fn from_vec(Vec<u8>) -> Result<Self>;
    fn from_hex(&str) -> Result<Self>;
    fn from_base64(&str) -> Result<Self>;

    fn as_bytes(&self) -> &[u8];
    fn into_vec(self) -> Vec<u8>;
    fn to_hex(&self) -> String;
    fn to_base64(&self) -> String;
}
```

SergioBenitez, line 213: It would be nice to provide a variant that allows the caller to provide associated data.

SergioBenitez (changes requested):

core/lib/src/config/secret_key.rs
```diff
@@ -178,6 +193,80 @@ impl SecretKey {
...
+        encrypted_data.extend_from_slice(nonce);
+        encrypted_data.extend_from_slice(&ciphertext);
+
+        Ok(encrypted_data)
+    }
+
+    /// Decrypts the given encrypted data.
+    /// Extracts the nonce from the data and uses it for decryption.
+    /// Returns the decrypted Vec<u8>.
+    pub fn decrypt<T: AsRef<[u8]>>(&self, encrypted: T) -> Result<Vec<u8>, Error> {
```
SergioBenitez, line 243: This would then take a `Cipher`, which is pre-validated.

core/lib/src/config/secret_key.rs
va-an: Great point! Should we do the same for [cookie-rs](https://github.com/rwf2/cookie-rs/blob/master/src/secure/private.rs#L60)?

va-an: I am going to use `XChaCha20Poly1305` because the crate `chacha20poly1305` has received a [security audit](https://docs.rs/chacha20poly1305/latest/chacha20poly1305/#security-notes), and the crate `aes_gcm_siv` has not been audited [yet](https://docs.rs/aes-gcm-siv/latest/aes_gcm_siv/#security-warning).

core/lib/src/config/secret_key.rs
```diff
@@ -1,15 +1,15 @@
...
+    aead::{Aead, AeadCore, KeyInit, OsRng, generic_array::GenericArray},
+    XChaCha20Poly1305, XNonce,
 };
 use cookie::Key;
 use serde::{de, ser, Deserialize, Serialize};
 
 use crate::request::{Outcome, Request, FromRequest};
 
-const NONCE_LEN: usize = 12;
+const NONCE_LEN: usize = 24; // 192-bit
```
SergioBenitez, line 12: I don't think we need this anymore. There's [`NonceSize`](https://docs.rs/chacha20poly1305/latest/chacha20poly1305/trait.AeadCore.html#associatedtype.NonceSize).

core/lib/src/config/secret_key.rs
```diff
@@ -1,15 +1,15 @@
...
 const KEY_LEN: usize = 32;
```
SergioBenitez, line 13: Same with this. We should be able to get it from the [`Key`](https://docs.rs/chacha20poly1305/latest/chacha20poly1305/type.Key.html) and [`ConstArrayLen`](https://docs.rs/generic-array/latest/generic_array/type.ConstArrayLength.html).

core/lib/src/config/secret_key.rs
SergioBenitez: That sounds good. `cookie-rs` has slightly different requirements. It may be worthwhile to change our algorithm there too, at least to `aes-gcm-siv` but the trade-off is different there. It's something I'm considering, but it's orthogonal to this PR.

SergioBenitez (changes requested):

va-an:
Rebased on master.

the10thWiz:
There has been an issue with CI on Windows, so you will need to merge the latest commit from master before CI will pass.

va-an:
> There has been an issue with CI on Windows, so you will need to merge the latest commit from master before CI will pass.

Understood, I will do it after I finish making changes based on the comments. 
Thanks!

core/lib/src/config/secret_key.rs
va-an: We can implement it in different ways.

- change the current method by adding an optional argument for AAD:
```rust
pub fn encrypt<T: AsRef<[u8]>, A: AsRef<[u8]>>(
    &self,
    value: T,
    aad: Option<A>
) -> Result<Cipher, Error> {
```

- or add separate method with additional argument (and accordingly a similar method
for `decrypt`):
```rust
pub fn encrypt_with_aad<T: AsRef<[u8]>, A: AsRef<[u8]>>(
    &self,
    value: T,
    aad: A
) -> Result<Cipher, Error> {
```

I prefer the second way. Despite requiring more code, it might be clearer for the user.
Which one do you like more?

core/lib/src/config/secret_key.rs
va-an: I have added encryption key generation using HKDF, so the original key is no longer stored on the stack. 
I'm not sure I understand about `enсryption()` - do we need to use a different method to obtain the secret key for KDF?

core/lib/src/config/secret_key.rs
va-an: Done!

core/lib/Cargo.toml
```diff
@@ -74,6 +74,11 @@ tokio-stream = { version = "0.1.6", features = ["signal", "time"] }
 cookie = { version = "0.18", features = ["percent-encode"] }
 futures = { version = "0.3.30", default-features = false, features = ["std"] }
 state = "0.6"
+chacha20poly1305 = "0.10.1"
+hkdf = "0.12.4"
+sha2 = "0.10.8"
+base64 = "0.22.1"
+hex = "0.4.3"
```
the10thWiz, line 81: All of these should probably be feature-gated behind the `secrets` feature.

core/lib/src/config/secret_key.rs
```diff
@@ -178,6 +228,70 @@ impl SecretKey {
...
+    /// Extracts the nonce from the data and uses it for decryption.
+    /// Returns the decrypted Vec<u8>.
+    pub fn decrypt(&self, encrypted: &Cipher) -> Result<Vec<u8>, Error> {
+        let encrypted = encrypted.as_bytes();
+
+        // Check if the length of decoded data is at least the length of the nonce
+        let nonce_len = <XChaCha20Poly1305 as AeadCore>::NonceSize::USIZE;
+        if encrypted.len() <= nonce_len {
+            return Err(Error::EncryptedDataLengthError);
+        }
```
the10thWiz, line 281: This should not need to be checked here. Rather, it should be checked when `Cipher` is constructed.

core/lib/src/config/secret_key.rs
```diff
@@ -269,3 +383,53 @@ impl fmt::Debug for SecretKey {
...
+    pub fn from_hex(hex: &str) -> Result<Self, Error> {
+        let decoded  = hex::decode(hex).map_err(|_| Error::HexDecodeError)?;
+        Ok(Cipher(decoded))
+    }
+
+    /// Create a `Cipher` from a base64 string.
+    pub fn from_base64(base64: &str) -> Result<Self, Error> {
+        let decoded = URL_SAFE.decode(base64).map_err(|_| Error::Base64DecodeError)?;
+        Ok(Cipher(decoded))
+    }
```
the10thWiz, line 408: All of these methods should validate that the length is longer than the Nonce, and that the length (minus the Nonce) is a multiple of the block size. Ideally, the only error that should be caught when decrypting the value is that the value was not created by encrypting with the same secret key.

core/lib/src/config/secret_key.rs
```diff
@@ -80,6 +98,38 @@ pub struct SecretKey {
...
+///
+/// Creating a `Cipher` from a base64 string:
+/// ```
+/// let base64_str = "c29tZSBlbmNyeXB0ZWQgZGF0YQ==";
+/// let cipher = Cipher::from_base64(base64_str).unwrap();
+/// ```
+///
+/// Converting a `Cipher` back to bytes:
+/// ```
+/// let bytes = cipher.as_bytes();
```
the10thWiz, line 128: Many of these doc-tests fail. Use `cargo test --doc` to run them locally.

core/lib/src/config/secret_key.rs
```diff
@@ -80,6 +98,38 @@ pub struct SecretKey {
...
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Cipher(Vec<u8>);
```
the10thWiz, line 131: To make Cipher more useful, it could implement `FromForm`, `FromParam`, `FromSegment`, and `FromData`. They should either check the format (which may not be possible, since hex could be re-interpreted as base-64), or always use one specific encoding (and provide a method like `encode` that encodes it using the same encoding).

the10thWiz: Similarly, we could implement `Serialize` and `Deserialize` if the `serde` feature is enabled.

the10thWiz (changes requested):
I think is getting closer to a complete PR, but there are still a number of changes that need to be made.

va-an:
@the10thWiz Thanks for the thorough review!
I merged the changes from master that you mentioned and moved the dependencies to the "secrets" feature. 
I will address the remaining comments a bit later.

SergioBenitez:
@va-an Checking in. Any update?

va-an:
I apologize for my absence. Please allow me to take a deep breath, update the branch, and finish with this PR.

vadim-anfv:
Sorry, I won't be able to continue with this task.
Thanks for the review and your patience!

May the Force be with you! :rocket:

PR #2806, opened 2024-06-06
Fix broken documentation links in README.md

onkoe:
This is a tiny fix that could be super helpful to some users, as it modifies the documentation links to be correct.

Since the frontend changed its links to be sectioned by Rocket version, the _Quickstart_ and _Getting Started_ pages moved. 

## Unrelated README Nit-Pick

Also, I noticed that the README example still uses `#[macro_use] extern crate` syntax, which is years out of date at this point. It was used to indicate to the compiler that there were macros to import, but past the 2021 Edition, it isn't necessary. 

Now, it just imports all macros from a crate, but these can be shadowed by any other item, and users won't really understand what's going on because it's uncommon. It's not great practice to include `#[macro_use]` for that reason.

I'm happy to replace that usage if you'd like! :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,8 +38,8 @@ Rocket is extensively documented:
   * [Guide]: A detailed guide and reference to Rocket.
   * [API Documentation]: The "rustdocs".
 
-[Quickstart]: https://rocket.rs/guide/quickstart
-[Getting Started]: https://rocket.rs/guide/getting-started
+[Quickstart]: https://rocket.rs/guide/v0.5/quickstart/
+[Getting Started]: https://rocket.rs/guide/v0.5/getting-started/
 [Overview]: https://rocket.rs/overview/
 [Guide]: https://rocket.rs/guide/
 [API Documentation]: https://api.rocket.rs


SergioBenitez:
Those links should actually be redirecting to the related documentation for the current release. That is, `rocket.rs/guide/foo` should redirect to `https://rocket.rs/guide/vx/foo` for the current release `x`. That doesn't appear to be happening; I'll see if I can push on fix on the site end.

> Also, I noticed that the README example still uses #[macro_use] extern crate syntax, which is years out of date at this point. It was used to indicate to the compiler that there were macros to import, but past the 2021 Edition, it isn't necessary.

It's not out-of-date, per se, and we explain why we prefer doing this in the [overview](https://rocket.rs/guide/v0.5/overview/#routing). I'll copy it here:

> Throughout this guide and the majority of Rocket's documentation, we import rocket explicitly with #[macro_use] even though the Rust 2018 edition makes explicitly importing crates optional. However, explicitly importing with #[macro_use] imports macros globally, allowing you to use Rocket's macros anywhere in your application without importing them explicitly [each time].
>
> You may instead prefer to import macros explicitly or refer to them with absolute paths: use rocket::get; or #[rocket::get].

onkoe:
Got it, I'm glad this helped to locate the issue! :)

As for the Overview page, thank you for the link! Would you consider a tiny PR that links to the [Rust Reference page on this](https://doc.rust-lang.org/reference/macros-by-example.html#the-macro_use-attribute)? I can imagine a scenario where someone does this for all their macros and hits that once-in-a-lifetime namespace case... 🥹

SergioBenitez:
Sure!

PR #2814, opened 2024-06-26
Typed catchers

### LINKED ISSUE #749: Filtering requests

the10thWiz:
Addresses #749, adds support for typed catchers via the `transient` crate.

When a `FromRequest`, `FromParam`, `FromQuery`, or `FromData` implementation returns an `Error` or `Forward`, if the returned error implements `Any<Co<'r>>`, it is boxed and travels with the `Outcome`.

A catcher can have a third parameter, now `(&E, Status, &Request)`, where `E` is some type that implements `Any<Co<'r>>`. If the error contained in the `Outcome` is of the same type, it is downcast to
the specific type, and passed to the catcher.

A working example can be seen in the `error-handling` example - simply visit `/hello/name/a` to see a `ParseIntError` returned by the `FromParam` of `i8`.

# TODO

- [x] Better syntax for passing the error type to catchers. This is likely going to look like a new `FromError` trait, and a `error = "<error>"` parameter for the upcast error.
  - [x] Requires new compile tests
- [x] Catchers should carry an `Option<TypeId>`, and use it to prevent false collisions.
  - [x] Update matcher code to also check `TypeId`
- [x] Special `TypedError` trait
- [x] `Responder` should return an `Outcome<..., Self::Error, ...>`
- [x] Improve example error message
- [x] `Any<Co<'r>>` should be `Any<Inv<'r>>`. Transient actually already protects us from casting `Any<Inv<'r>>` to a different `Any<Inv<'l>>`, by telling the compiler that `Inv<'a>` is invariant wrt `'a`.
- [x] Documentation needs to be written - this is a pretty big change, and somewhat difficult to use. There are a number of small gotchas - from types that don't implement `Transient`, to simply using the wrong error type in the catcher. I'm not sure what the right way to present this is.
- [x] We should consider logging or carrying the error type's name, to help users identify what the type is, and why their catcher doesn't catch it. (Also log if the type implements `Transient`, since many types don't).
  - [x] Tests (likely using a tracing subscriber) should validate this (I'm not sure how, since `type_name` isn't a stable name - nevermind, it is, since the test is compiled with the same version).
- [x] Export a `TypedError` derive macro.
  - [x] I'm current involved with @JRRudy1 in working on an improved derive macro, which would be safe to use.
  - [x] Also derive a `Transient` implementation at the same time.
- [x] `rocket::form::Errors` (and it's parts) should implement Transient. (Now non-issue)
- [x] `FromError` trait
- [x] Fairings should be able to reject requests (and include a `TypedError`)
- [x] Clean up TODOs
- [x] Contrib
  - [x] dyn_templates: The responder and fairing impls need to be updated.
  - [x] (sync_)db_pools: The error types should implement `TypedError`.
  - [x] ws: The error types should implement `TypedError`, and responder/fairing impls need to be updated.

### DIFF
--- a/core/codegen/src/derive/responder.rs
+++ b/core/codegen/src/derive/responder.rs
@@ -70,9 +70,7 @@ pub fn derive_responder(input: proc_macro::TokenStream) -> TokenStream {
         )
         .inner_mapper(MapperBuild::new()
             .with_output(|_, output| quote! {
-                fn respond_to(self, __req: &'r #Request<'_>)
-                    -> #_response::Result<'r, 'o>
-                {
+                fn respond_to(self, __req: &'r #Request<'_>) -> #_response::Result<'r, 'o> {
                     #output
                 }
             })
--- a/core/codegen/src/derive/responder.rs
+++ b/core/codegen/src/derive/responder.rs
@@ -1,3 +1,8 @@
+// use quote::ToTokens;
+// use crate::{exports::{*, Status as _Status}, syn_ext::IdentExt};
+// use devise::{*, ext::{TypeExt, SpanDiagnosticExt}};
+// use crate::http_codegen::{ContentType, Status};
+
 use quote::ToTokens;
 use devise::{*, ext::{TypeExt, SpanDiagnosticExt}};
 use proc_macro2::TokenStream;
@@ -6,12 +11,12 @@ use crate::exports::*;
 use crate::syn_ext::{TypeExt as _, GenericsExt as _};
 use crate::http_codegen::{ContentType, Status};
 
+
 #[derive(Debug, Default, FromMeta)]
 struct ItemAttr {
     content_type: Option<SpanWrapped<ContentType>>,
-    status: Option<SpanWrapped<Status>>,
+    status: Option<SpanWrapped<Status>>,            
 }
-
 #[derive(Default, FromMeta)]
 struct FieldAttr {
     ignore: bool,
@@ -65,7 +70,9 @@ pub fn derive_responder(input: proc_macro::TokenStream) -> TokenStream {
         )
         .inner_mapper(MapperBuild::new()
             .with_output(|_, output| quote! {
-                fn respond_to(self, __req: &'r #Request<'_>) -> #_response::Result<'o> {
+                fn respond_to(self, __req: &'r #Request<'_>)
+                    -> #_response::Result<'r, 'o>
+                {
                     #output
                 }
             })
--- a/core/codegen/src/derive/responder.rs
+++ b/core/codegen/src/derive/responder.rs
@@ -15,7 +15,7 @@ use crate::http_codegen::{ContentType, Status};
 #[derive(Debug, Default, FromMeta)]
 struct ItemAttr {
     content_type: Option<SpanWrapped<ContentType>>,
-    status: Option<SpanWrapped<Status>>,            
+    status: Option<SpanWrapped<Status>>,
 }
 #[derive(Default, FromMeta)]
 struct FieldAttr {
--- a/core/lib/src/catcher/catcher.rs
+++ b/core/lib/src/catcher/catcher.rs
@@ -154,22 +154,34 @@ impl Catcher {
     ///
     /// ```rust
     /// use rocket::request::Request;
-    /// use rocket::catcher::{Catcher, BoxFuture};
+    /// use rocket::catcher::{Catcher, BoxFuture, TypedError};
     /// use rocket::response::Responder;
     /// use rocket::http::Status;
     ///
-    /// fn handle_404<'r>(status: Status, req: &'r Request<'_>) -> BoxFuture<'r> {
-    ///    let res = (status, format!("404: {}", req.uri()));
-    ///    Box::pin(async move { res.respond_to(req) })
+    /// fn handle_404<'r>(
+    ///     status: Status,
+    ///     _: &'r dyn TypedError<'r>,
+    ///     req: &'r Request<'_>
+    /// ) -> BoxFuture<'r> {
+    ///     let res = (status, format!("404: {}", req.uri()));
+    ///     Box::pin(async move { res.respond_to(req).map_err(|e| e.into()) })
     /// }
     ///
-    /// fn handle_500<'r>(_: Status, req: &'r Request<'_>) -> BoxFuture<'r> {
-    ///     Box::pin(async move{ "Whoops, we messed up!".respond_to(req) })
+    /// fn handle_500<'r>(
+    ///     status: Status,
+    ///     _: &'r dyn TypedError<'r>,
+    ///     req: &'r Request<'_>
+    /// ) -> BoxFuture<'r> {
+    ///     Box::pin(async move{ "Whoops, we messed up!".respond_to(req).map_err(|e| e.into()) })
     /// }
     ///
-    /// fn handle_default<'r>(status: Status, req: &'r Request<'_>) -> BoxFuture<'r> {
+    /// fn handle_default<'r>(
+    ///     status: Status,
+    ///     _: &'r dyn TypedError<'r>,
+    ///     req: &'r Request<'_>
+    /// ) -> BoxFuture<'r> {
     ///    let res = (status, format!("{}: {}", status, req.uri()));
-    ///    Box::pin(async move { res.respond_to(req) })
+    ///    Box::pin(async move { res.respond_to(req).map_err(|e| e.into()) })
     /// }
     ///
     /// let not_found_catcher = Catcher::new(404, handle_404);
@@ -207,13 +219,17 @@ impl Catcher {
     ///
     /// ```rust
     /// use rocket::request::Request;
-    /// use rocket::catcher::{Catcher, BoxFuture};
+    /// use rocket::catcher::{Catcher, BoxFuture, TypedError};
     /// use rocket::response::Responder;
     /// use rocket::http::Status;
     ///
-    /// fn handle_404<'r>(status: Status, req: &'r Request<'_>) -> BoxFuture<'r> {
+    /// fn handle_404<'r>(
+    ///    status: Status,
+    ///    _e: &'r dyn TypedError<'r>,
+    ///    req: &'r Request<'_>
+    /// ) -> BoxFuture<'r> {
     ///    let res = (status, format!("404: {}", req.uri()));
-    ///    Box::pin(async move { res.respond_to(req) })
+    ///    Box::pin(async move { res.respond_to(req).map_err(|e| e.into()) })
     /// }
     ///
     /// let catcher = Catcher::new(404, handle_404);
@@ -233,14 +249,18 @@ impl Catcher {
     ///
     /// ```rust
     /// use rocket::request::Request;
-    /// use rocket::catcher::{Catcher, BoxFuture};
+    /// use rocket::catcher::{Catcher, BoxFuture, TypedError};
     /// use rocket::response::Responder;
     /// use rocket::http::Status;
     /// # use rocket::uri;
     ///
-    /// fn handle_404<'r>(status: Status, req: &'r Request<'_>) -> BoxFuture<'r> {
+    /// fn handle_404<'r>(
+    ///    status: Status,
+    ///    _e: &'r dyn TypedError<'r>,
+    ///    req: &'r Request<'_>
+    /// ) -> BoxFuture<'r> {
     ///    let res = (status, format!("404: {}", req.uri()));
-    ///    Box::pin(async move { res.respond_to(req) })
+    ///    Box::pin(async move { res.respond_to(req).map_err(|e| e.into()) })
     /// }
     ///
     /// let catcher = Catcher::new(404, handle_404);
@@ -287,13 +307,17 @@ impl Catcher {
     ///
     /// ```rust
     /// use rocket::request::Request;
-    /// use rocket::catcher::{Catcher, BoxFuture};
+    /// use rocket::catcher::{Catcher, BoxFuture, TypedError};
     /// use rocket::response::Responder;
     /// use rocket::http::Status;
     ///
-    /// fn handle_404<'r>(status: Status, req: &'r Request<'_>) -> BoxFuture<'r> {
+    /// fn handle_404<'r>(
+    ///     status: Status,
+    ///     _: &'r dyn TypedError<'r>,
+    ///     req: &'r Request<'_>
+    /// ) -> BoxFuture<'r> {
     ///    let res = (status, format!("404: {}", req.uri()));
-    ///    Box::pin(async move { res.respond_to(req) })
+    ///    Box::pin(async move { res.respond_to(req).map_err(|e| e.into()) })
     /// }
     ///
     /// let catcher = Catcher::new(404, handle_404);
@@ -321,7 +345,11 @@ impl Catcher {
 
 impl Default for Catcher {
     fn default() -> Self {
-        fn handler<'r>(status: Status, e: &'r dyn TypedError<'r>, req: &'r Request<'_>) -> BoxFuture<'r> {
+        fn handler<'r>(
+            status: Status,
+            e: &'r dyn TypedError<'r>,
+            req: &'r Request<'_>
+        ) -> BoxFuture<'r> {
             Box::pin(async move { Ok(default_handler(status, e, req)) })
         }
 
--- a/core/lib/src/catcher/catcher.rs
+++ b/core/lib/src/catcher/catcher.rs
@@ -1,12 +1,16 @@
 use std::fmt;
 use std::io::Cursor;
 
+use transient::TypeId;
+
 use crate::http::uri::Path;
 use crate::http::ext::IntoOwned;
 use crate::response::Response;
 use crate::request::Request;
 use crate::http::{Status, ContentType, uri};
 use crate::catcher::{Handler, BoxFuture};
+
+use super::TypedError;
 
 /// An error catching route.
 ///
@@ -128,6 +132,9 @@ pub struct Catcher {
     /// This is -(number of nonempty segments in base).
     pub(crate) rank: isize,
 
+    /// TypeId to match against
+    pub(crate) type_id: Option<TypeId>,
+
     /// The catcher's file, line, and column location.
     pub(crate) location: Option<(&'static str, u32, u32)>,
 }
@@ -188,6 +195,7 @@ impl Catcher {
             base: uri::Origin::root().clone(),
             handler: Box::new(handler),
             rank: rank(uri::Origin::root().path()),
+            type_id: None,
             code,
             location: None,
         }
@@ -313,8 +321,8 @@ impl Catcher {
 
 impl Default for Catcher {
     fn default() -> Self {
-        fn handler<'r>(s: Status, req: &'r Request<'_>) -> BoxFuture<'r> {
-            Box::pin(async move { Ok(default_handler(s, req)) })
+        fn handler<'r>(status: Status, e: &'r dyn TypedError<'r>, req: &'r Request<'_>) -> BoxFuture<'r> {
+            Box::pin(async move { Ok(default_handler(status, e, req)) })
         }
 
         let mut catcher = Catcher::new(None, handler);
@@ -331,7 +339,9 @@ pub struct StaticInfo {
     /// The catcher's status code.
     pub code: Option<u16>,
     /// The catcher's handler, i.e, the annotated function.
-    pub handler: for<'r> fn(Status, &'r Request<'_>) -> BoxFuture<'r>,
+    pub handler: for<'r> fn(Status, &'r dyn TypedError<'r>, &'r Request<'_>) -> BoxFuture<'r>,
+    /// TypeId to match against
+    pub type_id: Option<TypeId>,
     /// The file, line, and column where the catcher was defined.
     pub location: (&'static str, u32, u32),
 }
@@ -343,6 +353,7 @@ impl From<StaticInfo> for Catcher {
         let mut catcher = Catcher::new(info.code, info.handler);
         catcher.name = Some(info.name.into());
         catcher.location = Some(info.location);
+        catcher.type_id = info.type_id;
         catcher
     }
 }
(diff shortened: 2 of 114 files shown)


SergioBenitez (changes requested):
Thank you so, so much for this. As you know, this is a feature that's been
requested for quite some time now, so it's exciting to see some real progress
here.

While reviewing this PR, I found myself consistently asking:
  - Does this overcome the fundamental technical hurdles that have plagued this
    getting this feature into Rocket?
  - Does this meet the usability requirements we expect?
  - Is this feature-complete enough to avoid missing expectation due to missing
    features?

I _believe_ the answer to the first question is _yes_, which is truly exciting!
I temper this statement only because I also believe we're missing some key
usability-impacting features, and there's a concern that the approach won't
extend to these features, but at a first glance it seems probable.

## Feature-Completeness

This PR current allows errors from input guards (`FromParam`, `FromSegments`,
`FromForm`, and `FromData`) to be type-erased and later downcast into a concrete
type in an error catcher. The critical missing piece is allowing the same for
responders. In particular, we want the following to compile and work as one
might expect:

```rust
#[derive(Debug, Transient /* Error */)]
struct MyType<'r>(&'r str);

#[get("/<foo>")]
fn hello(foo: &str) -> Result<&'static str, MyType<'_>> {
    Err(MyType(foo))
}

#[catch(default)]
fn error(foo: &MyType) -> &str {
    foo.0
}
```

This means that `Responder` needs to change to accommodate an error type.

```rust
type Outcome<'o> = Outcome<Response<'o>, E, Status>;

pub trait Responder<'r, 'o: 'r> {
    type Error: Debug + 'o;

    fn respond_to(self, request: &'r Request<'_>) -> Outcome<'o, Self::Error>;
}
```

Where:

  - `Outcome::Success(Response)` is akin to today's `Ok(Response)`
  - `Outcome::Forward(Status)` is akin to today's `Err(Status)` and can be read
    as "forwarding to the catcher for `Status`"
  - `Outcome::Error(Self::Error)` is new. For many types, `Self:Error` will be
    set to `Infallible`, implying that the responder only succeeds or forwards.

One thing to consider is that we probably want to associate a status with the
error, and we also want to allow the error to respond directly if nothing
catches it.

Here we have two obvious approaches:

  * Require `Responder::Error: Responder`, apply the static dispatch
    specialization trick to check for `T: Transient` as this PR does. Check that
    we don't run into an infinite error-loop (i.e, a responder that returns
    itself as an error which then errors with itself, and so on).
  * Create a new trait, `Error`, and apply the static dispatch specialization
    trick to check for `T: Error`. The trait might look something like:

    ```rust
    trait Error<'r>: Transient + .. {
        fn respond_to(&self, request: &'r Request<'_>) -> Result<Response<'r>, Status> {
            Err(self.status())
        }

        /// Maybe? Ignore for now.
        fn status(&self) -> Status { Status::InternalServerError }

        /// Log on catcher failure: expected type Foo, have error Bar.
        fn type_name(&self) -> &'static str { .. }

        fn upcast(self) -> Box<dyn Transient<'r>> { .. }

        fn source(&self) -> Option<Box<dyn Error<'r>>> { .. }
    }
    ```

It's not obvious to me at this point which approach might be better, or if
there's another approach that's missing here. The main difference is that the
former _requires_ the error type to implement `Responder`, which may not be what
we want, and furthermore that the `Responder` implementation not vary between
being used as a regular response and as an error. As such, I lean towards the
latter approach.

## Usability

My main concerns on the usability front are two-fold:

  * the extensions to the `#[catch]` macro. In particular, I'd like to find a
    way to remove the need for additional attribute arguments of `status` and
    perhaps even `error`.
  * the difficulty in implementing custom downcastable errors

### `#[catch]`

In other attributes, like `#[route]`, we _need_ attribute parameters because
either they provide information about what to parse (path, segments, and query
params) or they operate in a fundamentally different manner at the type-system
level: (data and request). There's also a security and category-theoretical
argument to be made about creating different class types for parseable
parameters.

But with `status` and `error`, none of these argument need necessarily apply:
all of the information derives from references or copy-able types without and
particular security considerations. So the only reason* (there's one reason we
might want an `error` argument, actually) for the arguments to exist is for our
sake in implementing codegen.

More concretely, the `#[catch]` attribute separates types into three categories:

  * `T: FromRequest`
  * `T: Transient` via `error =`
  * `Status` via `status =`

I argue that in reality, these should all be one:

  * `T: FromError`

    ```rust
    trait FromError<'r> {
        fn from_error(request: &'r Request<'_>, error: Option<ErasedError>);
    }
    ```

Then we can recover all of the obvious implementations:

```rust
impl<'r> FromError<'r> for Status {
    fn from_error(request: &'r Request<'_>, error: Option<ErasedError>) {
        error.as_ref().map_or(Status::InternalServerError, |e| e.status())
    }
}
```

### Implementing Custom Errors

Consider the following complete program:

```rust
use rocket::*;
use rocket::request::{self, Request, FromRequest};
use rocket::outcome::Outcome;
use rocket::http::Status;

#[derive(Debug, transient::Transient)]
struct MyError<'r>(&'r str);

#[derive(Debug)]
struct MyGuard;

#[rocket::async_trait]
impl<'r> FromRequest<'r> for MyGuard {
    type Error = MyError<'r>;

    async fn from_request(req: &'r Request<'_>) -> request::Outcome<Self, Self::Error> {
        Outcome::Error((Status::InternalServerError, MyError(req.uri().path().as_str())))
    }
}

#[get("/")]
fn index(guard: MyGuard) -> &'static str {
    "Hello, world!"
}

#[catch(default, error = "<foo>")]
fn error<'a>(foo: &MyError<'a>) -> &'a str {
    foo.0
}

#[launch]
fn rocket() -> _ {
    rocket::build()
        .mount("/", routes![index])
        .register("/", catchers![error])
}
```

There are a few issues here:

  * I needed to depend on `transient` directly or else the derive wouldn't work
    since it references `transient::`. This shouldn't need to be the case.
  * The error catcher fails at runtime because, for whatever reason, the error
    type `MyError` is not meeting the specialization requirements. I would
    expect that if my type doesn't meet the requirements to be an error, then
    I wouldn't be allowed to use it as typed error at all, and my program would
    fail to compile.
  * Covariance needs to be declared: can we figure it out automatically, i.e,
    via by calling `rustc`, or for references syntactically (&T)?

This also begs the question: what exactly is required to be implemented of error
types? `Transient` does not appear to be sufficient. This further supports the
idea of a Rocket-specific `Error` trait that has all of the requirements
built-in. And if `Error` is a Rocket trait, then we can re-export the derive
to allow users to do the following, completely relegating `transient` to an
implementation detail.

```rust
#[derive(rocket::Error)]
struct MyError<'r> { .. }
```

-----


As an aside, the following gives an error message that we should look to improve:

```rust
#[derive(Debug, Responder, Transient)]
struct MyType<'r>(&'r str);

#[catch(default, error = "<foo>")]
fn error<'a>(uri: &Origin, foo: &'a MyType) -> &'a str {
    foo.0
}
```

core/lib/src/catcher/catcher.rs
```diff
@@ -132,10 +131,11 @@ pub struct Catcher {
...
 }
 
-// The rank is computed as -(number of nonempty segments in base) => catchers
+// The rank is computed as -(number of nonempty segments in base) *2 => catchers
 // with more nonempty segments have lower ranks => higher precedence.
+// Doubled to provide space between for typed catchers.
 fn rank(base: Path<'_>) -> isize {
-    -(base.segments().filter(|s| !s.is_empty()).count() as isize)
+    -(base.segments().filter(|s| !s.is_empty()).count() as isize) * 2
 }
```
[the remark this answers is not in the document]
the10thWiz: This logic has been fully removed. The preference for typed catchers is now implemented in lifecycle.rs, since it also needs to check the error's `.source()` method, it now always checks for a matching type before a catcher with any type.

the10thWiz:
@SergioBenitez I've made some significant progress, and I've found a couple ways to improve the API. I wanted to put them in writing, both so you can comment on them sooner rather than later, and for future reference.

## The 'final' API

I've opted to remove `Status` from anywhere a `TypedError` is returned, rather relying on the error to provide a status. To aid in this, there are `TypedError` implementations for `Status`, `(Status, T)` and `Custom<T>` (`where T: TypedError`), so the status can be customized without being forced to add a status member.

The final form of the `TypedError` trait:

```rust
// `AsAny` is a sealed/blanket impl for a `fn as_any(&self) -> &dyn Any`
pub trait TypedError<'r>: AsAny<Inv<'r>> + Send + Sync + 'r {
    /// Generates a default response for this type (or forwards to a default catcher)
    fn respond_to(&self, request: &'r Request<'_>) -> Result<Response<'r>, Status> {
        Err(self.status())
    }

    /// A descriptive name of this error type. Defaults to the type name.
    fn name(&self) -> &'static str { std::any::type_name::<Self>() }

    /// The error that caused this error. Defaults to None.
    ///
    /// # Warning
    /// A typed catcher will not attempt to follow the source of an error
    /// more than (TODO: exact number) 5 times.
    fn source(&'r self) -> Option<&'r (dyn TypedError<'r> + 'r)> { None }

    /// Status code
    fn status(&self) -> Status { Status::InternalServerError }
}
```

It should be fairly obvious what these methods get used for.

One change, is that `.source()` will be called multiple times (currently up to 5). I'm preventing loops by simply placing a reasonable upper limit on the number of times it can be called, since I don't think there is a better option. (I considered checking for pointer equality with a previous value, but I'm not confident we can rely on this, since ZSTs don't have a meaningful address, and false positives might be possible).

## Remaining decisions

Is `TypedError` too onerous of a trait bound for error types? I don't think so. Pretty much every std error is already a typed error, and `Status` can be used if you don't want to deal with an actual type. The derive macro has very few requirements (we only support a single lifetime parameter), so any type you control can pretty trivially implement `TypedError`. The one exception would be third-party errors (such as `sqlite::Error`, which obviously doesn't implement `TypedError`), but we could provide some kind of `Debug<E>` impl, that enables any `'static` to act as a `TypedError`. I also don't think it's a significant burden to ask rocket-specific crates to add `TypedError` to their error types, and wrap any error types they don't control.

With this in mind, if it isn't too much of a burden, we might be able to eliminate the specialization trick, since each of these error types would be required to implement `TypedError`. This would also allow `Responder` to return a `Result<Response, Self::Error>`, since it provides all the functionality we would require.

The main alternative would be to require some kind of `HasStatus` trait, with a single `fn status(&self) -> Status` method, and specialize on `TypedError`. However, I don't think there is much value in allowing types to implement `HasStatus` without `TypedError` - the only additional req is `Transient`, which is trivially derived with the `TypedError` derive.

I'm also considering removing `respond_to` from `TypedError`, since it's pretty unlikely it actually gets called. Any catcher (even an untyped one) prevents the `respond_to` method from getting called.

## Potential issues

The `respond_to` method is tried after untyped catchers, so an untyped catcher can prevent `respond_to` from ever getting called. Because of this, a default catcher mounted at the root will prevent `respond_to` from ever getting called, for any error.

We likely need to provide error wrappers for the contrib db crates, since there really isn't a good way to implement `TypedError` for the DB error types.

the10thWiz:
There are a few major updates with this. It's a partial rewrite, based on the latest version of main, along with a few other changes along the way.

First, the `TypedError`'s `source()` method now takes a parameter, so that it can return multiple errors. The only use-case in current code is `Either<A, B>`, where the error type is `(A, B)`.

Second, we now use a git dependency for `transient`, since there are a few changes that have not yet been merged into upstream. There are also some other issues I need to sort out with the CI for transient, but that shouldn't prevent reviewing this PR.

Third, `TypedError` is no longer optional, and the specialization trick is no longer used. We provide two primary options for implementing `TypedError`: the derive macro, which can be applied to almost any type, and `Debug<E>`, which provides a default implementation for any `'static` type.

Overall, I'm happy with this PR. From updating the tests, and a personal project I'm working on at the same time, there are only a couple of pain points. Since I'm deep in this project, I can't say with any level of certainty how easy it is to identify what type you should use in the catcher. I hope it's fairly easy with the `error_name` parameter in the trace logs. Also, a number of examples in Rocket itself use `Result` as a responder, in a way that no longer works, since `Result` now converts the `Err` variant into a `TypedError` that can be caught by a catcher. I rewrote the examples to use `Either`, but it did make them more verbose.

@SergioBenitez Please review this when you get a chance.

the10thWiz:
@SergioBenitez Awesome!

SergioBenitez:
@the10thWiz This needs to be rebased on master.

examples/cookies/src/session.rs
```diff
@@ -3,6 +3,7 @@ use rocket::request::{self, FlashMessage, FromRequest, Request};
 use rocket::response::{Redirect, Flash};
 use rocket::http::{CookieJar, Status};
 use rocket::form::Form;
+use rocket::either::Either;
```
SergioBenitez, line 6: I'd also import `Left` and `Right` to use them naturally.

examples/state/src/request_local.rs
```diff
@@ -1,6 +1,6 @@
 use std::sync::atomic::{AtomicUsize, Ordering};
 
-use rocket::State;
+use rocket::{State, StateMissing};
```
SergioBenitez, line 3: Should probably be named `StateError` or something? Looks to be what the stdlib prefers.

examples/tls/Cargo.toml
```diff
@@ -7,6 +7,7 @@ publish = false
 
 [dependencies]
 rocket = { path = "../../core/lib", features = ["tls", "mtls", "secrets", "http3-preview"] }
+tracing = "*"
```
SergioBenitez, line 10: We definitely don't want to do this. I fixed this upstream already - check it out.

examples/todo/src/main.rs
```diff
@@ -13,6 +13,7 @@ use rocket::response::{Flash, Redirect};
 use rocket::serde::Serialize;
 use rocket::form::Form;
 use rocket::fs::{FileServer, relative};
+use rocket::either::Either;
```
SergioBenitez, line 16:
```suggestion
use rocket::either::{Either, Left, Right};
```

examples/todo/src/main.rs
```diff
@@ -64,23 +65,27 @@ async fn new(todo_form: Form<Todo>, conn: DbConn) -> Flash<Redirect> {
...
+async fn toggle(id: i32, conn: DbConn) -> Either<Redirect, Template> {
     match Task::toggle_with_id(id, &conn).await {
-        Ok(_) => Ok(Redirect::to("/")),
+        Ok(_) => Either::Left(Redirect::to("/")),
         Err(e) => {
             error!("DB toggle({id}) error: {e}");
-            Err(Template::render("index", Context::err(&conn, "Failed to toggle task.").await))
+            Either::Right(
+                Template::render("index", Context::err(&conn, "Failed to toggle task.").await)
+            )
```
SergioBenitez, line 75:
```suggestion
            Right(Template::render("index", Context::err(&conn, "Failed to toggle task.").await))
```

scripts/mk-docs.sh
```diff
@@ -25,7 +25,7 @@ pushd "${PROJECT_ROOT}" > /dev/null 2>&1
       --crate-version ${DOC_VERSION} \
       --enable-index-page \
       --generate-link-to-definition" \
-      cargo doc -Zrustdoc-map --no-deps --all-features \
+      cargo +nightly doc -Zrustdoc-map --no-deps --all-features \
```
SergioBenitez, line 28: Is this actually necessary?

core/codegen/src/derive/responder.rs
```diff
@@ -65,7 +70,9 @@ pub fn derive_responder(input: proc_macro::TokenStream) -> TokenStream {
         )
         .inner_mapper(MapperBuild::new()
             .with_output(|_, output| quote! {
-                fn respond_to(self, __req: &'r #Request<'_>) -> #_response::Result<'o> {
+                fn respond_to(self, __req: &'r #Request<'_>)
+                    -> #_response::Result<'r, 'o>
+                {
```
SergioBenitez, line 75:
```suggestion
                fn respond_to(self, __req: &'r #Request<'_>) -> #_response::Result<'r, 'o> {
```

core/lib/src/catcher/catcher.rs
```diff
@@ -354,6 +393,7 @@ impl fmt::Debug for Catcher {
             .field("base", &self.base)
             .field("code", &self.code)
             .field("rank", &self.rank)
+            .field("type_id", &self.type_id.as_ref().map(|_| "TY"))
```
SergioBenitez, line 396: Why not the following?

```suggestion
            .field("type_id", &self.type_id)
```

core/lib/src/fairing/mod.rs
```diff
@@ -153,6 +157,22 @@ pub type Result<T = Rocket<Build>, E = Rocket<Build>> = std::result::Result<T, E
...
 ///     can potentially alter how a request is routed.
 ///
+///   * **<a name="request">RequestFilter</a> (`on_request_filter`)**
+///
+///     A request callback, represented by the [`Fairing::on_request_filter()`] method,
+///     is called just after a request is received, immediately after
+///     pre-processing the request and running all `Request` fairings. This method
+///     returns a `Result`, which can be used to terminate processing of a request,
+///     bypassing the routing process. The error value must be a [`TypedError`], which
+///     can then be caught by a typed catcher.
```
SergioBenitez, line 167: This is very interesting. I apologize if I missed it, but what's the rationale for this approach as opposed to changing request fairings to optionally return an error and abort the request? It seems desirable to be able to abort a request _before_ other request fairings are run, for example, or to be able to choose the order based on how you attach fairings.

SergioBenitez (changes requested):
This looks amazing. Seriously, incredible job @the10thWiz!

I'll need way more time than I budgeted this weekend to give this a proper review, but I wanted to leave a few comments first. At a high level, the typed catchers themselves look absolutely amazing. I think you've made really good decisions. I've left a few comments here and there, but nothing major. However, there's quite a bit of unsafe code being introduced without explanation, and that must be addressed before we can merge this. Better yet, whatever unsafe code can be removed should be.

The only interesting point related to design is the introduction of a new fairing class. I've left a comment on that and look forward to your response.

Finally, I think a bit of the code can be cleaned up, made pithier and more understandable, but for the most part, this is excellent. I'll have more on this when I can perform a thorough review.

scripts/mk-docs.sh
the10thWiz: The `-Z` flags are exclusive to Rust nightly. You might have set up a directory override on your machine (and iirc, the workflow also changes the default for the directory).

I think it's marginally more convenient for running it locally.

core/lib/src/catcher/catcher.rs
the10thWiz: The `TypeId` doesn't actually provide much (if any) useful information, since it looks something like `TypeId(0x474ccf3b5db264ef53916706f7d7bb2c)`. I've updated this to actually print the type name (and, since I have the actual type name, it's now printed as part of the catcher on launch).

SergioBenitez:
Okay, I've had a chance to play with this properly by porting an existing project. A couple of questions, and then some more general thoughts. Questions first:

1. `AsStatus` feels like it should be unnecessary. As far as I can tell, this only really exists to get a `From<Box<dyn TypedError<'_> + '_>> for Status` implementation, but we can implement that directly. Why not do this?
2. Earlier, you wrote:
   > The respond_to method is tried after untyped catchers, so an untyped catcher can prevent respond_to from ever getting called. Because of this, a default catcher mounted at the root will prevent respond_to from ever getting called, for any error.

   Is this still the case? If so, can you provide an example of what can go wrong?
   
 My primary thoughts center around forwarding:
 
 * `FromRequest` has been changed so that both the `Forward` and `Error` variants take a `TypedError`. I wonder if we should allow for the `Forward` type to be different from the `Error` type. 
 * How do I make my `TypedError` actually respond with something? Say I have a type that knows how to convert itself to JSON. What I would like to happen is that when I return it as an error from anywhere, I get the JSON output of that error value. How can I do that? I would expect perhaps something like the following to just work:
 
   ```rust
   #[derive(TypedError, Responder)]
   struct AppError { error: Json<SomeThing>, header: Foo }
   ```
   
   And if I return this as error, the `Responder` `impl` is used as needed. But instead, the above will always forward.

   I think we can fix this fairly transparently by applying the same [trick we use for `Sentinel` implementations](https://github.com/rwf2/Rocket/blob/504efef179622df82ba1dbd37f2e0d9ed2b7c9e4/core/lib/src/sentinel.rs#L461-L493). Basically, if the type implements `Responder`, then we use that implementation in the derived `TypedError` implementation. Otherwise, we generate an implementation that always forwards to the error catcher for the associated status.
   * The `status` from the responder and the `status()` method -- what is the relationship between them, if any? Can we avoid needing both?

Freddo3000:
I hate to ask but is there any plan on completing this for a 0.6 release? This currently seems to be the blocker for further development as it is a fairly large PR and I recall reading somewhere (but I can't remember where now) that this was a target for a 0.6 release. There is a risk of the project starting to splinter off into something such as https://github.com/rocket-rs-community/Rocket

martynp:
> I hate to ask but is there any plan on completing this for a 0.6 release? This currently seems to be the blocker for further development as it is a fairly large PR and I recall reading somewhere (but I can't remember where now) that this was a target for a 0.6 release. There is a risk of the project starting to splinter off into something such as https://github.com/rocket-rs-community/Rocket

I have moved on to a full fork with rename - should be ready in a few weeks.

It was mostly there until mkdocs-material got nuked by mkdocs 2.0 so I have had to rebuild all the documentation again. I am also considering moving to codeberg... GitHub is a mess these days.

But in any case, contributions will be welcome once v1 is released and assuming there are no breaking changes.

PR #2822, opened 2024-07-07
Drop Rocket inside the tokio async context when using launch

the10thWiz:
Moves error printing and the implicit drop inside the tokio runtime when using the `#[launch]` attribute.

Based on an issue reported on Element. See https://matrix.to/#/!kDIcCXWSVfdahNCJWq:mozilla.org/$2CSzlNC4rQw7Wd7Y7L2fV-2_fTxsvq93elL1foVfQ1s?via=mozilla.org&via=matrix.org&via=catgirl.cloud.

Basically, the Rocket instance gets dropped (after a successful shutdown or a `Shutdown` error) outside the tokio runtime. Some minor restructuring of the generated code ensures the tokio context outlives Rocket, if only slightly.

### DIFF
--- a/core/codegen/src/attribute/entry/launch.rs
+++ b/core/codegen/src/attribute/entry/launch.rs
@@ -113,7 +113,9 @@ impl EntryAttr for Launch {
             #[allow(dead_code)] #f
 
             #vis #sig {
-                #_error::Error::report(::rocket::async_main(#launch))
+                ::rocket::async_main(async move {
+                    #_error::Error::report(#launch.await)
+                })
             }
         ))
     }


SergioBenitez (changes requested):
What is this fixing/making possible? Can we add that as a test? If it's system-wide, perhaps an addition to the `testbench` would be reasonable.

the10thWiz (commented):
Needs tests

core/lib/tests/drop-in-async-context.rs
```diff
@@ -0,0 +1,63 @@
...
+    #[test]
+    fn test_execute() {
+        rocket::execute(async {
+            super::rocket()
+                .try_launch_on(TcpListener::bind("localhost:8002"))
+                .await
+                .unwrap();
+        });
+    }
+}
```
SergioBenitez, line 63: This is extremely fragile. The only correct way to start a TCP server in tests is to do so with an unnumbered port.

Given what this code is doing, I _really_ believe this needs to be a test in `testbench`. The tests are easier to write this way, too. All you need to do is introduce a new module in [testbench/src/servers](https://github.com/rwf2/Rocket/tree/db598be3a8e7ed3d5a01d8ef5eddfd6868c22333/testbench/src/servers). The returned handle allows you to shutdown the server as well.

SergioBenitez (changes requested):
The changes to `launch()` are good, but I want the test to live in `testbench`. It'll reduce the amount of code considerably while making the test robust.

core/lib/tests/drop-in-async-context.rs
the10thWiz: Yeah, I should switch to using an unnumbered port.

That being said, I'm not confident I can use testbench for this specific case - at least not without some issues. The challenge is that I'm trying to test the `#[launch]` macro itself - it looks like most cases pass a rocket instance to the spawn macro, which then attaches a fairing. I can't do that - since I'm directly executing the generated (sync) main function.;

the10thWiz: I've taken the time to make an attempt (it required some modifications to testbench, and still doesn't work). You can check it out from the 'drop-in-async-launch-testbench' branch of my fork.

core/lib/tests/drop-in-async-context.rs
```diff
@@ -0,0 +1,67 @@
...
+    fn test_execute() {
+        rocket::execute(async {
+            super::rocket().launch().await.unwrap();
+        });
+    }
+
+    #[test]
+    fn test_execute_directly() {
+        rocket::execute(super::rocket().launch()).unwrap();
+    }
```
SergioBenitez, line 66: I added this test, which I _think_ should pass for the same reasons?

SergioBenitez (changes requested):
Take a look at the new test I added, which currently fails.

core/lib/tests/drop-in-async-context.rs
the10thWiz: What's happening here, is Rocket is created and launched, on the async runtime. However, launch returns `Result<Rocket<Ignite>, Error>`. In the success case, Rocket is then returned by `rocket::execute`, and then dropped outside of it (and the async runtime). For 0.6, we should consider changing the interface of `rocket::execute` to `fn execute<P>(f: impl Future<Output = Rocket<P>>) -> Result<(), Error>`, and calling launch internally.

the10thWiz: This is why the one I wrote explicitly calls unwrap inside an async block.

SergioBenitez: The main use-case is exactly calling it with `rocket.launch()` directly, so it should be able to do that without any footguns or extra ceremony.

SergioBenitez: I'm wondering if there's a better approach altogether. In addition to trying to drop `Rocket` in an async context, why not also specifically drop managed state on shutdown? That is, as part of the shutdown procedure, which is async, drop managed state.

the10thWiz: I think that might be a valid approach, but the tradeoff would be that we can't support relaunching a Rocket application after it's been shutdown.

Given all this, what is the value to Rocket returning the instance after shutdown? The two values I can see is a) relaunching Rocket after shutdown, and b) inspecting managed state after shutdown. The fact that Rocket is dropped in the error case mean it's likely not super useful to inspect managed state, since that's when you would actually want to inspect it. I'm not sure what the exact use case of relaunching Rocket is, and I'm not convinced it wouldn't be better to recommend simply building a new instance of Rocket.

I think the best compromise would be changing launch to not return the Rocket instance, ever. This ensures Rocket is dropped inside an async context (since it's dropped before the launch future completes). Alternatively, we could create a `Shutdown` phase to return, which would have already dropped managed state, and be safe drop outside an async context.

PR #2823, opened 2024-07-10
Update rocket_ws dep to pick up security fix.

hcldan:
This may not be needed...   
This is the CVE that was reported in our code scans: https://nvd.nist.gov/vuln/detail/CVE-2023-43669

Rocket is already above that version, but our scans are still complaining.
I'm investigating, but will leave this open in case you just want to bump your deps.

### DIFF
--- a/contrib/ws/Cargo.toml
+++ b/contrib/ws/Cargo.toml
@@ -20,7 +20,7 @@ default = ["tungstenite"]
 tungstenite = ["tokio-tungstenite"]
 
 [dependencies]
-tokio-tungstenite = { version = "0.21", optional = true }
+tokio-tungstenite = { version = "0.23.1", optional = true }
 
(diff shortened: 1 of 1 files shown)


hcldan:
It may be that the change log didn't mention it being fixed until 0.23.0
https://github.com/snapview/tungstenite-rs/commit/8b3ecd3cc0008145ab4bc8d0657c39d09db8c7e2

SergioBenitez:
> Rocket is already above that version, but our scans are still complaining.

Do you know why? We can still update the dep, by the version we're depending on shouldn't be signaling issues.

hcldan:
I have not heard back from our vendor yet.  Will ping them again.

hcldan:
Vendor got back to me and determined it was an error in their data.
Will close.

PR #2828, opened 2024-07-22
Update mongodb integration to use version 3

abr-egn:
This updates the optional `mongodb` database support to use version 3 of the client crate.

### DIFF
--- a/contrib/db_pools/lib/Cargo.toml
+++ b/contrib/db_pools/lib/Cargo.toml
@@ -65,9 +65,9 @@ features = ["rt_tokio_1"]
 optional = true
 
 [dependencies.mongodb]
-version = "2"
+version = "3"
 default-features = false
-features = ["tokio-runtime"]
+features = ["compat-3-0-0", "rustls-tls"]
 optional = true
 
 [dependencies.diesel-async]
--- a/contrib/db_pools/lib/src/lib.rs
+++ b/contrib/db_pools/lib/src/lib.rs
@@ -131,7 +131,7 @@
 //! On shutdown, new connections are denied. Shutdown waits for connections to
 //! be returned.
 //!
-//! ## `mongodb` (v2)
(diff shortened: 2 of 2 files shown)


SergioBenitez:
Please see the build failures.

abr-egn:
I'm sorry, I don't know how I missed that.  Fixed!

SergioBenitez:
Can you point me to docs for the `compat-3-0-0` feature flag? I don't see it mentioned elsewhere. It seems like we shouldn't need _any_ feature flags - is that not the case?

abr-egn:
`compat-3-0-0` is to ensure forward compatibility - one of the issues we ran into repeatedly in 2.x was there was no way to make existing functionality optional (e.g. for supporting a new crate that provides functionality of an existing dependency).  Requiring a `compat` feature when default features are disabled means that if we needed to make something optional in 3.1 we could also add it to the previous release's `compat` feature and nobody's build would break.

It does look like we don't have this explicitly documented beyond the error you get with default features disabled and I'll fix that shortly 🙂 

As for `rustls-tls` - TLS functionality is required for the driver, so one of `rustls-tls` or `openssl-tls` is needed.  In 2.x, `rustls` was always pulled in as a dependency, even if openssl was used for TLS; in 3.x, that's no longer the case (and if both are enabled, it'll use openssl).

Ideally, that choice could be transparently propagated to the user of `rocket_db_pools`, but (AFAIK) there's no good way to do that.  Enabling `rustls-tls` here gives an equivalent situation to `mongodb` 2.x - `rustls` will always be pulled in, but openssl will be used if it's enabled.

The other option would be to split the `mongodb` feature in `rocket_db_pools` into `mongodb-rustls` and `mongodb-openssl`, which seemed worse to me and won't scale well if other TLS providers or similar features show up.  I'd be happy to update this PR to do that if that's what you'd prefer, though.

Issue #2543, opened 2023-05-18
`FromData` implementation for `Option<T>` causes malformed bodies to be silently ignored.


djrenren:
Rocket Version: 0.5.0-rc.3

**Description**

`FromData` implementation for `Option<T>` causes malformed bodies to be silently ignored.

**To Reproduce**

Sometimes a route handler needs to support an optional request body (for example, when making a backwards-compatible addition to an endpoint that previously required no request body).

So the definition of such a route looks like so:

```rust
#[post("/my_route", data = "<body>")]
pub async fn my_route_handler(body: Option<Json<Foo>>) {
    /* ... */
}
```

The desired behavior is that, if no request data is present, we will get `None` for the body, and if some data is provided, we'll try to parse it. This is mostly how it works except, in the case where the body does not represent a valid `Json<Foo>` that is, `Json::<Foo>::from_data` returns a `Failure` result. Rather than a failure, we get `None`.

**Expected Behavior**

If a request body is supplied, it should be validated. 

**Environment:**

 - OS Distribution and Kernel: all
 - Rocket Version: 0.5.0-rc.3

**Additional Context**

For prior art, we can look at what `serde` does. In general, `serde` follows the model that optional values are validated if supplied, so the following fails:

```rust
use serde::Deserialize;

#[derive(Debug, Deserialize)]
enum Foo {
    MyType
}

#[derive(Debug, Deserialize)]
struct Test {
    #[serde(default)]
    field: Option<Foo>
}

fn main() {
    serde_json::from_str::<Test>(r#"{"field": "garbage"}"#).unwrap();
}
```
[Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021)

That is, because a value was provided for `field`, an incorrect value results in an error, not a silent `None`.

It seems to me that rocket should follow this example because it provides stronger guarantees and more information.
If users want infallible parsing, they could use `Result` and just ignore any errors.

**Proposal for updating `FromData` impl**

Currently, the implementation of `FromData` for `Option<T>` clearly swallows all errors:

```rust
impl<'r, T: FromData<'r>> FromData<'r> for Option<T> {
    type Error = std::convert::Infallible;

    async fn from_data(req: &'r Request<'_>, data: Data<'r>) -> Outcome<'r, Self> {
        match T::from_data(req, data).await {
            Success(v) => Success(Some(v)),
            Failure(..) | Forward(..) => Success(None),
        }
    }
}
```

[...]

PR #2831, opened 2024-07-24
`Option` guards should return `None` only if the value is empty.

the10thWiz:
Based on #2543

Option is usable as pretty much any kind of guard. Right now, it's always infallible, and simply ignored errors. This may be somewhat counter-intuitive for newcomers (and old hats such as myself), and may have some other negative implications. This PR changes the behavior of `Option` to match what one might expect - `None` represents the empty case, as it makes sense for each type of guard.

## The Negative Implications

There are a couple of issues I can see with the current behavior. First, and most obvious, it's likely not what a newcomer to Rocket would expect. We should avoid sweeping errors under the rug, so to speak, and instead leave that choice fully up to the user. Second, there may be a non-trivial overhead to simply discarding data in the case we fail to parse. This is generally mitigated by data limits, but I don't think they represent a full solution. Finally, the current behavior may actually impede debugging efforts. To see how, consider the following route:

```rust
#[get("/?<age>")]
fn r(age: Option<u8>) -> ...
```

Right now, a request to `/?age=a` will call `r(None)` - with no mention (even in the logs) that the guard failed. With this PR, the above request will instead return an Error (which, with typed-catchers, could produce a nice error message for the consumer). This makes it much easier to work with APIs with optional parameters, since they won't just silently ignore parsing errors.

# New behavior

In general, the old behavior can be emulated by using `Result`, and ignoring the `Err` value. We could provide a separate `Option`-like enum that had the old behavior, but it might be easier to simply provide a set of type aliases that fill in the error type automatically, e.g. `type DataResult<T: FromData> = Result<T, T::Error>;`

## FromData

`FromData` now uses `peek` to check whether the data contains any bytes. Iff it is empty, it will succeed with `None`.

## FromParam

This implementation has been removed. This is a more obviously breaking change, but since Rocket ignores empty segments, we can't meaningfully provide this implementation anymore.

## FromSegements

This implementation now only succeeds with `None` if the `Segments` is empty.

## FromForm

This one has different behavior depending on the current strictness. In strict mode, it succeeds with `None` iff there were no values passed to the `FromForm` implementation. Otherwise, it also succeeds with `None` if all the `ErrorKind`s were `Missing`.

## FromRequest

This one has not been altered. There isn't really an easy way to check whether a request guard failed or was simply missing, but it may be reasonable to change this to only catch `Forwards`. In this case, the old behavior can be emulated by `Option<Result>`, (or `Result<Option>`).

# TODO

- [ ] Update documentation
  - [ ] FromData
  - [ ] FromParam
  - [ ] FromSegments
  - [ ] FromForm
  - [ ] FromRequest
- [ ] FromRequest implementation

### DIFF
--- a/core/lib/src/data/from_data.rs
+++ b/core/lib/src/data/from_data.rs
@@ -415,12 +415,14 @@ impl<'r, T: FromData<'r> + 'r> FromData<'r> for Result<T, T::Error> {
 
 #[crate::async_trait]
 impl<'r, T: FromData<'r>> FromData<'r> for Option<T> {
-    type Error = std::convert::Infallible;
+    type Error = T::Error;
 
-    async fn from_data(req: &'r Request<'_>, data: Data<'r>) -> Outcome<'r, Self> {
-        match T::from_data(req, data).await {
-            Success(v) => Success(Some(v)),
-            Error(..) | Forward(..) => Success(None),
-        }
+    async fn from_data(req: &'r Request<'_>, mut data: Data<'r>) -> Outcome<'r, Self> {
+        // Ask for at least one byte of data, if it's empty, there is no body.
+        if data.peek(1).await.is_empty() {
+            Outcome::Success(None)
+        } else {
+            T::from_data(req, data).await.map(Some)
+        }  
     }
 }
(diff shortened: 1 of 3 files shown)


the10thWiz:
@SergioBenitez This is a breaking change, so it needs to land before 0.6 (or be rejected).

Issue #2826, opened 2024-07-18
Implement derive FromParam for enum


loikki:
### API Docs to Existing Functionality

https://api.rocket.rs/v0.5/rocket/request/trait.FromParam

### Problems with Existing Functionality

It seems that FromParam is not implemented as a derive macro for enum. It means that you cannot receive an enum as a path variable.

Implementing it for any enum would not make any sense but why not providing it for enum without internal data (so kind of a string enum based on the variable name).

I can implement it if you are interested.

### Suggested Changes

Simply, we can declare an enum with a derive instruction and then add it as an option in a given route. When we query the endpoint, we can simply send the String corresponding to the enum.

```rust
#[derive(Debug, FromParamEnum)]
pub enum MyEnum {
  Option1,
  Option2
}

#[get("/<option>")]
pub async fn get_option(
    option: MyEnum,
)  {
  println!("{:?}", option);
}

// Query: GET /Option1
```

It requires a relatively simple code to do so:

```rust
extern crate proc_macro;
use proc_macro::TokenStream;
use quote::quote;
use syn::{parse_macro_input, Data, DeriveInput};

// Example of expansion
// impl<'a> FromParam<'a> for MyEnum {
//     type Error = &'a str;

//     fn from_param(param: &'a str) -> Result<Self, Self::Error> {
//         match param {
//             "Option1" => Ok(MyEnum::Option1),
//             "Option2" => Ok(MyEnum::Option2),
//             _ => Err("Failed to find enum"),
//         }
//     }
// }

#[proc_macro_derive(FromParamEnum)]
pub fn derive_macro_param(input: TokenStream) -> TokenStream {
    let input = parse_macro_input!(input as DeriveInput);
    let name = input.ident;

    let mut matches = quote!();

    match input.data {
        Data::Enum(ref data) => {
            for field in data.variants.iter() {
                let field_name = &field.ident;
                matches.extend(quote!(
                    stringify!(#field_name) => Ok(#name::#field_name),
                ))
            }
        }
        _ => panic!("Work only on enums"),
    };

    let expanded = quote! {
        impl<'a> FromParam<'a> for #name {
            type Error = &'a str;

            fn from_param(param: &'a str) -> Result<Self, Self::Error> {
                match param {
                    #matches
                    _ => Err("Failed to find enum")
                }
            }
        }
    };

    proc_macro::TokenStream::from(expanded)
}
```

### Alternatives Considered

The only alternative I found was to write it myself in my crate.

[...]

PR #2832, opened 2024-07-25
Derive FromParam for Enum

loikki:
Implement  #2826


core/codegen/src/derive/from_param.rs
```diff
@@ -0,0 +1,43 @@
...
+        .inner_mapper(MapperBuild::new()
+                      .enum_map(|_, data| {
+                          let mut matches = quote!();
+
+                          for field in data.variants() {
+                              let field_name = &field;
+                              matches.extend(quote!(
+                                  stringify!(#field_name) => Ok(Self::#field_name),
+                              ))
+                          }
```
the10thWiz, line 29: Matches can instead be a `Vec`, and the inserted with `#(#matches)*`, which more closely matches existing code.

core/codegen/src/lib.rs
```diff
@@ -774,6 +774,28 @@ pub fn derive_from_form(input: TokenStream) -> TokenStream {
...
+/// (zero-length) fields:
+///
+/// ```rust
+/// # #[macro_use] extern crate rocket;
+/// #
+/// #[derive(FromParam)]
+/// enum MyParam {
+///     A,
+///     B,
+/// }
```
the10thWiz, line 789: Can we make this a full doc-test?

core/codegen/src/lib.rs
```diff
@@ -774,6 +774,28 @@ pub fn derive_from_form(input: TokenStream) -> TokenStream {
     emit!(derive::from_form::derive_from_form(input))
 }
 
+/// Derive for the [`FromParam`] trait.
+///
+/// The [`FromParam`] derive can be applied to enums with nullary
+/// (zero-length) fields:
```
the10thWiz, line 780: Add `/// [`FromParam`]: ../rocket/request/trait.FromParam.html` to this doc comment, so these links will link to the trait.

core/codegen/tests/from_param.rs
```diff
@@ -0,0 +1,18 @@
+use rocket::request::FromParam;
+
+#[derive(Debug, FromParam, PartialEq)]
+enum Test {
+    Test1,
+    Test2
+}
+
+
```
the10thWiz, line 9: Nit: remove multiple blank lines.

core/codegen/tests/from_param.rs
```diff
@@ -0,0 +1,18 @@
...
+
+#[test]
+fn derive_from_param() {
+    let test1 = Test::from_param("Test1").expect("Should be valid");
+    assert_eq!(test1, Test::Test1);
+
+    let test2 = Test::from_param("Test2").expect("Should be valid");
+    assert_eq!(test2, Test::Test2);
```
the10thWiz, line 17: We should also check the case where the value is not one of the enum variants.

core/codegen/src/lib.rs
```diff
@@ -774,6 +774,28 @@ pub fn derive_from_form(input: TokenStream) -> TokenStream {
...
+/// # #[macro_use] extern crate rocket;
+/// #
+/// #[derive(FromParam)]
+/// enum MyParam {
+///     A,
+///     B,
+/// }
+/// ```
+///
+/// Now `MyParam` can be used in an endpoint and will accept either 'A' or 'B'.
```
the10thWiz, line 792: Nit: A and B would look better in inline code blocks.

the10thWiz (changes requested):
Overall, this is a great start.

Ideally, we should include some ui-fail tests with this macro, see core/codegen/tests/ui-fail/from_form.rs for some examples.

@SergioBenitez, I think this should be good to go with the changes I've recommended, let us know if you have any other comments.

core/codegen/src/lib.rs
loikki: I am not sure what you mean by that? You mean ensuring that it does compile with `rustdoc --test`?

loikki:
Hi @the10thWiz 

Thanks for the feedback. I have applied all your suggestions but I am not sure what is the point of `core/codegen/tests/ui-fail/from_form.rs`. 

From the name, I guess everything should fail in it right? How do you check that each structure fails? In my understanding, just one failing would be sufficient for the test to pass, right?

the10thWiz:
> In my understanding, just one failing would be sufficient for the test to pass, right?

Not in this case. We use `trybuild` to run the ui-fail test cases, which actually validates the compiler output generated by the build failures. When you run the test, you can set an evironment variable to have trybuild save the current outpuut.

Use `./scripts/test.sh --ui` to execute the ui tests locally. (They are ignored by default, and only run on one platforms in CI since they can be somewhat finnicky).

loikki:
Thanks for the explanations. `trybuild` looks like a nice project :)

I have updated the tests with the correct stderr. Should I add it to nightly too or just stable?

core/codegen/src/derive/from_param.rs
```diff
@@ -0,0 +1,43 @@
...
+                          quote! {
+                              type Error = &'a str;
+                              fn from_param(param: &'a str) -> Result<Self, Self::Error> {
+                                  match param {
+                                      #(#matches)*
+                                      _ => Err("Failed to find enum")
+                                  }
+                              }
+                          }
+                      })
```
the10thWiz, line 40: Can you take a look at the indentation here?

core/codegen/tests/ui-fail/from_param.rs
```diff
@@ -0,0 +1,23 @@
...
+struct Foo2 {}
+
+#[derive(FromParam)]
+enum Foo3 {
+    A(String),
+    B(String)
+}
+
+#[derive(FromParam)]
+enum Foo4 {}
```
the10thWiz, line 18: Looking at the compiler output, it doesn't look like this derive fails, and I'm not sure whether it should. I think we should consider this a supported case, since it's useful for prototyping, but I'm open to arguments either way.

core/codegen/src/lib.rs
the10thWiz: I was thinking something more like
```rust
assert_eq!(MyParam::from_param("A").unwrap(), MyParam::A);
```
This can also be nice for documentation, since it shows what the parameter value is that can be parsed.

the10thWiz (changes requested):
I think we're very close! In terms of the ui-fail test, I think we will need a nightly version, but I might just take care of it myself.

core/codegen/tests/ui-fail/from_param.rs
loikki: I don't really have any opinion on it so let's allow it.

core/codegen/src/derive/from_param.rs
loikki: I don't see any issue with it. I checked with the auto-indent in my IDE and with `cargo fmt --check`.

loikki:
I have pushed the requested changes and wrote a few comments

core/codegen/src/derive/from_param.rs
the10thWiz: Rocket intentionally disables rustfmt (`cargo fmt`), since it doesn't handle some of the more complex trait bounds well. I suspect your IDE's auto-indent is based on rustfmt as well (I don't think there is another well built option out there yet), so that's likely why it's not picking up on the issues here.

The specific issue here is that not all the indentations are the same size (This is extremely obvious in my editor, since I have indent guides turned on). If you would like, I can push a fixed version to this PR.

the10thWiz:
You're close, just a few things to fix. The indentation on the one file needs to be fixed, and the test failure in the one doc-test needs to be addressed.

Other than that, I'm quite happy with this feature.

core/codegen/tests/ui-fail-nightly/from_param.stderr
```diff
@@ -0,0 +1,57 @@
...
+9 | struct Foo2 {}
+  | ^^^^^^^^^^^^^^
+
+error: [note] error occurred while deriving `FromParam`
+ --> tests/ui-fail-stable/from_param.rs:8:10
+  |
+8 | #[derive(FromParam)]
+  |          ^^^^^^^^^
+  |
+  = note: this error originates in the derive macro `FromParam` (in Nightly builds, run with -Z macro-backtrace for more info)
```
SergioBenitez, line 29: It's surprising that this isn't an actual `note`, since they exist on nightly. Is this being tested? Is this the output that's actually being produced?

core/codegen/src/derive/from_param.rs
```diff
@@ -0,0 +1,42 @@
...
+use devise::ext::SpanDiagnosticExt;
+use crate::exports::*;
+
+pub fn derive_from_param(input: proc_macro::TokenStream) -> TokenStream {
+    DeriveGenerator::build_for(input, quote!(impl<'a> #_request::FromParam<'a>))
+        .support(Support::Enum)
+        .validator(ValidatorBuild::new()
+                   .fields_validate(|_, fields| {
+                       if !fields.is_empty() {
+                           return Err(fields.span().error("Only empty enums are accepted"));
```
SergioBenitez, line 14: Maybe the wording here should be "only enums without variants are supported"?

core/codegen/src/derive/from_param.rs
```diff
@@ -0,0 +1,42 @@
...
+        .inner_mapper(MapperBuild::new()
+                      .enum_map(|_, data| {
+                          let mut matches = vec![];
+
+                          for field in data.variants() {
+                              let field_name = &field;
+                              matches.push(quote!(
+                                  stringify!(#field_name) => Ok(Self::#field_name),
+                              ))
+                          }
```
SergioBenitez, line 28: This can probably be a `.iter().map()`. I don't recall if we quote can use the iterator directly, but if it can, then that would save us the allocation here.

core/codegen/src/lib.rs
```diff
@@ -774,6 +774,33 @@ pub fn derive_from_form(input: TokenStream) -> TokenStream {
...
+/// #
+/// use rocket::request::FromParam;
+///
+/// #[derive(FromParam)]
+/// enum MyParam {
+///     A,
+///     B,
+/// }
+///
+/// assert_eq!(MyParam::from_param("A").unwrap(), MyParam::A);
```
SergioBenitez, line 793: Exhaustive positive tests when possible are always useful, as are negative tests. Say:

```rust
assert_eq!(MyParam::from_param("B").unwrap(), MyParam::B);
assert!(MyParam::from_param("a").is_err());
assert!(MyParam::from_param("b").is_err());
assert!(MyParam::from_param("C").is_err());
```

core/codegen/src/lib.rs
```diff
@@ -774,6 +774,33 @@ pub fn derive_from_form(input: TokenStream) -> TokenStream {
     emit!(derive::from_form::derive_from_form(input))
 }
 
+/// Derive for the [`FromParam`] trait.
+///
+/// The [`FromParam`] derive can be applied to enums with nullary
```
SergioBenitez, line 779: We need some description of what the macro produces. Is it a parser that stringifies each variant and then case-sensitively matches the stringified variant to the respective enum variant? If so, we should say that.

SergioBenitez (commented):
Does this correctly handle raw idents? IE:

```rust
enum Foo {
    r#for,
}

// parse "for" as `Foo::r#for`
```

Finally, it want to note that this macro is almost exactly the [`FromFormField`](https://api.rocket.rs/v0.5/rocket/derive.FromFormField) macro except that macro also allows renaming fields. Is there a reason to have two implementations, or can we simply factor out the existing `FromFormField` implementation for `FromParam`?

core/codegen/src/derive/from_param.rs
the10thWiz: Probably `Only enums without data fields are supported`.

SergioBenitez: I think the language I was looking for was getting away form "enums with foo" and instead going for "enum variants with foo," since that's the actual constant. We should also make sure the error is on the fields themselves.

core/codegen/src/derive/from_param.rs
loikki: Even after enabling the indent guide, I don't see it. So it would be nice if you can do it.

core/codegen/tests/ui-fail-nightly/from_param.stderr
loikki: My bad. I supposed that it would produce the same output. Just tested and fixed it.

core/codegen/src/derive/from_param.rs
loikki: `We should also make sure the error is on the fields themselves.` Do you mean the error message?

loikki:
I agree that `FromFormField` is very similar. I have little experience with rust macros and especially with devise (which is extremely badly documented). So I am not sure how would be the best approach to do it.

I am just wondering if it makes sense to merge them together. Is there any chance that `FromFormField` will be extended to more complex structures? `FromParam` is really restricted to "object" that can be matched to string as it is used in the query path.

By the way, raw identifiers are not working. How would do you like to proceed? Removing the `r#` from the string or add a parameter to rename the field (e.g. `#[field(value="for")]` for `FromFormField`.

core/codegen/src/derive/from_param.rs
the10thWiz: He means the associated span, and it looks like you already have it correct.

the10thWiz (approved):
Looks great! I'll take care of formatting and raw idents (`Ident` has an `unraw` method that strips the `r#`), and merge these changes.

Looking forward, I think we will want to add support for a rename attribute, as well as refactor both this and the `FromFormField` derive to share code as much as possible. `FromFormField` may eventually wind up with a more complex macro, since it does have quite a bit more context (i.e. the name, and potentially a whole file). It also matches case-insensitively, but I think the difference in behavior is justified.

the10thWiz:
Merged in 38cebebb!

PR #2834, opened 2024-07-30
Fix wrong URL in guide (configuration)

Javex:
Found a URL that was going to the wrong place. This should fix it.


the10thWiz (approved):
Nice Catch!

the10thWiz:
Merged in 509a033!

PR #2841, opened 2024-08-06
Allow TlsListener::from() to specify a resolver.

hcldan:


the10thWiz:
This specific change doesn't really make much sense, for a number of reasons. First, we would prefer to add support for creating a `TlsConfig` with a `DynResolver` directly (which would allow us to avoid using making the internal type public). Second, we are also interested in refactoring this code to support resolver-only configurations.

hcldan:
@the10thWiz I understand the concerns and I agree where you want to go is better.

I'm just trying to use what's there today. As you can see with how I modified the test, while the initial TlsConfig ends up working for the test, it's currently impossible for the test to use the DynResolver.

Using a fairing for this seems like kind of a strange dance, and I would prefer to construct a TlsConfig directly with a resolver, but this is where we currently are. I *might* be able to hide this more and have it done by rocket directly when you hand it the TlsListener, but I don't know all the code paths here and I don't want to break anything that already works. 

I would like to get support for this in the product we are looking to ship. It looks like I'm going to have to deal with a breaking change, regardless, moving forward... I'd appreciate some cooperation in getting something that works for us in the interim, even if it "doesn't really make sense".

hcldan:
I've implemented my own TlsListener for my needs. This isn't necessary.

PR #2847, opened 2024-08-17
Allow HTTP extension methods

lobsterwise:
While you can create routes with HTTP methods such as PROPFIND and REPORT, Rocket will return 400 errors when handling these requests as it errors for any unknown method. This was fixed by just parsing unknown Hyper methods into Rocket methods.


SergioBenitez:
Fun! There's actually a test for this, but it doesn't go through the full server. Can you add a testbench test (testbench/src/servers) that dispatches an http request with an extension method and ensures it works?

PR #2852, opened 2024-08-23
Not enable the rustls feature by default for sqlx

guenhter:
DB's like sqlite not need the tls feature with sqlx at all. When cross compiling e.g. to i586, ring is not supported and therefore the db-pool cannot be used because the rustls feature is turned on by default even though not needed.


guenhter:
@SergioBenitez Is there anything else I should do to make this PR move along faster? 
The issue really hinders me and I have to patch all other libs too which I use and depend on rocket to make it working with cargo and this version of the PR.

PR #2853, opened 2024-08-24
Small change: Improve clarity in docs/06-responses.md

aznszn:
I was reading the docs and thought the wording here could be improved.


docs/guide/06-responses.md
```diff
@@ -33,8 +33,7 @@ where `R` is some type that implements `Responder`:
 struct WrappingResponder<R>(R);
 ```
 
-A wrapping responder modifies the response returned by `R` before responding
-with that same response. For instance, Rocket provides `Responder`s in the
+A wrapping responder modifies the response returned by `R` before sending the modified response. For instance, Rocket provides `Responder`s in the
```
SergioBenitez, line 36:
```suggestion
A wrapping responder takes a response from `R`, modifies it, and returns it. For instance, Rocket provides `Responder`s in the
```

SergioBenitez (commented):
How about this? Is this clearer? I think saying "sends" might imply that the response is actually sent to the client, which it isn't.

aznszn:
> How about this? Is this clearer? I think saying "sends" might imply that the response is actually sent to the client, which it isn't.

Yes this is much clearer

PR #2868, opened 2024-09-23
Update Cargo.toml

Dylan-DPC:
Bumps the dependency to at least 2.2.3 as per the advisory

[Advisory](https://rustsec.org/advisories/RUSTSEC-2024-0365.html)

(even though this crate may not be using the function mentioned in the advisory, it is better to bump it to ensure that there is no chance of other dependencies pulling it)

### DIFF
--- a/contrib/sync_db_pools/lib/Cargo.toml
+++ b/contrib/sync_db_pools/lib/Cargo.toml
@@ -23,7 +23,7 @@ r2d2 = "0.8"
 tokio = { version = "1.6.1", features = ["rt", "rt-multi-thread"] }
 serde = { version = "1.0", features = ["derive"] }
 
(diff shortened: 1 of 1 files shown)


the10thWiz:
It's unclear to me that this brings significant benefit to Rocket users. Rocket already enforces data limits for user requests, so it's highly unlikely a problematic sql query could be built from a Rocket request.

The main issue with upgrading right away, is that we also need a compatible version of `diesel-async`, which supports the newest version `diesel`. This is noted in the CI failures as well.

sawa-ko:
Please update diesel and diesel async 🙏

Andynopol:
Hi! I have to ask, diesel is on version 2.2.12 and rocket supports it until 2.1.6.

When can we expect a sync for the databases for the `rocket_db_pools`?

PR #2874, opened 2024-10-03
Add support for QUERY method

### LINKED ISSUE #2873: Implement QUERY Method
### What's missing?

I am a _huge_ fan of Rocket and thought that I could add support for the nearly-standardized QUERY method (https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-05.html). I have a _WIP_ PR that I will post to go along with this Feature Request. I have done my absolute best to make sure that I have followed all the proper steps.

### Ideal Solution

I believe that the ideal solution is to simply add `QUERY` in the appropriate spot in the call(s) to the `define_methods` macro. Because QUERY is not yet an RFC, there will be a few changes to the `define_methods` macro so that the documentation is generated correctly. 

N/A

N/A

### Alternative Solutions

_No response_

### Additional Context

Again, thank you for all the work that you are doing on Rocket. I understand that the PR that I will post does not have tests and that is absolutely a requirement. However, I wanted to post early/often to see whether support for `QUERY` is something that you would consider. If it is, then I will absolutely add all the tests necessary and clean up anything else that I have done incorrectly. 

### System Checks

- [X] I do not believe that this feature can or should be implemented outside of Rocket.
- [X] I was unable to find a previous request for this feature.

hawkinsw:
Closes #2873.

### DIFF
--- a/core/codegen/src/lib.rs
+++ b/core/codegen/src/lib.rs
@@ -293,6 +293,7 @@ route_attribute!(delete => Method::Delete);
 route_attribute!(head => Method::Head);
 route_attribute!(patch => Method::Patch);
 route_attribute!(options => Method::Options);
+route_attribute!(query => Method::Query);
 
 /// Attribute to generate a [`Catcher`] and associated metadata.
 ///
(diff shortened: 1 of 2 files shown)


SergioBenitez:
See my comment in https://github.com/rwf2/Rocket/issues/2873#issuecomment-2511104597.

PR #2878, opened 2024-10-22
Fix minor typos in guide and docstring

timmc:
Just some things I noticed while reading through the docs.


the10thWiz (approved):
Thanks for these corrections!

PR #2891, opened 2024-11-24
Fix percent-encode doctests to actually test the functions

timmc:
One doctest was still calling decode instead of encode, and the other was missing an assert.


SergioBenitez:
Merged in ea0a1af683e1e1330825963d6d68c518b29952e9.

PR #2892, opened 2024-12-02
Fix contrib package versions

grauwoelfchen:
Hello,

I just noticed that the version numbers of some `contrib` packages on the `master` branch are older than the latest versions actually released on on crates.io, while the `v0.5` branch already has the correct values.

I understand that the `master` branch contains unpublished changes (like dependencies etc.), but I was a bit confused by these relapse differences when reading the code at a glance 😅

## Changes

To catch up with the latest versions, I've updated the values in each `Cargo.toml` (and also docs/guide, doc comments and each `README.md`, according to the changes in the `v0.5` branch) as follows:

| contrib | version |
|---|---|
| rocket_db_pools | `0.1.0` -> [0.2.0](https://crates.io/crates/rocket_db_pools/versions) |
| rocket_db_pools_codegen | `0.1.0` -> [0.2.0](https://crates.io/crates/rocket_db_pools_codegen/versions) |
| rocket_dyn_templates | `0.1.0` -> [0.2.0](https://crates.io/crates/rocket_dyn_templates/versions) |
| rocket_ws | `0.1.0` -> [0.1.1](https://crates.io/crates/rocket_ws/versions) |

When the next version (`0.6`?) is released, these values perhaps may be different. But until then, I hope these changes make sense and avoid any misunderstanding 🙆🏽‍♀️

### DIFF
--- a/contrib/db_pools/README.md
+++ b/contrib/db_pools/README.md
@@ -17,7 +17,7 @@ full usage details.
 
    ```toml
    [dependencies.rocket_db_pools]
-   version = "0.1.0"
+   version = "0.2.0"
    features = ["sqlx_sqlite"]
    ```
 
--- a/contrib/db_pools/codegen/Cargo.toml
+++ b/contrib/db_pools/codegen/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rocket_db_pools_codegen"
-version = "0.1.0"
+version = "0.2.0"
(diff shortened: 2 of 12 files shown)


the10thWiz (approved):
This looks like the correct move, although I'll likely have to actually go through the git history to figure out what the final values should be.

If we have unreleased versions in master (which happens fairly often), we should bump the version (and potentially append `-dev`). Thanks for catching this!

the10thWiz:
This change was applied in `f70114a5973d7be4ac22d452162b5ac0771909ff`

PR #2898, opened 2025-01-01
Fix Dockerfile example in "Deploying" chapter

padenot:
This now matches the layout of a typical project, instead of erroring out.

### DIFF
--- a/docs/guide/11-deploying.md
+++ b/docs/guide/11-deploying.md
@@ -262,9 +262,9 @@ WORKDIR /app
 COPY --from=build /build/main ./
 
 ## copy runtime assets which may or may not exist
-COPY --from=build /build/Rocket.tom[l] ./static
-COPY --from=build /build/stati[c] ./static
-COPY --from=build /build/template[s] ./templates
+COPY --from=build /build/Rocket.tom[l] ./
+COPY --from=build /build/stati[c] ./static/
+COPY --from=build /build/template[s] ./templates/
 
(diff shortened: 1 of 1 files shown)


docs/guide/11-deploying.md
```diff
@@ -262,9 +262,9 @@ WORKDIR /app
 COPY --from=build /build/main ./
 
 ## copy runtime assets which may or may not exist
-COPY --from=build /build/Rocket.tom[l] ./static
-COPY --from=build /build/stati[c] ./static
-COPY --from=build /build/template[s] ./templates
+COPY --from=build /build/Rocket.tom[l] ./
+COPY --from=build /build/stati[c] ./static/
+COPY --from=build /build/template[s] ./templates/
```
SergioBenitez, line 267: Are these changes functional?

padenot: Yes and no. They aren't strictly needed and generally if the fix on the previous line is applied, it will work the same, but if the next two lines are applied and not the first, the error message can change, maybe depending on the `coreutils` version one uses.

I don't feel strongly about this, I thought it was a touch clearer. The first line is the important fix.

the10thWiz (approved):
I like these changes. Doing a quick test on one of my test projects, it looks like without the trailing slashes, it will silently fail by copying the Rocket.toml to a file called static (which was confusing to debug).

the10thWiz:
Merged in 1551b1bc

PR #2923, opened 2025-03-16
Fixing typo in 12-pastebin.md

JonathanMcCormickJr:
Correcting `crate` to instead be `create` due to the context of how it's used.


the10thWiz (approved):
Nice catch.

PR #2926, opened 2025-03-29
Fix rocket db pools config

dfego:
See: https://github.com/rwf2/Rocket/issues/2925


the10thWiz (approved):
Looks good to me. I'm not sure why the workflows didn't trigger (I'll have to look into it), but that should be the only thing left before we can merge this.

the10thWiz:
Merged in 207e1226

PR #2930, opened 2025-04-10
Add 'Request::version()' to return `hyper` version (non-local)

hcldan:
Adds a method (`Request::version()`) to return the hyper version of the request.

It is optional so that local clients don't need to deal with it. It seemed like overkill to add the extra arg to all the local client request methods. I'm not sure how to deal with an example or test for this unless you end up wanting to expose it to the local clients, but I don't know if it makes sense to in that context.


the10thWiz:
Merged (with changes) in 4358dae5

PR #2932, opened 2025-04-14
Improve db_pools init: do not crash if DB unavailable during startup (sqlx)

### LINKED ISSUE #2931: rocket_db_pools: Allow pool creation to succeed even if DB unavailable (sqlx, maybe others)
### What's missing?

The `sqlx` driver for `rocket_db_pools` currently crashes at startup if your database is offline or if anything else fails -- sslrootcert is invalid/missing, user doesn't exist, password wrong, etc etc. This is sometimes useful in development as you are reminded to start your database if you forgot, but a log would work just as well. In production it is a bad choice.

The [`deadpool`](https://docs.rs/deadpool/latest/deadpool/) crate's second-top-billed feature is that pool creation never fails even if the database is unavailable:

> **Identical startup and runtime behaviour.** When writing long running application there usually should be no difference between startup and runtime if a database connection is temporarily not available. Nobody would expect an application to crash if the database becomes unavailable at runtime. So it should not crash on startup either. Creating the pool never fails and errors are only ever returned when calling Pool::get().

I'd like that, basically for the reasoning given there. It is needlessly crashy to do so only at startup.

To motivate somewhat more, if you are not convinced:
- I am currently debugging the provision + rotation of database credentials and TLS certificates to a Rocket application, and it is quite annoying having to restart a crash-loop-backoff rocket application deployment to see if one of the changes I have made works. K8s is able to update a rotated sslrootcert certificate file in-place without rebooting a container -- because of the crash behaviour, Rocket does not.
- It would be much easier to recover from a failed database cluster if a crashed or recovering database didn't completely bring down everything else that used it. The startup behaviour means all the alarms go off during a deployment when in fact your database went down an hour ago and you already know that. It means you have to reboot more things more times, while you're already in an emergency, but *also, you must wait until your DB has finished recovering from a backup before rebooting everything else*, so this either
    - forces you to be really careful not to deploy to your HTTP servers, or allow them to crash, during a DB meltdown
    - or necessarily slows down the app recovery, by forcing it to happen serially

### Ideal Solution

The sqlx implementation of rocket_db_pools' Database trait [uses `Pool::connect_with`](https://docs.rs/rocket_db_pools/0.2.0/src/rocket_db_pools/pool.rs.html#316) instead of [Pool::connect_lazy_with](https://docs.rs/sqlx-core/0.8.3/sqlx_core/pool/struct.Pool.html#method.connect_lazy_with). You can make a one-line change to switch to `connect_lazy_with`, and make it behave like deadpool.

You could also make it configurable, but if people really want their app to crash in this case, they can follow deadpool's advice and add a fairing that tries to make a connection and crashes the app. People writing `.unwrap()` in a fairing is probably easier than making this configurable.

It has been implemented outside of rocket. The deadpool backend behaves this way, I presume. But rocket_db_pools does not configure sqlx to do it.

To work around this limitation, you could probably use a custom Database or Pool implementation, but you would need to make your own wrapper type for `sqlx::Pool` in order to get around the fact that rocket_db_pools implements its traits on that. I will probably do this myself for now, if it turns out to be easy I'll post a snippet.

### Alternative Solutions

_No response_

### Additional Context

_No response_

### System Checks

- [x] I do not believe that this feature can or should be implemented outside of Rocket.
- [x] I was unable to find a previous request for this feature.

cormacrelf:
Fixes #2931 

## Why?

When using `Pool::connect[_with]`, sqlx attempts to connect to the given
database immediately, and the fairing will fail if there are any
problems in that attempt (beyond obvious configuration problems that are
found before hitting the network), e.g.:

- the database is unavailable; or
- the username/password is incorrect; or
- the ssl configuration is invalid; or
- any other connection issue.

There are a few pros and cons to this approach:

Pros:

- In development, configuration errors are surfaced slightly faster

Cons:

- Databases are expected to be unavailable sometimes. It does not
  normally crash a server if one becomes unavailable after startup,
  so why should it prevent a server from starting at all? See
  [deadpool's justification]{https://docs.rs/deadpool} for not crashing.
- In production/testing, slower to debug configuration or networking
  errors as your edit-test loop now involves restarting an application
  rather than refreshing a page or trying a request again.
- Causes database or configuration issues to appear as "failed
  deployments" in standard deployment scenarios.
- Introduces hard ordering constraints on operator actions during
  database recovery, requiring reboots to follow a functioning database
  or applications not to be restarted at certain times

## Effect of change

The sqlx backend now behaves like the deadpool backend: no connection
issues are surfaced during startup. You will not see them until you
attempt to get a connection from the pool. That means rocket will launch
and you can find problems like these in smoke tests.


cormac-ainc:
Test failures are the same as on master)

the10thWiz (approved):
I think this is a good change. More generally, we should consider also attempting to detect database issues and report them, even if the server continues running.

This would likely take the form of spawning a task to pull a connection from the pool, and potentially run some kind of basic sql (e.g. `select 1;`). We could then provide an error log indicating that the db is currently inaccessible.

the10thWiz:
Merged in 64ef4e93

PR #2943, opened 2025-05-23
expose the sni server name.

### LINKED ISSUE #2942: HTTP spec compliance with TLS handshake
### What's missing?

Not sure if I should file this as a bug or a feature request... doing the latter for now.

It appears that Rocket may be violating the HTTP (TLS extension) spec by not validating (or optionally not letting the crate user validate) the host header provided to the server against the TLS handshake requested SNI.

conversation: https://matrix.to/#/!kDIcCXWSVfdahNCJWq:mozilla.org/$20txAsjUwc3wFfiaMRUflxYLSH1Q2KUgnLMYzSLCVwQ?via=mozilla.org&via=matrix.org&via=catgirl.cloud

relevant http/1.1 spec: https://www.rfc-editor.org/rfc/rfc6066#section-11.1
relevant http/2 spec: https://httpwg.org/specs/rfc7540.html#reuse

### Ideal Solution

_No response_

It appears only the TlsConfig is presented via the request.remote() function. It would be best if it provided the resolved `ServerConfig` instead... and also the handshake data which contains the requested SNI.

_No response_

### Alternative Solutions

_No response_

### Additional Context

_No response_

### System Checks

- [x] I do not believe that this feature can or should be implemented outside of Rocket.
- [x] I was unable to find a previous request for this feature.

hcldan:
Closes #2942


hcldan:
@the10thWiz Can you please review this?

Issue #2969, opened 2025-09-28
`./scripts/test.sh --contrib` fails on current git HEAD


mcclure:
### Rocket Version

ec44ab7e058105a2127bfd6d2e1109c4fbcf9faa

### Operating System

Debian 13

### Rust Toolchain Version

rustc 1.88.0 (6b00bc388 2025-06-23) also tested with rustc 1.90.0 (1159e78c4 2025-09-14)

### What happened?

  I am writing a PR. The PR contribution guidelines say to run `./scripts/test.sh --contrib`. It fails in my PR. But it **also** fails on current git head, ec44ab7e05810 . It fails with:

```rust
error[E0433]: failed to resolve: could not find `try_join` in `futures_util`
   --> /home/mcc/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/diesel-async-0.5.2/src/pg/mod.rs:429:23
    |
429 |         futures_util::try_join!(
    |                       ^^^^^^^^ could not find `try_join` in `futures_util`

For more information about this error, try `rustc --explain E0433`.
error: could not compile `diesel-async` (lib) due to 1 previous error
```

I don't quite understand what's happening here, and it's not clear to me the problem is in Rocket. If I had to guess, I'd guess cargo is for some reason picking an inappropriately old version of `futures_util`, and a Cargo.toml somewhere needs a more explicit version number.

### Test Case

```rust
./scripts/test.sh --contrib
```

### Log Output

```shell
n/a, rocket is not running
```

### System Checks

- [x] My bug report relates to functionality.
- [x] I have tested against the latest Rocket release or a recent git commit.
- [x] I have tested against the latest stable `rustc` toolchain.
- [x] I was unable to find this issue previously reported.

the10thWiz:
This is a weird issue. Doing a quick test, `diesel_async` cannot be built with the following set of features: `async-connection-wrapper,deadpool,pool,postgres,tokio,tokio-postgres`. (In fact, you only need to enable the `postgres` feature to cause the compilation failure).

The issue isn't present in `diesel_async` 0.6 and later (as far as I can tell), which might be the correct move going forward.

Issue #2967, opened 2025-09-26
Support diesel+sqlite in rocket_db_pools


mcclure:
### What's missing?

I am trying to develop a website with Rocket+Yew for the frontend and Diesel+Sqlite for the backend.

Rocket prefers database connections to be stored in rocket_db_pools. rocket_db_pools has drivers for Diesel+Postgres and Diesel+Mysql. However, it does *not* support Diesel+Sqlite: https://api.rocket.rs/v0.5/rocket_db_pools/#supported-drivers

### Ideal Solution

I believe the reason you do not support sqlite is that you use [diesel_async](https://docs.rs/diesel-async/latest/diesel_async/), which as of its 2023 release did not support sqlite. Since then however diesel_async has added a `SyncConnectionWrapper` class which is used to access Sqlite. As far as I know, your existing code for `diesel::PgPool` or `diesel::MysqlPool` could simply be duplicated adding support for this class.

### Why can't this be implemented outside of Rocket?

By providing rocket_db_pools and republishing the `::diesel` crate Rocket has essentially claimed responsibility for exposing diesel's feature set. The work appears to already be done outside Rocket, but plumbing within Rocket is needed to expose it.

### Are there workarounds usable today?

?

### Alternative Solutions

?

### System Checks

- [x] I do not believe that this feature can or should be implemented outside of Rocket.
- [x] I was unable to find a previous request for this feature.

mcclure:
I have an attempted PR for this #2970

https://github.com/mcclure/Rocket/tree/diesel_sqlite_feature

I will attempt to test it with my live code later this week.

On an related note, while looking through the codebase I realized this existed:

https://api.rocket.rs/v0.5/rocket_sync_db_pools/

But I don't understand whether it solves my problem, or how you'd decide whether to use `rocket_sync_db_pools` or `rocket_db_pools`.

PR #2970, opened 2025-09-28
Introduce diesel_sqlite feature (for rocket_db_pools support)

mcclure:
Attempt at #2967 . I am pretty uncertain about this and I do not yet know how to test this (aside from the basic smoke test in the PR— see below).

I ran `./scripts/test.sh --contrib` as requested in the testing guide. It failed because of #2969. diesel.rs has three tests in it for MysqlPool, PgPool, and (now) SqlitePool, but it appears only the Mysql test is run. I don't know if this is because of #2969 or some other reason.


mcclure:
I haven't heard any updates about this in almost a year. I bumped a merge with newest master.

I am making use of this patch in a project, and in this project, I found I needed at least diesel 0.6.1 and diesel-async 2.3.2 for feature reasons. I have created also a PR #3005 which bumps these projects to their actual newest versions (0.9.2 and 2.3.12). For ease of my own testing, I based 3005 atop this patch.

Any chance of movement on these PRs? They're straightforward changes and they're tested.

Issue #2946, opened 2025-06-29
Support gRPC


JasonHonKL:
### What's missing?

I understand that rocket is focusing on web framework. However as micro services become more popular these days we should also provide gRPC support in the framework. 

### Why can't this be implemented outside of Rocket?

It can be implement outside of rocket however if we want to use one library to acting as a micro service backend then we shall allow it to do so inside the rocket library. 

### System Checks

- [x] I do not believe that this feature can or should be implemented outside of Rocket.
- [x] I was unable to find a previous request for this feature.

jbcurtin:
+1 for gRPC support.

-  Something that makes it easy to configure multiple keys per deployment would be an awesome feature. For example, a tenant might be a part of a multi-tier CA. Root, Intermediate/Tenant, Service. Having the key details in a method handler would be a really nice auth feature.
- rustls before native-tls is a nice to have, in my opinion 
- I imagine there might need to be a different API than form handlers, for gRPC. However, I'd be impressed if you were able to integrate the two using traits
- I could help test this

the10thWiz:
I don't think this, as stated, belongs inside the Rocket repos. As noted, gRPC can be implemented outside of Rocket, and I don't think adding one (and supporting it) as a contrib library within the Rocket repository is worth the effort. Additionally, unless you have a very specific need to build both a full-fat HTTP api, as well as a micro-serivce interface, you're going to much better off using either Rocket or a gRPC library like Tonic - but not both.

PR #2971, opened 2025-09-28
Proposal for Support gRPC using Tonic

davthecodercom:
Adding Support for gRPC #2946 

## Test
```bash
cd examples/grpc
cargo run
```

```bash
# Install grpcurl
brew install grpcurl

# Test unary call
grpcurl -plaintext -proto proto/greeter.proto -d '{"name": "World"}' localhost:50051 greeter.Greeter/SayHello
{
  "message": "Hello World! (from gRPC)"
}

# Test streaming call  
grpcurl -plaintext -proto proto/greeter.proto -d '{"name": "World"}' localhost:50051 greeter.Greeter/SayHelloStream
{
  "message": "Hello World (stream message 1)!"
}
{
  "message": "Hello World (stream message 2)!"
}
{
  "message": "Hello World (stream message 3)!"
}
{
  "message": "Hello World (stream message 4)!"
}
{
  "message": "Hello World (stream message 5)!"
}
```

Test HTTP endpoints:

```bash
curl http://localhost:8000/
curl http://localhost:8000/stats
```

#### Key Features

1. **Dual Protocol Support**: Serve both HTTP and gRPC from the same application
2. **Shared State**: Access Rocket's managed state from gRPC services
3. **Streaming Support**: Full support for unary and streaming gRPC methods
4. **Easy Integration**: Uses Rocket's fairing system for seamless integration
5. **Tonic Integration**: Built on top of the tonic gRPC library

### DIFF
--- a/contrib/grpc/src/lib.rs
+++ b/contrib/grpc/src/lib.rs
@@ -35,6 +35,9 @@ use rocket::{Rocket, Build, fairing::{Fairing, Info, Kind}};
 #[cfg(feature = "tonic")]
 use tonic::transport::Server;
 
+#[cfg(feature = "tonic")]
+use std::error::Error;
+
 pub use self::service::*;
 
 mod service;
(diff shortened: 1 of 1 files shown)


the10thWiz:
For the same reasons outlined in my comment on #2946, this will not be merged into Rocket. You are more than welcome to publish this crate yourself, but it does not belong in the main Rocket repository.

PR #3007, opened 2026-08-14
Correct docs to reflect new log_level names in post-tracing Rocket

mcclure:
In 0.5.1, the log levels are "off", "normal", "debug", and "critical". In current `master`, due to adoption of the `tracing` crate, they are "off", "error", "warn", "info", "debug", "trace", or 0-5. This is not currently reflected in the documentation. (Which is bad, because if you check out `master` and follow the documentation currently in the `docs/` folder, you will set your log_level to "normal" or "critical", and the program will fail to boot.) Patch addresses this.

Quirks of patch:

- After switching to tracing, it appears the numbers 0-5 are also allowed as values for `log_level`. I don't know what strings these correspond to, so I didn't try to document them. I will fix this if you ask.
- I changed one line to fix an ambiguity caused by the fact `info` and `debug` are now separate log levels.

### DIFF
--- a/docs/guide/03-getting-started.md
+++ b/docs/guide/03-getting-started.md
@@ -92,7 +92,7 @@ run`. You should see the following:
    >> limits: [..]
    >> tls: disabled
    >> temp dir: /tmp
-   >> log level: normal
+   >> log level: info
    >> cli colors: true
 🛰  Routes:
    >> (index) GET /
--- a/docs/guide/04-overview.md
+++ b/docs/guide/04-overview.md
@@ -176,7 +176,7 @@ Running the application, the console shows:
    >> limits: [..]
    >> tls: disabled
    >> temp dir: /tmp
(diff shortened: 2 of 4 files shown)


martynp:
Thank you for the patch - I had already addressed most of these issues on my fork (https://github.com/rustfoo/rkt/commit/a6aaece88fd2024136ab14eea6630fe2c843b121 - went in a while ago, but my git foo is bad this morning I re-wrote the history fixing a typo)

But you have highlighted a few areas that I missed. Your patch is not directly applicable to my fork as I have moved a lot of things around - also the banner blocks have changed significantly - I have created a new PR incorporating those fixes here https://github.com/rustfoo/rkt/pull/20.

mcclure:
All right— in case we are eventually able to get hold of the original Rocket maintainer, are there additional documentation fixes from your fork I should try to port back into this PR?

PR #3010, opened 2026-08-23
Add Tera v2 support

axoking:
I added a feature `tera2` which will use tera v2 instead of the deprecated v1. I left the old `tera` feature working for compatibility.

An `if` condition that checks if the template exists could be omitted, because tera 2 early returns before that function is even called if the template does not exist.

I did not write any documentation or testing, because I'm not familiar with that. But I tested both `tera` and `tera2` with a seperate crate I wrote, and they work and render using the correct tera version.


contrib/dyn_templates/src/lib.rs
```diff
@@ -180,13 +180,19 @@
 #![doc(html_favicon_url = "https://rocket.rs/images/favicon.ico")]
 #![doc(html_logo_url = "https://rocket.rs/images/logo-boxed.png")]
 
+#[cfg(all(feature = "v1", feature = "v2"))]
+compile_error!("features \"v1\" and \"v2\" are mutually exclusive — enable only one");
```
peddermaster2: Shoudn't this be `tera` & `tera2`

```suggestion
#[cfg(all(feature = "tera", feature = "tera2"))]
compile_error!("features \"tera\" and \"tera2\" are mutually exclusive — enable only one");
```

axoking: fuck yeah you're right, I copied those two lines from somewhere because I didn't know how to forbid using two features together in rust.

That's really unprofessional I'm sorry

axoking:
I just realized that this also breaks tera autoescaping for some reason.

I understand that the .html.tera files are not auto-escaped on version 2, because they are registered to tera without their file extension. However, I don't understand why this wasn't a problem with tera v1.

martynp:
@axoking Thanks for the PR!

It wasn't an issue with Tera v1 because it had a fallback to look for the files the extensions - this functionality is different in v2.

A few points to note:

* We should avoid mutually exclusive features (https://doc.rust-lang.org/cargo/reference/features.html#feature-unification, and https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features) - so show preference to v2 over v1 if both are selected.
* The new feature has no test coverage.
* The documentation build will fail because it forces `all-features = true` - so the compile_error! guard will trigger.

I have integrated a tera_v2 feature into github.com/rustfoo/rkt with the above fixed. I have some more revisions to make before making it public - for rkt I am contemplating making the tera feature use tera_v2 and tera_v1 point to the now deprecated features.

axoking:
@martynp Thank you!

So you got a solution for auto escaping? I'm also currently looking into the tera crate to see if you could add auto escaping functionality there, e.g. getting the extension from the file path, or giving other crates an option to manually enable auto-escaping on files (maybe this is already an option, I will have a look later)

martynp:
@axoking Yes — it's working in rustfoo/rkt now.

Rather than change how templates are registered I kept the names and made the escape list explicit: pick out the escapable templates by their source path and pass their names to `autoescape_on()`.

axoking:
Ok nice! I will check that out

I'm also working on the autoescaping stuff here: https://github.com/Keats/tera/issues/1028

